PR #12, opened 2024-03-06
docker/explorer: add explorer container image

Fadude:


docker/explorer/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM node:20-slim AS build
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
+RUN mkdir /app
```
zeevmoney, line 5:
```suggestion
```

docker/explorer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+COPY ./Cargo.lock /app/Cargo.lock
+
+COPY ./package.json /app/package.json
+COPY ./pnpm-lock.yaml /app/pnpm-lock.yaml
+COPY ./pnpm-workspace.yaml /app/pnpm-workspace.yaml
+COPY ./turbo.json /app/turbo.json
+
+COPY ./sdk /app/sdk
+COPY ./apps /app/apps
+COPY ./dapps /app/dapps
```
zeevmoney, line 19:
```suggestion
COPY /sdk /app/sdk
COPY /apps /app/apps
COPY /dapps /app/dapps
COPY Cargo.toml Cargo.lock /app/
COPY package.json /app/package.json
COPY pnpm-lock.yaml /app/pnpm-lock.yaml
COPY pnpm-workspace.yaml /app/pnpm-workspace.yaml
COPY .eslintrc.js .npmrc turbo.json /app/
```

docker/explorer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+WORKDIR /app
+
+RUN pnpm install
+RUN pnpm explorer build
```
zeevmoney, line 24:
```suggestion
# --mount=[...] will ensure a persistent cache volume is mounted at /pnpm_store.
# This is like running docker run -v /tmp/cache:/pnpm_store ... except that mount is
# only available at build time. Our resulting Docker image won't have this mount available after build.
RUN --mount=type=cache,id=pnmcache,target=/pnpm_store \
  # pnpm config set store-dir /pnpm_store instructs pnpm to use /pnpm_store directory (mounted as cache volume)
  # to save downloaded packages and re-use existing one if possible.
  pnpm config set store-dir /pnpm_store && \
  # `pnpm config set package-import-method copy` instructs pnpm to copy packages directly in `node_modules/`
  # instead of symlinking to store. Otherwise pnpm would create symlinks from `node_modules` to
  # `/pnpm_store` which wouldn't exist after build time.
  # By copying we ensure that packages are effectively saved in resulting Docker image.
  pnpm config set package-import-method copy && \
  pnpm install --prefer-offline --ignore-scripts --frozen-lockfile
RUN pnpm explorer build
```

docker/explorer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+FROM nginx:alpine
+COPY --from=build /app/apps/explorer/build /usr/share/nginx/html
```
zeevmoney, line 27:
```suggestion
FROM nginx:alpine as release
COPY --from=build /app/apps/explorer/build /usr/share/nginx/html
```

zeevmoney (changes requested):
see comments, I added build optimization with docker buildkit.
Also in the file:

`dockerignore`, add the following:

node_modules
*/node_modules

zeevmoney (approved):

PR #23, opened 2024-04-17
feat: sui state proofs

alexanderludwig:


crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
ycscaly, line 1: format

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,145 @@
...
+            event_type_layout: event_type_layout,
+            message_field_name: message_field_name,
+        };
+
+        let first_committee = EpochCommittee {
+            id: object::new(ctx),
+            committee: init_committee,
+        };
+
+        event::emit(EpochCommitteeSubmitted {
```
ycscaly, line 64: You implicitly assume that this is the genesis committee, but I'm not conviced that is the case and I'm also uncertain that's what we'll initialize the module with; it's possible to just say "I choose the current epoch, trust the current committee, and since dWallets are not in use yet I can start from here"

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,145 @@
...
+        });
+        
+        transfer::share_object(registry);
+        transfer::freeze_object(config);
+        transfer::freeze_object(first_committee);
+    }
+
+    
+
+    native fun sui_state_proof_verify_committee(prev_committee: vector<u8>, checkpoint_summary: vector<u8>): vector<u8>;
```
ycscaly, line 76: Place all native imports in a single place (either in the top or bottom of the file)

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,145 @@
...
+        ctx: &mut TxContext,
+    ) {
+        let committee_verified_bytes = sui_state_proof_verify_committee(prev_committee.committee, new_checkpoint_summary);
+
+        let committee_new = EpochCommittee {
+                                    id: object::new(ctx),
+                                    committee: committee_verified_bytes,
+                                    };
+
+        registry.highest_epoch = registry.highest_epoch + 1;
```
ycscaly, line 92: why is this true?
this proof only passes for sequtnial changes? there's no way to prove an epoch that's further in time than the immediately next epoch?

ycscaly: is there no way to get the epoch from the proof anyways? seems like a best practice. Perhaps after that you could assert it is indeed `prev_epoch + 1`

ycscaly: In any case this needs proper documentation

ycscaly: Here's an attack for you. 
I start in epoch `x`, proving it correctly yields an `EpochCommittee`. 
Now I prove epoch `x + 1` using this `EpochCommittee` object. This increases the registry's highest epoch to `x+1`. It creates an `EpochCommittee` for epoch `x+1`.
Now I submit the same proof again against the old `EpochCommittee` for epoch `x`. This proof verifies, and a new `EpochCommittee` object with the same value is created, again for epoch `x+1`. But now, the registry's highest epoch is set to `x+2`, wheras there's not a single `EpochCommittee` object for that epoch. I can do this indefinetely. Also, I can overflow the u64 potentially (perhaps the fees won't make it possible tho) and even in the same transaction.

ycscaly: Notice that in this attack you are emitting events that has a wrong `EpochCommittee` to `epoch` relation

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,145 @@
...
+        checkpoint_summary: vector<u8>,
+        checkpoint_contents: vector<u8>,
+        transaction: vector<u8>,
+        ctx: &mut TxContext
+    ){        
+        
+        let (cap_id_bytes, _) = sui_state_proof_verify_transaction(committee.committee, checkpoint_summary, checkpoint_contents, transaction, config.event_type_layout, config.package_id );
+        let wrapper = CapWrapper {
+            id: object::new(ctx),
+            cap_id_sui: object::id_from_bytes(cap_id_bytes),
```
ycscaly, line 116: Notice that we are calling the `object::id_from_bytes()` function, that takes a *sui* serialized object, and attempts to parse it using the *dwallet network* code. 

This works now because we are a sui fork, and this structure hasn't changed in the version differences.

However, this is the wrong approach.

Instead, we should somehow import the *sui* `object::id_from_bytes` function and call it.
We should also pay attention to version updates to see it won't be changed.

Alternatively, we should be in touch with the sui team to assure this object structure will never be changed.

In any case this needs proper thinking

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,145 @@
...
+            cap: dwallet_cap,
+        };
+
+        transfer::share_object(wrapper);
+    }
+
+
+
+    native fun sui_state_proof_verify_transaction(committee: vector<u8>, checkpoint_summary: vector<u8>, checkpoint_contents: vector<u8>, transaction: vector<u8>,  event_type_layout: vector<u8>,  package_id: vector<u8>): (vector<u8>, vector<u8>);
```
ycscaly, line 125: put this somewhere else

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,145 @@
...
+        committee: &EpochCommittee,
+        checkpoint_summary: vector<u8>,
+        checkpoint_contents: vector<u8>,
+        transaction: vector<u8>,
+        ): vector<MessageApproval>{
+        
+        let (cap_id_bytes, messages_serialised_bytes) = sui_state_proof_verify_transaction(committee.committee, checkpoint_summary, checkpoint_contents, transaction, config.event_type_layout, config.package_id );
+
+        let messages = bcs::peel_vec_vec_u8(&mut bcs::new(messages_serialised_bytes));
+        let cap_id_address = bcs::peel_address(&mut bcs::new(cap_id_bytes));
```
ycscaly, line 140: I don't really understand this. What does `peel_address` do?
There's an implicit assumption on the structure of the cap id object that isn't properly documented here. Is the address the first value in the structure?

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,145 @@
...
+        
+        assert!(object::id_from_address(cap_id_address)  == cap_wrapper.cap_id_sui, 0);
```
ycscaly, line 142: Why not call `object::id_from_bytes` like you extracted the ID previously? what's different here?

crates/sui-light-client/src/light-client/light-client.rs
ycscaly, line 1: This file is copied as-is from sui?

crates/sui-light-client/src/main.rs
ycscaly, line 1: This file is copied as-is from sui?

crates/sui-light-client/src/service/apisigner.rs
ycscaly, line 1: This file is copied as-is from sui?

crates/sui-protocol-config/src/lib.rs
```diff
@@ -1222,7 +1222,7 @@ impl ProtocolConfig {
             max_arguments: Some(512),
             max_type_arguments: Some(16),
             max_type_argument_depth: Some(16),
-            max_pure_argument_size: Some(16 * 1024),
+            max_pure_argument_size: Some(1024 * 1024), // TODO change back to 16
```
ycscaly, line 1225:
```suggestion
            max_pure_argument_size: Some(16 * 1024),
```
Should we?

sdk/typescript/src/signature-mpc/sui_stateproof.ts
ycscaly, line 1: This should probably be in its own crate

ycscaly: @omersadika

sdk/typescript/src/signature-mpc/sui_stateproof.ts
```diff
@@ -0,0 +1,217 @@
...
+import { Keypair } from '../cryptography';
+
+type TxDataResponse = {
+	ckp_epoch_id: number;
+	checkpoint_summary_bytes: Uint8Array;
+	checkpoint_contents_bytes: Uint8Array;
+	transaction_bytes: Uint8Array;
+};
+
+export async function submitDwalletCreationProof(
```
ycscaly, line 17:
```suggestion
export async function submitDWalletCreationProof(
```

state-proofs/sui_state_proof/sources/sui_state_proof.move
```diff
@@ -0,0 +1,37 @@
...
+    use sui::object::{Self, ID, UID};
+    use sui::event;
+    use sui::tx_context::TxContext;
+
+
+    struct DWalletCap has key, store {
+        id: UID,
+    }
+
+    struct DWalletNetworkRequest has copy, drop {
```
ycscaly, line 12: This name isn't clear. Perhaps `SignatureRequest`?

state-proofs/sui_state_proof/sources/sui_state_proof.move
```diff
@@ -0,0 +1,37 @@
...
+        cap_id: ID,
+        message: vector<u8>,
+    }
+
+    public fun create_cap(ctx: &mut TxContext): DWalletCap {
+        let cap = DWalletCap {
+            id: object::new(ctx),
+        };
+
+        event::emit(DWalletNetworkRequest {
```
ycscaly, line 22: It is a very ackward design choice you made here, choosing to use the same event for two functionally independent events: creation and message approval. These should be seperated, as well as the native code that checks for them. It was also confusing in the dwallet side's move code.

state-proofs/sui_state_proof/sources/sui_state_proof.move
```diff
@@ -0,0 +1,37 @@
+module sui_state_proof::sui_state_proof {
+    use std::vector;
+    use sui::object::{Self, ID, UID};
+    use sui::event;
+    use sui::tx_context::TxContext;
+
+
+    struct DWalletCap has key, store {
```
ycscaly, line 8: The fact there is no backward link here, and that there is just a one-way link from dWallet network here and not the opposite, means that two different dWallets could be controlled by the same capability.

This might not be a problem, but is a conscious design decision that needs to be made intentionally, and not as an unintentional byproduct.

I can see a reason why we'd want two different dWallets to be controlled by the same sui module, but I'm not sure that this is the right way to do it: instead, a single object could wrap/hold as child objects two different capabilities. IMO this is clearer, and it achieves the same objective.

@omersadika thoughts?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+ **************************************************************************************************/
+pub fn sui_state_proof_verify_committee(
+    context: &mut NativeContext,
+    ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(ty_args.is_empty());
+    debug_assert!(args.len() == 2);
+
+    // println!("giga debug 1");
```
ycscaly, line 52:
```suggestion
```
Please remove all of these

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    // Load the cost parameters from the protocol config
+    let sui_state_proof_cost_params = &context
+        .extensions()
+        .get::<NativesCostTable>()
+        .twopc_mpc_dkg_cost_params // TODO
```
ycscaly, line 58: Please fix this

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+        .clone();
+
+    // Charge the base cost for this oper
```
ycscaly, line 61:
```suggestion
    // Charge the base cost for this operation
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    let checkpoint_summary_bytes = pop_arg!(args, Vec<u8>);
+    let prev_committee_bytes = pop_arg!(args, Vec<u8>);
+
+    let Ok(prev_committee) = bcs::from_bytes::<Committee>(&prev_committee_bytes) else {
+        return Ok(NativeResult::err(
+            cost,
+            INVALID_COMMITTEE
+        ));
+    };
```
ycscaly, line 78: Nice, I didn't know this coding pattern in Rust. @zeevmoney

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    // println!("giga debug 2");
```
ycscaly, line 80:
```suggestion
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    
+    let Ok(checkpoint_summary) = bcs::from_bytes::<CertifiedCheckpointSummary>(&checkpoint_summary_bytes) else {
+        return Ok(NativeResult::err(
+            cost,
+            INVALID_CHECKPOINT_SUMMARY
+        ));
+    };
+
+    // println!("giga debug 3");
```
ycscaly, line 90:
```suggestion
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    match checkpoint_summary.clone().verify(&prev_committee) {
+        Ok((_)) => (),
+        Err(e) => return Ok(NativeResult::err(cost, INVALID_TX)),
+    }
+
+    // println!("giga debug 4");
```
ycscaly, line 97:
```suggestion
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+    }) = &checkpoint_summary.end_of_epoch_data
+    {
+        let next_committee = next_epoch_committee.iter().cloned().collect();
+        next_committee_epoch =
+            Committee::new(checkpoint_summary.epoch().checked_add(1).unwrap(), next_committee);
+    } else {
+        return Ok(NativeResult::err(cost, INVALID_TX))
+    }
+
+    // println!("giga debug 5");
```
ycscaly, line 113:
```suggestion
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+
+
+
+
+
+/***************************************************************************************************
+ * native fun dkg_verify_decommitment_and_proof_of_centralized_party_public_key_share
+ * Implementation of the Move native function `dwallet_2pc_mpc_ecdsa_k1::dkg_verify_decommitment_and_proof_of_centralized_party_public_key_share(commitment_to_centralized_party_secret_key_share: vector<u8>, secret_key_share_encryption_and_proof: vector<u8>, centralized_party_public_key_share_decommitment_and_proofs: vector<u8>): (vector<u8>, vector<u8>, vector<u8>);`
+ *   gas cost: dkg_verify_decommitment_and_proof_of_centralized_party_public_key_share_cost_base   | base cost for function call and fixed opers
+ **************************************************************************************************/
```
ycscaly, line 126:
```suggestion
```
?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(ty_args.is_empty());
+    debug_assert!(args.len() == 6);
+
+    // Load the cost parameters from the protocol config
+    let sui_state_proof_cost_params = &context
+        .extensions()
+        .get::<NativesCostTable>()
+        .twopc_mpc_dkg_cost_params // TODO
```
ycscaly, line 139: fix

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+        .clone();
+
+
+    // Load the cost parameters from the protocol config
+    // let object_runtime = context
+    //     .extensions()
+    //     .get::<ObjectRuntime>();
+
```
ycscaly, line 147:
```suggestion
```
?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+    let checkpoint_contents_bytes = pop_arg!(args, Vec<u8>);
+    let summary_bytes = pop_arg!(args, Vec<u8>);
+    let committee_bytes = pop_arg!(args, Vec<u8>);
+
+    let Ok(committee) = bcs::from_bytes::<Committee>(&committee_bytes) else {
+        return Ok(NativeResult::err(
+            cost,
+            INVALID_COMMITTEE
+        ));
+    };
```
ycscaly, line 169: I'm a little confused. The `Committee` struct, where is that defined, who is creating that?
Is it a Rust or Move struct?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+
+
+    tx_events.iter().for_each(|event| {
+        println!("{:?}", event);
+    });
+    let mut messages = Vec::new();
+    let mut cap_id_final = SuiAddress::ZERO;
+
+    for event in tx_events {
+        if !(event.clone().package_id == package_id_target && event.clone().type_.module.into_string() == "sui_state_proof" && event.clone().type_.name.into_string() == "DWalletNetworkRequest") {
```
ycscaly, line 236: What does the module name check actually do? I think the package id is a unique identifier no?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+            continue;
+        }
+
+        let json_val = SuiJsonValue::from_bcs_bytes(Some(&type_layout), &event.contents).unwrap().to_json_value();
+
+
+        let approve_message = match json_val.clone() {
+            JsonValue::Object(map) => {
+                // println!("mapppp{:?}", map);
```
ycscaly, line 245:
```suggestion
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+                }
+                else {
+                    return Ok(NativeResult::err(cost, INVALID_TX));
+                }
+            },
+            _ => return Ok(NativeResult::err(cost, INVALID_TX))
+        };
+
+
+        // TOOD no need to read it several times
```
ycscaly, line 257:
```suggestion
```
Fix please

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+                }
+    
+            },
+            _ => return Ok(NativeResult::err(cost, INVALID_TX))
+        };
+
+        let cap_id = match SuiAddress::from_str(&cap_id) {
+            Ok(address) => address,
+            Err(_) => return Ok(NativeResult::err(cost, INVALID_TX)),
+        };
```
ycscaly, line 274: Why is the ID a SuiAddress?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+    let tx_events = &transaction.events.as_ref().unwrap().data;
+
+
+    tx_events.iter().for_each(|event| {
+        println!("{:?}", event);
+    });
+    let mut messages = Vec::new();
+    let mut cap_id_final = SuiAddress::ZERO;
+
+    for event in tx_events {
```
ycscaly, line 235: btw its not very common in Rust to write for loops, usually you do functional stuff

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,287 @@
...
+    
+            },
+            _ => return Ok(NativeResult::err(cost, INVALID_TX))
+        };
+
+        let cap_id = match SuiAddress::from_str(&cap_id) {
+            Ok(address) => address,
+            Err(_) => return Ok(NativeResult::err(cost, INVALID_TX)),
+        };
+        cap_id_final = cap_id;
```
ycscaly, line 275: Yeah so I think there's a vulnerability here too. I can create a tx that approves two messages, m1 and m2. m1 I approve for a dWallet1, and m2 I approve for a dWallet2.
Now dWallet2 never did any policy check for m1, and in-fact it can be unallowed by its policy: e.g., you were given access to spend 1 eth in dwallet2. You create a new dwallet1 without any policy (doesn't even have to hold any funds) and approve a 100eth withdrawl from dwallet1 and 0.1 eth from dwallet2 - and because you say the cap id is the last one, and you don't check they are all equal, than the result would be you'll say dwallet2 approve a 100.1 eth withdrawl....
If true this completely breaks the entire security model

ycscaly: @alexanderludwig

ycscaly:
Thanks @alexanderludwig ! Overall good job. 
This code is far from production grade atm (but that was expected), and I made many remarks that could help it get there. 
Most crucially, I think I found a critical vuln. here, so to your attention.

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
ycscaly: This fixes it

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -82,15 +90,18 @@ module dwallet_system::sui_state_proof {
...
 
         let committee_new = EpochCommittee {
                                     id: object::new(ctx),
-                                    committee: committee_verified_bytes,
+                                    committee: new_committee_verified_bytes,
                                     };
 
         registry.highest_epoch = registry.highest_epoch + 1;
 
+        assert!(committee_epoch == registry.highest_epoch, EWrongEpochSubmitted);
```
ycscaly, line 102: This should be a good fix, however, it is best practice to perform security checks _before_ updating state etc. 

Although it should be ok as a failed assert will revert state changes, I suggest moving this check right after verifying the state proof, and before creating a new committee object and mutating the registry.

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -278,12 +255,17 @@ pub fn sui_state_proof_verify_committee(
             Ok(address) => address,
             Err(_) => return Ok(NativeResult::err(cost, INVALID_TX)),
         };
+
+
+        // Ensure that the cap id is the same for all messages.
+        if cap_id_final != SuiAddress::ZERO && cap_id_final != cap_id {
+            return Ok(NativeResult::err(cost, INVALID_TX));
+        }
```
ycscaly, line 263: This fixes the vulnerability, however, I wonder if this is the best design; an alternative option would be for the native function to get a list of messages and a cap ID, and to check that all messages are approved by that cap ID. This would allow, for example, a single transaction in Sui to approve messages for different dWallets. @alexanderludwig

crates/sui-protocol-config/src/lib.rs
alexanderludwig: we need this bc otherwise we can't submit the state proofs

ycscaly: Yeah I’m just saying that you had a TODO there, so handle it and delete it

integrations/sui/sources/dwallet_cap.move
```diff
@@ -7,29 +7,39 @@ module dwallet_network::dwallet_cap {
...
         cap
     }
 
 
     public fun approve_message(cap: &DWalletCap, message: vector<u8>){
-        event::emit(DWalletNetworkRequest{
+        event::emit(DWalletNetworkApproveRequest{
             cap_id: object::id(cap),
             message: message,
         });
```
ycscaly: great!

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,182 @@
...
+
+
+    public fun init_module(
+        init_committee: vector<u8>, 
+        package_id: vector<u8>, 
+        init_cap_event_type_layout: vector<u8>, 
+        approve_event_type_layout: vector<u8>, 
+        epoch_id_committee: u64, 
+        ctx: &mut TxContext) {
+        let registry = StateProofRegistry {
```
ycscaly, line 64:
```suggestion
        // TODO: handle initialization authorization/once
        let registry = StateProofRegistry {
```
We will handle this in a future PR, but for now let's at least put a TODO here.

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,182 @@
...
+            registry_id: object::id(&registry),
+            package_id: package_id,
+            init_cap_event_type_layout: init_cap_event_type_layout,
+            approve_event_type_layout: approve_event_type_layout,
+        };
+
+        let first_committee = EpochCommittee {
+            id: object::new(ctx),
+            committee: init_committee,
+        };
```
ycscaly, line 80: I'm wondering if there could be a sanity check here that the committee is of the right structure. In any case, this function is an authorized one and users should check it was initialized correctly, so this isn't a big deal security wise (e.g. if it was attempted to be initialized maliciously one could just put a wrong committee) but if we could at least make sure that this deserializes into the correct structure it could be good

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,182 @@
...
+
+    public fun submit_new_state_committee(
+        registry: &mut StateProofRegistry,
+        prev_committee: &EpochCommittee,
+        new_checkpoint_summary: vector<u8>,
+        ctx: &mut TxContext,
+    ) {
+        let (new_committee_verified_bytes, new_committee_epoch) = sui_state_proof_verify_committee(prev_committee.committee, new_checkpoint_summary);
+
+        assert!(new_committee_epoch - 1 == registry.highest_epoch, EWrongEpochSubmitted);
```
ycscaly, line 104:
```suggestion
        assert!(new_committee_epoch == registry.highest_epoch + 1, EWrongEpochSubmitted);
```
Let's make it consistent with line 112

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -0,0 +1,182 @@
...
+        ctx: &mut TxContext
+    ){        
+        
+        let (sui_cap_id_bytes, dwallet_cap_id_bytes) = sui_state_proof_verify_link_cap(committee.committee, checkpoint_summary, checkpoint_contents, transaction, config.init_cap_event_type_layout, config.package_id );
+        
+
+        // check if the cap id used on SUI is the same as the id of dwallet_cap
+        let sui_cap_id_address = bcs::peel_address(&mut bcs::new(sui_cap_id_bytes));
+        let dwallet_cap_id_address = bcs::peel_address(&mut bcs::new(dwallet_cap_id_bytes));
+        assert!(object::id_from_address(dwallet_cap_id_address) == object::id(&dwallet_cap), EWrongDWalletCapId);
```
ycscaly, line 140: Good!

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,387 @@
...
+    let next_committee_epoch;
+    // Extract the new committee information
+    if let Some(EndOfEpochData {
+        next_epoch_committee,
+        ..
+    }) = &checkpoint_summary.end_of_epoch_data
+    {
+        let next_committee = next_epoch_committee.iter().cloned().collect();
+        next_committee_epoch =
+            Committee::new(checkpoint_summary.epoch().checked_add(1).unwrap(), next_committee);
```
ycscaly, line 103:
```suggestion
        // epochs won't overflow since epoch is a 64-bit number and an epoch happens every few hours.
        next_committee_epoch =
             Committee::new(checkpoint_summary.epoch() + 1, next_committee);
```
There's no point in doing a `checked_add()` if you don't handle the failure (you just `unwrap()`).

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,387 @@
...
+    // Verify the checkpoint summary using the committee
+    let res = summary.verify_with_contents(&committee, Some(&checkpoint_contents));
+    if let Err(_) = res {
+        return Ok(NativeResult::err(cost, INVALID_TX));
+    }
+
+
+    // Ensure the tx is part of the checkpoint
+    let is_valid_checkpoint_tx = checkpoint_contents.iter().any(|&digest| digest == transaction.effects.execution_digests());
+    if !is_valid_checkpoint_tx {
```
ycscaly, line 207:
```suggestion
    if !checkpoint_contents.iter().any(|&digest| digest == transaction.effects.execution_digests()) {
```
@zeevmoney which why would you prefer?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
ycscaly, line 1: Did this file go through formatting? 
@zeevmoney do we have clippy & rustfmt configured for this repo?

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,387 @@
...
+        if event.package_id == package_id_target && event.type_.module.clone().into_string() == "dwallet_cap" && event.type_.name.clone().into_string() == "DWalletNetworkApproveRequest" {
+            let json_val = SuiJsonValue::from_bcs_bytes(Some(&type_layout), &event.contents).unwrap().to_json_value();
+    
+            let cap_id_str = json_val.get("cap_id").and_then(JsonValue::as_str);
+            let cap_id = cap_id_str.and_then(|s| SuiAddress::from_str(s).ok());
+    
+            let approve_message = json_val.get("message").and_then(JsonValue::as_array);
+            let approve_msg_vec: Option<Vec<u8>> = approve_message.map(|msg_array| {
+                msg_array.iter().map(|msg| {
+                    msg.as_u64().unwrap() as u8
```
ycscaly, line 361: can you explain the logic of this line? @alexanderludwig

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -0,0 +1,387 @@
...
+        } else {
+            None
+        }
+    }).collect();
+    
+    let (cap_ids, messages): (Vec<_>, Vec<_>) = results.into_iter().unzip();
+    
+    if cap_ids.len() != messages.len() {
+        return Ok(NativeResult::err(cost, INVALID_TX));
+    }
```
ycscaly, line 379: This check is redundant since you already construct it in a way that assures this condition will always be fulfilled (as pairs -> unwrap). I don't see a problem with leaving this sanity check anyways though.

integrations/sui/sources/dwallet_cap.move
```diff
@@ -0,0 +1,47 @@
...
+
+    struct DWalletCap has key, store {
+        id: UID,
+        dwallet_network_cap_id: ID,
+    }
+
+
+    struct DWalletNetworkInitCapRequest has copy, drop {
+        cap_id: ID,
+        dwallet_network_cap_id: ID,
```
ycscaly, line 16: Note that here we're using the same type for both IDs, wheras one of them comes from a different network. This could cause confusion and problems in the future. Perhaps a better idea would be to import the dwallet Network crate and use its ID. For now we can leave it as-is, but at least document this so the assumptions and coupling will be explicitly mentioned

ycscaly (approved):
I reviewed the "blockchain-side" of things: the SUI `dwallet_cap` module, and the dWallet Network Move & Rust code that handles the light-client.
Apart from cosmetics, all looks good, and particularly I find no security issues and affirm that the ones I previously mentioned were fixed.

There could still be work in this PR: I haven't thoroughly reviewed the client side and @zeevmoney should do so. Besides that, Zeev should make sure the directory structures all makes sense, and that we have formatting & clippy run as from what I've seen it wasn't. Also, @alexanderludwig mentioned there is no formatter for Move, so Zeev please decide how to handle that.

Thanks for a great contribution @alexanderludwig !

zeevmoney (approved):

PR #26, opened 2024-05-08
mpc and sui light client

zeevmoney:


ycscaly (approved):
As a temporary measure, this should be pushed to main, although this is not production grade code yet

PR #38, opened 2024-06-16
dwallet-network: add `dwallet`, `dwallet_2pc_mpc_ecdsa_k1` move modules 

yuvalelmo:


crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,238 @@
...
+    use dwallet::tx_context;
+    use dwallet::tx_context::{TxContext};
+
+    friend dwallet_system::dwallet_2pc_mpc_ecdsa_k1;
+
+    // <<<<<<<<<<<<<<<<<<<<<<<< Error codes <<<<<<<<<<<<<<<<<<<<<<<<
+    const ENotSystemAddress: u64 = 0;
+    const EMesssageApprovalDWalletMismatch: u64 = 1;
+
+    // <<<<<<<<<<<<<<<<<<<<<<<< Error codes <<<<<<<<<<<<<<<<<<<<<<<<
```
ycscaly, line 18: All these comments are Sui standards? @omersadika

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,238 @@
...
+
+    public fun dwallet_cap_id(message_approval: &MessageApproval): ID {
+        message_approval.dwallet_cap_id
+    }
+
+    public fun message(message_approval: &MessageApproval): vector<u8> {
+        message_approval.message
+    }
+
+    public fun remove(message_approval: MessageApproval): (ID, vector<u8>) {
```
ycscaly, line 129: since this is a module that serves multiple objects, I think it's better to have fully-qualified names like `remove_message_approval`. But let's follow standards if there are.

ycscaly: Also note that we have `message` and then `sign_messages_messages`. This is confusing. Let's have fully quallified names for all

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,238 @@
...
+        id: UID,
+        message_approvals: vector<MessageApproval>,
+    }
+
+    struct MessageApproval has store {
+        dwallet_cap_id: ID,
+        message: vector<u8>,
+    }
+
+    struct SignMessages<S: store, E: store + copy + drop> has key, store {
```
ycscaly, line 52: Why is this object named liked this? its a very confusing name. @omersadika 
We should rename. Perhaps to `MessagesForSigning` or `ApprovedMessagesToSign` or something like that.

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,238 @@
...
+    public fun message(message_approval: &MessageApproval): vector<u8> {
+        message_approval.message
+    }
+
+    public fun remove(message_approval: MessageApproval): (ID, vector<u8>) {
+        let MessageApproval {
+            dwallet_cap_id,
+            message
+        } = message_approval;
+        (dwallet_cap_id, message)
```
ycscaly, line 134: Isn't there a `object::delete(id)` missing here?

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,238 @@
...
+        let messages_len = vector::length(&messages);
+        let approval_len = vector::length(&message_approvals);
+        assert!(messages_len == approval_len, EMesssageApprovalDWalletMismatch);
+
+        while (i < messages_len) {
+            let message_approval = vector::pop_back(&mut message_approvals);
+            let (message_approval_dwallet_cap_id, approved_message) = remove(message_approval);
+            assert!(dwallet_cap_id == message_approval_dwallet_cap_id, EMesssageApprovalDWalletMismatch);
+            let message = vector::borrow(&messages, i);
+            assert!(message == &approved_message, EMesssageApprovalDWalletMismatch);
```
ycscaly, line 198: This is how string comparison (byte-by-byte equality) is done in move?

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,238 @@
...
+            i = i +1;
+        };
```
ycscaly, line 200: Wouldn't it be more clean to have `create_sign_messages` get `message_approvals` so that we never even create a `SignMessages` object for un-approved messages, or am I missing something?

sui-execution/latest/sui-move-natives/src/crypto/twopc_mpc.rs
ycscaly, line 1: Why do native unimplemented code a part of this PR? I think we only need the move code? @zeevmoney

ycscaly (changes requested):
Left some design questions for discussion @omersadika

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
omersadika: Yes

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
omersadika: It has no id

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
omersadika: Seems like that - https://move-book.com/reference/equality.html
Not very clear, so it good if you will also double check that

sui-execution/latest/sui-move-natives/src/crypto/twopc_mpc.rs
omersadika: We have to have the native functions rust code so it think it's better to have it

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
ycscaly: @yuvalelmo please add a test that checks this:
1. that an approved message that also has the sign message works (i.e. the equality check passes)
2. that an approved message that does not match the sign message does not work (i.e. the equality check fails)

ycscaly (changes requested):
Please add tests, and documentation across the board

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
zeevmoney: 1. Vector len mismatch.
2. Vectors hold messages not in the same order (i.e. message approval != sign message).
3. Approved message == signed message.
3. Cap ID is the same.
4. Cap ID mismatch.

See 
https://examples.sui.io/testing/index.html
https://docs.sui.io/guides/developer/first-app/build-test

yuvalelmo: @ycscaly @zeevmoney 
pushed tests just now

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,277 @@
...
+        let message_approvals = vector::empty<MessageApproval>();
+        while (vector::length(&messages) > 0) {
+            let message = vector::pop_back(&mut messages);
+            vector::push_back(&mut message_approvals, MessageApproval {
+                dwallet_cap_id,
+                message,
+            });
+        };
+        message_approvals
+    }
```
ycscaly, line 134: Remove this buggy function

ycscaly (changes requested):
Test properly without adding unnecessary mock functions ti the codebase and document all public code

yuvalelmo:
@ycscaly I've fixed the tests. 
@zeevmoney will add the documentation for these modules.

crates/sui-framework/packages/dwallet-system/tests/dwallet_tests.move
```diff
@@ -0,0 +1,144 @@
...
+            dwallet_id,
+            object::id(&dwallet_cap),
+            sign_messages,
+            sign_data,
+            sign_data_event,
+            ctx
+        );
+
+        dwallet::sign(partial_user_signed_messages, message_approvals, ctx);
+        test_utils::destroy(dwallet_cap);
```
ycscaly, line 75: Please test the opposite: that if we have message approval on more than just one message, but we try to sign just one of them, it works (or fails, depending on what we decided to do)

ycscaly (changes requested):
Tests looking good. I added one test case, and will try to think if any more are needed

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,265 @@
...
+        dwallet_cap_id: ID,
+        messages: vector<vector<u8>>,
+        sign_data: S,
+        sign_data_event: E,
+    }
+
+    struct SharedPartialUserSignedMessages<S: store, E: store + copy + drop> has key {
+        id: UID,
+        partial_user_signed_messages: PartialUserSignedMessages<S, E>,
+    }
```
ycscaly, line 67: why is this a different type? can't we just share `PartialUserSignedMessages` directly?

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,265 @@
...
+
+    struct PartialUserSignedMessages<S: store, E: store + copy + drop> has key, store {
+        id: UID,
+        dwallet_id: ID,
+        dwallet_cap_id: ID,
+        messages: vector<vector<u8>>,
+        sign_data: S,
+        sign_data_event: E,
+    }
```
ycscaly, line 62: please explain why generics are needed here in a comment (i.e. supporting different algorithms/curves)

ycscaly: In the future actually these generics might become unneeded, for now lets keep them

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,265 @@
...
+    // <<<<<<<<<<<<<<<<<<<<<<<< Events <<<<<<<<<<<<<<<<<<<<<<<<
+
+    struct DWalletCap has key, store {
+        id: UID,
+    }
+
+    struct MessageApprovalsHolder has key {
+        id: UID,
+        message_approvals: vector<MessageApproval>,
+    }
```
ycscaly, line 48: I'm guessing its meaningful for the struct to hold a vector of message approvals, but I don't see why it is different from sending `vector<MessageApproval>` directly.

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,265 @@
...
+        ctx: &mut TxContext
+    ): PartialUserSignedMessages<S, E> {
+        PartialUserSignedMessages {
+            id: object::new(ctx),
+            dwallet_id,
+            dwallet_cap_id,
+            messages,
+            sign_data,
+            sign_data_event,
+        }
```
ycscaly, line 179: @omersadika can we take the dwallet_cap by reference here instead of the ID?
I know that the caller is a friend and will perform security validations there, but it is safer to have it anyway if possible.

ycscaly: Or is the caller of this function not necessairly an owner of the cap?

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,265 @@
...
+            sign_data_event,
+        } = partial_user_signed_messages;
+
+        object::delete(id);
+        let i = 0;
+        let messages_len = vector::length(&messages);
+        let approval_len = vector::length(&message_approvals);
+        assert!(messages_len == approval_len, EMesssageApprovalDWalletMismatch);
+
+        while (i < messages_len) {
```
ycscaly, line 220: no `for` loops in move?

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -0,0 +1,265 @@
...
+            dwallet_id,
+            dwallet_cap_id,
+            messages,
+            sign_data,
+            sign_data_event,
+        } = partial_user_signed_messages;
+
+        object::delete(id);
+        let i = 0;
+        let messages_len = vector::length(&messages);
```
ycscaly, line 216: When doing integer arithmetics, it is much clearer and safe to explicitly note the types. Is it possible to do so in move?
E.g. `let messages_len: u64 = vector::length(&messages);`
It is important to see types are of the same or compatible size and sign, e.g. for making sure no overflow occurrs

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -0,0 +1,302 @@
...
+            commitment_to_centralized_party_secret_key_share,
+            sender,
+        });
+        transfer::freeze_object(session);
+        cap
+    }
+
+    #[allow(unused_function)]
+    fun create_dkg_output(session: &DKGSession, commitment_to_centralized_party_secret_key_share: vector<u8>, secret_key_share_encryption_and_proof: vector<u8>, ctx: &mut TxContext) {
+        assert!(tx_context::sender(ctx) == @0x0, ENotSystemAddress);
```
ycscaly, line 147: I'm not sure what this assert is for, but shouldn't it be for `create_dkg_session()` and many others in case it is actually needed here?

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -0,0 +1,302 @@
...
+        let DKGSessionOutput {
+            id,
+            session_id,
+            dwallet_cap_id,
+            commitment_to_centralized_party_secret_key_share,
+            secret_key_share_encryption_and_proof,
+        } = output;
+        object::delete(id);
+
+        let (output, public_key) = dkg_verify_decommitment_and_proof_of_centralized_party_public_key_share(commitment_to_centralized_party_secret_key_share, secret_key_share_encryption_and_proof, centralized_party_public_key_share_decommitment_and_proof);
```
ycscaly, line 168: what happens if verification of zk proofs fails? does the native function fail with an exception that aborts the execution of this function?

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -0,0 +1,302 @@
...
+
+        let result = DWallet {
```
ycscaly, line 170:
```suggestion
        let dwallet = DWallet {
```

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -0,0 +1,302 @@
...
+            id: object::new(ctx),
+            session_id,
+            dwallet_cap_id,
+            output,
+            public_key,
+        };
+        transfer::freeze_object(result);
```
ycscaly, line 177:
```suggestion
        transfer::freeze_object(dwallet);
```

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
ycscaly, line 1: why does this file have to be in this pr?

crates/sui-framework/packages/dwallet-system/tests/dwallet_tests.move
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    #[test]
+    #[expected_failure(abort_code = EMesssageApprovalDWalletMismatch)]
+    public fun test_sign_with_different_messages_order_should_fail() {
```
ycscaly, line 113: please add test where its the same length, and just one message is different in the approvals (do one where its the first, one where last, one where middle)

ycscaly (changes requested):
see comments, good progress

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
omersadika: Cant

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
omersadika: no

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
omersadika: This function should be called only by the system, and this is what we check for

zeevmoney (approved):

PR #49, opened 2024-06-17
eth-light-clinet: Add utils.rs and sui-types/eth_dwallet_cap.rs *ready for review*

shaymalichi:


Cargo.toml
```diff
@@ -320,6 +320,7 @@ ed25519-consensus = { version = "2.0.1", features = ["serde"] }
 either = "1.8.0"
 petgraph = "0.5.1"
 enum_dispatch = "^0.3"
+ethers = "2.0.14"
```
zeevmoney, line 323: Do we need this as global dep or can it be placed inside the `eth-light-client`?

shaymalichi: i think its necessary if we want to work with the convention *.workspace = true

ycscaly (changes requested):
A bit puzzled about the purpose of this PR, and why this is the first PR for this feature

PR #51, opened 2024-06-17
eth-light-client: Add proof.rs And config.rs files *ready for review*

shaymalichi:


crates/eth-light-client/src/config.rs
ycscaly, line 1: Please, when you add public structs add documentation to explain their purpose.
I have no idea, reading this file, what these structs are for.

ycscaly, line 1: Please, when you add public structs add documentation to explain their purpose.
I have no idea, reading this file, what these structs are for.

crates/eth-light-client/src/proof.rs
ycscaly, line 1: These structs looks copy-pasted. Please explain why that is necessairy.

ycscaly (changes requested):
I don't understand the purpose of this PR. Why did you choose to submit these files together as a coherent unit of code?

Also, it seems that the logic the verifies the proof is missing from here. I'd expect that if we are doing a PR to add the proof, it'll be included

PR #54, opened 2024-06-17
sui-framework/dwallet-network: Add eth_dwallet.move & ethereum_state.move + native functions placeholders

shaymalichi:


crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -44,7 +44,7 @@ module dwallet_system::dwallet {
         message_approvals: vector<MessageApproval>,
     }
 
-    struct MessageApproval has store {
+    struct MessageApproval has store, drop {
```
yuvalelmo, line 47: @omersadika is it ok to add this drop ability here?  
if i call `eth_dwallet::approve_message` without it it would fail even though i am returning it

PR #64, opened 2024-06-23
EthereumLightClient: Add command `dwallet-connect-eth` to dwallet client 

shaymalichi:
This command is creating an Ethereum dWallet.

### DIFF
--- a/crates/sui/src/ethereum_client.rs
+++ b/crates/sui/src/ethereum_client.rs
@@ -1,5 +1,25 @@
+use anyhow::anyhow;
+use anyhow::Error;
 use clap::Subcommand;
+use move_core_types::ident_str;
+use move_core_types::identifier::IdentStr;
+use serde_json::{Number, Value};
+
+use shared_crypto::intent::Intent;
+use sui_json::SuiJsonValue;
+use sui_json_rpc_types::SuiExecutionStatus;
+use sui_keys::keystore::AccountKeystore;
+use sui_sdk::wallet_context::WalletContext;
 use sui_types::base_types::ObjectID;
+use sui_types::SUI_SYSTEM_PACKAGE_ID;
+use sui_types::transaction::{SenderSignedData, Transaction, TransactionDataAPI};
+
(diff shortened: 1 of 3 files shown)


crates/sui/src/ethereum_client.rs
```diff
@@ -1,5 +1,26 @@
+use anyhow::anyhow;
+use anyhow::Error;
 use clap::Subcommand;
+use move_core_types::ident_str;
+use move_core_types::identifier::IdentStr;
+use serde_json::{Number, Value};
+
+use shared_crypto::intent::Intent;
+use sui_json::SuiJsonValue;
+use sui_json_rpc_types::{SuiExecutionStatus, SuiTransactionBlockEffectsAPI}; // todo!(ide keeps deleting SuiTransactionBlockEffectsAPI)
```
yuvalelmo, line 10: the formatter removes this this package because it is a dependency of `serialize_or_execute` macro function - so while formatting the macro is not expanded and the formatter thinks it is redundant, when it is not.

yuvalelmo (approved):

crates/sui/src/ethereum_client.rs
```diff
@@ -78,5 +99,50 @@ pub enum EthClientCommands {
...
+    serialize_unsigned_transaction: bool,
+    serialize_signed_transaction: bool,
+) -> Result<SuiClientCommandResult, Error> {
+    // Serialize to the Move TX format.
+    let smart_contract_address = bcs::to_bytes(&smart_contract_address).unwrap();
+    let mut smart_contract_address: Vec<Value> = smart_contract_address
+        .iter()
+        .map(|v| Value::Number(Number::from(*v)))
+        .collect();
+    smart_contract_address.remove(0);
```
ycscaly, line 121: why do you remove the first number?

ycscaly: document this

ycscaly (changes requested):

PR #65, opened 2024-06-23
EthereumLightClient: Add command `init-eth-state` to dwallet client

shaymalichi:


crates/sui/src/ethereum_client.rs
```diff
@@ -20,6 +20,11 @@ use crate::serialize_or_execute;
 
 pub const ETH_DWALLET_MODULE_NAME: &IdentStr = ident_str!("eth_dwallet");
 pub const CREATE_ETH_DWALLET_CAP_FUNC_NAME: &IdentStr = ident_str!("create_eth_dwallet_cap");
+pub const INIT_STATE_FUNC_NAME: &IdentStr = ident_str!("init_state");
+pub const ETHEREUM_STATE_MODULE_NAME: &IdentStr = ident_str!("ethereum_state");
+pub const LATEST_ETH_STATE_STRUCT_NAME: &IdentStr = ident_str!("LatestEthereumState");
```
yuvalelmo, line 25: After we add the move modules, I think these should move to where we implement the rust versions of the move objects we use(inside sui-types).

yuvalelmo (approved):

crates/sui/src/ethereum_client.rs
```diff
@@ -146,3 +151,60 @@ pub(crate) async fn create_eth_dwallet(
...
+
+//todo(yuval): in future, we should load also the sync committee from binary data file.
+pub(crate) async fn init_ethereum_state(
+    checkpoint: String,
+    context: &mut WalletContext,
+    gas: Option<ObjectID>,
+    gas_budget: u64,
+    serialize_unsigned_transaction: bool,
+    serialize_signed_transaction: bool,
+) -> Result<SuiClientCommandResult, Error> {
```
ycscaly, line 163: Please explain the purpose of this function

ycscaly (changes requested):

PR #75, opened 2024-07-01
SuiLightClient: Add light_client.rs And service.rs files + Dependencies 

shaymalichi:
This PR adjusts the dependencies to ensure that the light_client and service files in crates/sui-light-client compile. It also adds these modules to the dwallet-network project.

I followed Alex’s PR, and he changed some of the crates to be fetched from the git repository instead of using the .workspace pattern. 

example:

``
move-core-types = { git = "https://github.com/MystenLabs/sui", package = "move-core-types", rev="f58faa4" }
instead of 
move-core-types.workspace = true
``

I haven’t changed that because I don’t know the motive yet.


zeevmoney:

PR #76, opened 2024-07-01
SuiLightClient: Add sui_state_proof.rs and its supported move files  

shaymalichi:


zeevmoney:

PR #77, opened 2024-07-01
SuiLightClient: Modify crates/sui-protocol-config/src/lib.rs

shaymalichi:
Modify lib.rs file as Alex's PR.

Also, sui-rest-rpc/client.rs as the PR.


zeevmoney:

PR #87, opened 2024-07-03
dwallet-network/eth-light-client: added eth_dwallet & ethereum_state move modules

yuvalelmo:
…ceholders.


crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,50 @@
...
+    /// Creates an Ethereum dWallet capability object by wrapping an existing DWalletCap.
+    public entry fun create_eth_dwallet_cap(
+        dwallet_cap: DWalletCap,
+        ctx: &mut TxContext
+    ) {
+        let eth_dwallet_cap = EthereumDWalletCap {
+            id: object::new(ctx),
+            dwallet_cap,
+        };
+        transfer::share_object(eth_dwallet_cap);
```
ycscaly, line 31: Should this be a shared or immutable object? The reason for it to be shared is if you could sometime "unwrap" the capability, but that is not supported now. I think we do want to support it in the future though.
@omersadika

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,50 @@
...
+    }
+
```
ycscaly, line 33: Do we also want a function that creates a dWallet that is already tied to a policy? @omersadika

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
ycscaly, line 31: Should this be a shared or immutable object? The reason for it to be shared is if you could sometime "unwrap" the capability, but that is not supported now. I think we do want to support it in the future though.
@omersadika

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,50 @@
...
+    /// Verifies the provided proof using the `verify_message_proof` native function.
+    /// If the proof is valid, the message is approved.
+    public fun approve_message(
```
ycscaly, line 36: Shouldn't this be an `approve_messages` so it matches the signature? Then, you need to change the lightclient side to fetch multiple messages per proof.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,112 @@
...
+    /// Ethereum state object.
+    struct EthereumState has key {
+        id: UID,
+        /// serialised ConsensusStateManager
+        data: vector<u8>,
+        time_slot: u64,
+    }
+
+    /// Latest Ethereum state object.
+    /// Holds the ID of the latest Ethereum state object that is verified by the dWallet network.
```
ycscaly, line 20:
```suggestion
    /// Holds the ID of the latest Ethereum state object that is verified by the dWallet network.
    /// This object should have exactly one instance per `network`.
```

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,112 @@
...
+    struct LatestEthereumState has key, store {
+        id: UID,
+        eth_state_id: ID,
+        last_slot: u64,
+        eth_smart_contract_addr: vector<u8>,
+        eth_smart_contract_slot: u64,
```
ycscaly, line 26: You're missing a `network` field

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /// Initializes the first Ethereum state with the given checkpoint.
+    /// Creates an EthereumState object, shares a LatestEthereumState object pointing to it,
+    /// and freezes the EthereumState object.
```
ycscaly, line 31:
```suggestion
    /// and freezes the EthereumState object.
    /// NOTE: this function performs no verification on the `checkpoint`, 
    /// and it serves as an initial "trusted" state which users should verify externally (once) before using.
```

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,112 @@
...
+        });
+        transfer::freeze_object(state);
+    }
+
+    /// Updates the latest Ethereum state reference if the provided EthereumState object has a newer time slot.
+    public fun update_latest_eth_state(
+        self: &mut LatestEthereumState,
+        eth_state: &EthereumState,
+    ) {
+        if (eth_state.time_slot > self.last_slot) {
```
ycscaly, line 61: Here I'm missing a network check: so that updates for Ethereum `testnet` could not be set to the state of the `mainnet` object.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,112 @@
...
+            state_bytes
+        );
+
+        let new_state = EthereumState {
+            id: object::new(ctx),
+            data,
+            time_slot,
+        };
+
+        if (new_state.time_slot > latest_ethereum_state.last_slot) {
```
ycscaly, line 90: Also here check networks match

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,112 @@
...
+            eth_state_id: object::id(&state),
+            last_slot: state.time_slot,
+            eth_smart_contract_addr,
+            eth_smart_contract_slot,
+        });
+        transfer::freeze_object(state);
+    }
+
+    /// Updates the latest Ethereum state reference if the provided EthereumState object has a newer time slot.
+    public fun update_latest_eth_state(
```
ycscaly, line 57: Why do you need this function if it is already being updated in `verify_new_state`?

ycscaly: Also there's duplicate code

ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,50 @@
...
+        eth_dwallet_cap: &EthereumDWalletCap,
+        message: vector<u8>,
+        proof: vector<u8>,
+    ): vector<MessageApproval> {
+        let is_valid = verify_message_proof(proof);
+        assert!(is_valid, EInvalidStateProof);
+        dwallet::approve_messages(&eth_dwallet_cap.dwallet_cap, vector[message])
```
ycscaly, line 43: There is no check here that `message` and `proof` are connected! I can create a valid proof over some message `m1`, and submit this `proof`, with `m2` as `message`, and have you approve `m2` which was never approved in Ethereum??

yuvalelmo: Good catch. fixed.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
yuvalelmo: This function was necessary in the beginning but is no longer, I'll remove.

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
yuvalelmo: This is a good idea, but it will take some time to implement.
I've opened an issue for this one.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,106 @@
...
+        state_bytes: vector<u8>,
+        latest_ethereum_state: &mut LatestEthereumState,
+        ctx: &mut TxContext,
+    ) {
+        let (data, time_slot) = verify_eth_state(
+            updates_bytes,
+            finality_update_bytes,
+            optimistic_update_bytes,
+            state_bytes
+        );
```
ycscaly, line 76: You must always verify the new state against the old one: You are only passing user-supplied arguments to `verify_eth_state`, and not using `latest_ethereum_state` at all! This means I can just bypass the proof completely by sending a proof for some forked state, or state from another network, or anything like that.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
ycscaly: I do. Let's try to find how this is enforced in `helios`.

ycscaly: When you say it is "pulled" using the ID from `LatestEthereumState`, this implies you are trusting the user honestly followed protocol. But there is no reason for they to do so, nor for you to trust so. Trust nothing, verify everything.

ycscaly (changes requested):

zeevmoney:
closing this one, everything will be handled in #88

PR #88, opened 2024-07-04
dwallet-network:eth-light-client: added native functions implementations

yuvalelmo:
implemented the following functions:
create_initial_eth_state_data
verify_eth_state
verify_message_proof


sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -47,7 +47,48 @@ pub fn verify_message_proof(
...
+            StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT.into(),
+        ));
+    };
+
+    // Verify the account proof against the state root.
+    let is_valid_account_proof = verify_proof(
+        &proof.account_proof.proof.as_slice(),
+        &proof.account_proof.root,
+        &proof.account_proof.path,
+        &proof.account_proof.value,
```
ycscaly, line 78: Here and everywhere @yuvalelmo :
The entire idea of verifying proofs, is that you verify against some statement and then you trust *that same statement* for something. Even the notion of bundling up proofs & statements together in one struct is already "eye-catching" (there are situations in more complex protocols where this is applicable, but generally this is dangerous unless you have a good reason to do so.)

The correct way to design it is to call the verification function with the proof and *statements from the verifier* as opposed to the way it is currently done where the statements come from the prover.

This is a recurring theme in every place we would verify any kind of proof. A good example of where it was done correctly by @yaelAbergel @ItayLevyOfficial is in #123. There, they first [constructed the statement](https://github.com/dwallet-labs/dwallet-network/pull/123/files/e3e7e8963c1181ad84bb8e3b42288f6816e34468#r1705216946) from the `public_key_share` that [comes from the dkg output](https://github.com/dwallet-labs/dwallet-network/blob/e7dd19fbf55816cc567bd8e0f3ad726fc092ed3b/sui-execution/latest/sui-move-natives/src/crypto/twopc_mpc.rs#L96). Then, they [verified the proof](https://github.com/dwallet-labs/dwallet-network/blob/e7dd19fbf55816cc567bd8e0f3ad726fc092ed3b/crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs#L227) against that statement, which also includes the `encrypted_user_share` the prover sent. Constructed this way, the proof verification assures the verifier that a relationship exists between the two statements: the public key share and the encrypted user share, thus proving a correct encryption. This is done because the verifier supplied the public key share that it took from a previously known-good state - the DKG output.

Here, verifying against a prover-supplied statements (i.e. `path`, `value`, and `root`) will yield nothing. The correct way is for the verifier to construct these values so that when the proof verifies, it assures them of their validity. 

Actually, a proof usually proves a relationship: so like the encrypted user share proof proves that the encryption is of the secret key share that matches the public key share, here the proof proves a relationship between `path`,`value` and `root`: i.e. that the Merkle tree root of Ethereum state in `root` (that has to be previously verify against the lightclient) has `value` in `path`.

So you must:
1. remove `root`, `path` and `value` from the proof structure
2. take `root` from the `LatestEthereumState` move object (that was previously verified in lightclient)
3. construct `path` from the `message` and `dwallet id` (and `slot`)
4. use `true` for `value`

ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -39,8 +44,22 @@ module dwallet_system::eth_dwallet {
...
         data_slot: u64,
         proof: vector<u8>,
+        latest_ethereum_state: &LatestEthereumState,
+        current_wrapped_eth_state: &EthereumState,
     ): vector<MessageApproval> {
-        let is_valid = verify_message_proof(proof, message, dwallet_id, data_slot);
+        // Validate that the current Ethereum state from the user is the latest Ethereum state.
+        let current_ethereum_state_id = object::id(current_wrapped_eth_state);
+        let ethereum_state_id_from_latest = ethereum_state::get_ethereum_state_id_from_latest(latest_ethereum_state);
+        assert!(current_ethereum_state_id == ethereum_state_id_from_latest, 0x1);
```
ycscaly, line 53: Maybe this check is too restrictive. It could be that someone else has updated `LatestEthereumState` whilst you are trying to approve a message and it will fail even though it should have passed. This would require you to retry. Depending on the frequency of Ethereum light client updates, it could be that this edge case would not be neglible.

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -39,8 +44,22 @@ module dwallet_system::eth_dwallet {
         dwallet_id: vector<u8>,
```
ycscaly, line 44:
```suggestion
```
Take this from the `EthereumDWalletCap`

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -39,8 +44,22 @@ module dwallet_system::eth_dwallet {
...
         data_slot: u64,
```
ycscaly, line 45:
```suggestion
```
Take this from the `EthereumState`

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -39,8 +44,22 @@ module dwallet_system::eth_dwallet {
...
         proof: vector<u8>,
+        latest_ethereum_state: &LatestEthereumState,
```
ycscaly, line 47:
```suggestion
```
I'm not sure that we care it is latest

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -39,8 +44,22 @@ module dwallet_system::eth_dwallet {
...
     ): vector<MessageApproval> {
-        let is_valid = verify_message_proof(proof, message, dwallet_id, data_slot);
+        // Validate that the current Ethereum state from the user is the latest Ethereum state.
+        let current_ethereum_state_id = object::id(current_wrapped_eth_state);
+        let ethereum_state_id_from_latest = ethereum_state::get_ethereum_state_id_from_latest(latest_ethereum_state);
+        assert!(current_ethereum_state_id == ethereum_state_id_from_latest, 0x1);
+
+        // Validate that the Ethereum dWallet capability is for the correct network.
+        let state_network = ethereum_state::get_ethereum_state_network(latest_ethereum_state);
+        assert!(eth_dwallet_cap.network == state_network, 0x2);
```
ycscaly, line 57: What is `0x2`?

ycscaly: Replace with error code

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -45,6 +48,7 @@ module dwallet_system::ethereum_state {
             id: object::new(ctx),
             data: state_data,
             time_slot: 0u64,
+            state_root: vector::empty(),
```
ycscaly, line 51: No no, the purpose of the `init` function is to initialize to a trusted starting point from which verification should occur. This means that there should be a value here for the starting point `state_root` which points to some block in Ethereum that we have as a sort of "later genesis" we trust. Same on `time_slot`.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -79,6 +83,7 @@ module dwallet_system::ethereum_state {
             id: object::new(ctx),
             data,
             time_slot,
+            state_root
         };
 
         if (new_state.time_slot > latest_ethereum_state.last_slot) {
```
ycscaly, line 89: This should be a restrictive check: you don't allow creation of the `EthereumState` objects unless they are newer than the latest

ycscaly: Also, you must verify they are both in the same network.

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -47,7 +53,67 @@ pub fn verify_message_proof(
...
+        .map_err(|_| PartialVMError::new(StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT))?;
+
+    let message_map_index = get_message_storage_slot(message.clone(), dwallet_id.clone(), map_slot)
+        .map_err(|_| PartialVMError::new(StatusCode::UNKNOWN_STATUS))?;
+
+    let contract_address = Address::from_slice(contract_address.as_slice());
+    let state_root = Bytes32::try_from(state_root.as_slice())
+        .map_err(|_| PartialVMError::new(StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT))?;
+
+    let account_proof = create_account_proof(&contract_address, &state_root, &proof);
```
ycscaly, line 93: I don't like this. Why is this needed?

ycscaly: I mean, why do you need to put everything into a struct? this function should just compute the path

ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
yuvalelmo: I moved the check of the time slot so that if the time slot is not bigger than the latest it would not create the state.
Furthermore, there is a check for that inside the state verification function.

Also, as we saw together on last Thursday, the verification would fail if the network is not the same network, as this is checked when validating the sync committee signatures. 
The network is defined inside the `ConsensusStateManager` object that is serialized into the `data` field of this object, and it's configuration is hard-coded into helios.

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
yuvalelmo: It is not there.
inside `EthereumDWalletCap` we only have the `DWalletCap` ID.

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
yuvalelmo: It is still needed since it has the contract address & data slot

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
yuvalelmo: There is absolutely no use for the `state_root` in the state verification process.
The only purpose of this field is to have it later when we verify a message.

Let's say that this is the first time we use the initialized state object.
The first thing that would be done is bootstrapping the state using the `checkpoint` only (later we can add sync committee and some more data to initializer function, then we would not need to bootstrap).
Later, after verifying the state, we keep the updated `state_root` and `time_slot` in these fields, and use them later for the message verification.

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
ycscaly: So get `DWallet`, and check that it matches the ID of the `EthereumDWalletCap.dwallet_cap`. But you must verify the user-passed argument `dwallet_id` somehow, or better even, not get it from the user and extract from a Move struct.

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
ycscaly: Ok, but then you must verify that `LatestEthereumState` and `EthereumState` match the same network

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -42,22 +43,17 @@ module dwallet_system::eth_dwallet {
         eth_dwallet_cap: &EthereumDWalletCap,
         message: vector<u8>,
         dwallet_id: vector<u8>,
-        data_slot: u64,
         proof: vector<u8>,
         latest_ethereum_state: &LatestEthereumState,
         current_wrapped_eth_state: &EthereumState,
```
ycscaly, line 48:
```suggestion
        eth_state: &EthereumState,
```

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -90,35 +93,46 @@ pub fn verify_message_proof(
...
     // Verify the account proof against the state root.
     let is_valid_account_proof = verify_proof(
-        &account_proof.proof.as_slice(),
-        &account_proof.root,
-        &account_proof.path,
-        &account_proof.value,
+        &proof.clone().account_proof,
+        &state_root.as_slice().to_vec(),
+        &account_path,
+        &account_encoded,
```
ycscaly, line 104: good

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -90,35 +93,46 @@ pub fn verify_message_proof(
     let state_root = Bytes32::try_from(state_root.as_slice())
         .map_err(|_| PartialVMError::new(StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT))?;
 
-    let account_proof = create_account_proof(&contract_address, &state_root, &proof);
+    let account_path = keccak256(contract_address.as_bytes()).to_vec();
+    let account_encoded = encode_account(&proof);
```
ycscaly, line 97: good

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -90,35 +93,46 @@ pub fn verify_message_proof(
...
+        .iter()
+        .find(|p| p.key == U256::from(message_map_index.as_bytes()))
+        .ok_or_else(|| PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
+
+    let storage_value = [TRUE_VALUE].to_vec();
+    let mut msg_storage_proof_key_bytes = [0u8; 32];
+    msg_storage_proof
+        .key
+        .to_big_endian(&mut msg_storage_proof_key_bytes);
+    let storage_key_hash = keccak256(msg_storage_proof_key_bytes);
```
ycscaly, line 122: good

ycscaly: document what you're doing here please

ycscaly (changes requested):
I'm still not convinced that we are checking proofs belong to the correct network

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
yuvalelmo: added some doc, hope it's better now

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -42,20 +45,35 @@ module dwallet_system::eth_dwallet {
...
+        let latest_state_network = ethereum_state::get_latest_ethereum_state_network(latest_ethereum_state);
+        assert!(eth_dwallet_cap.network == latest_state_network, ENetworkMismatch);
+
+        // Validate that the EthereumState is for the correct network.
+        let eth_state_network = ethereum_state::get_ethereum_state_network(eth_state);
+        assert!(eth_state_network == latest_state_network, ENetworkMismatch);
+
+        // Validate that the Ethereum dWallet capability is for the correct DWallet.
+        let dwallet_cap_id = dwallet_2pc_mpc_ecdsa_k1::get_dwallet_cap_id(dwallet);
+        assert!(object::id(&eth_dwallet_cap.dwallet_cap) == dwallet_cap_id, EInvalidDWalletCap);
```
ycscaly, line 63: good

ycscaly (changes requested):
Ok, I'm sorry but this is all wrong, I made a wrong suggestion.

An attacker could simply initialize a new `LatestEthereumState`, and since we trust the initial values, the attacker can state whatever states they wish in there, and bypass verification.

This means that the `network` string has no security value.

In fact the only thing of importance here is the `LatestEthereumState`, and its update rules.

What needs to be done:
1. `LatestEthereumState` is intiialized to a trusted setup (this is OK as it is now)
2. When verifying a new state, it is verified against the `state_bytes` taken from `LatestEthereumState`, not a parameter. You add a new `latest_ethereum_state_id` field to the `EthereumState` object, and reflect it there.
3. The `EthereumDWalletCap` should hold `latest_ethereum_state_id` instead of `network`.
4. Now when approving a message, you don't check networks, you check that `latest_ethereum_state_id` match.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -126,10 +140,23 @@ module dwallet_system::ethereum_state {
         state.state_root
     }
 
-    public(friend) fun get_ethereum_state_network(
+    public(friend) fun set_latest_ethereum_state_id(
```
ycscaly, line 143: why does this has to be a function? it should be only set once during the initialization of the state object

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -51,12 +51,11 @@ module dwallet_system::eth_dwallet {
...
         // Validate that the Ethereum dWallet capability is for the correct network.
-        let latest_state_network = ethereum_state::get_latest_ethereum_state_network(latest_ethereum_state);
-        assert!(eth_dwallet_cap.network == latest_state_network, ENetworkMismatch);
+        assert!(eth_dwallet_cap.latest_ethereum_state_id == object::id(latest_ethereum_state), EStateObjectMismatch);
 
         // Validate that the EthereumState is for the correct network.
-        let eth_state_network = ethereum_state::get_ethereum_state_network(eth_state);
-        assert!(eth_state_network == latest_state_network, ENetworkMismatch);
+        let eth_state_id = object::id(eth_state);
+        assert!(eth_state_id == ethereum_state::get_ethereum_state_id_from_latest(latest_ethereum_state), EStateObjectMismatch);
```
ycscaly, line 58: this checks that the latest state is used, I think we said we don't have to. Lets do the opposite check: that the `latest_ethereum_state_id == eth_state.latest_ethereum_state_id`

ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -64,28 +75,88 @@ module dwallet_system::ethereum_state {
...
-            state_bytes
+            eth_state_bytes
         );
 
-        let new_state = EthereumState {
-            id: object::new(ctx),
-            data,
-            time_slot,
-        };
+        if (time_slot > latest_ethereum_state.last_slot) {
```
ycscaly, line 98: let's move this `if` to the beginning of the function, and actually make it an `assert`. No other purpose to call this function

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -64,28 +75,88 @@ module dwallet_system::ethereum_state {
...
+        assert!(latest_ethereum_state_id == object::id(latest_ethereum_state), EStateObjectMismatch);
+
+        let eth_state_bytes = get_ethereum_state_data(eth_state);
+        let (data, time_slot, state_root, _network) = verify_eth_state(
             updates_bytes,
             finality_update_bytes,
             optimistic_update_bytes,
-            state_bytes
+            eth_state_bytes
         );
```
ycscaly, line 97: assert `network == latest_ethereum_state.network`

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
ycscaly: delete unused function

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -73,12 +222,36 @@ pub(crate) fn create_initial_eth_state_data(
...
+    let network = String::from_utf8(network)
+        .map_err(|_| PartialVMError::new(StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT))?;
+
+    let file_path = format!("crates/eth-initial-state/eth_state_{}.bcs", network);
+
+    let mut file = File::open(file_path)
+        .map_err(|_| PartialVMError::new(StatusCode::VALUE_SERIALIZATION_ERROR))?;
+    let mut eth_state_bytes = Vec::new();
+    File::read_to_end(&mut file, &mut eth_state_bytes)
+        .map_err(|_| PartialVMError::new(StatusCode::VALUE_SERIALIZATION_ERROR))?;
```
ycscaly, line 252: Wait wait, I'm really uncertain that we can do this from the validator side. You are reading a file from the file system here? it's not common to do I/O like this in a native function. Let's think where else we can put this. One idea is indeed to do this in Genesis. If this is the plan, let's document this with a `TODO` here, and in an issue.

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
ycscaly: In this case, please in this function also call `verify_new_state`. I don't want to have unused, uninitialized values.
An alternative is to say we will remove this function and create these objects in Genesis, in which case we should add a `TODO` here and document this explicit in the issue. (Link issue in `TODO`)

ycscaly (changes requested):
Good progress

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
yuvalelmo: this is not the case anymore, i've changed the function to also return the state_root and set it to the correct one.

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
yuvalelmo: As we talked with Omer yesterday, there is no plan to insert this into genesis process.
Do you have any idea how else we can do this?
This is a constant data that we need to have hardcoded somehow, and I can't figure out another way to do this (this is huge data that we can't just hardcode into the .rs file)

ycscaly: @omersadika

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,89 @@
+/*
+ * eth_dwallet module provides functionality for managing Ethereum dWallets within the dwallet_system.
+ * It includes structures to represent dWallet capabilities, functions to create and manage Ethereum dWallets,
+ * and native functions to verify message proofs.
+*/
```
zeevmoney, line 5:
```suggestion
// 
// `eth_dwallet` module provides functionality for managing Ethereum dWallets within the dwallet_system.
// It includes structures to represent dWallet capabilities, functions to create and manage Ethereum dWallets,
// and native functions to verify message proofs.
// 
```

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -0,0 +1,265 @@
...
+/// The value of `True` in the contract's storage map.
+const TRUE_VALUE: u8 = 1;
+
+#[derive(Clone)]
+pub struct EthDWalletCostParams {
+    /// Base cost for invoking the `verify_eth_state` function
+    pub verify_eth_state_cost_base: InternalGas,
+    /// Base cost for invoking the `verify_message_proof` function
+    pub verify_message_proof_cost_base: InternalGas,
+    /// Base cost for invoking the `create_initial_eth_state_data` function
```
zeevmoney, line 47:
```suggestion
    /// Base cost for invoking the `verify_eth_state` function.
    pub verify_eth_state_cost_base: InternalGas,
    /// Base cost for invoking the `verify_message_proof` function.
    pub verify_message_proof_cost_base: InternalGas,
    /// Base cost for invoking the `create_initial_eth_state_data` function.
```

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+/***************************************************************************************************
+ * native fun verify_message_proof
+ * Implementation of the Move native function `verify_message_proof(p0: &mut NativeContext, p1: Vec<Type>, p2: VecDeque<Value>) -> PartialVMResult<NativeResult>`
+ * gas cost: verify_message_proof_cost_base   | base cost for function call and fixed operations.
+ **************************************************************************************************/
+pub fn verify_message_proof(
+    context: &mut NativeContext,
+    _ty_args: Vec<Type>,
```
zeevmoney, line 58:
```suggestion
```

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -0,0 +1,265 @@
...
+        &storage_value,
+    );
+    Ok(NativeResult::ok(cost, smallvec![Value::bool(is_valid)]))
+}
+
+/***************************************************************************************************
+ * native fun verify_eth_state
+ * Implementation of the Move native function `eth_dwallet::verify_eth_state(proof: vector<vector<u8>>, proof_len: u64, root: vector<u8>, eth_smart_contract_address: vector<u8>, eth_smart_contract_slot: u64, message: vector<u8>): bool;`
+ * gas cost: verify_eth_state_cost_base   | base cost for function call and fixed operations.
+ **************************************************************************************************/
```
zeevmoney, line 141:
```suggestion
/***************************************************************************************************
 * native fun verify_eth_state
 * Implementation of the Move native function
 * `eth_dwallet::verify_eth_state(proof: vector<vector<u8>>,
 * proof_len: u64,
 * root: vector<u8>,
 * eth_smart_contract_address: vector<u8>,
 * eth_smart_contract_slot: u64,
 * message: vector<u8>): bool;`
 * gas cost: verify_eth_state_cost_base   | base cost for function call and fixed operations.
 **************************************************************************************************/
```
Also, the params are wrong.

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -0,0 +1,265 @@
...
+            Value::vector_u8(network),
+        ],
+    ))
+}
+
+/***************************************************************************************************
+ * native fun create_initial_eth_state_data
+ * Implementation of the Move native function `eth_dwallet::create_initial_eth_state_data(checkpoint: vector<u8>): vector<u8>;`
+ * gas cost: create_initial_eth_state_data_cost_base   | base cost for function call and fixed operations.
+ **************************************************************************************************/
```
zeevmoney, line 219:
```suggestion
/***************************************************************************************************
* native fun create_initial_eth_state_data
* Implementation of the Move native function
* `eth_dwallet::create_initial_eth_state_data(checkpoint: vector<u8>): vector<u8>;`
* gas cost:
* create_initial_eth_state_data_cost_base | base cost for function call and fixed operations.
**************************************************************************************************/
```

zeevmoney (changes requested):
see comments

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
yuvalelmo: I think this is a part of a native function signature.

crates/sui/src/ethereum_client_commands.rs
```diff
@@ -138,9 +145,16 @@ pub enum EthClientCommands {
 /// This function should only be called once to initialize the Ethereum state.
 /// After the state is initialized, the Ethereum state object ID is saved in the configuration,
 /// and the state is updated whenever a new state is successfully verified.
+///
+/// There are only two acceptable checkpoints for the Ethereum state to initialize it:
+/// MAINNET Checkpoint: 0x8bfa089414dc5fe78dadc8b160a097fe744f17a80251f08eed0a3cdcc60b42f4
+/// HOLESKY Checkpoint: 0x8f867e31e2c55d9257dcd83effa0b7b74d7566a08bf2aabc5e133e91ffd11e2f
+///
+/// NOTE: Every other checkpoint would not create the initial state.
```
yuvalelmo, line 153:
```suggestion
/// NOTE: Any other checkpoint would end up failing to create the initial state.
```

ycscaly (approved):
Great work! Please rename the PR to a more appropriate name after we closed #87 and unified it.

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,90 @@
+// 
+// `eth_dwallet` module provides functionality for managing Ethereum dWallets within the dwallet_system.
+// It includes structures to represent dWallet capabilities, functions to create and manage Ethereum dWallets,
+// and native functions to verify message proofs.
+// 
+
+module dwallet_system::eth_dwallet {
+    use dwallet::object::{Self, ID, UID};
+    use dwallet::transfer;
```
zeevmoney, line 9:
```suggestion
```

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,90 @@
+// 
+// `eth_dwallet` module provides functionality for managing Ethereum dWallets within the dwallet_system.
+// It includes structures to represent dWallet capabilities, functions to create and manage Ethereum dWallets,
+// and native functions to verify message proofs.
+//
```
zeevmoney, line 5:
```suggestion
//
// `eth_dwallet` module provides functionality for managing Ethereum dWallets within the dwallet_system.
// It includes structures to represent dWallet capabilities, functions to create and manage Ethereum dWallets,
// and native functions to verify message proofs.
//
```

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,90 @@
...
+        proof: vector<u8>,
+        latest_ethereum_state: &LatestEthereumState,
+        eth_state: &EthereumState,
+    ): vector<MessageApproval> {
+        // Validate that the Ethereum dWallet capability is for the correct network.
+        assert!(eth_dwallet_cap.latest_ethereum_state_id == object::id(latest_ethereum_state), EStateObjectMismatch);
+
+        // Validate that the EthereumState is for the correct network.
+        let latest_ethereum_state_id = object::id(latest_ethereum_state);
+        assert!(latest_ethereum_state_id == ethereum_state::get_ethereum_state_latest_state_id(eth_state), EStateObjectMismatch);
```
zeevmoney, line 59:
```suggestion
        assert!(
            latest_ethereum_state_id == ethereum_state::get_ethereum_state_latest_state_id(eth_state),
            EStateObjectMismatch
        );
```

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,174 @@
+//
+// `ethereum_state` module provides functionality for managing and verifying the state of Ethereum within
+// the dwallet_system.
+// It includes structures to represent Ethereum state, functions to initialize and update state,
+// and native functions to verify state updates and create initial state data.
+//
+module dwallet_system::ethereum_state {
+    use dwallet::object::{Self, UID, ID};
```
zeevmoney, line 8:
```suggestion
//

module dwallet_system::ethereum_state {
    use dwallet::object::{Self, ID, UID};
```

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,174 @@
...
+        last_slot: u64,
+        eth_smart_contract_address: vector<u8>,
+        eth_smart_contract_slot: u64,
+        network: vector<u8>,
+    }
+
+    /// Initializes the first Ethereum state with the given checkpoint.
+    /// Creates an EthereumState object, shares a LatestEthereumState object pointing to it,
+    /// and freezes the EthereumState object.
+    /// NOTE: this function performs no verification on the `checkpoint`,
```
zeevmoney, line 43:
```suggestion
    /// NOTE: this function performs no verification on the `checkpoint`,
```

crates/sui-framework/packages/dwallet-system/sources/eth_dwallet.move
```diff
@@ -0,0 +1,90 @@
...
+        };
+        transfer::share_object(eth_dwallet_cap);
+    }
+
+    /// Verifies the provided proof using the `verify_message_proof` native function.
+    /// If the proof is valid, the message is approved.
+    public fun approve_message(
+        eth_dwallet_cap: &EthereumDWalletCap,
+        message: vector<u8>,
+        dwallet: &DWallet,
```
zeevmoney, line 49:
```suggestion
```

crates/sui-protocol-config/src/lib.rs
```diff
@@ -874,6 +874,13 @@ pub struct ProtocolConfig {
     /// Maximum allowed precision loss when reducing voting weights for the random beacon
     /// protocol.
     random_beacon_reduction_allowed_delta: Option<u16>,
+
+    // eth_dwallet::verify_eth_state_cost_base
+    verify_eth_state_cost_base: Option<u64>,
+    // eth_dwallet::verify_message_proof_cost_base
+    verify_message_proof_cost_base: Option<u64>,
+    // eth_dwallet::create_initial_eth_state_data_cost_base
+    create_initial_eth_state_data_cost_base: Option<u64>,
```
zeevmoney, line 883:
```suggestion
    // eth_dwallet::verify_eth_state_cost_base.
    verify_eth_state_cost_base: Option<u64>,
    // eth_dwallet::verify_message_proof_cost_base.
    verify_message_proof_cost_base: Option<u64>,
    // eth_dwallet::create_initial_eth_state_data_cost_base.
    create_initial_eth_state_data_cost_base: Option<u64>,
```

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -0,0 +1,174 @@
...
+            network,
+        };
+
+        state.latest_ethereum_state_id = object::uid_to_inner(&latest_ethereum_state.id);
+        transfer::freeze_object(state);
+        transfer::share_object(latest_ethereum_state);
+    }
+
+    /// Verifies the new Ethereum state according to the provided updates, and updates the LatestEthereumState object
+    /// if the new state is valid and has a newer time slot.
```
zeevmoney, line 76:
```suggestion
    /// Verifies the new Ethereum state according to the provided updates,
    /// and updates the LatestEthereumState object
    /// if the new state is valid and has a newer time slot.
```

zeevmoney (commented):
see comments

zeevmoney (approved):

PR #104, opened 2024-07-24
Fix broken file paths

sukrucildirr:


zeevmoney:
Thanks for the contribution, but we will change this readme entirely very soon.

PR #123, opened 2024-08-05
feat: Allow users to transfer their encrypted secret share with on-chain verification

ItayLevyOfficial:


crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+    let mut padded_vec = vec![0; target_length - vec.len()];
+    padded_vec.extend(vec);
+    padded_vec
+}
+
+pub fn generate_keypair() -> (Vec<u8>, Vec<u8>) {
+    let (public_key, private_key) = DecryptionKey::generate(&mut OsRng).unwrap();
+    let private_key = bcs::to_bytes(&private_key.secret_key).unwrap();
+    let public_key = bcs::to_bytes(&public_key).unwrap();
+    (public_key, private_key)
```
ycscaly, line 89: This is coupled to paillier/tiresias, maybe we should use generic traits instead

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+        paillier_public_parameters,
+        generator,
+    )
+}
+
+fn parse_plaintext(
+    plaintext: Vec<u8>,
+    public_parameters: &tiresias::encryption_key::PublicParameters,
+) -> PlaintextSpaceGroupElement {
+    let plaintext = pad_vector(plaintext);
```
ycscaly, line 124: What is this?

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+        plaintext,
+        public_parameters.plaintext_space_public_parameters(),
+    )
+    .unwrap();
+    plaintext
+}
+
+pub fn generate_proof(
+    encryption_key: Vec<u8>,
+    plaintext: Vec<u8>,
```
ycscaly, line 136:
```suggestion
    user_share: Vec<u8>,
```

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+        Lang,
+    >(
+        unbounded_witness_public_parameters,
+        language_public_parameters,
+    );
+
+    let witnesses = generate_witnesses(randomness, &enhanced_language_public_parameters, plaintext);
+
+    let (proofs, statements) = SecretShareProof::prove(
+        &PhantomData,
```
ycscaly, line 174: This should be `ProtocolContext`. Need to think what to put there

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+        &enhanced_language_public_parameters,
+        witnesses,
+        &mut OsRng,
+    )
+    .unwrap();
```
ycscaly, line 179: No `unwrap`

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+    NUM_RANGE_CLAIMS,
+    { COMMITMENT_SCHEME_MESSAGE_SPACE_SCALAR_LIMBS },
+    range::bulletproofs::RangeProof,
+    UnboundedWitnessSpaceGroupElement,
+    Lang,
+>;
+
+pub struct ProofPublicOutput {
+    pub proof: SecretShareProof,
+    pub encrypted_discrete_log: CiphertextSpaceValue,
```
ycscaly, line 67:
```suggestion
    pub encrypted_user_share: CiphertextSpaceValue,
```

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+> = EnhancedLanguage<
+    REPETITIONS,
+    NUM_RANGE_CLAIMS,
+    { COMMITMENT_SCHEME_MESSAGE_SPACE_SCALAR_LIMBS },
+    range::bulletproofs::RangeProof,
+    UnboundedWitnessSpaceGroupElement,
+    Lang,
+>;
+
+pub struct ProofPublicOutput {
```
ycscaly, line 65:
```suggestion
pub struct EncryptedUserShareAndProof {
```

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    let witnesses =
+        EnhancedLanguage::<
+            { maurer::SOUND_PROOFS_REPETITIONS },
+            RANGE_CLAIMS_PER_SCALAR,
+            { COMMITMENT_SCHEME_MESSAGE_SPACE_SCALAR_LIMBS },
+            bulletproofs::RangeProof,
+            RandomnessSpaceGroupElement,
+            Lang,
+        >::generate_witnesses(witnesses, &enhanced_language_public_parameters, &mut OsRng)
```
ycscaly, line 224: Use `generate_witness`

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+            .encryption_scheme_public_parameters
+            .ciphertext_space_public_parameters(),
+    )
+    .unwrap();
+
+    let statement = (
+        range_proof_commitment,
+        (encrypted_secret_share, public_key_share.clone()).into(),
+    )
+        .into();
```
ycscaly, line 310: Good

ycscaly (changes requested):
Use meaningful names please, and where possible write the underlying functions generically, using `ahe` traits

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
ycscaly: Actually it seems I didn't add a `generate` API there, so for now we can keep it as-is

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
```diff
@@ -0,0 +1,361 @@
...
+    )
+    .unwrap()
+}
+pub fn is_valid_proof(
+    language_public_parameters: LangPublicParams,
+    proof_public_output: ProofPublicOutput,
+    centralized_public_keyshare: group::Value<secp256k1::GroupElement>,
+) -> bool {
+    let secp256k1_group_public_parameters = secp256k1::group_element::PublicParameters::default();
+    let protocol_public_parameters = ProtocolPublicParameters::new(DUMMY_PUBLIC_KEY);
```
ycscaly, line 263: No no no, don't instantiate this struct with a wrong public key. This cannot be a part of production code - you should have been alarmed by the use of the word `DUMMY` inside a production non-test function.

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+    use dwallet::object::{Self, ID, UID};
+    use dwallet::token::recipient;
+    use dwallet::transfer;
+    use dwallet::tx_context;
+    use dwallet::tx_context::TxContext;
+    use dwallet_system::dwallet_2pc_mpc_ecdsa_k1::{DWallet, output};
+
+    const DWALLET_TRANSFER_ERROR: u64 = 0x1;
+
+    struct DwalletTransfer has key{
```
ycscaly, line 15:
```suggestion
    struct EncryptedUserShare has key {
```

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+        id: UID,
+        dwallet_id: ID,
+        encrypted_secret_share: vector<u8>,
+    }
+
+    struct PublicKey has key {
```
ycscaly, line 21:
```suggestion
    /// An Additively Homomorphic Encryption (AHE) public key that can be used to encrypt a user share in order to prove to the network that the recipient could sign with a dWallet when it is transferred or access is granted to it. 
    /// The recipient `address` must hold the corresponding decryption key.
    struct AHEncryptionKey has key {
```

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+        id: UID,
+        public_key: vector<u8>,
```
ycscaly, line 23:
```suggestion
        encryption_key: vector<u8>,
```

ycscaly: also please change everywhere else, encryption key is more appropriate

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public fun store_public_key(key: vector<u8>, ctx: &mut TxContext): ID {
+        let pk = PublicKey {
+            id: object::new(ctx),
+            public_key: key,
+            key_owner_address: tx_context::sender(ctx),
+        };
+        let pk_id = object::id(&pk);
+        transfer::freeze_object(pk);
```
ycscaly, line 34: A few consequences of this design:
1. There could be more than one encryption key per-address. There isn't really any need for that.
2. In case the decryption key has been lost, there is no way to "update" or delete this object as it is frozen. This means that in edge cases like these a dWallet could be transferred but the user could not access its secret share.
3. How do encrypters choose which encryption key to encrypt to?
4. Not every address is guranteed to hold an encryption key (I'm not sure actually that we can solve this, at least not without really changing SUI, so this might be ok)

ycscaly: Alt #1: make this a shared object

ycscaly: Alt #2: create a "registry" of encryption keys, which holds the current encryption key for every address. This could be a shared object, which is fine IMO since dWallet transfers don't seem to be a super time-critical task.

ycscaly: @omersadika

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+        pk_id
+    }
+
+    public fun transfer_dwallet(
```
ycscaly, line 38: So this is not transferring the dWallet, as the dWallet capability isn't even an argument of this function.
This function grants access to a new user for a dWallet.

In any case, @omersadika we should think from the user-experience perspective what kind of guarantee we want here.

sdk/signature-mpc-wasm/src/lib.rs
```diff
@@ -219,6 +219,32 @@ pub fn recovery_id_sha256(
...
+            public_key,
+            secret_share,
+            language_public_parameters,
+        );
+
+    let proof = bcs::to_bytes(&proof_public_output.proof).unwrap();
+    let encypted_discrete_log = bcs::to_bytes(&proof_public_output.encrypted_discrete_log).unwrap();
+    let range_proof_commitment = bcs::to_bytes(&proof_public_output.range_proof_commitment).unwrap();
+
+    serde_wasm_bindgen::to_value(&(proof, encypted_discrete_log, range_proof_commitment)).unwrap()
```
ycscaly, line 245: Why not serialize the `proof_public_output` directly? That's the whole purpose of this struct to hold it together no?

ycscaly:
Please move the `move` code to the `dwallet` module, no need to seperate

ycscaly:
Please add a step to the `create_dwallet` function that encrypts the user share before creating the `dwallet` i.e. as part of the dkg

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
ycscaly, line 23: Please add a `scheme` member that is an enum, that you now only have one option `Paillier`. Make sure to send this as a parameter to the `verify_dwallet_transfer` function. This is so that in the future we can use other algorithms

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
yaelAbergel: Wdym? Aren't we already using `generate_witness`?

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
yaelAbergel: I agree that the name `DUMMY` should not be in the production code, but it used to be `N` in many parts of the production code.

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+    ): ID {
+        let is_valid = verify_dwallet_transfer(
+            range_proof_commitment_value,
+            proof,
+            public_key.public_key,
+            encrypted_secret_share,
+            output(dwallet),
+        );
+
+        if (!is_valid) abort DWALLET_TRANSFER_ERROR;
```
yaelAbergel, line 55: change abort to assert

sui-execution/latest/sui-move-natives/src/crypto/twopc_mpc.rs
```diff
@@ -12,35 +11,96 @@ use move_vm_types::{
...
+
+#[derive(Clone)]
+pub struct TransferDWalletCostParams {
+    pub transfer_dwallet_gas: InternalGas,
+}
+
+/***************************************************************************************************
+ * native fun transfer_dwallet
+ **************************************************************************************************/
+pub fn verify_dwallet_transfer(
```
yaelAbergel, line 45: rename

crates/signature-mpc/Cargo.toml
```diff
@@ -32,3 +32,4 @@ thiserror = "1.0"
 workspace-hack.workspace = true
 bcs.workspace = true
 sha3 = "0.10.8"
+hex = "0.4.3"
```
yaelAbergel, line 35: dev-deps

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+
+        let dwallet_transfer = DwalletTransfer {
+            id: object::new(ctx),
+            dwallet_id: object::id(dwallet),
+            encrypted_secret_share,
+        };
+
+        let dt_id = object::id(&dwallet_transfer);
+        transfer::transfer(dwallet_transfer, recipient);
```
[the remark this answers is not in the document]
ycscaly: I believe it should be immutable

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
```diff
@@ -0,0 +1,76 @@
...
+        encrypted_secret_share: vector<u8>,
+    }
+
+    struct EncryptionKey has key {
+        id: UID,
+        encryption_key: vector<u8>,
+        key_owner_address: address,
+    }
+
+    public fun store_encryption_key(key: vector<u8>, ctx: &mut TxContext): ID {
```
ycscaly, line 27:
```suggestion
    public fun register_encryption_key(key: vector<u8>, ctx: &mut TxContext): ID {
```
Seems like a better name

yaelAbergel:
> Please add a step to the `create_dwallet` function that encrypts the user share (to the owner's own key, as backup) before creating the `dwallet` i.e. as part of the dkg

This will be addressed in a separate task.

crates/signature-mpc/src/twopc_mpc_protocols/dwallet_transfer.rs
ycscaly: following a conversation with our cryptographers, it seems fine to leave it as-is for this purpose

crates/sui-framework/packages/dwallet-system/sources/dwallet_transfer.move
yaelAbergel: I'm not sure if it is necessary to send the `scheme` to the `verify_dwallet_transfer` function at this stage. Perhaps there is a better design we can consider, and make it generic in a separate task?

yaelAbergel: Also, it seems like there are no emuns in move

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -321,6 +325,32 @@ pub fn is_valid_proof(
...
+    let ciphertext = CiphertextSpaceGroupElement::new(
+        encrypted_user_share_and_proof.encrypted_user_share,
+        language_public_parameters
+            .encryption_scheme_public_parameters
+            .ciphertext_space_public_parameters(),
+    )
+    .unwrap();
+
+    let paillier_public_parameters: tiresias::encryption_key::PublicParameters =
+        bcs::from_bytes(&encryption_key).unwrap();
```
ycscaly, line 343: This is the same as `encryption_scheme_public_parameters`

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -321,6 +325,32 @@ pub fn is_valid_proof(
         .is_ok()
 }
 
+pub fn decrypt_user_share(
+    encryption_key: Vec<u8>,
+    decryption_key: Vec<u8>,
+    encrypted_user_share_and_proof: EncryptedUserShareAndProof,
+) -> Vec<u8> {
+    let language_public_parameters = get_proof_public_parameters(encryption_key.clone());
```
ycscaly, line 333:
```suggestion
```
use `paillier_public_parameters`

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -321,6 +325,32 @@ pub fn is_valid_proof(
...
+            .ciphertext_space_public_parameters(),
+    )
+    .unwrap();
+
+    let paillier_public_parameters: tiresias::encryption_key::PublicParameters =
+        bcs::from_bytes(&encryption_key).unwrap();
+    let decryption_key = bcs::from_bytes(&decryption_key).unwrap();
+    let decryption_key = DecryptionKey::new(decryption_key, &paillier_public_parameters).unwrap();
+
+    let plaintext = decryption_key
```
ycscaly, line 347:
```suggestion
    // safe to `unwrap` as decryption in Paillier always succeeds
    let plaintext = decryption_key
```

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -310,4 +311,77 @@ module dwallet_system::dwallet {
...
+    /// that can be used to encrypt a user share in order to prove to the network that
+    /// the recipient can sign with a dWallet when it is transferred or access is granted to it.
+    struct EncryptionKey has key {
+        id: UID,
+        scheme: u8,
+        encryption_key: vector<u8>,
+        key_owner_address: address,
+    }
+
+    const PaillierEncryptionKey: u8 = 0;
```
ycscaly, line 336:
```suggestion
    const Paillier: u8 = 0;
```

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -310,4 +311,77 @@ module dwallet_system::dwallet {
...
+
+        let encrypt_user_share = EncryptedUserShare {
+            id: object::new(ctx),
+            dwallet_id: object::id(dwallet),
+            encrypted_secret_share_and_proof,
+            encryption_key_id: object::id(encryption_key),
+        };
+
+        let encrypt_user_share_obj_id = object::id(&encrypt_user_share);
+        transfer::transfer(encrypt_user_share, recipient);
```
ycscaly, line 377: I prefer having it as an immutable object, but in any case if it is owned you must check that `address == encryption_key.key_onwer_address`

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -240,15 +227,9 @@ module dwallet_system::dwallet_2pc_mpc_ecdsa_k1 {
...
-            id: object::new(ctx),
-            session_id,
-            dwallet_cap_id,
-            output,
-            public_key,
-        };
+        let dwallet = new_dwallet(session_id, dwallet_cap_id, output, public_key, ctx);
         // Create dwallet + make it immutable.
-        transfer::freeze_object(dwallet);
+        transfer::public_freeze_object(dwallet);
```
ycscaly, line 232: Please explain this change

crates/sui-framework/packages/dwallet-system/sources/dwallet_object.move
ycscaly, line 1: I think this object is already generic and can be kept in `dwallet.move` 
@omersadika

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -0,0 +1,406 @@
...
+
+fn pad_vector(vec: Vec<u8>) -> Vec<u8> {
+    let target_length = 256;
+    if vec.len() >= target_length {
+        return vec;
+    }
+    let mut padded_vec = vec![0; target_length - vec.len()];
+    padded_vec.extend(vec);
+    padded_vec
+}
```
ycscaly, line 84:
```suggestion
```

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -0,0 +1,406 @@
...
+        paillier_public_parameters,
+        generator,
+    )
+}
+
+fn parse_plaintext(
+    plaintext: Vec<u8>,
+    public_parameters: &tiresias::encryption_key::PublicParameters,
+) -> PlaintextSpaceGroupElement {
+    let plaintext = pad_vector(plaintext);
```
ycscaly, line 125:
```suggestion
```

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -0,0 +1,406 @@
...
+    let plaintext: LargeBiPrimeSizedNumber = LargeBiPrimeSizedNumber::from_be_slice(&plaintext);
```
ycscaly, line 126:
```suggestion
    let plaintext: LargeBiPrimeSizedNumber = U256::from_be_slice(&plaintext).into();
```

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -0,0 +1,406 @@
...
+    let witnesses =
+        EnhancedLanguage::<
+            { maurer::SOUND_PROOFS_REPETITIONS },
+            RANGE_CLAIMS_PER_SCALAR,
+            { COMMITMENT_SCHEME_MESSAGE_SPACE_SCALAR_LIMBS },
+            bulletproofs::RangeProof,
+            RandomnessSpaceGroupElement,
+            Lang,
+        >::generate_witnesses(witnesses, &enhanced_language_public_parameters, &mut OsRng)
+        .unwrap();
```
ycscaly, line 230: Use `EnhancedLanguage::generate_witness`

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -0,0 +1,406 @@
...
+    )
+    .unwrap()
+}
+pub fn is_valid_proof(
+    language_public_parameters: LangPublicParams,
+    proof_public_output: EncryptedUserShareAndProof,
+    centralized_public_keyshare: group::Value<secp256k1::GroupElement>,
+) -> bool {
+    let secp256k1_group_public_parameters = secp256k1::group_element::PublicParameters::default();
+    let protocol_public_parameters = ProtocolPublicParameters::new(N);
```
ycscaly, line 268: Don't use test constants in production code: https://github.com/dwallet-labs/dwallet-network/pull/123/files/e3e7e8963c1181ad84bb8e3b42288f6816e34468#r1705313423

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -310,4 +311,77 @@ module dwallet_system::dwallet {
...
+    public fun register_encryption_key(key: vector<u8>, scheme: u8, ctx: &mut TxContext): ID {
+        assert!(is_valid_encryption_key_scheme(scheme), EInvalidEncryptionKeyScheme);
+        let encryption_key = EncryptionKey {
+            id: object::new(ctx),
+            scheme: scheme,
+            encryption_key: key,
+            key_owner_address: tx_context::sender(ctx),
+        };
+        let encryption_key_id = object::id(&encryption_key);
+        transfer::freeze_object(encryption_key);
```
ycscaly, line 350: didn't we also said to create a `table`? and verify in `encrypt_user_share()` that it is the update object?

ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
yaelAbergel: I've refactored the `DWallet` object into a separate module. As a result, we need to utilize public transfers. Alternatively, I can implement a `freeze` method that uses `transfer::freeze_object`, and call `dwallet_object::freeze` from here. However, I believe it's best to proceed with the current version. [https://docs.sui.io/concepts/transfers]

crates/sui-framework/packages/dwallet-system/sources/dwallet_object.move
yaelAbergel: Lmk

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
yaelAbergel: cool

crates/signature-mpc/src/twopc_mpc_protocols/encrypt_user_share.rs
```diff
@@ -205,30 +183,28 @@ fn generate_witnesses(
...
             bulletproofs::RangeProof,
             RandomnessSpaceGroupElement,
             Lang,
-        >::generate_witnesses(witnesses, &enhanced_language_public_parameters, &mut OsRng)
-        .unwrap();
-    witnesses
+        >::generate_witness(witnesses, &enhanced_language_public_parameters, &mut OsRng)
+            .unwrap();
+    witness
 }
```
ycscaly, line 208: this function is now simple enough to be more of a bloatware than descriptive

ycscaly (changes requested):
Remove unwraps

zeevmoney:
Fix conflicts

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
zeevmoney: In a separate PR, she is working on it.

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -386,4 +419,77 @@ module dwallet_system::dwallet {
...
+            encryption_key_id: object::id(encryption_key),
+        };
+
+        let encrypt_user_share_obj_id = object::id(&encrypt_user_share);
+        transfer::freeze_object(encrypt_user_share);
+        encrypt_user_share_obj_id
+    }
+
+    #[allow(unused_function)]
+    native fun validate_encrypted_user_secret_share(
```
zeevmoney, line 490: change to verify.

zeevmoney (changes requested):
see comment

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -45,6 +45,39 @@ module dwallet_system::dwallet {
...
+    struct DWallet<phantom T> has key, store {
+        id: UID,
+        session_id: ID,
+        dwallet_cap_id: ID,
+        // `output` is output for `verify_decommitment_and_proof_of_centralized_party_public_key_share()`
+        output: vector<u8>,
+        public_key: vector<u8>,
+    }
+
+    public fun new_dwallet<T: drop>(
```
omersadika, line 59: rename it to create_dwallet and make it public(friend)

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -386,4 +419,80 @@ module dwallet_system::dwallet {
...
+    }
+
+    /// Encrypt a user share with an AHE encryption key.
+    public fun encrypt_user_share<T: drop>(
+        dwallet: &DWallet<T>,
+        encryption_key: &EncryptionKey,
+        encrypted_secret_share_and_proof: vector<u8>,
+        ctx: &mut TxContext,
+    ): ID {
+        let is_valid = verify_encrypted_user_secret_share(
```
omersadika, line 472: You have to move the verification to the implementation module

crates/sui-framework/packages/dwallet-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -456,6 +468,37 @@ module dwallet_system::dwallet_2pc_mpc_ecdsa_k1 {
...
+        create_encrypted_user_share(
+            object::id(dwallet),
+            encrypted_secret_share_and_proof,
+            object::id(encryption_key),
+            ctx
+        );
+    }
+
+    #[allow(unused_function)]
+    native fun verify_encrypted_user_secret_share(
```
ycscaly, line 495:
```suggestion
    native fun verify_encrypted_user_secret_share_secp256k1(
```
I'm not sure, but I think that the native function's should have unique names right, so probably better to have a unique name so it wouldn't cause compilation problems in the future

ycscaly (changes requested):
Great progress with generics. Getting close

sdk/signature-mpc-wasm/src/lib.rs
```diff
@@ -213,10 +228,50 @@ pub fn recovery_id_sha256(
...
+#[wasm_bindgen]
+pub fn generate_keypair() -> Result<JsValue, JsErr> {
+    let (public_key, private_key) =
+        signature_mpc::twopc_mpc_protocols::encrypt_user_share::generate_keypair()
+            .map_err(|e| to_js_err(e))?;
+    Ok(serde_wasm_bindgen::to_value(&(public_key, private_key))?)
+}
+
+#[wasm_bindgen]
+pub fn generate_proof(secret_share: Vec<u8>, public_key: Vec<u8>) -> Result<JsValue, JsErr> {
```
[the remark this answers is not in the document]
yaelAbergel: Could you please clarify where this should be used? I'm referring to [this test](https://github.com/dwallet-labs/enhanced-maurer/blob/main/src/proof.rs#L450) as an example. Can you guide me on the appropriate usage or location for it? Additionally, we've not used this until now, and the verification has been passing successfully.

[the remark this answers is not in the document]
zeevmoney: @ycscaly why?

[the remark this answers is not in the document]
ycscaly: Because otherwise how will users use it? They don't have access to the secret share, only the DKG ouptut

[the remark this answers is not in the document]
zeevmoney: We already talked about it.
Currently this code is used like this:

```typescript
		const dwallet = await createDWallet(toolbox.keypair, toolbox.client);
		const dwalletID = dwallet?.dwalletId!;
		// Extract the secret share from the DKGOutput.
		// It's the first 32 bytes because it's the first field in the Output Rust struct.
		let secretShareBytes = dwallet?.dkgOutput!.slice(0, 32);
		let secretShare = new Uint8Array(secretShareBytes);
		const encryptedUserShareAndProof = generate_proof(secretShare, recipientData?.encryptionKey!);
```

The `secretShare` is taken from the DKG output.
If we pass only the DKG output then we will need to do the same thing as above, but in Rust.

In order to avoid it we have to move the `signature_mpc::twopc_mpc_protocols::encrypt_user_share::generate_proof` code to the 2PCMPC lib.

I think for now let's keep it and if we decide to move the code to the 2PCMPC lib, then we can change this code appropriately.

zeevmoney:
Also fmt.

zeevmoney (approved):

sdk/signature-mpc-wasm/src/lib.rs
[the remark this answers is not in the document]
ycscaly: https://github.com/dwallet-labs/2pc-mpc/pull/26 approved, now you can fix it so that there are no weird bypasses. Make the rust code take `dkg_output`, or create a new wasm function that takes the `secret_key_share` out of it.

ycscaly (approved):
gj

zeevmoney:
Good job!

PR #153, opened 2024-08-15
chore: Add TypeScript CI flow

ItayLevyOfficial:


.github/workflows/ts-ci.yaml
zeevmoney, line 1: Looks good, just add cache

zeevmoney (approved):

PR #156, opened 2024-08-18
feat: Add a TS function to verify `secret_share` matches `public_share`

ItayLevyOfficial:


zeevmoney (changes requested):

ycscaly (changes requested):
Final thing to add: when encrypting, verify that the decentralized party DKG output holds correct values for the public key shares

ycscaly (approved):

PR #198, opened 2024-09-01
fix docker & Remove redundant `sdk/create-dapp/templates/react-e2e-counter` folder

zeevmoney:
see #197 
Change default docker stack.


ItayLevyOfficial (approved):
LGTM!

PR #199, opened 2024-09-01
feat: Create encrypted user shares table to store all encrypted user shares encrypted by user

yaelAbergel:


crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -454,6 +455,49 @@ module dwallet_system::dwallet {
         encryption_key_id: ID,
     }
 
+    /// Owned object that holds all encrypted user shares that were sent to the user,
```
ycscaly, line 458:
```suggestion
    /// Owned object that holds all encrypted user shares for the dWallet the user has access to. 
    /// These are those that were encrypted after the user created the dWallet or sent to the user,
```

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -454,6 +455,49 @@ module dwallet_system::dwallet {
...
+        transfer::transfer(holder, tx_context::sender(ctx));
+    }
+
+    public fun save_encrypted_user_share(
+        encrypted_user_shares: &mut EncryptedUserShares,
+        encrypted_user_share: &EncryptedUserShare,
+        encryption_key: &EncryptionKey,
+        ctx: &mut TxContext
+    ) {
+        // TODO: add signature verification
```
ycscaly, line 480:
```suggestion
```
What signature needs to be verified here? I believe only when the `EncryptedUserShare` object is created it should be verified

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -454,6 +455,49 @@ module dwallet_system::dwallet {
...
+        let validate_parameters = !table::contains(&encrypted_user_shares.encrypted_user_shares, encrypted_user_share.dwallet_id)
```
ycscaly, line 481: what is this pattern? I think its better to have the condition inside the `assert`

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -454,6 +455,49 @@ module dwallet_system::dwallet {
...
+            && object::id(encryption_key) == encrypted_user_share.encryption_key_id
```
ycscaly, line 482:
```suggestion
```
I think this check is not needed

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -454,6 +455,49 @@ module dwallet_system::dwallet {
...
+            id: object::new(ctx),
+            encrypted_user_shares: table::new(ctx),
+        };
+        transfer::transfer(holder, tx_context::sender(ctx));
+    }
+
+    public fun save_encrypted_user_share(
+        encrypted_user_shares: &mut EncryptedUserShares,
+        encrypted_user_share: &EncryptedUserShare,
+        encryption_key: &EncryptionKey,
```
ycscaly, line 477:
```suggestion
```

sdk/typescript/src/signature-mpc/dwallet.ts
ycscaly, line 1: I think I'm missing:
1. after DKG, you encrypt the user share and save it in the table
2. when you accept a user-share, you add it to the table
If they are handled in different PRs please note which

ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
yaelAbergel: Why not? Don't we want to enforce users to save only user shares encrypted by their own key?

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
yaelAbergel: Ok. I thought it'll be clearer this way since its a long condition

sdk/typescript/src/signature-mpc/dwallet.ts
yaelAbergel: This is just the table intended for the two features you mentioned. We haven't used it in this PR yet.

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
ycscaly: Encryptions are not said to be "by a key" but "to a key" - since this key is public and anyone can encrypt to it. You are right, we should have this check.

ycscaly: But I am missing another check: that the `EncryptedUserShare` is actually encrypted *by* the user. This doesn't exist here because you didn't base the PR on the branch that includes the signature inside `EncryptedUserShare`. Do that pleas

sdk/typescript/src/signature-mpc/dwallet.ts
ycscaly: Yes but you should also make sure to add these when these operations happen. This is just the move & ts code for using the move code, but we should also actually call this code from relevant code paths

sdk/typescript/src/signature-mpc/dwallet.ts
```diff
@@ -265,3 +309,25 @@ export const transferEncryptedUserShare = async (
 		},
 	});
 };
+
+export const getEncryptedUserShare = async (
+	client: DWalletClient,
+	keypair: Keypair,
+	encrptedUserSharesObjID: string,
```
ItayLevyOfficial, line 316: Pull it from the chain, remove parameter

ycscaly (approved):
gj

ItayLevyOfficial (approved):
LGTM

zeevmoney (approved):

PR #203, opened 2024-09-02
feat: Create an SDK function that retrieves an encryption key object for a given keypair if exists

ItayLevyOfficial:


ycscaly (approved):
Good, now we need to incorporate this (in a seperate PR) in the normal API so that users don't need to remember any secrets besides their keeper, and can simply use dWallets by IDs.

zeevmoney (approved):

PR #213, opened 2024-09-05
`dwallet-network`: Ethereum light client fix

yuvalelmo:
…beacon block sent from the user (compared to the beacon block hash we keep in state).


crates/sui-framework/packages/dwallet-system/sources/dwallet.move
```diff
@@ -59,7 +59,7 @@ module dwallet_system::dwallet {
 
     /// `MessageApproval` represents a message that was approved.
     /// Bound to a `DWalletCap`.
-    struct MessageApproval has store {
+    struct MessageApproval has store, drop {
```
ycscaly, line 62: why did you add this? @omersadika

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -93,17 +114,64 @@ pub fn verify_message_proof(
...
+
+    let verified_block_hash = verified_block_header
+        .hash_tree_root()
+        .map_err(|_| PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
+
+    if beacon_block_hash != verified_block_hash {
+        return Ok(NativeResult::ok(cost, smallvec![Value::bool(false)]));
+    }
+
+    let state_root = beacon_block.body.execution_payload().state_root();
```
ycscaly, line 166: Please explain the change and why this is required

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -93,17 +114,64 @@ pub fn verify_message_proof(
...
+        .hash_tree_root()
+        .map_err(|_| PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
+
+    let mut verified_block_header = eth_state_data.get_finalized_header();
+
+    let verified_block_hash = verified_block_header
+        .hash_tree_root()
+        .map_err(|_| PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
+
+    if beacon_block_hash != verified_block_hash {
```
ycscaly, line 162: don't you need to actually compute the hash yourself?

ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/dwallet.move
yuvalelmo: last line of the `approve_message` is calling:
`dwallet::approve_messages(&eth_dwallet_cap.dwallet_cap, vector[message])`.
When there is no drop ability I get a runtime error whenever i approve a message, even though my function also returns `vec<MessageApproval>`

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
yuvalelmo: When a user approves a message, he sends the relevant `EthereumState` move object ID, with the whole beacon block that he fetched from an rpc.
In the `EthereumState` object we have an already verified state(otherwise the object wouldn't have been created).
What we do is take the beacon block that is sent from the user, and compute the hash of it, then we want to make sure that the hash is equal to the hash of the header we already have verified in the `EthereumState` object.

If so - then we know we have the correct block, and we can trust it's execution payload input, and then we get the current Execution layer `state_root`.

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
yuvalelmo: This is what I do.
method call `.hash_tree_root()` gives us the hash of the block/header.

I compare the hash of the input beacon block, and the hash of the verified block header that is saved in the `EthereumState` object.

ycscaly: I'm not sure that this method computes the hash, it might retrtieve it from some field. Can you link the implementation?

yuvalelmo: https://docs.rs/ssz_rs/latest/ssz_rs/trait.Merkleized.html

`Merkleized` is a trait that is being implemented in our code by using the `SimpleSerialize` derive, provided by `ssz_rs` lib.

yuvalelmo: You can also see it's specs here:
[ssz specs](https://github.com/ethereum/consensus-specs/blob/fa09d896484bbe240334fa21ffaa454bafe5842e/ssz/simple-serialize.md#merkleization)

[ssz readme](https://github.com/ralexstokes/ssz-rs/blob/1df4cd9b849a48c44a6105abc3a38d21cd4fd8d3/README.md?plain=1#L29)

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
ycscaly: Please document this in the function

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -93,17 +114,69 @@ pub fn verify_message_proof(
...
+    let beacon_block_body = BeaconBlockBody::new_from_existing_with_execution_payload(
+        beacon_block_body,
+        beacon_block_execution_payload,
+    );
+    beacon_block.body = beacon_block_body;
+
+    let eth_state_data = bcs::from_bytes::<ConsensusStateManager<NimbusRpc>>(&eth_state_data)
+        .map_err(|_| PartialVMError::new(StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT))?;
+
+    // Get the beacon block hash.
```
ycscaly, line 152:
```suggestion
    // Compute the root hash of the Merkle tree using the unverified (user-specified) beacon block.
```

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -93,17 +114,69 @@ pub fn verify_message_proof(
...
+    let beacon_block_hash = beacon_block
+        .hash_tree_root()
+        .map_err(|_| PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR))?;
+
+    let mut verified_block_header = eth_state_data.get_finalized_header();
+
+    // Get the verified beacon header hash.
```
ycscaly, line 159:
```suggestion
    // Compute the root hash of the Merkle tree using the verified beacon header.
```

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
ycscaly: So definitely the Merkle roots get computed and it makes sense that the header contains just the hash of the beacon body. Two things are left for us to check:
1. can we access the beacon body hash from the verified header directly, and then hash the block ourselves to see if it matches (this probably is an overkill, but I'm good with doing excess checks)?
I think we can, using `finalized_header.body_root` and comparing it to the hash of the `beacon_block_body`. let's try to add this check
2. I'm not sure about the data structure of the verified header - it should be a `BeaconBlockHeader`, but I see it is a `Header`: `finalized_header: Header`. I see its not only a naming issue, but these two types exist. They do seem to hold the exact same values, so there could be some overloading here, but I'm not sure that the `finalized_header` is necessairly a `BeaconBlockHeader` yet

ycscaly: In fact, `BeaconBlockHeader` is not actually used anywhere in the code, only `Header` does. 
So that leaves the question: do you know (and can point me out) to where in the protocol it says the header of the `ConsensusStateManager` i.e. `finalized_header` is said to be a header of a beacon block exclusively?

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
```diff
@@ -82,6 +80,7 @@ module dwallet_system::ethereum_state {
         optimistic_update_bytes: vector<u8>,
         latest_ethereum_state: &mut LatestEthereumState,
         eth_state: &EthereumState,
+        is_init_state: bool,
```
ycscaly, line 83: Please explain this channe. This is a public function, meaning that the user can freely specify this value at will. What is its purpose? what will happen if a user says its an init state when its not?

ycscaly: document this

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -191,12 +265,20 @@ pub(crate) fn verify_eth_state(
         optimistic_update,
     };
 
-    eth_state.verify_and_apply_updates(&updates).map_err(|e| {
-        error!("failed to verify updates: {:?}", e);
-        PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
-    })?;
+    if (is_init) {
+        eth_state
+            .verify_and_apply_initial_updates(&updates)
```
ycscaly, line 270: can you link the code for this function `verify_and_apply_initial_updates()`? I can't find it.

ycscaly (changes requested):
I am closer to believing this change works, but still some clarifications are needed

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
yuvalelmo: 1. this is exactly what i thought. did some research yesterday and i think it is necessary to also validate the body root as you stated. I am adding this right now.
2. there you go:
https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#lightclientheader

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
yuvalelmo: This is probably because the helios PR we did is not merged yet.
You can find it here:
https://github.com/dwallet-labs/helios/pull/23/files

crates/sui-framework/packages/dwallet-system/sources/ethereum_state.move
yuvalelmo: is_init_state is not a very good name, ill change it.
This is a flag to figure how to apply the updates.
If it is not the first time we apply the update(which only happens the first time we use and update the `LatestEthereumState`), then we must only verify and apply the finality update & optimistic update, and only if the "regular" update has a new sync committee - we would verify and apply it as well(otherwise we do not use it).

If you put the wrong value in this flag - the state verification would fail.

ycscaly (approved):
Great work!

zeevmoney (approved):

PR #237, opened 2024-09-16
chore: Implement start flow message handling

ItayLevyOfficial:
This PR only create the initial party upon receiving a "start flow" message

### DIFF
--- a/crates/mpc-protocols/src/mpc_service.rs
+++ b/crates/mpc-protocols/src/mpc_service.rs
@@ -1,5 +1,9 @@
 use crate::mpc_events::{CreatedProofMPCEvent, MPCEvent};
+use group::{secp256k1, GroupElement};
+use maurer::knowledge_of_discrete_log::PublicParameters;
+use rand_core::OsRng;
 use std::collections::{HashMap, VecDeque};
+use std::marker::PhantomData;
 use sui_types::base_types::ObjectID;
 use sui_types::event::Event;
 use tokio::sync::{mpsc, Mutex};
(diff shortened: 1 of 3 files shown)


crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -51,8 +70,30 @@ impl MPCService {
...
+            let party: proof::aggregation::asynchronous::Party<
+                maurer::Proof<{ maurer::SOUND_PROOFS_REPETITIONS }, Lang, PhantomData<()>>,
+            > = proof::aggregation::asynchronous::Party::new_proof_round_party(
+                language_public_parameters.clone(),
+                PhantomData,
+                3,
+                1,
+                &mut OsRng,
+            )
+            .unwrap();
```
cdesaintguilhem, line 84: These public parameters look like they're independent of the instance, and will be the same for every MPC Session. Is it not cheaper to hold only one copy of them in the RAM and only give an `Arc` reference (or whatever type of reference is suitable in the async environment) to them to each instance?

yaelAbergel: We changed the code to initiate the public parameters just once. `Party::new_proof_round_party` function takes ownership so we must clone it.

cdesaintguilhem: Okay, that saves the computation of the public parameters. But when you're cloning it you're also creating a new copy and therefore using more RAM, no?

(I say this from an efficiency perspective, maybe we don't care about saving that much RAM.)

PR #241, opened 2024-09-17
chore (proof-mpc): re-create `CreatedProofMPCSessionEvent` and matching rust type

ItayLevyOfficial:
Need to open this PR again as wer'e now working on `pera-v1.32.2`, that contain the latest version of Sui


cdesaintguilhem (approved):
Fine for me, but I don't know how much my approval is worth here.

ycscaly (approved):

PR #243, opened 2024-09-17
fix: Remove Pera rename duplications 

yaelAbergel:
We saw that Omer's renaming had redundant leftovers, this PR removes them


zeevmoney (approved):
run fmt

PR #244, opened 2024-09-18
chore (proof-mpc): re-implement init event handling & structure for the proof MPC flow

ItayLevyOfficial:
This PR is identical to PR #234 just for the sui latest version (1.32.2)


cdesaintguilhem (approved):

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+            }
+        });
+    }
+
+    /// Filter the relevant MPC events from the transaction events & handle them
+    pub fn handle_mpc_events(&mut self, events: &Vec<Event>) -> anyhow::Result<()> {
+        for event in events {
+            if CreatedProofMPCEvent::type_() == event.type_ {
+                let deserialized_event: CreatedProofMPCEvent = bcs::from_bytes(&event.contents)?;
+                let _ = self.handle_proof_init_event(deserialized_event);
```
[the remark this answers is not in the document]
yaelAbergel: good catch, I fixed it

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Handles a proof initialization event
+    /// Spawns a new MPC instance if the number of active instances is below the limit
+    /// Otherwise, adds the instance to the pending queue
+    fn handle_proof_init_event(&mut self, event: CreatedProofMPCEvent) {
+        println!(
+            "Received start flow event for session ID {:?}",
+            event.session_id
+        );
+        // let active_instances_counter = self.active_instances_counter.clone().into_inner();
```
erik-3milabs, line 84: dead code? please remove

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /// Handles a proof initialization event
+    /// Spawns a new MPC instance if the number of active instances is below the limit
+    /// Otherwise, adds the instance to the pending queue
+    fn handle_proof_init_event(&mut self, event: CreatedProofMPCEvent) {
+        println!(
+            "Received start flow event for session ID {:?}",
+            event.session_id
+        );
```
erik-3milabs, line 83: Is it OK to print directly to `stdout` here? Do you not use loggers / debug flags for printing?

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+                MPCInstance {
+                    status: MPCStatus::Pending,
+                    messages_handler_sender: None,
+                    pending_messages: vec![],
+                },
+            );
+            self.pending.push_back(event.session_id.bytes);
+            return;
+        }
+        let (messages_handler_sender, messages_handler_receiver) = mpsc::channel(100);
```
erik-3milabs, line 98: Alert: hard coding `100`.

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/// The `MPCService` is responsible for managing MPC instances:
+/// - It keeps track of all MPC instances
+/// - Runs the MPC session for each active instance
+/// - Ensures that the number of active sessions does not go over `MAX_ACTIVE_MPC_INSTANCES` at the same time
+pub struct MPCService {
+    mpc_instances: HashMap<ObjectID, MPCInstance>,
+    pending: VecDeque<ObjectID>,
+    active_instances_counter: usize,
```
erik-3milabs, line 42: Keeping this counter updated is an error prone chore. Would it be possible to make this a computed property that filters `mpc_instances` to count the active instances?

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+                status: MPCStatus::Active,
+                messages_handler_sender: Some(messages_handler_sender),
+                pending_messages: vec![],
+            },
+        );
+        self.active_instances_counter += 1;
+        println!(
+            "Added MPCInstance to service for session_id {:?}",
+            event.session_id
+        );
```
erik-3milabs, line 113: 1. Consider making this an `if-else` instead.
2. Consider moving the `self.mpc_instances.insert(` outside the `if-else` to avoid unnecessary duplication.
3. How unique are these session_ids? This function implicitly assumes that the `session_id.bytes` are not yet in the map.

crates/mpc-protocols/src/mpc_events.rs
```diff
@@ -1,13 +1,35 @@
+use move_core_types::ident_str;
+use move_core_types::identifier::IdentStr;
+use move_core_types::language_storage::StructTag;
```
erik-3milabs, line 3: Consider optimizing imports (putting them in one line)?

crates/mpc-protocols/src/mpc_events.rs
```diff
@@ -1,13 +1,35 @@
...
 /// Rust version of the Move [`pera_system::dwallet::CreatedProofMPCSessionEvent`] type.
 #[derive(Debug, Serialize, Deserialize, Clone, JsonSchema, Eq, PartialEq)]
-pub struct CreatedProofMPCSessionEvent {
+pub struct CreatedProofMPCEvent {
     pub session_id: ID,
     pub sender: PeraAddress,
 }
+
+pub const PROOF_MODULE_NAME: &IdentStr = ident_str!("proof");
+pub const CREATED_PROOF_STRUCT_NAME: &IdentStr = ident_str!("CreatedProofMPCSessionEvent");
```
erik-3milabs, line 20: Could you please explain what the upside to making these `const &IdentStr` is? Just curious

crates/mpc-protocols/src/mpc_events.rs
```diff
@@ -1,13 +1,35 @@
...
+
+pub trait MPCEvent {
+    fn type_() -> StructTag;
+}
+
+impl MPCEvent for CreatedProofMPCEvent {
+    fn type_() -> StructTag {
+        StructTag {
+            address: PERA_SYSTEM_ADDRESS,
```
erik-3milabs, line 29: This confuses me. Is the address hardcoded / constant? If so, why would that not be problematic?

crates/pera-core/src/authority.rs
```diff
@@ -1522,6 +1533,28 @@ impl AuthorityState {
...
+            TransactionEffects::V1(effects) => effects.status(),
+            TransactionEffects::V2(effects) => effects.status(),
+        };
+        if status.is_err() {
+            return Ok(());
+        }
+        self.mpc_state
+            .lock()
+            .handle_mpc_events(&inner_temporary_store.events.data)?;
+        Ok(())
```
erik-3milabs, line 1555: IIUC, you can simplify this to

```suggestion
        self.mpc_state
            .lock()
            .handle_mpc_events(&inner_temporary_store.events.data)
```

crates/pera-core/src/authority.rs
```diff
@@ -1522,6 +1533,28 @@ impl AuthorityState {
...
+        if !self.is_validator(epoch_store) {
+            return Ok(());
+        }
+        let status = match &effects {
+            TransactionEffects::V1(effects) => effects.status(),
+            TransactionEffects::V2(effects) => effects.status(),
+        };
+        if status.is_err() {
+            return Ok(());
+        }
```
erik-3milabs, line 1551: This feels problematic. You are returning an OK in case of an error? Should this not return `Err`?

crates/pera-core/src/authority.rs
```diff
@@ -803,6 +806,8 @@ pub struct AuthorityState {
     pub overload_info: AuthorityOverloadInfo,
 
     pub validator_tx_finalizer: Option<Arc<ValidatorTxFinalizer<NetworkAuthorityClient>>>,
+
+    mpc_state: Mutex<MPCService>,
```
erik-3milabs, line 810: Is there a reason this `MPCService` object is called the `mpc_state` here? I would expect `mpc_service` as a name.

crates/pera-framework/docs/pera-system/proof.md
```diff
@@ -0,0 +1,122 @@
+---
+title: Module `0x3::proof`
+---
+
+The proof module
+Responsible to start and manage the Proof generation MPC flow
```
erik-3milabs, line 6:
```suggestion
Responsible for starting and managing the Proof generation for the MPC flow.
```

crates/pera-framework/docs/pera-system/proof.md
```diff
@@ -0,0 +1,122 @@
...
+Used only for testing the way we launch & manage an MPC flow.
```
erik-3milabs, line 7:
```suggestion
Used only for testing launching & managing an MPC flow.
```

sdk/typescript/test/e2e/signature-mpc.test.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
```
erik-3milabs, line 2: Why is this copyright for Mysten Labs?

sdk/typescript/src/signature-mpc/proof.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Mysten Labs, Inc.
```
erik-3milabs, line 1: Why is this copyright for Mysten Labs?

erik-3milabs (changes requested):
Looks like functional code. Takes some refinement to be excellent code :rocket:

crates/mpc-protocols/src/mpc_service.rs
erik-3milabs: image

ItayLevyOfficial:
will be moved to https://github.com/dwallet-labs/dwallet-network/pull/248

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
+use crate::mpc_events::{CreatedProofMPCEvent, MPCEvent};
+use pera_types::base_types::ObjectID;
+use pera_types::event::Event;
+use std::cell::RefCell;
+use std::collections::{HashMap, VecDeque};
+use tokio::sync::{mpsc, Mutex, RwLock};
+
+const MAX_ACTIVE_MPC_INSTANCES: usize = 100;
```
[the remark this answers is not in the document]
ItayLevyOfficial: Created an issue & added todo

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+pub enum MPCInput {
+    InitEvent(CreatedProofMPCEvent),
+    Message,
+    OutputEvent,
+}
+
+struct MPCInstance {
+    status: MPCStatus,
+    /// The channel to send message to this instance's message handler thread when this instance is active
+    messages_handler_sender: Option<mpsc::Sender<MPCInput>>,
```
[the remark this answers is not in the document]
ItayLevyOfficial: I accepted your suggestion, I'm not a fan of comments either but Zeev asked me to over-document the code so it will be easier for him to grasp it when he comes back. (He's on a holiday rn)

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+/// - Active: The session is currently running, new messages will be forwarded to the session
+/// - Pending: The session is waiting for a slot to become active, when the init received there were already `MAX_ACTIVE_MPC_INSTANCES` active sessions
+/// - Finished: The session has finished, no more messages will be forwarded. The session will be removed from the service after a timeout.
+/// We want to keep it for some time so leftover messages related to it won't be treated as malicious.
+#[derive(Clone, Copy)]
+enum MPCStatus {
+    Active,
+    Pending,
+    Finished,
+}
```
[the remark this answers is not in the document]
ItayLevyOfficial: I'm not a fan of comments either but Zeev asked me to over-document the code so it will be easier for him to grasp it when he comes back. (He's on a holiday rn)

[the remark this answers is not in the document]
ItayLevyOfficial: Re "Pending to AwaitingInput" - I prefer pending

crates/pera-core/src/authority.rs
yaelAbergel: If the transaction failed, we don't need to handle MPC events. added comment

crates/mpc-protocols/src/mpc_events.rs
yaelAbergel: `PERA_SYSTEM_ADDRESS` is the package address not the users. This will always be the same for all MPC events.

crates/mpc-protocols/src/mpc_events.rs
yaelAbergel: We don't know why, this is just the type being used in `StructTag`.

crates/mpc-protocols/src/mpc_events.rs
yaelAbergel: These imports are from different modules .

crates/mpc-protocols/src/mpc_service.rs
yaelAbergel: 1. Zeev repeatedly asks us to return early instead of using `if-else`.
2. They have different fields. 
3. `session_id` is move `UID` type so it is unique.

crates/mpc-protocols/src/mpc_events.rs
erik-3milabs: I don't see how them being from different modules is relevant:

```suggestion
use move_core_types::{ident_str, identifier::IdentStr, language_storage::StructTag};
```

erik-3milabs:
Since I cannot unresolve, I would like a response on this: https://github.com/dwallet-labs/dwallet-network/pull/244#discussion_r1772731298

crates/mpc-protocols/src/mpc_service.rs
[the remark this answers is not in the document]
erik-3milabs: I see that `AwaitingInput` is indeed not a good name. Will start a new comment to address my new feedback.

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// The channel to send message to this instance's message handler thread when this instance is active
+    input_receiver: Option<mpsc::Sender<MPCInput>>,
+    pending_messages: Vec<MPCInput>,
+}
+
+/// Possible status of an MPC session
+/// - Active: The session is currently running, new messages will be forwarded to the session
+/// - Pending: The session is waiting for a slot to become active, when the init received there were already `MAX_ACTIVE_MPC_INSTANCES` active sessions
+/// - Finished: The session has finished, no more messages will be forwarded. The session will be removed from the service after a timeout.
+/// We want to keep it for some time so leftover messages related to it won't be treated as malicious.
```
erik-3milabs, line 28: - what is a "slot"? please be more specific about this
- "when the init received there were already `MAX_ACTIVE_MPC_INSTANCES` active sessions". This is business logic, which is prone to change. Drop it and let the interface speak about that.
- "The session will be removed from the service after a timeout." again, implementation. Perhaps compact that a bit with the first sentence.
- "/// We want to keep it for some time so leftover messages related to it won't be treated as malicious." again, implementation. This comment should be closer to the logic taking care of awaiting removal.

```suggestion
/// Possible statuses of an MPC session:
/// - Active: The session is currently running; new messages will be forwarded to the session.
/// - Pending: The session is awaiting a slot <WHAT IS A SLOT?> to become active.
/// - Finished: The session is finished and pending removal; incoming messages will not be forwarded.
```

erik-3milabs: Please do not accept the above change immediately but update it to deal with the question.

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,117 @@
...
+enum MPCSessionStatus {
+    Active,
+    Pending,
+    Finished,
+}
+
+/// The `MPCService` is responsible for managing MPC instances:
+/// - It keeps track of all MPC instances
+/// - Runs the MPC session for each active instance
+/// - Ensures that the number of active sessions does not go over `MAX_ACTIVE_MPC_INSTANCES` at the same time
```
erik-3milabs, line 39:
```suggestion
/// The `MPCService` is responsible for managing MPC instances:
/// - keeping track of all MPC instances,
/// - executing all active instances, and
/// - (de)activating instances.
```

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,117 @@
...
+pub struct MPCService {
+    mpc_instances: HashMap<ObjectID, MPCInstance>,
+    /// Being used to keep track of the order of the received pending instances, so we'll know which one to launch first.
```
erik-3milabs, line 42: Again: over-documentation.

```suggestion
    /// Used to keep track of the order in which pending instances are received so they are activated in order of arrival.
```
I don't even like this being here, but OK.

crates/mpc-protocols/src/mpc_service.rs
erik-3milabs: Yes, please keep the counter. But we also have to ensure that the counter is not out of sync with the actual number of active instances (which I'm not convinced of yet). Perhaps this should be wrapped in a struct whose sole purpose is keeping track of that?

crates/mpc-protocols/src/mpc_service.rs
erik-3milabs: 1. Reasonable or not, I'm allergic to 99% of hardcoding (outside of tests) :wink: It obscures behavior, which makes the code hard to understand. By giving a constant a name (e.g., by extracting it into a constant), the reader can realize/double-check _why_ this constant is used and verify it.
2. Can we perhaps derive this from existing constants? For example, does a constant exist that captures the number of active validators? We know that the number of messages to an instance should be, at most, that number multiplied by the maximum number of rounds for a protocol.
3. To close: I don't think 100 is reasonable. If you want to scale this project beyond 100 validators, you'll likely overflow this channel from time to time.

crates/mpc-protocols/src/mpc_service.rs
erik-3milabs: I don't want to go against Zeev's wishes, but I think we can resolve this there being a need for returning early.
What do you think of the following code?

```rs
/// Handles a proof initialization event
/// Spawns a new MPC instance if the number of active instances is below the limit
/// Otherwise, adds the instance to the pending queue
fn handle_proof_init_event(&mut self, event: CreatedProofMPCEvent) {
    info!(
        "Received start flow event for session ID {:?}",
        event.session_id
    );

    let mut new_instance = MPCInstance {
        status: MPCSessionStatus::Pending,
        input_receiver: None,
        pending_messages: vec![],
    };
    self.mpc_instances
        .insert(event.session_id.clone().bytes, new_instance);

    // Activate the instance if possible
    if self.active_instances_counter < MAX_ACTIVE_MPC_INSTANCES {
        new_instance.status.set_active();
        new_instance.init_input_receiver();
        self.active_instances_counter += 1;        
    } else {
        self.pending_instances_queue
            .push_back(event.session_id.bytes);
    };

    info!(
        "Added MPCInstance to service for session_id {:?}",
        event.session_id
    );
}
```

crates/pera-core/src/authority.rs
```diff
@@ -1522,6 +1533,28 @@ impl AuthorityState {
...
 
+    async fn handle_mpc_events(
+        &self,
+        inner_temporary_store: &InnerTemporaryStore,
+        effects: &TransactionEffects,
+        epoch_store: &Arc<AuthorityPerEpochStore>,
+    ) -> anyhow::Result<()> {
+        if !self.is_validator(epoch_store) {
+            return Ok(());
+        }
```
erik-3milabs, line 1544: It feels to me like this should've been captured upstream. Is there a reason for (only) putting it here?

crates/pera-core/src/authority.rs
```diff
@@ -1522,6 +1533,28 @@ impl AuthorityState {
...
+            TransactionEffects::V1(effects) => effects.status(),
+            TransactionEffects::V2(effects) => effects.status(),
+        };
+        if status.is_err() {
+            // If the transaction failed, we don't need to handle MPC events.
+            return Ok(());
+        }
+        self.mpc_service
+            .lock()
+            .handle_mpc_events(&inner_temporary_store.events.data)
```
erik-3milabs, line 1555: something to consider

```suggestion
        if status.is_ok() {
            self.mpc_service
                .lock()
                .handle_mpc_events(&inner_temporary_store.events.data)
        } else {
            // If the transaction failed, we don't need to handle MPC events.
            Ok(())
        }
```

erik-3milabs (changes requested):

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,115 @@
...
+use std::collections::{HashMap, VecDeque};
+use tokio::sync::{mpsc, Mutex, RwLock};
+
+const MAX_ACTIVE_MPC_INSTANCES: usize = 100;
+
+pub enum MPCInput {
+    InitEvent(CreatedProofMPCEvent),
+    Message,
+    OutputEvent,
+}
```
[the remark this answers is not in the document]
ItayLevyOfficial: We simply removed the OutputEvent as it's necessary

crates/mpc-protocols/src/mpc_service.rs
ItayLevyOfficial: Your'e right; I created an issue and added a TODO. The reason I didn't fix it right now is that we change the initialization place of the mpc service in a later PR, and the constructor already receives the number of parties.

crates/mpc-protocols/src/mpc_service.rs
ItayLevyOfficial: Right now your'e right. I don't understand how a struct solves the issue; Anyhow, the way to go imo is:
1. When we implement instance completion, make sure we decrease the counter upon a completed instance (we'll have it's completed event).

2. Start an async task after any instance starts to change it's status after some reasonable timeout (10 minutes let's say) to "failed" and decrease the counter. 

Wdyt?

ItayLevyOfficial: I created an issue & added a todo

crates/pera-core/src/authority.rs
ItayLevyOfficial: AFAIK each full node & validator should execute each of the certificates in the consensus output, but only validators should care about signature-MPC related stuff, like those event, so we put it here. This is where Omer put it as well. Where exactly would you want to put it?

crates/mpc-protocols/src/mpc_service.rs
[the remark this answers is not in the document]
erik-3milabs: ... as it is _not_ necessary?

crates/mpc-protocols/src/mpc_service.rs
erik-3milabs: Hmm, perhaps "wrapping [`mpc_instances`] with struct" is not the solution either. We just have to make sure that `mpc_instances` stays private.

crates/pera-core/src/authority.rs
erik-3milabs: I supposed there would be a point upstream that distinguished between full-nodes and validators. I could not tell you where would be a better place to put it. Let's leave it here.

erik-3milabs:
@ItayLevyOfficial, you were too eager to accept [this](https://github.com/dwallet-labs/dwallet-network/pull/244#discussion_r1772746419) proposal: it was not ready to be accepted as such! See the `<What is a slot?>` in [here](https://github.com/dwallet-labs/dwallet-network/pull/244/commits/f9a6d2fc67fe87b8e4cc7b2deb6e5d39a5481c91).

ItayLevyOfficial:
> @ItayLevyOfficial, you were too eager to accept [this](https://github.com/dwallet-labs/dwallet-network/pull/244#discussion_r1772746419) proposal: it was not ready to be accepted as such! See the `<What is a slot?>` in [here](https://github.com/dwallet-labs/dwallet-network/pull/244/commits/f9a6d2fc67fe87b8e4cc7b2deb6e5d39a5481c91).

I replaced it and updated this doc afterward, check the actual code again

ItayLevyOfficial:
> > @ItayLevyOfficial, you were too eager to accept [this](https://github.com/dwallet-labs/dwallet-network/pull/244#discussion_r1772746419) proposal: it was not ready to be accepted as such! See the `<What is a slot?>` in [here](https://github.com/dwallet-labs/dwallet-network/pull/244/commits/f9a6d2fc67fe87b8e4cc7b2deb6e5d39a5481c91).
> 
> I replaced it and updated this doc afterward, check the actual code again

https://github.com/dwallet-labs/dwallet-network/pull/244/commits/5e2bdd53bd23bacb18f2a8aec8d33040389cec44 @erik-3milabs

erik-3milabs (approved):
looks good now!

crates/mpc-protocols/src/mpc_events.rs
```diff
@@ -1,13 +1,33 @@
...
     pub sender: PeraAddress,
 }
+
+pub const PROOF_MODULE_NAME: &IdentStr = ident_str!("proof");
+pub const CREATED_PROOF_STRUCT_NAME: &IdentStr = ident_str!("CreatedProofMPCSessionEvent");
+
+pub trait MPCEvent {
+    fn type_() -> StructTag;
+}
+
```
ycscaly, line 23: What is this trait? why no documentation for something this unobvious?

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+    InitEvent(CreatedProofMPCEvent),
+    Message,
+}
+
+struct MPCInstance {
+    status: MPCSessionStatus,
+    /// The channel to send message to this instance
+    input_receiver: Option<mpsc::Sender<MPCInput>>,
+    pending_messages: Vec<MPCInput>,
+}
```
ycscaly, line 22: please document every struct, and every trait.

ycscaly: also every enum

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl MPCInstance {
+    fn set_active(&mut self) {
+        self.status = MPCSessionStatus::Active;
+        // TODO (#256): Replace hard coded 100 with the number of validators times 10
```
ycscaly, line 27: why times 10?

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+
+pub enum MPCInput {
+    InitEvent(CreatedProofMPCEvent),
+    Message,
+}
+
+struct MPCInstance {
+    status: MPCSessionStatus,
+    /// The channel to send message to this instance
+    input_receiver: Option<mpsc::Sender<MPCInput>>,
```
ycscaly, line 20: why is this an option?

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+                    let _ = input_receiver.send(message);
+                }
+            }
+            MPCSessionStatus::Pending => {
+                self.pending_messages.push(message);
+            }
+            MPCSessionStatus::Finished => {
+                // Do nothing
+            }
+        }
```
ycscaly, line 56: No documentation, no references to the paper.. 
This is a key function. Where is the actual logic of checking that no one sent two messages, or when we reach the amount of messages we need etc?

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+/// Possible statuses of an MPC session:
+/// - Active: The session is currently running; new messages will be forwarded to the session.
+/// - Pending: Too many active instances are running atm; incoming messages will be queued. The session
```
ycscaly, line 62:
```suggestion
/// - Pending: Too many active instances are running at the moment; incoming messages will be queued. The session
```
Please don't use shortwords in documentation

crates/mpc-protocols/src/mpc_service.rs
ycscaly: Or if that is not the purpose of this function, explain what is.

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+    pending_messages: Vec<MPCInput>,
+}
+
+impl MPCInstance {
+    fn set_active(&mut self) {
+        self.status = MPCSessionStatus::Active;
+        // TODO (#256): Replace hard coded 100 with the number of validators times 10
+        let (messages_handler_sender, messages_handler_receiver) = mpsc::channel(100);
+        self.input_receiver = Some(messages_handler_sender);
+        self.spawn_mpc_messages_handler(messages_handler_receiver);
```
ycscaly, line 30: please explain (document) the purpose of these channels

crates/mpc-protocols/src/mpc_service.rs
```diff
@@ -0,0 +1,141 @@
...
+        info!(
+            "Received start flow event for session ID {:?}",
+            event.session_id
+        );
+
+        let mut new_instance = MPCInstance {
+            status: MPCSessionStatus::Pending,
+            input_receiver: None,
+            pending_messages: vec![],
+        };
```
ycscaly, line 121: Implement `Defauolt` for `MPCInstance` and call `default()` here

ycscaly (changes requested):
Missing a lot of documentation

crates/mpc-protocols/src/mpc_events.rs
ItayLevyOfficial: removed it

crates/mpc-protocols/src/mpc_service.rs
ItayLevyOfficial: Because it can be none when the instance is pending. This entire channels logic has been removed in [258](https://github.com/dwallet-labs/dwallet-network/pull/258) anyway

crates/mpc-protocols/src/mpc_service.rs
ItayLevyOfficial: done

crates/mpc-protocols/src/mpc_service.rs
ItayLevyOfficial: Seemed reasonable for me at the time, the entire channels logic has been removed in a later PR anyway as it's not needed

PR #247, opened 2024-09-18
chore: Add TS format & lint CI for `sdk/typescript` folder

ItayLevyOfficial:
Most file changes are line breaks, as Pera is one letter longer then Sui. Changes also include imports reordering, as they are ordered alphabetically and P comes before S.


.eslintrc.js
yaelAbergel, line 1: It seems that you don't need to change this file. You only changed `pera` to `sui`, and the license should be `BSD-3-Clause-Clear`, as Omer did.

yaelAbergel (changes requested):

yaelAbergel (approved):

zeevmoney (approved):

PR #249, opened 2024-09-18
upgrade Sui state proof

alexanderludwig:
features
- adapted state proof logic so we can create several CapWrappers from one Sui transaction 
- fixed interface on Sui package so that several messages can be approved at once (required for BTC)
- adapted TS state proof SDK to work with new signing flow

bugfixes
- removed some unwraps which could have potentially crashed the node
- added missing checks which could have led to funds controlled by the light client being stolen


ycscaly (changes requested):

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -126,29 +127,40 @@ module dwallet_system::sui_state_proof {
...
-        
-        let wrapper = CapWrapper {
-            id: object::new(ctx),
-            cap_id_sui: object::id_from_address(sui_cap_id_address),
-            cap: dwallet_cap,
-        };
+        let sui_cap_id_address_vec = bcs::peel_vec_address(&mut bcs::new(sui_cap_ids_bytes));
+        let dwallet_cap_id_address_vec = bcs::peel_vec_address(&mut bcs::new(dwallet_cap_ids_bytes));
+
+        assert!(vector::length(&sui_cap_id_address_vec) == vector::length(&dwallet_caps), EWrongAmountOfDWalletCaps);
```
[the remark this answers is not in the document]
alexanderludwig: dont need this check anymore as we don't pass a vec of caps anymore

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -164,9 +176,8 @@ module dwallet_system::sui_state_proof {
         
         let (cap_ids_serialised_bytes, messages_serialised_bytes) = sui_state_proof_verify_transaction(committee.committee, checkpoint_summary, checkpoint_contents, transaction, config.approve_event_type_layout, config.package_id );
 
-
-        let messages = bcs::peel_vec_vec_u8(&mut bcs::new(messages_serialised_bytes));
         let cap_ids = bcs::peel_vec_address(&mut bcs::new(cap_ids_serialised_bytes));
```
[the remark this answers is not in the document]
alexanderludwig: technically it could be that in the same txn several dwallet caps on sui requested signing of a message. then it would be wrong to just return a single one. would keep it like this.

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -164,9 +176,8 @@ module dwallet_system::sui_state_proof {
...
+        let messages = bcs::peel_vec_vec_u8(&mut bcs::new(messages_serialised_bytes));
```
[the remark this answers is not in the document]
alexanderludwig: ok

crates/sui-framework/packages/dwallet-system/sources/sui_state_proof.move
```diff
@@ -164,9 +176,8 @@ module dwallet_system::sui_state_proof {
...
```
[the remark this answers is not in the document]
alexanderludwig: did it not 100% clean, addressed this now

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -185,60 +186,74 @@ pub fn sui_state_proof_verify_link_cap(
         return Ok(NativeResult::err(cost, INVALID_TX));
     };
 
-    let tx_events = &transaction.events.as_ref().unwrap().data;
+    let events_digest = transaction.events.as_ref().map(|events| events.digest());
+    // Ensure that the execution digest matches the events digest of the passed transaction
+    if !(events_digest.as_ref() == transaction.effects.events_digest()) {
```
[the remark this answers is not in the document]
alexanderludwig: events_digest is part of the execution digest. see here the mysten implementation: https://github.com/MystenLabs/sui/blob/27b07b7bc214ee97b2af60a5e9a9f210f9e87e24/crates/sui-light-client/src/proof.rs#L178

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -185,60 +186,74 @@ pub fn sui_state_proof_verify_link_cap(
         return Ok(NativeResult::err(cost, INVALID_TX));
     };
```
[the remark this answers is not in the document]
alexanderludwig: should be clear from the Move smart contracts, right?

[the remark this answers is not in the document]
ycscaly: Better to add here too, would ease reviewing and future changes

zeevmoney (commented):
Please explain what is the purpose of this upgrade, write a few lines in the PR description.

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -185,60 +183,74 @@ pub fn sui_state_proof_verify_link_cap(
...
-        .into_iter()
-        .filter_map(|event| {
-            if event.type_.address.to_hex() == package_id_target.to_hex()
-                && event.type_.module.clone().into_string() == "dwallet_cap"
-                && event.type_.name.clone().into_string() == "DWalletNetworkInitCapRequest"
-            {
-                let json_val = SuiJsonValue::from_bcs_bytes(Some(&type_layout), &event.contents)
-                    .unwrap()
-                    .to_json_value();
+    let events_digest = transaction.events.as_ref().map(|events| events.digest());
```
ycscaly, line 186: Seems you are missing this check that the transaction digest is right?
https://github.com/MystenLabs/sui/blob/27b07b7bc214ee97b2af60a5e9a9f210f9e87e24/crates/sui-light-client/src/proof.rs#L158

ycscaly: Also here you are using `.iter()` when thyey use `.enumerate_transactions()`?
https://github.com/MystenLabs/sui/blob/27b07b7bc214ee97b2af60a5e9a9f210f9e87e24/crates/sui-light-client/src/proof.rs#L167

alexanderludwig: Addressed both of your comments.

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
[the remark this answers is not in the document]
alexanderludwig: ok added

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -332,16 +380,22 @@ pub fn sui_state_proof_verify_transaction(
...
                         .collect()
                 });
-
-                if let (Some(cap_id), Some(approve_msg_vec)) = (cap_id, approve_msg_vec) {
-                    Some((cap_id, approve_msg_vec))
+                if let (Some(cap_id), Some(approve_msgs_vec)) = (cap_id, approve_msgs_vec) {
+                    Some((cap_id, approve_msgs_vec))
                 } else {
                     None
                 }
```
ycscaly, line 401:
```suggestion
                cap_id.and_then(|cap_id| approve_msgs_vec.map(|messages| (cap_id, messages)))
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -308,17 +339,34 @@ pub fn sui_state_proof_verify_transaction(
...
-    let tx_events = &transaction.events.as_ref().unwrap().data;
+    let tx_events = match transaction.events.as_ref() {
+        Some(events) => &events.data,
+        None => {
+            return Ok(NativeResult::err(cost, INVALID_TX));
+        }
+    };
 
-    let results: Vec<(SuiAddress, Vec<u8>)> = tx_events
+    let results: Vec<(SuiAddress, Vec<Vec<u8>>)> = tx_events
```
ycscaly, line 369: This will yield a vector in which a single cap_id could appear multiple times with different values. I believe that the way it should be is a `HashMap` (you can always turn it back to a `Vec` before outputting, but this will guarantee uniquness). It doesn't look buggy to me, but it probably is the better approach anyways. If you choose to leave it as-is, just convince yourself there's no bug here.

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
```diff
@@ -308,17 +339,34 @@ pub fn sui_state_proof_verify_transaction(
...
         return Ok(NativeResult::err(cost, INVALID_TX));
     };
 
-    let tx_events = &transaction.events.as_ref().unwrap().data;
+    let tx_events = match transaction.events.as_ref() {
+        Some(events) => &events.data,
+        None => {
+            return Ok(NativeResult::err(cost, INVALID_TX));
+        }
+    };
```
ycscaly, line 367: This code is duplicated in both `sui_state_proof_verify_transaction()` and `sui_state_proof_verify_link_cap()` please DRY it by moving it to a function.

ycscaly (approved):
Good work, please address my comments before merging.

I would have felt much more at ease if we'd have tests that would try to bypass the light-client verification, see that it passed in the previous PR (with the vuln.) and fails here. If you can do it now, it's best. Otherwise, open an issue for that please

integrations/sdk-test/src/index.ts
zeevmoney, line 1: This is broken since @omersadika pushed a new version of the SDK, you are importing in your code the SDK from `npm`, import it locally, no need to use the external version, you are inside the project.

integrations/sdk-test/package.json
```diff
@@ -21,8 +21,8 @@
 	"author": "",
 	"license": "ISC",
 	"dependencies": {
-		"@dwallet-network/dwallet.js": "workspace:*",
```
zeevmoney, line 24: why did you change this? use the workspace version.

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
zeevmoney, line 1: pushed some minor changes here.

zeevmoney (changes requested):
See comments

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
alexanderludwig: As a dWallet user on Sui, I would expect that if I use the same dWallet cap on Sui to sign several messages within the same PTB, I should be able to submit a state proof for all of them. Using a HashMap would only allow one of these transactions to go through, which is less predictable and creates a worse developer experience IMO.

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
alexanderludwig: Introduced a helper fun

integrations/sdk-test/src/index.ts
alexanderludwig: changed the import in package.json which should address this

integrations/sdk-test/src/index.ts
```diff
@@ -135,17 +241,22 @@ async function main() {
 		}
 
 		if (
-			typeof resultFinal.effects?.created == 'object' &&
-			'reference' in resultFinal.effects?.created?.[0]
+			typeof resultFinal1.effects?.created == 'object' &&
+			'reference' in resultFinal1.effects?.created?.[0]
```
zeevmoney, line 245:
```suggestion
			resultFinal1.effects &&
			Array.isArray(resultFinal1.effects.created) &&
			typeof resultFinal1.effects.created[0] === 'object' &&
			'reference' in resultFinal1.effects.created[0]
```

sui-execution/latest/sui-move-natives/src/crypto/sui_state_proof.rs
zeevmoney, line 1: Still a lot of unwraps() here, I will approve it, but deploy a fix for it later.

zeevmoney (approved):

PR #252, opened 2024-09-23
chore (proof-mpc): Move `mpc-protocols` crate into `pera-core`

ItayLevyOfficial:
We needed to do it to be able to inject the validators signature MPC logic into Pera while avoiding import cycles.


crates/pera-analytics-indexer/Cargo.toml
```diff
@@ -51,7 +51,7 @@ pera-json-rpc-types.workspace = true
 pera-package-resolver.workspace = true
 simulacrum.workspace = true
 arrow.workspace = true
-gcp-bigquery-client = "0.18.0"
+gcp-bigquery-client = "0.23.0"
```
ItayLevyOfficial, line 54: Needed to change this as this package fails to compile otherwise

erik-3milabs (approved):
LGTM!

cdesaintguilhem (approved):
Looks clean to me.

ycscaly:
Why is this PR needed? why not change `244` so that the packages would be created correctly instead of creating something just to change it?

ItayLevyOfficial:
> Why is this PR needed? why not change `244` so that the packages would be created correctly instead of creating something just to change it?

Because two people already reviewed it and changing it would discard their comments. Refactor PRs, like this one, are common, and this PR is quite clean

PR #258, opened 2024-09-25
feat (proof-mpc): Implement Proof MPC flow message handling

ItayLevyOfficial:
Waits for threshold messages & advancing until we reach a Finalize result.


erik-3milabs (changes requested):

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,116 +23,190 @@ use std::sync::{Arc, Weak};
...
+                        Ok(advance_result) => match advance_result {
+                            AdvanceResult::Advance((message, new_party)) => {
+                                let message_tx = ConsensusTransaction::new_signature_mpc_message(
+                                    epoch_store.name,
+                                    bcs::to_bytes(&message).unwrap(),
+                                    session_id,
+                                );
+                                let _ = consensus_adapter
+                                    .submit_to_consensus(&[message_tx], &epoch_store)
+                                    .await;
```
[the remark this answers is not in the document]
cdesaintguilhem: I agree, if there is no Result to catch and then ignore, there should need to be a `let _`.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,92 +23,156 @@ use std::sync::{Arc, Weak};
...
 
-    async fn handle_mpc_proof_init_event(
+    fn insert_mpc_message(
+        message: &ProofMPCMessage,
+        mut messages: &mut HashMap<PartyID, (Proof<1, Lang, PhantomData<()>>, Vec<Value>)>,
+        epoch_store: Arc<AuthorityPerEpochStore>,
+    ) -> anyhow::Result<()> {
+        let party_id = authority_name_to_party_id(message.authority, &epoch_store)?;
+        if messages.contains_key(&party_id) {
+            // TODO(#260): Punish an authority that sends multiple messages in the same round
```
cdesaintguilhem, line 140: Yes!

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,92 +23,156 @@ use std::sync::{Arc, Weak};
...
+                epoch_store.name
+            ));
+        }
+
+        match bcs::from_bytes(&message.message) {
+            Ok(message) => {
+                messages.insert(party_id, message);
+                Ok(())
+            }
+            Err(err) => Err(anyhow!("Error deserializing the first message: {:?}", err)),
```
cdesaintguilhem, line 152: Why does this error say "the first" message? Are there others?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -227,18 +295,38 @@ impl SignatureMPCManager {
...
         &mut self,
         message: &[u8],
         authority_name: AuthorityName,
+        session_id: ObjectID,
     ) -> anyhow::Result<()> {
-        // TODO (#235): Implement MPC messages handling
+        let mut instance = self
+            .mpc_instances
+            .get_mut(&session_id)
+            // TODO (#261): Punish a validator that sends a message related to a non-existing mpc instance
```
cdesaintguilhem, line 307: Yep.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,116 +23,190 @@ use std::sync::{Arc, Weak};
...
-            let mut party: ProofParty;
+            let mut messages = HashMap::new();
+
+            let Some(epoch_store) = epoch_store.upgrade() else {
+                // TODO: (#259) Handle the case when the epoch switched in the middle of the MPC instance
+                return;
+            };
+
+            let _ =
+                Self::insert_mpc_message(&first_message, &mut messages, Arc::clone(&epoch_store));
```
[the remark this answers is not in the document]
ItayLevyOfficial: I think the name maybe was misleading - we're simply storing the message here until we reach threshold. I renamed `insert_mpc_message` to `store_message` - is it make more sense now?

crates/pera-core/src/signature_mpc/mpc_manager.rs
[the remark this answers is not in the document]
ItayLevyOfficial: I removed it, although it returns a result we're ignoring rn. I added a todo & an issue.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,116 +23,190 @@ use std::sync::{Arc, Weak};
...
-
-                // TODO (#235): Implement MPC messages handling
             }
         });
     }
 
-    async fn handle_mpc_proof_init_event(
+    fn insert_mpc_message(
+        message: &Vec<u8>,
+        mut messages: &mut HashMap<PartyID, (Proof<1, Lang, PhantomData<()>>, Vec<Value>)>,
```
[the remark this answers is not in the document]
ItayLevyOfficial: it's not using any of it's fields as you can see.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,116 +23,190 @@ use std::sync::{Arc, Weak};
...
+        epoch_store: Arc<AuthorityPerEpochStore>,
+    ) -> anyhow::Result<()> {
+        let party_id = authority_name_to_party_id(epoch_store.name, &epoch_store)?;
```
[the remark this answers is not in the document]
ItayLevyOfficial: You can see it [here](https://github.com/dwallet-labs/dwallet-network/blob/31aa308e707282e4fa82124e8d9cd428e56be8a8/crates/pera-core/src/authority/authority_per_epoch_store.rs#L257), this is the name (public key) of this authority. Using it is actually a common way to get the public key, as it's being done in the same way [here](https://github.com/dwallet-labs/dwallet-network/blob/31aa308e707282e4fa82124e8d9cd428e56be8a8/crates/pera-core/src/epoch/randomness.rs#L482) & [here](https://github.com/dwallet-labs/dwallet-network/blob/31aa308e707282e4fa82124e8d9cd428e56be8a8/crates/pera-core/src/epoch/randomness.rs#L413)

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,116 +23,190 @@ use std::sync::{Arc, Weak};
...
                     consensus_adapter
                         .submit_to_consensus(&vec![message_tx], &epoch_store)
                         .await?;
-                    new_party
+                    Ok((new_party, bcs::to_bytes(&message)?))
                 }
                 AdvanceResult::Finalize(output) => {
-                    return Err(anyhow!("Finalization reached unexpectedly"));
+                    Err(anyhow!("Finalization reached unexpectedly"))
                 }
```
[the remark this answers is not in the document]
ItayLevyOfficial: tried, lmk if this is sufficient. It's a bit challenging as there are plenty of generics.

ItayLevyOfficial:
> .handle_mpc_message(message, *authority, *session_id)

From the paper: 
_Optimisation 3. Perform the quorum satisfaction check each time a new message is saved, instead of waiting until the end of the delivery.
This means the processing of the next round’s message can begin (slightly) sooner, but requires checking that new information coming from the rest of the delivery isn’t missed._

We basically did it, do you want us to remove it?

ItayLevyOfficial:
> The big question for this PR is that the MPC Protocol moves forward as soon as it has received enough messages, which is in contradiction with the spec that says that MPC Instances should wait until the end of the consensus delivery before moving forward if they have enough messages.

From the paper:
Optimisation 3. Perform the quorum satisfaction check each time a new message is saved, instead of waiting until the end of the delivery.
This means the processing of the next round’s message can begin (slightly) sooner, but requires checking that new information coming from the rest of the delivery isn’t missed.

We basically did it, do you want us to remove it?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -17,92 +23,156 @@ use std::sync::{Arc, Weak};
...
 use tokio::sync::{mpsc, Mutex, RwLock};
 use tokio::time::sleep;
-use tracing::{debug, info};
+use tracing::{debug, error, info};
 
-#[derive(Debug)]
-pub enum MPCInput {
-    InitEvent(CreatedProofMPCEvent),
-    Message,
+struct ProofMPCMessage {
```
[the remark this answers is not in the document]
ItayLevyOfficial: Yea the problem is there's also Randomness MPC protocol built into the chain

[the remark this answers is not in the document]
cdesaintguilhem: I see, then maybe `SignatureMPCMessage`?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
+        // This should never happen, as the validator only accepts messages from committee members
+        .ok_or_else(|| {
+            PeraError::InvalidCommittee(
+                "Received a proof MPC message from a validator that is not in the committee"
+                    .to_string(),
+            )
+        })? as PartyID)
 }
 
 struct MPCInstance {
```
cdesaintguilhem, line 54: Can we have some documentation describing what this is and what it does?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
     maurer::language::PublicParameters<{ maurer::SOUND_PROOFS_REPETITIONS }, Lang>;
 
 impl MPCInstance {
-    fn set_active(&mut self, public_parameters: ProofPublicParameters) {
-        self.status = MPCSessionStatus::Active;
-        // TODO (#256): Replace hard coded 100 with the number of validators times 10
-        let (messages_handler_sender, messages_handler_receiver) = mpsc::channel(100);
-        self.input_receiver = Some(messages_handler_sender);
-        self.spawn_mpc_messages_handler(public_parameters, messages_handler_receiver);
+    async fn new(
```
cdesaintguilhem, line 69: This function does a lot of things (such as setting the status to `Active` immediately and generating the first
message); can we have some documentation for the function describing this?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
-                                // This should never happen, as there should be on-chain verification on the init transaction
-                                return;
-                            }
-                        };
-                    }
-                    MPCInput::Message => {
-                        // TODO (#235): Implement MPC messages handling
-                    }
-                }
+    fn advance(&mut self) -> Option<ConsensusTransaction> {
```
cdesaintguilhem, line 100: Can we have some documentation describing this?

cdesaintguilhem: Also, it's weird that "advancing an MPC Instance" returns something of the type `ConsensusTransaction`; can we have
maybe a type alias here?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
+                // TODO (#238): Verify the output and write it to the chain
+                println!("Finalized output: {:?}", output);
+                self.status = MPCSessionStatus::Finished;
+            }
+        }
+        None
+    }
 
-                // TODO (#235): Implement MPC messages handling
+    fn new_signature_mpc_message(&self, message: ProofMessage) -> Option<ConsensusTransaction> {
```
cdesaintguilhem, line 129: Can we have some documentation describing this function?

cdesaintguilhem: Also, same comment about a function called `new_signature_mpc_message` that returns something of the type `ConsensusTransaction`.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
+            return None;
+        };
+        Some(ConsensusTransaction::new_signature_mpc_message(
+            epoch_store.name,
+            bcs::to_bytes(&message).unwrap(),
+            self.session_id.clone(),
+        ))
+    }
+
+    fn store_message(
```
cdesaintguilhem, line 141: Can we have some documentation here?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
-                let message_tx = ConsensusTransaction::new_signature_mpc_message(
-                    epoch_store.name,
-                    bcs::to_bytes(&message)?,
-                );
-                consensus_adapter
-                    .submit_to_consensus(&vec![message_tx], &epoch_store)
+        let party: ProofParty = proof::aggregation::asynchronous::Party::new_proof_round_party(
+            public_parameters,
+            PhantomData,
+            self.mpc_threshold_number_of_parties as PartyID,
```
cdesaintguilhem, line 171: The `threshold` field being of type `PartyID` doesn't make any sense; @ycscaly it looks like this is comming from the
Maurer proof code.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
+        }
+        None
+    }
 
-                // TODO (#235): Implement MPC messages handling
+    fn new_signature_mpc_message(&self, message: ProofMessage) -> Option<ConsensusTransaction> {
+        let Some(epoch_store) = self.epoch_store.upgrade() else {
+            // TODO: (#259) Handle the case when the epoch switched in the middle of the MPC instance
+            return None;
+        };
```
cdesaintguilhem, line 133: This code is repeated in other locations (e.g., lines 175–178), maybe during solving
[#259](https://github.com/dwallet-labs/dwallet-network/issues/259) this could be reformated as a single function so that
it is handled in the same way in different locations.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
+            ));
+        };
+        match advance_result {
+            AdvanceResult::Advance((message, party)) => {
+                let message_tx = self.new_signature_mpc_message(message).ok_or_else(|| {
+                    anyhow!(
+                        "Epoch store switched in the middle of session: {:?}",
+                        self.session_id
+                    )
+                })?;
```
cdesaintguilhem, line 193: Because there's no documentation for `new_signature_mpc_message()`, I can't immediately know that "epoch store switched
in the middle of session" is the only possible reason for it to return `None`, can you justify why that is the case?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -2,163 +2,231 @@ use crate::authority::authority_per_epoch_store::AuthorityPerEpochStore;
...
-            Err(err) => Err(anyhow!("Error while advancing the MPC instance: {:?}", err)),
+            AdvanceResult::Finalize(_) => Err(anyhow!(
+                "Finalization reached unexpectedly: {:?}",
+                self.session_id
+            )),
         }
     }
 
-    async fn handle_message(&mut self, message: MPCInput) {
+    fn handle_message(&mut self, message: ProofMPCMessage) -> PeraResult<()> {
```
cdesaintguilhem, line 206: Documentation please.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -232,49 +300,82 @@ impl SignatureMPCManager {
...
     }
 
-    pub fn handle_mpc_message(
+    pub async fn handle_end_of_delivery(&mut self) -> PeraResult {
+        let txs_to_send: Vec<ConsensusTransaction> = self
+            .mpc_instances
+            .iter_mut()
+            .filter(|(_, instance)| {
+                instance.status == MPCSessionStatus::Active
+                    && instance.pending_messages.len() >= self.mpc_threshold_number_of_parties
```
cdesaintguilhem, line 309: This should be changed to something like `instance.can_advance()` which returns `True` if it has received enough
messages, whatever the definition of "enough" is. Because right now this will break if we forget to change it after
changing to a voting power-based threshold.

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -232,49 +300,82 @@ impl SignatureMPCManager {
...
+            })
+            .collect::<Vec<_>>()
+            .par_iter_mut()
+            .filter_map(|(_, ref mut instance)| instance.advance())
```
cdesaintguilhem, line 313: Two questions about this line:
1. Does it sequentially advance each Instance one-by-one? Why not spin off a new thread for each Instance so that the
   current thread can continue to the next job?
2. The `filter_map` documentation says that it returns "a new iterator with only the items from `Some` results", are we
   not forgetting about any errors by dropping the Instances for which `advance()` returned `None`? Or are these caught
   later or differently?

crates/pera-core/src/signature_mpc/mpc_manager.rs
```diff
@@ -232,49 +300,82 @@ impl SignatureMPCManager {
...
+        let Some(epoch_store) = self.epoch_store.upgrade() else {
+            // TODO: (#259) Handle the case when the epoch switched in the middle of the MPC instance
+            return Ok(());
+        };
+        self.consensus_adapter
+            .submit_to_consensus(&txs_to_send, &epoch_store)
+            .await
+    }
+
+    pub fn handle_message(
```
cdesaintguilhem, line 324: Documentation please.

cdesaintguilhem (changes requested):
Nice to see that advancing is happening at the end of the delivery now, but I'm not sure that this is done in a very efficient manner at the moment.

crates/pera-core/src/signature_mpc/mpc_manager.rs
ItayLevyOfficial: yep, will do in #259

crates/pera-core/src/signature_mpc/mpc_manager.rs
ItayLevyOfficial: Added docs

crates/pera-core/src/signature_mpc/mpc_manager.rs
ItayLevyOfficial: They run in parallel, I'll add a todo here on error handling & punishing malicious actors.

PR #265, opened 2024-09-26
`ethereum-light-client`: add light client interface to typescript sdk

yuvalelmo:


sdk/typescript/src/eth-light-client/eth-light-client.ts
```diff
@@ -0,0 +1,305 @@
...
+	tx.setGasBudget(200000000);
+
+	let result = await client.signAndExecuteTransactionBlock({
+		signer: keypair,
+		transactionBlock: tx,
+		options: { showEffects: true },
+	});
+
+	if (result.effects?.created?.at(0)?.reference.objectId === undefined) {
+		console.log(result.effects);
```
ItayLevyOfficial, line 58: you are logging the effects and then returning the undefined object id? Seems like a weird behavior, maybe throw a descriptive error?

sdk/typescript/src/eth-light-client/eth-light-client.ts
```diff
@@ -0,0 +1,305 @@
...
+			break;
+		}
+		case 'holesky': {
+			checkpoint = '0x12e6b81891d23c90502dbc2354de9cb52afe4ff823ca00fd41d411213c6e7bbb';
+			break;
+		}
+		default: {
+			throw new Error('Invalid network');
+		}
+	}
```
ItayLevyOfficial, line 107: I think having those checkpoints hard coded is problematic, as after some time those checkpoints will become old, and syncing from them will take a long time.

sdk/typescript/src/eth-light-client/eth-light-client.ts
```diff
@@ -0,0 +1,305 @@
...
+		size: stateBytes.length,
+		maxSize: stateBytes.length * 2,
+		allocateSize: stateBytes.length,
+	});
+
+	let contractAddressArrayU8 = ethers.getBytes(contractAddress);
+	let contractAddressBcs = bcs.vector(bcs.u8()).serialize(contractAddressArrayU8);
+
+	const tx = new TransactionBlock();
+	tx.moveCall({
```
ItayLevyOfficial, line 137: IIUC, this light client is keep being update constantly after this in the background by the validators right?

sdk/typescript/src/eth-light-client/eth-light-client.ts
```diff
@@ -0,0 +1,305 @@
...
+ * @param {string} ethDwalletCapId - The ObjectID of the Ethereum dWallet capability, representing the link between the dWallet and Ethereum.
+ * @param {string} message - The Ethereum transaction message to be approved.
+ * @param {string} dWalletID - The ObjectID of the dWallet to which the transaction belongs.
+ * @param {string} latestStateObjectID - The ObjectID of the latest Ethereum state.
+ * @param {string} executionRpc - The Ethereum execution RPC endpoint.
+ * @param {string} consensusRpc - The Ethereum consensus RPC endpoint.
+ * @param {Keypair} keypair - The keypair used to sign the transaction.
+ * @param {DWalletClient} client - The dWallet client instance.
+ */
+export const approveEthereumMessage = async (
```
ItayLevyOfficial, line 192: This function is 100 LOC - maybe break it to two internal functions - one that updates the light client state and one that approves the message?

sdk/typescript/src/eth-light-client/eth-light-client.ts
```diff
@@ -0,0 +1,305 @@
...
+		options: { showEffects: true },
+	});
+
+	if (txResult.effects?.status.status !== 'success') {
+		throw new Error(
+			'Failed to verify Ethereum state. txResult: ' + JSON.stringify(txResult.effects),
+		);
+	}
+
+	// Get the latest Ethereum state after verification.
```
ItayLevyOfficial, line 247: instead of comments that easily become outdated, create small functions with descriptive names

sdk/typescript/src/eth-light-client/rpc.ts
```diff
@@ -0,0 +1,283 @@
+// Copyright (c) dWallet Labs, Ltd.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+import { ssz } from '@lodestar/types';
+import { ethers } from 'ethers';
+
+import { calculateMessageStorageSlot, compareUint8Arrays, keysToSnakeCase } from './utils.js';
+
+const MAX_REQUEST_LIGHT_CLIENT_UPDATES = 128;
```
ItayLevyOfficial, line 9: Add doc

sdk/typescript/src/eth-light-client/rpc.ts
```diff
@@ -0,0 +1,283 @@
...
+ *
+ * **Errors**
+ * Throws an error if the beacon block cannot be retrieved or verified.
+ */
+export async function getBeaconBlockData(
+	consensusRpc: string,
+	finalityUpdateJson: VerifiedFinalityHeader,
+) {
+	let block = await getVerifiedBeaconBlock(consensusRpc, finalityUpdateJson);
+	// Convert keys to snake case
```
ItayLevyOfficial, line 76:
```suggestion
```

sdk/typescript/src/eth-light-client/rpc.ts
```diff
@@ -0,0 +1,283 @@
...
+		if (key === 'body') {
+			return undefined;
+		}
+		return value;
+	});
+
+	let blockBody = block.body;
+	let blockBodyJsonString = JSON.stringify(blockBody, (key, value) => {
+		if (key === 'execution_payload') {
+			return undefined;
```
ItayLevyOfficial, line 89: Doc this behavior in the returns doc section

sdk/typescript/src/eth-light-client/rpc.ts
```diff
@@ -0,0 +1,283 @@
...
+
+	let latestFinalizedBlockNumber = blockExecutionPayload.block_number;
+
+	return {
+		blockJsonString: blockJsonString,
+		blockBodyJsonString: blockBodyJsonString,
+		blockExecutionPayloadJsonString: blockExecutionPayloadJsonString,
+		blockType: finalityUpdateJson['version'],
+		latestFinalizedBlockNumber: latestFinalizedBlockNumber,
+	};
```
ItayLevyOfficial, line 105: Create a `BeaconBlockData` type

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -343,8 +344,12 @@ pub(crate) fn create_initial_eth_state_data(
     // hardcoded hashes for verifying network states
     const MAINNET_STATE_HASH: &str =
         "9fb325c6f66a0f98b57f4b8117c193982c622ee4eb0f6373c84cfc46821091de";
+    const MAINNET_STATE_HASH_TS: &str =
```
ItayLevyOfficial, line 347: Again, can be problematic

ItayLevyOfficial (commented):
Let's sit on Sunday morning so I'll understand the flow a bit better

sdk/typescript/eth-light-client/package.json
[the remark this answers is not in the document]
yuvalelmo: as i understand, this is used to separate the commonJS module and the ESM exports, when using the sdk.

crates/sui/src/ethereum_client_commands.rs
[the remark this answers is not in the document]
yuvalelmo: it is here by mistake, but let's keep it here, it is redundant code that should be deleted.

sdk/typescript/src/eth-light-client/eth-light-client.ts
yuvalelmo: Look again. This is not instead, this is in addition to a function call.
It explains that this is the second time we fetch that same `LatestEthereumState` object as we did before, but we do it again to get the updated inner state of it.
I'll update the comment to something more explainable.

sdk/typescript/src/eth-light-client/eth-light-client.ts
yuvalelmo: No, the state is only updated upon calling `verify_new_state` function, that verifies the updates that are sent by the users and applies them to the state if they are valid.

sdk/typescript/src/eth-light-client/eth-light-client.ts
yuvalelmo: This is correct, but after discussing with Scaly about that, this function should be called once close to genesis of the network, and then all state managing would be handled using the created object.

Yet, another thing I want to add is iterating calls to `verify_new_state` with smaller batches of updates, since at some point the updates array gets too big to pass as a parameter.

sdk/typescript/src/eth-light-client/eth-light-client.ts
yuvalelmo: this is a mistake

sdk/typescript/src/eth-light-client/eth-light-client.ts
```diff
@@ -0,0 +1,305 @@
...
+ * and be managed through the specified smart contract.
+ *
+ * **Arguments**
+ * @param {string} dwalletCapId - The ObjectID of the dWallet capability.
+ * @param {string} latestEthereumStateId - The ObjectID of the latest Ethereum state.
+ * @param {Keypair} keypair - The keypair used to sign the transaction.
+ * @param {DWalletClient} client - The dWallet client instance.
+ */
+export const createEthereumDWallet = async (
+	dwalletCapId: string,
```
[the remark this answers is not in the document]
yuvalelmo: I'll add this as an issue since it would require changes also in the cli command

ItayLevyOfficial (approved):
LGTM!

yaelAbergel (approved):
Please note that I left two comments. Overall, the code looks good!

sdk/typescript/src/eth-light-client/eth-light-client.ts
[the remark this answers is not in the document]
yuvalelmo: I tested in another branch that I forked from `sign-ia-wasm`.
Cannot have the same tests here because it is using the signature mpc

sui-execution/latest/sui-move-natives/src/eth_state_proof.rs
```diff
@@ -124,22 +124,23 @@ pub fn verify_message_proof(
         .map_err(|_| PartialVMError::new(StatusCode::UNKNOWN_STATUS))?;
 
     let contract_address = String::from_utf8(contract_address.clone())
-        .map_err(|_| PartialVMError::new(StatusCode::FAILED_TO_DESERIALIZE_ARGUMENT))?;
+        .unwrap_or(hex::encode(contract_address.clone()));
```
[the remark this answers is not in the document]
yuvalelmo: Had to do this because the input is a little different between the rust cli and the ts sdk, so if it fails to convert to string, then the hex encoding would do the work

ycscaly (approved):

PR #272, opened 2024-10-01
`ethereum-light-client`: update state only when necessary

yuvalelmo:
Updating the state now only happens if the message cannot be verified using the current state.
If the current state includes the message then it won't be updated.


ItayLevyOfficial (commented):
Are you sure this code is working? AFAIK, the code will fail to generate the proof if it tries to create it with an outdated checkpoint. From [here](https://github.com/dwallet-labs/dwallet-network/blob/4572daf620d8921fef4a04711959f26e44c57e75/crates/sui/src/ethereum_client_commands.rs#L422), it seems that when it fails to do so, you simply return the error instead of updating the state and trying again. Am I missing something?

yuvalelmo:
> Are you sure this code is working? AFAIK, the code will fail to generate the proof if it tries to create it with an outdated checkpoint. From [here](https://github.com/dwallet-labs/dwallet-network/blob/4572daf620d8921fef4a04711959f26e44c57e75/crates/sui/src/ethereum_client_commands.rs#L422), it seems that when it fails to do so, you simply return the error instead of updating the state and trying again. Am I missing something?

Yes, I am pretty sure this code works. :)
The code does not generate any proof. 
It makes an RPC call to the ethereum execution rpc and gets the proof. 
If it fails - then something is wrong with either the rpc or the response form - not with the checkpoint.

If you call the proof endpoint with an outdated state it would not fail. you would still get a response with a proof, it would just have some of the fields empty.

ItayLevyOfficial:
> > Are you sure this code is working? AFAIK, the code will fail to generate the proof if it tries to create it with an outdated checkpoint. From [here](https://github.com/dwallet-labs/dwallet-network/blob/4572daf620d8921fef4a04711959f26e44c57e75/crates/sui/src/ethereum_client_commands.rs#L422), it seems that when it fails to do so, you simply return the error instead of updating the state and trying again. Am I missing something?
> 
> Yes, I am pretty sure this code works. :) The code does not generate any proof. It makes an RPC call to the ethereum execution rpc and gets the proof. If it fails - then something is wrong with either the rpc or the response form - not with the checkpoint.
> 
> If you call the proof endpoint with an outdated state it would not fail. you would still get a response with a proof, it would just have some of the fields empty.

gotcha cool - maybe add a comment with this explanation? Was not intuitive for me.

crates/sui/src/ethereum_client_commands.rs
```diff
@@ -376,14 +378,15 @@ pub(crate) async fn eth_approve_message(
         None => return Err(anyhow!("ETH State object ID configuration not found")),
     };
```
ycscaly, line 379: can you get this from `eth_dwallet_cap_id` - by querying the network?

ycscaly (approved):

crates/sui/src/ethereum_client_commands.rs
yuvalelmo: sure.

yaelAbergel (approved):

PR #289, opened 2024-10-14
Binder with the old code version

yuvalelmo:


zeevmoney:
public user share is here: https://github.com/dwallet-labs/dwallet-network/pull/401

zeevmoney:
not needed for now.

PR #316, opened 2024-11-04
feat: Allow users to create dWallets on Pera (DKG)

yaelAbergel:


.cargo/config.toml
ItayLevyOfficial, line 1: need to undo this change

crates/pera-core/src/execution_cache/cache_types.rs
```diff
@@ -113,12 +113,6 @@ impl<I: Iterator> AssertOrdered<I> {
     }
 }
 
-impl<I: IntoIterator> From<I> for AssertOrdered<I::IntoIter> {
```
ItayLevyOfficial, line 116: This is needed because this implementation conflicts with another implementation from the cryptography-private repository

crates/telemetry-subscribers/Cargo.toml
```diff
@@ -26,7 +26,7 @@ tokio = { workspace = true, features = ["full"] }
 futures.workspace = true
 clap.workspace = true
 bytes.workspace = true
-bytes-varint = { version = "1" }
+bytes-varint = { version = "=1.0.3" }
```
ItayLevyOfficial, line 29: This is needed to use this exact version, as the latest version breaks the project compilation atm

sdk/typescript/tsconfig.json
```diff
@@ -5,7 +5,8 @@
 		"module": "CommonJS",
 		"outDir": "dist/cjs",
 		"rootDir": "src",
-		"isolatedModules": true
+		"isolatedModules": true,
+		"sourceMap": true
```
ItayLevyOfficial, line 9: Needed for debugging TS code in the IDE

crates/telemetry-subscribers/Cargo.toml
erik-3milabs: 1. Consider turning this into an issue: this will need to be resolved one way or another, even if this PR is not the place for it.
2. Instead of _fixing_ the version at exactly `=1.0.3`, consider using something like `<1.1.0` instead. This has a couple upsides:
    - If a `1.0.4` patch is released, you will still automatically make use of it,
    - You now know precisely which version is introducing the bug/problem (it is `1.1.0`, not `1.0.4` or something like that), even when new patches are put in place.

cdesaintguilhem (commented):
As decided by @ycscaly, I'm pausing my review of this PR for now. There are still several core modules of the `pera-core` crate that I have _not_ reviewed, but I also made several comments in other places which might be helpful before we resume reviewing.

sdk/typescript/tsconfig.json
zeevmoney: It's impossible to comment on this file.

crates/pera-analytics-indexer/Cargo.toml
```diff
@@ -51,7 +51,7 @@ pera-json-rpc-types.workspace = true
 pera-package-resolver.workspace = true
 simulacrum.workspace = true
 arrow.workspace = true
-gcp-bigquery-client = "0.19.0"
+gcp-bigquery-client = "=0.18.0"
```
zeevmoney, line 54: reason for the downgrade?

crates/pera-core/src/execution_cache/cache_types.rs
zeevmoney: No, there is no point to comment in places where code was removed.

zeevmoney:
Merged in #351

PR #326, opened 2024-11-05
feat: PreSign on Pera 

yaelAbergel:


crates/pera-core/src/dwallet_mpc/bytes_party.rs
```diff
@@ -45,18 +50,6 @@ pub enum AdvanceResult {
     Finalize(Vec<u8>),
 }
 
-/// Holds information about the current MPC session.
```
yaelAbergel, line 48: Moved into pera-types crate

zeevmoney:
Merged in #351

PR #329, opened 2024-11-06
feat: Identify malicious validators in DWallet MPC flow and ignore their future messages

ItayLevyOfficial:


crates/pera-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2561,15 +2568,15 @@ impl AuthorityPerEpochStore {
...
-        let verified_transactions: Vec<_> = transactions
-            .into_iter()
-            .filter_map(|transaction| {
-                self.verify_consensus_transaction(
-                    transaction,
-                    &authority_metrics.skipped_consensus_txns,
-                )
-            })
-            .collect();
+        let mut verified_transactions: Vec<VerifiedSequencedConsensusTransaction> = vec![];
```
ItayLevyOfficial, line 2571: Replaced the filter_map with this less-elegant alternative since async closures are not yet supported in Rust

crates/pera-core/src/consensus_handler.rs
```diff
@@ -371,25 +372,30 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
+                        let Some(dwallet_mpc_manager) = self.epoch_store.dwallet_mpc_manager.get()
+                        else {
+                            error!("MPC manager was not initialized when verifying DWalletMPCOutput output from session {:?}", session_id);
+                            return;
+                        };
+                        let mut dwallet_mpc_manager = dwallet_mpc_manager.lock().await;
+                        if dwallet_mpc_manager.malicious_actors.contains(
+                            &self.committee
+                                .authority_pubkey_by_index(authority_index)
+                                .unwrap(),
```
ItayLevyOfficial, line 384: Safe to unwrap here as it's being done by Sui [here](https://github.com/dwallet-labs/dwallet-network/blob/29929ded135f05578b6ce33b52e6ff5e894d0487/crates/pera-core/src/consensus_handler.rs#L406-L409)

crates/pera-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2395,7 +2395,7 @@ impl AuthorityPerEpochStore {
     /// Important: This function can potentially be called in parallel and you can not rely on order of transactions to perform verification
     /// If this function return an error, transaction is skipped and is not passed to handle_consensus_transaction
     /// This function returns unit error and is responsible for emitting log messages for internal errors
-    fn verify_consensus_transaction(
+    async fn verify_consensus_transaction(
```
ItayLevyOfficial, line 2398: Changed this function to be async so I'll be able to acquire the `dwallet_mpc_manager` lock

crates/pera-core/src/dwallet_mpc/dkg.rs
```diff
@@ -48,27 +54,57 @@ impl FirstDKGBytesParty {
...
+    let malicious_parties: Vec<PartyID> = parsing_results
+        .clone()
+        .into_iter()
+        .filter_map(|result| {
+            if let Err(PeraError::DWalletMPCMaliciousParties(malicious_parties)) = result {
+                Some(malicious_parties)
+            } else {
+                None
+            }
+        })
```
erik-3milabs, line 80: Would this work? Do you agree it is clear to read?
```suggestion
        .filter(|result| result.is_err())
        .map(|err| {
            let Err(PeraError::DWalletMPCMaliciousParties(malicious_parties)) = err;
            malicious_parties
        })
```

crates/pera-core/src/dwallet_mpc/dkg.rs
```diff
@@ -48,27 +54,57 @@ impl FirstDKGBytesParty {
...
+        .flatten()
+        .collect();
+    if !malicious_parties.is_empty() {
+        return Err(PeraError::DWalletMPCMaliciousParties(malicious_parties));
+    }
+    parsing_results.into_iter().collect()
```
erik-3milabs, line 86:
```suggestion
    if malicious_parties.is_empty() {
        parsing_results.into_iter().collect()
    } else {
        Err(PeraError::DWalletMPCMaliciousParties(malicious_parties))
    }

```

crates/pera-core/src/dwallet_mpc/dkg.rs
```diff
@@ -48,27 +54,57 @@ impl FirstDKGBytesParty {
...
+/// Any value that fails to deserialize is considered to be sent by a malicious party.
+/// Returns the deserialized messages or an error including the IDs of the malicious parties.
+pub fn deserialize_mpc_messages<M: DeserializeOwned + Clone>(
+    messages: HashMap<PartyID, Vec<u8>>,
+) -> PeraResult<HashMap<PartyID, M>> {
+    let parsing_results: Vec<PeraResult<(PartyID, _)>> = messages
+        .iter()
+        .map(|(k, v)| {
+            let value =
+                bcs::from_bytes(&v).map_err(|_| PeraError::DWalletMPCMaliciousParties(vec![*k]))?;
```
erik-3milabs, line 67: Also, what is the reason for putting `*k` into a `vec![]` here?

crates/pera-core/src/dwallet_mpc/dkg.rs
```diff
@@ -48,27 +54,57 @@ impl FirstDKGBytesParty {
...
+
         let result = self
             .party
-            .advance(messages, &auxiliary_input, &mut rand_core::OsRng)?;
+            .advance(
+                deserialize_mpc_messages(messages)?,
+                &auxiliary_input,
+                &mut rand_core::OsRng,
+            )
+            .map_err(twopc_error_to_pera_error)?;
```
erik-3milabs, line 107: beautiful! I love it :smile:  :heart_eyes:

crates/pera-core/src/dwallet_mpc/dkg.rs
```diff
@@ -176,22 +212,17 @@ impl BytesParty for SecondDKGBytesParty {
...
-            .into_iter()
-            .map(|(k, v)| {
-                let message = bcs::from_bytes(&v).map_err(|_| twopc_mpc::Error::InvalidParameters);
-                return match message {
-                    Ok(message) => Ok((k, message)),
-                    Err(err) => Err(err),
-                };
-            })
-            .collect::<Result<HashMap<_, _>, _>>()?;
+            bcs::from_bytes(&auxiliary_input).map_err(|_| PeraError::DWalletMPCInvalidUserInput)?;
```
erik-3milabs, line 217: Given that this is the second DKG round, are we sure this is again `InvalidUserInput`?

crates/pera-core/src/dwallet_mpc/dkg.rs
```diff
@@ -176,22 +212,17 @@ impl BytesParty for SecondDKGBytesParty {
...
         let result = self
             .party
-            .advance(messages, &auxiliary_input, &mut rand_core::OsRng)?;
+            .advance(
+                deserialize_mpc_messages(messages)?,
+                &auxiliary_input,
+                &mut rand_core::OsRng,
+            )
+            .map_err(twopc_error_to_pera_error)?;
```
erik-3milabs, line 225: Code duplication from `impl ... for FirstDKGByteParty`. Please see whether you can extract this into a function :)

crates/pera-core/src/dwallet_mpc/mpc_instance.rs
```diff
@@ -112,6 +112,33 @@ impl DWalletMPCInstance {
         Ok(())
     }
 
+    fn restart(&mut self) {
```
erik-3milabs, line 115: missing docstring :smile:

crates/pera-core/src/dwallet_mpc/mpc_instance.rs
```diff
@@ -112,6 +112,33 @@ impl DWalletMPCInstance {
...
+        self.pending_messages.clear();
+        self.status = MPCSessionStatus::FirstExecution;
+        self.party = match &self.session_info.mpc_round {
+            MPCRound::DKGFirst => {
```
erik-3milabs, line 119: I'm confused: we're setting the `status` to `FirstExecution`, but then select the `party` based on the `session_info.mpc_round`? Can you explain what is supposed to happen here?

crates/pera-core/src/dwallet_mpc/mpc_instance.rs
```diff
@@ -144,10 +171,8 @@ impl DWalletMPCInstance {
     ) -> PeraResult<()> {
         let party_id = authority_name_to_party_id(message.authority, &epoch_store)?;
         if self.pending_messages.contains_key(&party_id) {
```
erik-3milabs, line 173: To better self-document the code, consider this:
```suggestion
        let already_received_message_from_party = self.pending_messages.contains_key(&party_id);
        if already_received_message_from_party {
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -122,18 +141,54 @@ impl DWalletMPCManager {
...
+    ) -> PeraResult {
+        let malicious_parties_names = malicious_parties
+            .into_iter()
+            .map(|party_id| {
+                Ok(self
+                    .epoch_store()?
+                    .committee()
+                    .authority_by_index(party_id as u32)
+                    .ok_or(PeraError::InvalidCommittee("".to_string()))?
+                    .clone())
```
erik-3milabs, line 183: why not
```suggestion
                self.epoch_store()?
                    .committee()
                    .authority_by_index(party_id as u32)
                    .ok_or(PeraError::InvalidCommittee("".to_string()))
```
?

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -122,18 +141,54 @@ impl DWalletMPCManager {
...
+            })
+            .collect::<PeraResult<Vec<AuthorityName>>>()?;
+        warn!(
+            "found malicious parties while advancing, {:?}",
+            malicious_parties_names
+        );
```
erik-3milabs, line 189: This is presumptuous. Who is to say this function was called during "advancing"? Consider 
```suggestion
        warn!(
            "flagged the following parties as malicious: {:?}",
            malicious_parties_names
        );
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -122,18 +141,54 @@ impl DWalletMPCManager {
...
+            self.store_malicious_parties_by_indices(malicious_parties)?;
+            return Ok(());
+        };
+        handle_message_response
+    }
+
+    /// Convert the indices of the malicious parties to their addresses and store them
+    /// in the malicious actors set
+    /// New messages from these parties will be ignored
+    fn store_malicious_parties_by_indices(
```
erik-3milabs, line 171: 1. IIRC, there is no function `store_malicious_parties` (i.e., without using indices). Consider renaming to `store_malicious_parties`.

2. "storing" is perhaps not the right phrase here. Consider using `flag_parties_as_malicious` or `report_parties_as_malicious`

crates/pera-core/src/dwallet_mpc/presign.rs
```diff
@@ -56,23 +59,19 @@ impl BytesParty for FirstPresignBytesParty {
...
         let result = self
             .party
-            .advance(messages, &auxiliary_input, &mut rand_core::OsRng)?;
-
+            .advance(
+                deserialize_mpc_messages(messages)?,
+                &auxiliary_input,
+                &mut rand_core::OsRng,
+            )
+            .map_err(twopc_error_to_pera_error)?;
```
erik-3milabs, line 74: code duplication (3/*). Can this be removed?

crates/pera-core/src/dwallet_mpc/presign.rs
```diff
@@ -170,23 +169,19 @@ impl BytesParty for SecondPresignBytesParty {
...
```
erik-3milabs, line 184: code duplication (4/*). Can this be removed?

erik-3milabs (changes requested):
I found some minor things only. Aside from that: nice code :ok_hand: 
I am missing tests, though :thinking: Do we not have to write these here?

ItayLevyOfficial:
> I found some minor things only. Aside from that: nice code 👌 I am missing tests, though 🤔 Do we not have to write these here?

Thanks! I created a separate issue for the tests beforehand, we'll do it later as we want to get a working version asap --> https://github.com/dwallet-labs/dwallet-network/issues/332

crates/pera-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -2437,11 +2437,16 @@ impl AuthorityPerEpochStore {
...
+                        error!("dwallet_mpc_manager is not initialized when verifying dwallet_mpc_message");
+                        return None;
+                    };
                     warn!(
                         "DWalletMPCMessage authority {} does not match its author from consensus {}",
                         authority, transaction.certificate_author_index
                     );
+                    let mut dwallet_mpc_manager = dwallet_mpc_manager.lock().await;
+                    dwallet_mpc_manager.malicious_actors.insert(*authority);
                     return None;
```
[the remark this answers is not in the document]
ItayLevyOfficial: not sure it's necessary as it seems to be the only time it happens in this file, wdyt

crates/pera-core/src/consensus_handler.rs
```diff
@@ -369,21 +369,38 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-                                OutputVerificationResult::Duplicate
+                        let Some(origin_authority) =
+                            self.committee.authority_pubkey_by_index(authority_index)
+                        else {
+                            error!(
+                                "Malicious output from unknown authority index {:?}",
+                                authority_index
+                            );
+                            return;
+                        };
```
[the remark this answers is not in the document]
ItayLevyOfficial: this is the only time it's being called in this file so not sure it's necessary

crates/pera-core/src/dwallet_mpc/dkg.rs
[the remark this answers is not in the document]
ItayLevyOfficial: it will only return from the map, not from the entire function - so we will catch all the bad folks

crates/pera-core/src/dwallet_mpc/dkg.rs
ItayLevyOfficial: idk seems the same for me

crates/pera-core/src/dwallet_mpc/dkg.rs
ItayLevyOfficial: Zeev prefer to return early instead of using if else

crates/pera-core/src/dwallet_mpc/dkg.rs
ItayLevyOfficial: The DWalletMPCMaliciousParties error type receives a vec of malicious parties so it will be able to handle the case we have more than 1 malicious party

crates/pera-core/src/dwallet_mpc/dkg.rs
ItayLevyOfficial: Yeah. Also in this case, the auxiliary input deserialization will only fail if the user provided invalid input

crates/pera-core/src/dwallet_mpc/dkg.rs
ItayLevyOfficial: This is needed as this is the place where the actual types are being specified, and as the code duplication is only this function call it will be challenging to reduce it further

crates/pera-core/src/dwallet_mpc/mpc_instance.rs
ItayLevyOfficial: private functions doesn't require docstring & i think the name is quite descriptive

crates/pera-core/src/dwallet_mpc/mpc_instance.rs
ItayLevyOfficial: We need to restart that instance, which require setting it's status to first execution, so it will be executed again in the next round without any pending messages, and also re-create the first party, as it's being consumed by the advance call

crates/pera-core/src/dwallet_mpc/mpc_instance.rs
ItayLevyOfficial: Not crucial imo

crates/pera-core/src/dwallet_mpc/presign.rs
ItayLevyOfficial: I don't think so, see my previous reply

crates/pera-core/src/dwallet_mpc/presign.rs
ItayLevyOfficial: I don't think so, see my previous reply

zeevmoney:
Merged in #351

PR #345, opened 2024-11-17
chore: Update the `cryptography-private` repository 

ItayLevyOfficial:
This PR includes taking the voting power into account when dealing with the MPC threshold, & mark malicious actors when an MPC advancement was successful despite some actors behaved maliciously.


zeevmoney:
Merged in #351

PR #347, opened 2024-11-19
feat: Sign

yaelAbergel:
We set the voting power to 4 because the `deal_trusted_shares` takes significant time, and the real DKG has not yet been implemented (See #349).


crates/pera-bridge/src/server/mod.rs
yaelAbergel, line 1: The `From` implementation was modified due to conflicting implementations with the cryptography library. This change required adjustments to unrelated code to handle errors in the updated manner.

sdk/typescript/src/dwallet-mpc/mock.ts
yaelAbergel, line 1: These mock values are used to bypass the full DKG and Presign flows.

PR #361, opened 2024-11-27
feat: Support signing a batch of messages in one TX & re-implement message approvals logic

ItayLevyOfficial:
1. before emitting start sign session for each of the messages, i emit a startBatchSignSession event
2. when catching this event, I create a new batchSign object in the mpc manager
3. When receiving a sign output, instead of replacing it with a system TX, i store it in the batched sign session
4. When the batched sign session is completed, we create the system TX to emit the completed sign batch session event


crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -46,6 +49,10 @@ pub struct DWalletMPCManager {
 /// The possible results of verifying an incoming output for an MPC session.
 /// We need to differentiate between a duplicate & a malicious output, as the output can be sent twice by honest parties.
 pub enum OutputVerificationResult {
+    /// When working on a batch, e.g. signing on a batch of messages, we write the output to the chain only once - when the entire batch is ready.
+    ValidWithNewOutput(Vec<u8>),
```
yaelAbergel, line 53: I think `ValidWithOutput` is a enough or maybe `ValidOutput` & `WaitingForOutputs`

crates/pera-core/src/authority.rs
```diff
@@ -1556,14 +1558,35 @@ impl AuthorityState {
...
-            let res = MPCParty::from_event(
-                event,
-                &dwallet_mpc_manager,
-                authority_name_to_party_id(&epoch_store.name, &epoch_store)?,
-            );
-            if let Ok((party, auxiliary_input, session_info)) = res {
-                dwallet_mpc_manager.push_new_mpc_instance(auxiliary_input, party, session_info)?;
-            };
+            if event.type_ == StartBatchedSignEvent::type_() {
+                let deserialized_event: StartBatchedSignEvent = bcs::from_bytes(&event.contents)?;
```
yaelAbergel, line 1562: Maybe move this logic into a function (like the other events)

yaelAbergel (changes requested):
Please also address the quotations I asked in Slack.
I dislike the separation of the signing protocol from the other protocols. Additionally, we should verify whether each signing session (per message) should count as an active instance, or if one batch should be considered a single active instance.

crates/pera-framework/packages/pera-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -259,92 +291,141 @@ module pera_system::dwallet_2pc_mpc_ecdsa_k1 {
...
+    public fun create_sign_output(signed_messages: vector<vector<u8>>, batch_session_id: ID, ctx: &mut TxContext) {
         assert!(tx_context::sender(ctx) == @0x0, ENotSystemAddress);
-        let output = SignOutput {
-            id: object::new(ctx),
-            session_id,
-            dwallet_id,
-            output,
-        };
-        transfer::transfer(output, initiating_user);
+        event::emit(CompletedSignEvent {
```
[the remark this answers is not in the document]
ItayLevyOfficial: becuase it's redundant

sdk/dwallet-mpc-wasm/src/lib.rs
```diff
@@ -14,30 +14,33 @@ pub fn create_dkg_centralized_output(
...
+    ))?)
 }
 
 #[wasm_bindgen]
 pub fn create_sign_centralized_output(
     centralized_party_dkg_output: Vec<u8>,
     presign_first_round_output: Vec<u8>,
     presign_second_round_output: Vec<u8>,
-    message: Vec<u8>,
+    messages: Vec<u8>,
```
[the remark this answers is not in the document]
ItayLevyOfficial: I think ye, not crucial atm

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
ItayLevyOfficial: I prefer the current names

PR #376, opened 2024-12-05
feat: Add logic to lock the next committee before epoch switch

ItayLevyOfficial:
Knowing the next epoch's committee before the current epoch ends is crucial for enabling the chain to re-share its DWallet secret with the committee for the next epoch


crates/pera-json-rpc-types/src/pera_transaction.rs
```diff
@@ -467,6 +468,7 @@ impl Display for PeraTransactionBlockKind {
             Self::DWalletMPCOutput(_) => {
                 writeln!(writer, "Transaction Kind: dwallet mpc Output")?;
             }
+            _ => {}
```
yaelAbergel, line 471: Don't use `_` as it may cause potential debugging difficulties in the future.

```suggestion
           PeraTransactionBlockKind::LockNextCommittee => {
                writeln!(writer, "Transaction Kind: Lock Next Committee")?;
            }
```

crates/pera-types/src/pera_system_state/pera_system_state_inner_v1.rs
```diff
@@ -626,6 +627,7 @@ impl PeraSystemStateTrait for PeraSystemStateInnerV1 {
             system_state_version,
             validators:
                 ValidatorSetV1 {
+                    locked: _,
```
yaelAbergel, line 630: Why are you ignoring the locked value?

crates/pera-types/src/pera_system_state/pera_system_state_inner_v2.rs
```diff
@@ -197,6 +197,7 @@ impl PeraSystemStateTrait for PeraSystemStateInnerV2 {
             system_state_version,
             validators:
                 ValidatorSetV1 {
+                    locked: _,
```
yaelAbergel, line 200: same as v1.

pera-execution/v1/pera-adapter/src/execution_engine.rs
```diff
@@ -671,6 +671,7 @@ mod checked {
             TransactionKind::DWalletMPCOutput(_) => {
                 panic!("DWalletMPCOutput should not exist in v1");
             }
+            TransactionKind::LockNextCommittee(..) => Ok(Mode::empty_results()),
```
yaelAbergel, line 674: Why do you support the lock transaction in this version? And if it’s supported in v1, why not in v2?

crates/pera-types/src/transaction.rs
```diff
@@ -297,7 +297,7 @@ pub enum TransactionKind {
     /// Used to send the output of the dwallet MPC flow to the other validators so they will be able to
     /// create a system transaction that writes it to the chain
     DWalletMPCOutput(DWalletMPCOutput),
-    // .. more transaction types go here
+    LockNextCommittee(EpochId), // .. more transaction types go here
```
yaelAbergel, line 300:
```suggestion
    LockNextCommittee(EpochId), 
    // .. more transaction types go here
```

crates/pera-types/src/pera_system_state/pera_system_state_inner_v1.rs
ItayLevyOfficial: no need to maintain v1

pera-execution/v1/pera-adapter/src/execution_engine.rs
ItayLevyOfficial: removed

yaelAbergel (approved):

cdesaintguilhem (commented):
This code looks good to me, with just one question:

It looks like a validator will send a `LockNextCommittee` transaction every time close_epoch() is called (which is after every checkpoint once the epoch is 24h old); is that intentional?

PR #378, opened 2024-12-08
feat: Implement network DKG before running dWallet MPC sessions

yaelAbergel:


crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -88,10 +96,25 @@ impl DWalletMPCManager {
...
+        // Start the network DKG if this is the first epoch
+        let (status, mpc_instances) = if epoch_id == FIRST_EPOCH_ID {
+            (
+                ManagerStatus::WaitingForNetworkDKGCompletion,
+                NetworkDkg::init(epoch_store.clone())?,
+            )
+        } else {
+            (ManagerStatus::Active, HashMap::new())
+        };
+        let mut outputs_manager = epoch_store.get_dwallet_mpc_outputs_manager().await?;
```
ItayLevyOfficial, line 109: Also if this is not the first epoch we need to fetch the validator secret from the chain here, so maybe add an additional status & a TODO here for that

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -104,13 +127,18 @@ impl DWalletMPCManager {
             weighted_threshold_access_structure,
             weighted_parties,
             batched_sign_sessions: HashMap::new(),
-            outputs_manager: DWalletMPCOutputsManager::new(&epoch_store),
+            outputs_manager: outputs_manager.clone(),
```
ItayLevyOfficial, line 130:
```suggestion
```
No! This should be a new outputs manager, so when the manager handle the outputs one by one, it will catch all the malicious actors.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -104,13 +127,18 @@ impl DWalletMPCManager {
...
+            status,
         };
+        if let Err(err) = manager.handle_end_of_delivery().await {
+            error!("Failed to advance network DKG first round: {:?}", err);
+        }
```
ItayLevyOfficial, line 135: This is redundant as a new checkpoint is being created every 0.5s, and then this function would get called anyway.
```suggestion
```

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -0,0 +1,155 @@
...
+use mpc::WeightedThresholdAccessStructure;
+use pera_types::base_types::{EpochId, ObjectID};
+use pera_types::error::{PeraError, PeraResult};
+use pera_types::messages_dwallet_mpc::{MPCRound, SessionInfo};
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+const NONE_OBJ_ID: ObjectID = ObjectID::from_single_byte(0);
+const SECP256K1_DKG_SESSION_ID: ObjectID = ObjectID::from_single_byte(0);
```
ItayLevyOfficial, line 19: Ye this is problematic, let's go with something like
```suggestion
const SECP256K1_DKG_SESSION_ID: ObjectID = ObjectID::bytes("chain-dkg".to_bytes());
```

ItayLevyOfficial, line 19: Ye this is problematic, let's go with something like
```suggestion
const SECP256K1_DKG_SESSION_ID: ObjectID = ObjectID::bytes("chain-dkg".to_bytes());
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
ItayLevyOfficial, line 109: Also if this is not the first epoch we need to fetch the validator secret from the chain here, so maybe add an additional status & a TODO here for that

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
ItayLevyOfficial, line 130:
```suggestion
outputs_manager: DWalletMPCOutputsManager::new(&epoch_store),
```
No! This should be a new outputs manager, so when the manager handle the outputs one by one, it will catch all the malicious actors.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
ItayLevyOfficial, line 135: This is redundant as a new checkpoint is being created every 0.5s, and then this function would get called anyway.
```suggestion
```

ItayLevyOfficial (approved):
LGTM

PR #405, opened 2024-12-12
fix: Prevent using the same Presign object for two separate sign sessions

ItayLevyOfficial:
Using the same presign object for two separate sign sessions can allow a malicious actor to break the chain's secret key


zeevmoney:
Will be handled in the batch pre-sign #408

PR #422, opened 2024-12-22
feat: Implement a naive state sync

ItayLevyOfficial:
This PR also includes passing a shared object to every TX that starts an MPC flow, so we will receive them in the same order at all validators.

I changed also the behavior of dWallet MPC functions that should never fail to retry instead of returning a result, as a failure can cause the node to go out of sync.


crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -131,6 +144,7 @@ impl DWalletMPCManager {
     }
 
     async fn handle_incoming_channel_message(&mut self, message: DWalletMPCChannelMessage) {
+        self.received_messages += 1;
```
ItayLevyOfficial, line 147: remove before openning pr

crates/pera/src/pera_commands.rs
```diff
@@ -53,7 +53,7 @@ use std::sync::Arc;
 use std::{fs, io};
 use tempfile::tempdir;
 use tracing;
-use tracing::info;
+use tracing::{error, info, warn};
```
ItayLevyOfficial, line 56: undo changes before openning a pr

crates/pera-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -3776,29 +3855,22 @@ impl AuthorityPerEpochStore {
...
-                    .ok_or(DwalletMPCError::MissingDWalletMPCSender)?
-                    .send(DWalletMPCChannelMessage::Output(
-                        output.clone(),
-                        *authority,
-                        session_info.clone(),
-                    ))
-                    .map_err(|err| DwalletMPCError::DWalletMPCSenderSendFailed(err.to_string()))?;
-                Ok(ConsensusCertificateResult::ConsensusMessage)
-            }
+            }) => Ok(ConsensusCertificateResult::ConsensusMessage),
```
ItayLevyOfficial, line 3860: moved to the same place where we try to verify the output before creating the system tx

crates/pera/src/pera_commands.rs
```diff
@@ -806,13 +806,36 @@ async fn start(
             config,
         });
 
-        start_faucet(app_state, CONCURRENCY_LIMIT, &prometheus_registry).await?;
+        tokio::spawn(start_faucet(
```
ItayLevyOfficial, line 809: A bug fix for Sui - needed to allow the test in line 820

PR #432, opened 2024-12-24
dwallet-mpc: encrypt & send dWallet user share and Network DKG

zeevmoney:


pera-execution/latest/pera-adapter/src/execution_engine.rs
```diff
@@ -1130,66 +1130,134 @@ mod checked {
...
+                        )
+                    })?),
                     CallArg::Pure(data.session_info.session_id.to_vec()),
                     CallArg::Pure(batch_session_id.to_vec()),
                 ],
             ),
             MPCRound::PresignSecond(dwallet_id, _first_round_output, batch_session_id) => {
-                let presigns: Vec<(ObjectID, Vec<u8>)> = bcs::from_bytes(&data.output).unwrap();
-                let keys: Vec<ObjectID> = presigns.clone().into_iter().map(|(k, _)| k).collect();
+                // todo(zeev): need to clean this, also why is it different from batch sign?
```
zeevmoney, line 1188: @ItayLevyOfficial we need to talk about this one

crates/pera-node/src/lib.rs
```diff
@@ -1296,9 +1296,10 @@ impl PeraNode {
         }
 
         // Start the dWallet MPC manager on epoch start.
+        // todo(zeev): Itay, we should doc this and explain the diff from Manager.
```
zeevmoney, line 1299: @ItayLevyOfficial we need to talk about this

crates/pera-types/src/messages_dwallet_mpc.rs
```diff
@@ -15,18 +15,17 @@ pub enum MPCRound {
...
-    /// Contains the `PartyID` associated with the decryption share,
-    /// the `ObjectID` of the batched sign session,
-    /// and the hashed message being signed.
-    Sign(ObjectID, Vec<u8>),
+    /// Contains the `ObjectID` of the batched sign session,
+    /// and the hashed messages being signed.
+    Sign(ObjectID, MPCPublicOutput),
     /// A batched sign session, contains the list of messages that are being signed.
-    BatchedSign(Vec<Vec<u8>>),
+    BatchedSign(Vec<MPCMessage>),
```
ItayLevyOfficial, line 28: I think MPCMessage is not apt as this is a message the user provided to sign on, not a message sent between validators

ItayLevyOfficial: not a big fan of this "type to explain field" practice in general, we'll better have structs for each round's inner data with descriptive field names

ItayLevyOfficial (approved):
lgtm

crates/pera-core/src/dwallet_mpc/encrypt_user_share.rs
```diff
@@ -0,0 +1,125 @@
...
+/// Verify the signature on the public share of the DWallet,
+/// and that the public key that signed the public user share is matching the address that initiated this TX.
+fn verify_signatures(
+    verification_data: &StartEncryptedShareVerificationEvent,
+) -> DwalletMPCResult<()> {
+    let dkg_output: <AsyncProtocol as twopc_mpc::dkg::Protocol>::DecentralizedPartyDKGOutput =
+        bcs::from_bytes(&verification_data.dwallet_output)?;
+    let signature =
+        <Ed25519Signature as ToFromBytes>::from_bytes(&verification_data.signed_public_share)
+            .map_err(|e| DwalletMPCError::EncryptedUserShareVerificationFailed)?;
```
ycscaly, line 68: I think we wanted more than just the public user share signed, no?

crates/pera-core/src/dwallet_mpc/encrypt_user_share.rs
```diff
@@ -0,0 +1,125 @@
...
+    let public_key =
+        <Ed25519PublicKey as ToFromBytes>::from_bytes(&verification_data.encryptor_ed25519_pubkey)
```
ycscaly, line 70: Why it is a different field from `verify_encryption_key()`? And why is the logic duplicated snd the function not called?

crates/pera-core/src/dwallet_mpc/encrypt_user_share.rs
```diff
@@ -0,0 +1,125 @@
...
+        )
+        .map_err(|e| DwalletMPCError::EncryptedUserShareVerificationFailed)?;
+    let derived_sui_addr = PeraAddress::from(&public_key);
+    if derived_sui_addr != verification_data.initiator {
+        return Err(DwalletMPCError::EncryptedUserSharePublicKeyDoesNotMatchAddress);
+    }
+    Ok(())
+}
+
+/// Verifies that the given secret encryption is the encryption of the given dwallet's secret share.
```
ycscaly, line 85:
```suggestion
/// Verifies that the given secret encryption is the encryption of the given dwallet's secret share by verifying a ZK proof that assures that.
```

crates/pera-core/src/dwallet_mpc/encrypt_user_share.rs
```diff
@@ -0,0 +1,125 @@
...
+        CiphertextSpaceValue<SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS>,
+    ) = bcs::from_bytes(encrypted_share_and_proof)?;
+    let encrypted_secret_share = CiphertextSpaceGroupElement::new(
+        encrypted_secret_share,
+        &language_public_parameters
+            .encryption_scheme_public_parameters
+            .ciphertext_space_public_parameters(),
+    )?;
+    let public_key_share = secp256k1::GroupElement::new(
+        dkg_output.centralized_party_public_key_share,
```
ycscaly, line 117: Good

crates/pera-core/src/dwallet_mpc/mod.rs
```diff
@@ -65,6 +58,18 @@ pub(crate) fn authority_name_to_party_id(
...
 }
 
+/// Convert a given [`PartyID`] to it's corresponding authority name (address).
+pub(crate) fn party_id_to_authority_name(
+    party_id: PartyID,
+    epoch_store: &AuthorityPerEpochStore,
+) -> DwalletMPCResult<AuthorityName> {
+    Ok(epoch_store
+        .committee()
+        .authority_by_index(party_id as u32 - 1)
```
ycscaly, line 68: Good, just add a comment

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -74,6 +85,12 @@ pub enum DWalletMPCChannelMessage {
     /// This starts when the current epoch time has ended, and it's time to start the
     /// reconfiguration process for the next epoch.
     StartLockNextEpochCommittee,
+    /// A validator's public key and proof for the network DKG protocol.
+    /// Each validator's data is being emitted separately because the proof size is
+    /// almost 250 KB, which is the maximum event size in Sui.
+    /// The manager accumulates the data until it receives such an event for all validators,
+    /// and then it starts the network DKG protocol.
+    ValidatorDataForDKG(ValidatorDataForDWalletSecretShare),
```
ycscaly, line 93: `ForDWalletSecretShare`? This name is confusing. Maybe `ForNetworkDKG`

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -167,9 +171,35 @@ impl DWalletMPCManager {
...
+        let party_id = authority_name_to_party_id(
+            &AuthorityPublicKeyBytes::from_bytes(&data.protocol_pubkey_bytes)
+                .map_err(|e| DwalletMPCError::InvalidPartyPublicKey(e))?,
+            &epoch_store,
+        )?;
+        if self.validators_data_for_network_dkg.contains_key(&party_id) {
+            debug!("Received duplicate data for party_id: {:?}", party_id);
+        } else {
+            self.validators_data_for_network_dkg.insert(party_id, data);
+        }
```
ycscaly, line 199: Probably better to this the opposite way: just save the mappying key by the authority pubkey.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -167,9 +171,35 @@ impl DWalletMPCManager {
...
         }
     }
 
+    fn handle_validator_data_for_dkg(
+        &mut self,
+        data: ValidatorDataForDWalletSecretShare,
+    ) -> DwalletMPCResult<()> {
+        let epoch_store = self.epoch_store()?;
+        let party_id = authority_name_to_party_id(
+            &AuthorityPublicKeyBytes::from_bytes(&data.protocol_pubkey_bytes)
```
ycscaly, line 191: Where do we verify that the sender who signed this message equals the pubkey that is written inside?

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -189,39 +219,67 @@ impl DWalletMPCManager {
...
+    pub(super) fn get_protocol_public_parameters(
+        &self,
+        key_scheme: DWalletMPCNetworkKeyScheme,
+        key_version: u8,
+    ) -> DwalletMPCResult<Vec<u8>> {
+        if let Some(self_decryption_share) = self.epoch_store()?.dwallet_mpc_network_keys.get() {
+            return self_decryption_share.get_protocol_public_parameters(key_scheme, key_version);
+        }
+        Err(DwalletMPCError::TwoPCMPCError(
+            "Decryption share not found".to_string(),
```
ycscaly, line 236: I don't like these names

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -189,39 +219,67 @@ impl DWalletMPCManager {
...
+        ))
+    }
+
+    pub(super) fn get_decryption_public_parameters(
```
ycscaly, line 240:
```suggestion
    pub(super) fn get_decryption_key_share_public_parameters(
```
You might think it helps using these shortened names, but it has the opposite effect of confusing auditors. Please use my original names every where

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -189,39 +219,67 @@ impl DWalletMPCManager {
...
+    ///
+    /// This function accesses the current epoch's store and determines the party ID for the
+    /// authority using its name.
+    /// It then retrieves the corresponding decryption share from
+    /// the node configuration.
+    /// The decryption share is combined with the public parameters
+    /// to build a [`DecryptionKeyShare`].
+    /// If any required data is missing or invalid, an
+    /// appropriate error is returned.
+    fn get_decryption_share(
```
ycscaly, line 263:
```suggestion
    fn get_decryption_key_shares(
```
`decryption_share` means something else entirely

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -240,55 +298,53 @@ impl DWalletMPCManager {
...
+                            self.weighted_threshold_access_structure
+                                .party_to_weight
+                                .get(authority_index)
+                        })
+                        .sum(),
+                    _ => 0,
+                };
+
+                let is_ready = match session.status {
+                    MPCSessionStatus::Active => received_weight as StakeUnit >= threshold,
```
ycscaly, line 317: In previous pr I said to use the access structure method `is_authorized_subset` for this

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -240,55 +298,53 @@ impl DWalletMPCManager {
...
+                        .pending_messages
+                        .get(session.round_number)
+                        .unwrap_or(&HashMap::new())
+                        .keys()
+                        .filter_map(|authority_index| {
+                            self.weighted_threshold_access_structure
+                                .party_to_weight
+                                .get(authority_index)
+                        })
+                        .sum(),
```
ycscaly, line 312: Code duplication

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -240,55 +298,53 @@ impl DWalletMPCManager {
...
+                    _ => false,
+                };
+
+                let is_valid_network_dkg_transaction =
+                    matches!(session.session_info.mpc_round, MPCRound::NetworkDkg(..))
+                        && self.validators_data_for_network_dkg.len()
+                            == self
+                                .weighted_threshold_access_structure
+                                .party_to_weight
+                                .len();
```
ycscaly, line 328: We don't want validators to be able to DOS by preventing a DKG from happening by simply not supplying you with their `data_for_network_dkg`. Instead, you should pad it with default values for every validator that didn't send you this data; my code has validity checks, and will simply report that validator as malicious.

ycscaly: Alternatively, we can change my code to be able to get a partial mapping, or have the values as `Option` and report every missing player as malicious.

crates/pera-core/src/dwallet_mpc/mpc_outputs_verifier.rs
```diff
@@ -60,21 +70,26 @@ impl DWalletMPCOutputsVerifier {
...
+    /// that have voted exceeds or equals the quorum threshold, it returns `true`.
+    /// Otherwise, it returns `false`.
     pub(crate) fn should_lock_committee(&mut self, authority_name: AuthorityName) -> bool {
         self.voted_to_lock_committee.insert(authority_name);
         self.voted_to_lock_committee
             .iter()
-            .map(|voter_name| self.weighted_parties.get(voter_name).unwrap_or(&0))
+            .map(|voter| self.weighted_parties.get(voter).unwrap_or(&0))
             .sum::<StakeUnit>()
             >= self.quorum_threshold
```
ycscaly, line 94: Code duplication, use my method

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -167,9 +171,35 @@ impl DWalletMPCManager {
                     );
                 }
             }
+            DWalletMPCChannelMessage::ValidatorDataForDKG(data) => {
```
ycscaly, line 174: Remind self to think why it should be sent as such

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -143,12 +132,132 @@ impl DWalletMPCSession {
...
+                let public_input = bcs::from_bytes(&self.public_input)?;
+                let result = crate::dwallet_mpc::advance::<DKGSecondParty>(
+                    session_id,
+                    self.party_id,
+                    &self.weighted_threshold_access_structure,
+                    self.pending_messages.clone(),
+                    public_input,
+                    (),
+                )?;
+                if let AsynchronousRoundResult::Finalize { public_output, .. } = &result {
```
ycscaly, line 163: Please... add documentation. You're doing something very non-standard, that goes aside of every other party flow: explain why!

I know this is for encrypted user share, but most wouldn't get it immediately. This needs to be explained, and why it is here, and what is the purpose of this check in the grand scheme of things in relation to that feature.

@ItayLevyOfficial

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -143,12 +132,132 @@ impl DWalletMPCSession {
...
+                    verify_encrypted_share(&StartEncryptedShareVerificationEvent {
+                        dwallet_output: public_output.clone(),
+                        encrypted_secret_share_and_proof: event_data
+                            .encrypted_secret_share_and_proof
+                            .clone(),
+                        encryption_key: event_data.encryption_key.clone(),
+                        encryption_key_id: event_data.encryption_key_id.clone(),
+                        initiator: event_data.initiator.clone(),
+                        signed_public_share: event_data.signed_public_share.clone(),
```
ycscaly, line 172: Missing sig on network pub key share

ycscaly: Probably on its encrypted user share too. Refer to the original issue

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -143,12 +132,132 @@ impl DWalletMPCSession {
...
+                        encryptor_ed25519_pubkey: event_data.encryptor_ed25519_pubkey.clone(),
+
+                        // Fields not relevant for verification; passing empty values.
```
ycscaly, line 175: So why are they there?

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -143,12 +132,132 @@ impl DWalletMPCSession {
...
+                self.pending_messages.clone(),
+                bcs::from_bytes(
+                    &self
+                        .private_input
+                        .clone()
+                        .ok_or(DwalletMPCError::MissingMPCPrivateInput)?,
+                )?,
+            ),
+            MPCRound::EncryptedShareVerification(verification_data) => {
+                match verify_encrypted_share(verification_data) {
```
ycscaly, line 230: I wanted to ask: why is this here twice, once after DKG and once by its own. I know the design so I understood that this is for dWallet transfer where the other is for dWallet creation: but other won't! This must be documented

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -143,12 +132,132 @@ impl DWalletMPCSession {
...
+                verify_encryption_key(verification_data)
+                    .map(|_| AsynchronousRoundResult::Finalize {
+                        public_output: vec![],
+                        private_output: vec![],
+                        malicious_parties: vec![],
+                    })
+                    .map_err(|err| err)
+            }
+            MPCRound::BatchedPresign(..) | MPCRound::BatchedSign(..) => {
+                unreachable!("advance should never be called on a batched session")
```
ycscaly, line 249: Remind self to check this can never be reached even if other validators ir users behave maliciously

dolevmu (commented):
I release this before reviewing everything because I may be busy with 2pcmpc Scaly CRs. I viewed a lot of it.

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -182,31 +289,39 @@ impl DWalletMPCSession {
...
-    fn store_message(&mut self, round: usize, message: &DWalletMPCMessage) -> DwalletMPCResult<()> {
-        let epoch_store = self.epoch_store()?;
-        let party_id = authority_name_to_party_id(&message.authority, &epoch_store)?;
-        if self.pending_messages[round].contains_key(&party_id) {
-            return Err(DwalletMPCError::MaliciousParties(vec![party_id]));
+    /// Every new message received for a session is stored.
+    /// When a threshold of messages is reached, the session advances.
+    fn store_message(&mut self, message: &DWalletMPCMessage) -> DwalletMPCResult<()> {
+        let source_party_id =
+            authority_name_to_party_id(&message.authority, &*self.epoch_store()?)?;
```
ycscaly, line 296: Signature verification happens before this right? We know `message.authority` signed this message?

ycscaly: Point us to where this occurs please

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -182,31 +289,39 @@ impl DWalletMPCSession {
...
+
+        let msg_len = self.pending_messages.len();
+        match self.pending_messages.get_mut(message.round_number) {
```
ycscaly, line 299: Good

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -182,31 +289,39 @@ impl DWalletMPCSession {
...
-        let party_id = authority_name_to_party_id(&message.authority, &epoch_store)?;
-        if self.pending_messages[round].contains_key(&party_id) {
-            return Err(DwalletMPCError::MaliciousParties(vec![party_id]));
+    /// Every new message received for a session is stored.
+    /// When a threshold of messages is reached, the session advances.
+    fn store_message(&mut self, message: &DWalletMPCMessage) -> DwalletMPCResult<()> {
+        let source_party_id =
+            authority_name_to_party_id(&message.authority, &*self.epoch_store()?)?;
+
+        let msg_len = self.pending_messages.len();
```
ycscaly, line 298:
```suggestion
        let current_round = self.pending_messages.len();
```

ycscaly: This is definitely a wrong name: `msg_len` means the length of a message, not of a vector of messages.

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -182,31 +289,39 @@ impl DWalletMPCSession {
...
+        match self.pending_messages.get_mut(message.round_number) {
+            Some(party_to_msg) => {
+                if party_to_msg.contains_key(&source_party_id) {
+                    // Duplicate.
+                    return Err(DwalletMPCError::MaliciousParties(vec![source_party_id]));
+                }
+                party_to_msg.insert(source_party_id, message.message.clone());
+            }
+            // If next round.
+            None if message.round_number == msg_len => {
```
ycscaly, line 308: This line exemplifies how bad of a name it is - look how confusing reading this is.

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -182,31 +289,39 @@ impl DWalletMPCSession {
...
+                return Err(DwalletMPCError::MaliciousParties(vec![source_party_id]));
+            }
         }
-        self.pending_messages[round].insert(party_id, message.message.clone());
         Ok(())
     }
 
     /// Handles a message by either forwarding it to the session
     /// or ignoring it if the session is not active.
     pub(crate) fn handle_message(&mut self, message: &DWalletMPCMessage) -> DwalletMPCResult<()> {
```
ycscaly, line 323: I think this function is useless now, just delete it and call the inner function directly

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -36,23 +48,51 @@ pub struct DwalletMPCNetworkKeyVersions {
...
-    pub fn new(epoch_store: &AuthorityPerEpochStore) -> Self {
-        let decryption_key_share = epoch_store
-            .load_validator_decryption_key_shares_from_system_state()
-            .unwrap_or(HashMap::new());
+    pub fn new(
+        epoch_store: &AuthorityPerEpochStore,
+        weighted_threshold_access_structure: &WeightedThresholdAccessStructure,
+        class_groups_decryption_key: ClassGroupsDecryptionKey,
+    ) -> Self {
+        // Safe to unwrap because the authority name is always present in the epoch store.
```
ycscaly, line 69: When you this kind of thing, it entails us needing to verify such complex assumptions, and I don't see the reason why not to return an error instead.

Think: an `unwrap` that fails throws an exception that halts the entire blockchain. This is a potential critical vulnerability. Auditors cannot ignore this. So we would have to go into Sui, and try to make sure under no condition this is possible - why waste our time on such useless endeavor, when you could have simply returned `Result`?

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -36,23 +48,51 @@ pub struct DwalletMPCNetworkKeyVersions {
...
+        let class_groups_decryption_key =
+            dwallet_classgroups_types::mock_class_groups::mock_cg_private_key().unwrap();
+
         let encryption = epoch_store
             .load_decryption_key_shares_from_system_state()
             .unwrap_or(HashMap::new());
+        let decryption_key_share = Self::validator_decryption_key_shares(
+            &encryption,
+            weighted_threshold_access_structure,
+            class_groups_decryption_key,
```
ycscaly, line 92: Where is this being taken from if the mock class group feature is disabled?

crates/pera-core/src/dwallet_mpc/network_dkg.rs
ycscaly, line 1: All the names used throughout this file are completely false and misleading. This makes the code unreadable. I cannot audit this - please, stop inventing names for cryptographic variables, use their original names, I picked them for a purpose!

ycscaly (changes requested):
I'm stopping the review until suitable names are chosen for variables and types; the code is impossible to review in its current state. Use the names given in the original crypto crates.

Also, there is much missing documentation.

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -27,6 +53,19 @@ enum Hash {
...
 type HashedMessages = Vec<u8>;
 type SignedMessages = Vec<u8>;
+type SecretShareEncryptionProof = EncryptionOfDiscreteLogProofWithoutCtx<
+    SCALAR_LIMBS,
+    SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+    SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+    secp256k1::GroupElement,
+>;
+
+type SpecificEncryptionKey = EncryptionKey<
```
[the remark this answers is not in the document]
zeevmoney: Please suggest a name.

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -27,6 +53,19 @@ enum Hash {
 
 type HashedMessages = Vec<u8>;
 type SignedMessages = Vec<u8>;
+type SecretShareEncryptionProof = EncryptionOfDiscreteLogProofWithoutCtx<
```
ycscaly, line 56:
```suggestion
type EncryptionOfSecretShareProof = EncryptionOfDiscreteLogProofWithoutCtx<
```

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+    match key_scheme {
+        DWalletMPCNetworkKeyScheme::Secp256k1 => {
+            Ok(bcs::to_bytes(&ProtocolPublicParameters::new::<
+                { secp256k1::SCALAR_LIMBS },
+                { FUNDAMENTAL_DISCRIMINANT_LIMBS },
+                { NON_FUNDAMENTAL_DISCRIMINANT_LIMBS },
+                secp256k1::GroupElement,
+            >(bcs::from_bytes(
+                &protocol_public_parameters,
+            )?))?)
```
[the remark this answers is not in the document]
ItayLevyOfficial: @yaelAbergel

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -137,21 +192,29 @@ pub fn create_sign_output(
...
+            let round_result = SignCentralizedParty::advance(
+                (),
+                &bcs::from_bytes(&centralized_party_secret_key_share)?,
+                &centralized_party_public_input,
+                &mut OsRng,
+            )
+            .context("advance() failed on the SignCentralizedParty")?;
 
-            let signed_message = bcs::to_bytes(&sign_message)?;
+            let signed_message = bcs::to_bytes(&round_result.outgoing_message)?;
```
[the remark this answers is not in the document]
zeevmoney: These are the messages (vector) signed by the centralized party, i think it should be named: "centralized_signed_messages"

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+    protocol_public_parameters: Vec<u8>,
+    key_scheme: u8,
+) -> anyhow::Result<Vec<u8>> {
+    let key_scheme = DWalletMPCNetworkKeyScheme::try_from(key_scheme)?;
+
+    match key_scheme {
+        DWalletMPCNetworkKeyScheme::Secp256k1 => {
+            Ok(bcs::to_bytes(&ProtocolPublicParameters::new::<
+                { secp256k1::SCALAR_LIMBS },
+                { FUNDAMENTAL_DISCRIMINANT_LIMBS },
```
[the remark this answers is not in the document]
zeevmoney: Can't be done, it's from @ycscaly code.

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+    bcs::to_bytes(&dkg_output.centralized_party_public_key_share).map_err(Into::into)
+}
+
+/// Encrypts the given secret share to the given encryption key.
+/// Returns a tuple of the encryption key and proof of encryption.
+pub fn encrypt_secret_share_and_prove(
+    secret_share: Vec<u8>,
+    encryption_key: Vec<u8>,
+) -> anyhow::Result<Vec<u8>> {
+    let protocol_public_params = protocol_public_parameters();
```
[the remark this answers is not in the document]
zeevmoney: I don't think we can do someone about it in this code, we have to use these functions from the crypto libs.

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+        protocol_public_params.group_public_parameters.clone(),
+        bcs::from_bytes(&encryption_key)?,
+    );
+    let randomness = class_groups::RandomnessSpaceGroupElement::<
+        { SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS },
+    >::sample(
+        language_public_parameters
+            .encryption_scheme_public_parameters
+            .randomness_space_public_parameters(),
+        &mut OsRng,
```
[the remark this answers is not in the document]
zeevmoney: Yes

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+    Ok(bcs::to_bytes(&(proof, encryption_of_discrete_log.value()))?)
+}
+
+/// Verifies the given secret share matches the given DWallet's DKG output centralized_party_public_key_share.
+pub fn verify_secret_share(secret_share: Vec<u8>, dkg_output: Vec<u8>) -> anyhow::Result<bool> {
+    let expected_public_key = cg_public_share_from_secret_share(secret_share)?;
+    let dkg_output: <AsyncProtocol as twopc_mpc::dkg::Protocol>::DecentralizedPartyDKGOutput =
+        bcs::from_bytes(&dkg_output)?;
+    Ok(dkg_output.centralized_party_public_key_share == expected_public_key.value())
+}
```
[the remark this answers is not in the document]
zeevmoney: This is called by the user, note that we are in the centralized party lib (Typescript->wasm)

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+
+/// Decrypts the given encrypted user share using the given decryption key.
+pub fn decrypt_user_share_inner(
```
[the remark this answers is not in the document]
zeevmoney: This is called by the user, note that we are in the centralized party lib (Typescript->wasm)

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+    )?;
+
+    let decryption_key = bcs::from_bytes(&decryption_key)?;
+    let decryption_key: DecryptionKey<
+        SCALAR_LIMBS,
+        SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+        SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+        secp256k1::GroupElement,
+    > = DecryptionKey::new(decryption_key, &public_parameters)?;
+    let Some(plaintext): Option<<SpecificEncryptionKey as AdditivelyHomomorphicEncryptionKey<SCALAR_LIMBS>>::PlaintextSpaceGroupElement> = decryption_key
```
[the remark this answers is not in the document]
zeevmoney: Not in this case.

│ error[E0282]: type annotations needed
│    --> /Users/zeevmanilovich/Dev/dwallet-network-new-main/crates/dwallet-mpc-centralized-party/src/lib.rs:364:10
│     |
│ 364 |         .into()
│     |          ^^^^
│ ...
│ 368 |     let secret_share_bytes = U256::from(&plaintext.value()).to_be_bytes().to_vec();
│     |                                          --------- type must be known at this point
│     |
│ help: try using a fully qualified path to specify the expected types
│     |
│ 362 ~     let Some(plaintext): Option<_> = <CtOption<group::secp256k1::Scalar> as Into<Option<_>>>::into(decryption_key
│ 363 ~         .decrypt(&ciphertext, &public_parameters))

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -161,3 +224,145 @@ pub fn create_sign_output(
...
+/// Derives a DWallet's public share from a private share.
+fn cg_public_share_from_secret_share(
+    secret_share: Vec<u8>,
+) -> anyhow::Result<group::secp256k1::GroupElement> {
+    let public_parameters = group::secp256k1::group_element::PublicParameters::default();
+    let generator_group_element =
+        group::secp256k1::group_element::GroupElement::generator_from_public_parameters(
+            &public_parameters,
+        )?;
+    Ok(generator_group_element.scale(&Uint::<{ SCALAR_LIMBS }>::from_be_slice(&secret_share)))
```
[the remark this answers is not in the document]
zeevmoney: it's not just bcs, need to change the scale function as I understand, please do a suggestion.

crates/class-groups-constants/src/lib.rs
```diff
@@ -1,18 +1,47 @@
...
 
-pub fn protocol_public_parameters() -> class_groups::ProtocolPublicParameters<
-    { secp256k1::SCALAR_LIMBS },
-    { twopc_mpc::secp256k1::class_groups::DISCRIMINANT_LIMBS },
-    secp256k1::GroupElement,
-> {
+pub fn network_dkg_final_output() -> <Secp256k1Party as mpc::Party>::PublicOutput {
     // Safe to unwrap as we're using a hardcoded constant.
-    let protocol_public_parameters = STANDARD.decode(PROTOCOL_PUBLIC_PARAMETERS).unwrap();
+    let protocol_public_parameters = STANDARD.decode(&NETWORK_DKG_OUTPUT).unwrap();
```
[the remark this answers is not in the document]
ycscaly: rename at least

crates/dwallet-mpc-centralized-party/src/lib.rs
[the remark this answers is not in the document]
ycscaly: why do you need to change it?

crates/dwallet-mpc-centralized-party/src/lib.rs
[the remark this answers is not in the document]
ycscaly: impossible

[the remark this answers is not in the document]
ycscaly: my code uses `SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS`. I dont export `FUNDAMENTAL_DISCRIMINANT_LIMBS` @zeevmoney

crates/pera-framework/packages/pera-system/sources/dwallet.move
```diff
@@ -33,26 +32,41 @@ module pera_system::dwallet {
...
+    /// - `centralized_output`: Centralized public output.
+    /// - `dwallet_mpc_network_key_version`: Version of the MPC network key.
     public struct DWallet<phantom T> has key, store {
         id: UID,
         session_id: ID,
         dwallet_cap_id: ID,
-        // The output of the second DKG round.
-        output: vector<u8>,
+        decentralized_output: vector<u8>,
+        centralized_output: vector<u8>,
```
ycscaly, line 50: Why is this saved here? It definitely shouldn't be.

crates/pera-framework/packages/pera-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -457,81 +507,132 @@ module pera_system::dwallet_2pc_mpc_ecdsa_k1 {
...
     ///
+    /// This function handles the creation of a new DKG session ID and emits an event containing
+    /// all the necessary parameters to continue the DKG process.
     /// ### Parameters
-    /// - `dwallet_cap`: The capability for the associated dWallet.
-    /// - `public_key_share_and_proof`: Public key share and proof from the first round.
-    /// - `first_round_output`: Output from the first DKG round.
-    /// - `first_round_session_id`: Session ID of the first DKG round.
+    /// - `dwallet_cap`: A reference to the `DWalletCap`, representing the capability associated with the dWallet.
+    /// - `centralized_public_key_share_and_proof`: The user (centralized) public key share and proof from the first round.
```
ycscaly, line 517: you can simply say its the user message

crates/pera-framework/packages/pera-system/sources/dwallet_2pc_mpc_ecdsa_k1.move
```diff
@@ -457,81 +507,132 @@ module pera_system::dwallet_2pc_mpc_ecdsa_k1 {
...
             first_round_output: first_round_output.output,
-            public_key_share_and_proof,
+            public_key_share_and_proof: centralized_public_key_share_and_proof,
             dwallet_cap_id: object::id(dwallet_cap),
             first_round_session_id,
             encrypted_secret_share_and_proof,
             encryption_key: get_encryption_key(encryption_key),
             encryption_key_id: object::id(encryption_key),
-            signed_public_share,
+            signed_public_share: signed_centralized_public_output,
```
ycscaly, line 551: No this is definitely wrong, why did you change it?

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -54,7 +54,7 @@ impl ClassGroupsKeyPairAndProof {
     }
 
     pub fn decryption_key(&self) -> ClassGroupsDecryptionKey {
```
ycscaly, line 56: rename it also to `decryption_key_per_crt_prime` so its consistent

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -1,55 +1,133 @@
...
-        ClassGroupsPublicKeyAndProof {
-            encryption_key: self.encryption_key,
-            proof: self.proof.clone(),
-        }
+    pub fn public_bytes(&self) -> ClassGroupsPublicKeyAndProofBytes {
+        // Safe to unwrap because the serialization should never fail.
+        bcs::to_bytes(&self.public()).unwrap()
+    }
+
+    pub fn public(&self) -> ClassGroupsEncryptionKeyAndProof {
```
ycscaly, line 52: What does this name mean

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -1,55 +1,133 @@
...
+
+        let mut rng = rand_chacha::ChaCha20Rng::from_seed(seed);
+        let decryption_key =
+            generate_keypairs_per_crt_prime(setup_parameters_per_crt_prime.clone(), &mut rng)
+                .unwrap();
+
+        let encryption_key_and_proof = generate_knowledge_of_decryption_key_proofs_per_crt_prime(
+            language_public_parameters_per_crt_prime.clone(),
+            decryption_key,
+        )
```
ycscaly, line 104: Where is `rng`

crates/pera-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -143,12 +132,132 @@ impl DWalletMPCSession {
...
+                            .clone(),
+                        encryption_key: event_data.encryption_key.clone(),
+                        encryption_key_id: event_data.encryption_key_id.clone(),
+                        initiator: event_data.initiator.clone(),
+                        signed_public_share: event_data.signed_public_share.clone(),
+                        encryptor_ed25519_pubkey: event_data.encryptor_ed25519_pubkey.clone(),
+
+                        // Fields not relevant for verification; passing empty values.
+                        dwallet_id: ID::new(ObjectID::new([0; 32])),
+                        session_id: ID::new(ObjectID::new([0; 32])),
```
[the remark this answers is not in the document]
ycscaly: What transcript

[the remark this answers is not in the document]
ycscaly: You mean the protocol context? We've talked it through and you said we don't need anything external to the 2pc-mpc in there. Don't get me started

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -36,23 +48,51 @@ pub struct DwalletMPCNetworkKeyVersions {
...
+            return Self::mock_network_dkg(
+                epoch_store,
+                &weighted_threshold_access_structure,
+                party_id,
+            );
+        }
+
+        #[cfg(feature = "mock-class-groups")]
+        // This is used only for development purposes, the key is valid, so it is safe to unwrap.
+        let class_groups_decryption_key =
```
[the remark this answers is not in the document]
ycscaly: +1

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -36,67 +48,248 @@ pub struct DwalletMPCNetworkKeyVersions {
...
+                let dkg_public_output = <Secp256k1Party as CreatePublicOutput>::new(
+                    &share.encryption_key,
+                    &share.reconstructed_commitments_to_sharing,
+                    &share.current_epoch_encryptions_of_shares_per_crt_prime,
+                )?;
+
+                let secret_share = dkg_public_output.default_decryption_key_shares::<
+                    SECP256K1_SCALAR_LIMBS,
+                    FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                    secp256k1::GroupElement,
```
ycscaly, line 217: This of course would make this code non-generic, and would make it hard to integrate `ristretto` support. Generics are there to be used, not ignored.

ycscaly: Write your code using the same generics as I have, and then instantiate it once for `secp256k1` and once for `ristretto`

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -36,67 +48,248 @@ pub struct DwalletMPCNetworkKeyVersions {
 /// Encapsulates all the fields in a single structure for atomic access.
 pub struct DwalletMPCNetworkKeyVersionsInner {
     /// The validators' decryption key shares.
-    pub validator_decryption_key_share: HashMap<DWalletMPCNetworkKey, Vec<Vec<u8>>>,
+    pub validator_decryption_key_share: HashMap<
+        DWalletMPCNetworkKeyScheme,
+        Vec<HashMap<PartyID, <AsyncProtocol as Protocol>::DecryptionKeyShare>>,
```
ycscaly, line 53: When you don't take `AsyncProtocol` to be generic here, you lose the entire purpose of this structure; you make a structure that is supposed to support multiple types of keys, but you use the type of a single specific key for it. How can it work?

ycscaly: Probably you should just save the `SecretKeyShareSizedNumber` here and create the decryption key share when you know the type for it

crates/dwallet-mpc-centralized-party/src/lib.rs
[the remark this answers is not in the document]
zeevmoney: From you code:

image
but you also export `class_groups::SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS`
So I changed it.

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -162,103 +461,219 @@ pub(crate) fn advance_network_dkg(
...
+            class_groups_decryption_key,
         ),
-    }?)
-}
+    }?;
 
-pub(super) fn network_dkg_party(
+    match &res {
+        AsynchronousRoundResult::Finalize {
+            malicious_parties: _,
```
ycscaly, line 490: LINK TO ISSUE

crates/pera-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -162,103 +461,219 @@ pub(crate) fn advance_network_dkg(
...
+            private_output,
+            public_output,
+        } => {
+            // Update the network dWallet MPC keys with the new one.
+            // Todo (#507): Save the output only after it has been verified first by quorum of validators.
```
ycscaly, line 495: Why aren't you doing this already?

ycscaly: This has to be done immediately

ycscaly (commented):
Network DKG looks overall good. Two big things:
1. The code now is not generic and wouldn't work for Ristretto because you are keeping the `DecryptionKeyShare` type in the mapping, which is specific to Secp256k1 as you took it from a given instantiation of `secp256k1::AsyncProtocol`.
You must instead save the raw `SecretKeyShareSizedNumber`.
2. You left as deferred work to only save the output after consesnsus was reached - can't defer this, has to be done.

crates/pera-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1063,15 +1076,22 @@ impl AuthorityPerEpochStore {
     ///   which contains all versions of the encrypted decryption key shares.
     pub(crate) fn load_decryption_key_shares_from_system_state(
         &self,
-    ) -> DwalletMPCResult<HashMap<DWalletMPCNetworkKey, Vec<NetworkDecryptionKeyShares>>> {
-        let decryption_key_shares = (match self.epoch_start_state() {
+    ) -> DwalletMPCResult<HashMap<DWalletMPCNetworkKeyScheme, Vec<NetworkDecryptionKeyShares>>>
+    {
+        let decryption_key_shares = match self.epoch_start_state() {
             EpochStartSystemState::V1(data) => data.get_decryption_key_shares(),
```
ycscaly, line 1082: Is this only for the first epoch?

crates/pera-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1081,26 +1101,21 @@ impl AuthorityPerEpochStore {
...
+    ) -> DwalletMPCResult<HashMap<DWalletMPCNetworkKeyScheme, Vec<Vec<u8>>>> {
         let decryption_key_shares = self.load_decryption_key_shares_from_system_state()?;
-        let party_id = authority_name_to_party_id(&self.name, self)? as usize;
         decryption_key_shares
             .into_iter()
             .map(|(key_type, encryption_shares)| {
                 let shares = encryption_shares
                     .iter()
                     .map(|share| {
                         // TODO (#382): Decrypt the decryption key share
```
ycscaly, line 1115: How is this not implemented, yet the network DKG feature is considered done?

crates/pera-core/src/consensus_handler.rs
```diff
@@ -484,16 +486,62 @@ impl<C: CheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
...
-                                        .create_dwallet_mpc_output_system_tx(session_info, output);
-                                    transactions.push((
-                                        empty_bytes.as_slice(),
-                                        SequencedConsensusTransactionKind::System(transaction),
-                                        consensus_output.leader_author_index(),
-                                    ));
+                                    // Extract the final network DKG transaction parameters from
+                                    // the verified output.
+                                    // We can't preform this within the execution engine,
+                                    // as it requires the class-groups crate from crypto-private lib.
```
ycscaly, line 492: I don't understand

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -189,39 +246,67 @@ impl DWalletMPCManager {
...
-        let public_parameters = self
-            .node_config
-            .dwallet_mpc_decryption_shares_public_parameters
-            .as_ref()
-            .ok_or(DwalletMPCError::MissingDwalletMPCDecryptionSharesPublicParameters)?;
-
-        DecryptionKeyShare::new(party_id, share_value, public_parameters)
-            .map_err(|e| DwalletMPCError::TwoPCMPCError(e.to_string()))
+
+        let decryption_shares = epoch_store
```
ycscaly, line 297:
```suggestion
        let decryption_key_shares = epoch_store
```

ycscaly: @zeevmoney

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -189,39 +246,67 @@ impl DWalletMPCManager {
...
     }
 
-    // todo(zeev): doc this.
-    pub fn get_decryption_share(&self) -> DwalletMPCResult<DecryptionKeyShare> {
+    pub(super) fn get_protocol_public_parameters(
+        &self,
+        key_scheme: DWalletMPCNetworkKeyScheme,
+        key_version: u8,
+    ) -> DwalletMPCResult<Vec<u8>> {
+        if let Some(self_decryption_share) = self.epoch_store()?.dwallet_mpc_network_keys.get() {
```
ycscaly, line 259:
```suggestion
        if let Some(self_decryption_key_share) = self.epoch_store()?.dwallet_mpc_network_keys.get() {
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -189,39 +246,67 @@ impl DWalletMPCManager {
...
+            return self_decryption_share.get_protocol_public_parameters(key_scheme, key_version);
+        }
+        Err(DwalletMPCError::TwoPCMPCError(
+            "Decryption share not found".to_string(),
```
ycscaly, line 263:
```suggestion
            "Decryption key share not found".to_string(),
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -366,30 +422,22 @@ impl DWalletMPCManager {
...
-    pub(crate) fn handle_message(
-        &mut self,
-        message: &[u8],
-        authority_name: AuthorityName,
-        session_id: ObjectID,
-    ) -> DwalletMPCResult<()> {
-        if self.malicious_actors.contains(&authority_name) {
+    pub(crate) fn handle_message(&mut self, message: DWalletMPCMessage) -> DwalletMPCResult<()> {
+        if self.malicious_actors.contains(&message.authority) {
             return Ok(());
```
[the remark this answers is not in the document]
ycscaly: I don't agree with this at all

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -1,55 +1,133 @@
...
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct ClassGroupsPublicKeyAndProof {
-    pub encryption_key: CompactIbqf<{ SECRET_KEY_SHARE_DISCRIMINANT_LIMBS }>,
-    pub proof: MockProof,
-}
+pub type ClassGroupsDecryptionKey = [Uint<{ CRT_FUNDAMENTAL_DISCRIMINANT_LIMBS }>; MAX_PRIMES];
+pub type ClassGroupsEncryptionKeyAndProof = [(
+    CompactIbqf<{ CRT_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS }>,
+    ClassGroupsProof,
+); MAX_PRIMES];
```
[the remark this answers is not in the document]
yaelAbergel: Did you do it?
Notice that I am using different `ClassGroupsProof` types for mocking

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -1,55 +1,133 @@
...
+        });
+
+        return ClassGroupsKeyPairAndProof::new(decryption_key, encryption_key_and_proof);
+    }
+
+    #[cfg(not(feature = "mock-class-groups"))]
+    {
+        let setup_parameters_per_crt_prime =
+            construct_setup_parameters_per_crt_prime(DEFAULT_COMPUTATIONAL_SECURITY_PARAMETER)
+                .unwrap();
```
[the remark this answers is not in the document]
yaelAbergel: When could the Class Groups key generation fail?
In any case, we only use this function before the validator joins the network, and the validator cannot join the network without the Class Groups encryption key and proof.

crates/pera-framework/packages/pera-system/sources/dwallet.move
```diff
@@ -2,73 +2,313 @@
...
         dwallet_mpc_network_key_version: u8,
     }
 
-    /// `DWalletCap` holder controls a corresponding `DWallet`.
+    /// todo(zeev): check why we transfer both public key and address.
+    /// An Encryption key that is used to encrypt a dWallet centralized (user) secret key share.
+    /// Encryption keys facilitate secure data transfer between accounts on the
+    /// dWallet Network by ensuring that sensitive information remains confidential during transmission.
+    /// Each address on the dWallet Network is associated with a unique encryption key.
```
cdesaintguilhem, line 67: Is "dWallet Network" the right name?

crates/pera-framework/packages/pera-system/sources/dwallet.move
cdesaintguilhem, line 1: Why is all the code about the `EncryptionKey` objects in the same module as the `DWallet` objects? Should they not be in their own module?

cdesaintguilhem (commented):
In `dwallet-mpc-centralized-party/lib.rs` there is an encryption key passed into an argument that is called "public parameters". This might be correct from the code's perspective, but it's confusing to read.

sdk/typescript/test/e2e/encrypt-user-share.test.ts
```diff
@@ -0,0 +1,157 @@
...
+			encryptedUserKeyShareAndProofOfEncryption,
+		);
+		expect(decrypted).toEqual(dwalletSecretKeyShare);
+		const is_valid = verify_user_share(
+			decrypted,
+			new Uint8Array(Array.from(Buffer.from(DKGDecentralizedPublicOutput, 'base64'))),
+		);
+		expect(is_valid).toBeTruthy();
+	});
+});
```
dolevmu, line 157: Do we want to run a test also for non-happy flows? Like, a client shares a wrong share, or invalid proof, or a validator states the proof to be invalid although it is valid, to try to frame, etc.? I think currently tests are mostly to see everything is working when everyone is honest, correct me if I am wrong.

The most important thing to test of course is state sync, and that malicious parties cannot stall the network or make a session fail to finish.

crates/pera-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -572,6 +585,17 @@ pub struct AuthorityEpochTables {
     pub(crate) randomness_highest_completed_round: DBMap<u64, RandomnessRound>,
     /// Holds the timestamp of the most recently generated round of randomness.
     pub(crate) randomness_last_round_timestamp: DBMap<u64, TimestampMs>,
+    /// Holds all the DWallet MPC related messages that have been
+    /// received since the beginning of the epoch.
+    /// The key is the consensus round number,
+    /// the value is the dWallet-mpc messages that have been received in that
+    /// round.
```
dolevmu, line 592: Maybe mention why it is okay and we are not afraid a malicious party will bloat the memory of validators.

dolevmu (commented):
I focused on the state sync.

Cargo.toml
```diff
@@ -246,10 +254,11 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(msim)', 'cfg(fail_points)'
...
 [workspace.dependencies]
-mpc = { git = "https://github.com/dwallet-labs/cryptography-private", subdir = "mpc", features = ["parallel"], branch = "new-pera-api" }
-class_groups = { git = "https://github.com/dwallet-labs/cryptography-private", branch = "new-pera-api", subdir = "class-groups", features = ["default", "threshold"] }
-twopc_mpc = { git = "https://github.com/dwallet-labs/cryptography-private", subdir = "2pc-mpc", features = ["secp256k1", "bulletproofs", "paillier", "class_groups", "benchmarking"], branch = "new-pera-api" }
-group = { git = "https://github.com/dwallet-labs/cryptography-private", subdir = "group", branch = "new-pera-api" }
+mpc = { git = "https://github.com/dwallet-labs/cryptography-private", subdir = "mpc", features = ["parallel"], branch = "pera-network-dkg" }
+class_groups = { git = "https://github.com/dwallet-labs/cryptography-private", branch = "pera-network-dkg", subdir = "class-groups", features = ["default", "threshold", "parallel"] }
+twopc_mpc = { git = "https://github.com/dwallet-labs/cryptography-private", subdir = "2pc-mpc", features = ["secp256k1", "bulletproofs", "paillier", "class_groups", "benchmarking"], branch = "pera-network-dkg" }
```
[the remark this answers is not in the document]
yaelAbergel: https://github.com/dwallet-labs/dwallet-network/issues/605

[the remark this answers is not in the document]
yaelAbergel: https://github.com/dwallet-labs/dwallet-network/issues/605

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -1,55 +1,133 @@
...
-    let proof = vec![1u8; 32];
-    // Todo (#369): Uncomment this lines once the class groups
-    // Todo (#369): keygen is ready and doesn't take forever
-    // let mut rng = rand_chacha::ChaCha20Rng::from_seed(seed);
-    // let _ = class_groups::dkg::proof_helpers::generate_secret_share_sized_keypair_and_proof(rng);
-    ClassGroupsKeyPairAndProof {
-        encryption_key,
-        decryption_key,
-        proof,
+    #[cfg(feature = "mock-class-groups")]
```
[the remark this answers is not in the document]
yaelAbergel: Added todo... We must use it for development so that testing the code won't take forever.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
yaelAbergel: Why? Eventually we pass the map as a parameter to the `advance` function of the network DKG and it uses `PartyID`...

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -181,155 +332,270 @@ impl DWalletMPCManager {
...
+
+    /// Spawns all ready MPC cryptographic computations using Rayon.
+    /// If no local CPUs are available, computations will execute as CPUs are freed.
+    pub(crate) fn perform_cryptographic_computation(&mut self) {
+        while self
+            .cryptographic_computations_orchestrator
+            .currently_running_sessions_count
+            < self
+                .cryptographic_computations_orchestrator
+                .available_cores_for_cryptographic_computations
```
cdesaintguilhem, line 467: Couldn't this be written as

```suggestion
        while self
            .cryptographic_computations_orchestrator
            .has_available_core()
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -401,17 +449,11 @@ impl DWalletMPCManager {
     /// Convert the indices of the malicious parties to their addresses and store them
     /// in the malicious actors set.
     /// New messages from these parties will be ignored.
-    /// todo(zeev): clarify if it's restarted on epoch change.
+    /// Restarted for each epoch.
```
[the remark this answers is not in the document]
cdesaintguilhem: I think it's relatively okay since a party can only be malicious once per epoch. If they are malicious many times, then they should be losing a lot of rewards and the market will remove them from the set of validators.

crates/pera-core/src/dwallet_mpc/sign.rs
```diff
@@ -1,11 +1,21 @@
...
 
+/// The index of the last sign cryptographic round.
+/// Needed to be known in advance as this cryptographic step should ideally get computed only once
+/// by the `sign aggregation` protocol.
+pub(crate) const LAST_SIGN_ROUND_INDEX: usize = 1;
+/// The time a validator waits for each other validator to produce the result of the last sign
+/// computation round.
+/// Used to determine how long a validator should wait before running the final step of the sign
+/// MPC flow.
+pub(crate) const SIGN_LAST_ROUND_COMPUTATION_CONSTANT_SECONDS: usize = 15;
```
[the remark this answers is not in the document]
cdesaintguilhem: Indeed using SUI's built-in mechanism would be cleaner.

cdesaintguilhem (commented):
A session that fails to spawn is lost because it was already popped from the list of sessions waiting for computation.

cdesaintguilhem (approved):

cdesaintguilhem (approved):

zeevmoney:
This is an old version, before changing the architecture of the blockchain, it is now `ika`.

PR #451, opened 2024-12-30
fix: Decrease the party ID by 1 when mapping it back to the public key

ItayLevyOfficial:
image

### DIFF
--- a/crates/pera-core/src/dwallet_mpc/mpc_manager.rs
+++ b/crates/pera-core/src/dwallet_mpc/mpc_manager.rs
@@ -397,7 +397,7 @@ impl DWalletMPCManager {
             .map(|party_id| {
                 self.epoch_store()?
                     .committee()
-                    .authority_by_index(*party_id as u32)
+                    .authority_by_index(*party_id as u32 - 1)
                     .cloned()
                     .ok_or(DwalletMPCError::AuthorityIndexNotFound(*party_id))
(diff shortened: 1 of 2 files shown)


yaelAbergel (approved):

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -397,7 +397,7 @@ impl DWalletMPCManager {
             .map(|party_id| {
                 self.epoch_store()?
                     .committee()
-                    .authority_by_index(*party_id as u32)
+                    .authority_by_index(*party_id as u32 - 1)
```
zeevmoney, line 400: extract this to a function that does the conversion: party_id_to_authority_index.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -397,7 +397,7 @@ impl DWalletMPCManager {
...
                     .cloned()
                     .ok_or(DwalletMPCError::AuthorityIndexNotFound(*party_id))
```
zeevmoney, line 402: use the function mentioned above in here also.

zeevmoney (changes requested):
comments

zeevmoney (approved):

PR #469, opened 2025-01-05
feat: Use the network MPC key for the MPC sessions

yaelAbergel:
This is the third PR for #382. We now utilize the real network DKG output, which is used by both the decentralized and centralized parties. The network DKG can be skipped by compiling without the `with-network-dkg` feature. 
Note: We must compile with `mock-class-groups` feature.

To generate the RPC schema, run the main function from `crates/pera-open-rpc/src/generate_json_rpc_spec.rs`, and then run `pnpm tsx scripts/generate.ts` from `sdk/typescript`.


sdk/typescript/src/client/types/generated.ts
yaelAbergel, line 1: generated file

crates/pera-open-rpc/spec/openrpc.json
yaelAbergel, line 1: generated file

zeevmoney:
> This is the third PR for #382. We now utilize the real network DKG output, which is used by both the decentralized and centralized parties. The network DKG can be skipped by compiling without the `with-network-dkg` feature. Note: We must compile with `mock-class-groups` feature.

Add this to the readme we talked about.

zeevmoney (approved):

PR #474, opened 2025-01-06
feat: Support veriosned network MPC keys

yaelAbergel:
This is the final PR for #382. The key version is now retrieved from the emitted event and determined by the dWallet network key version.
Note: We most compile with `mock-class-groups` feature.


crates/pera-types/src/messages_dwallet_mpc.rs
```diff
@@ -17,7 +17,7 @@ pub enum MPCRound {
     /// The first round of the Presign protocol.
     /// Contains the `ObjectId` of the dWallet object,
     /// the DKG decentralized output, and the batch session ID.
```
zeevmoney, line 19:
```suggestion
    /// The first round of the Presign protocol.
    /// Contains the `ObjectId` of the dWallet object,
    /// the DKG decentralized output, the batch session ID,
    /// and the dWallets' network key version.
```

zeevmoney (approved):
see small fix

PR #496, opened 2025-01-09
feat: Allow signing with only the DWallet ID

ItayLevyOfficial:
This is allowed by fetching & decrypting the encrypted secret share from the chain.

This PR also changes the signing behavior, during the encrypt secret share flow, to sign on the entire DKG centralized public output instead on only the public user share, as the entire output is being used in the centralized sign process.

This PR also includes minor refactors, such as moving the TS mocks to a separate file.

Additionally, this PR includes a bug fix to the encryption key flow, that is broken in the `cleanup2` branch.


crates/pera-core/src/dwallet_mpc/mod.rs
ItayLevyOfficial, line 1: seems to be a bugfix for the main branch. We need to have e2e test CI ASAP

PR #508, opened 2025-01-13
chore: Remove `private_output` from MPC session status

yaelAbergel:
To reduce complexity in advancing the asynchronous MPC session, the `advance` function should accept an immutable session reference. This PR is the first step in extracting and isolating the mutability logic from within the `advance` function.


ItayLevyOfficial (approved):
lgtm

zeevmoney (approved):

PR #511, opened 2025-01-13
chore: Make the `MPCSession::advance` function receive immutable reference  

yaelAbergel:
To reduce complexity in advancing the asynchronous MPC session, the advance function should accept an immutable session reference. This PR is the last step in extracting and isolating the mutability logic from within the advance function.

Extract the restart, Failed status change & updated round number.


crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -188,6 +191,23 @@ impl DWalletMPCManager {
...
             }
+            DWalletMPCChannelMessage::MPCSessionFailed(session_id, err) => {
+                if let Some(session) = self.mpc_sessions.get_mut(&session_id) {
+                    match err {
+                        DwalletMPCError::MaliciousParties(malicious_parties) => {
+                            session.restart();
+                            error!(
+                                "MPC session failed with malicious parties: {:?}",
+                                malicious_parties
+                            );
```
ItayLevyOfficial, line 202: why are you not flagging them as malicious?

ItayLevyOfficial (approved):
I don't think this pr should include the message failed consensus message - consider removing it

zeevmoney (approved):

PR #528, opened 2025-01-20
chore: Remove the DWalletMPCManager's redundant channel

ItayLevyOfficial:
As the DWalletMPCManager is now fully synchronous, there is no longer need in that channel.


zeevmoney (approved):

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -143,28 +138,20 @@ impl DWalletMPCManager {
...
-                manager.handle_incoming_channel_message(message).await;
-            }
-        });
-
-        Ok(sender)
+        })
     }
 
-    async fn handle_incoming_channel_message(&mut self, message: DWalletMPCChannelMessage) {
+    pub(crate) async fn handle_dwallet_db_message(&mut self, message: DWalletMPCDBMessage) {
```
zeevmoney, line 144:
```suggestion
    pub(crate) async fn handle_dwallet_mpc_db_message(&mut self, message: DWalletMPCDBMessage) {
```

zeevmoney (approved):

PR #532, opened 2025-01-21
chore: Split the local DWalletMPCMessages DB table to four separate DB tables

ItayLevyOfficial:
Now the manager doesn't need to process all the outputs, and only get updated from the completed sessions table when a session gets completed.

Creating a separate events table is needed to make integrating Omer's code easier.


crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -70,8 +70,8 @@ impl fmt::Display for MPCSessionStatus {
         match self {
             MPCSessionStatus::Pending => write!(f, "Pending"),
             MPCSessionStatus::Active => write!(f, "Active"),
-            MPCSessionStatus::Finished(public_output) => {
-                write!(f, "Finished({:?})", public_output)
+            MPCSessionStatus::Finished => {
+                write!(f, "Finished")
             }
```
yaelAbergel, line 75:
```suggestion
            MPCSessionStatus::Finished => write!(f, "Finished"),
```

zeevmoney (approved):

PR #535, opened 2025-01-22
chore: Read new DWallet MPC data from the database every 100 millis instead of every five seconds

ItayLevyOfficial:
We now manage a small computations task queue at in the manager, and we avoid launching tasks that can't get parallelized in the current moment on the machine, because there are no available CPUs. When a CPU core is freed, right before actually launching the Rayon task, we make sure the computation has not become redundant by the messages that have been received since it has been added to the queue. This allows decreasing the read delay from the local DB without making the state sync slow.


zeevmoney (changes requested):
Add a summary before I review it, since this is not only changing the time, the titile is misleading.

ItayLevyOfficial:
> Add a summary before I review it, since this is not only changing the time, the titile is misleading.

There's more context in the issue

crates/pera-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -1,5 +1,5 @@
 ///! This module contains the DWalletMPCService struct.
```
zeevmoney, line 1: I can't mark it, but it should be like this:
```Rust
//! This module defines the `DWalletMPCService` struct, which handles reading 
//! DWallet MPC messages from the local database at intervals defined by [`READ_INTERVAL_MS`].
//! These messages are then forwarded to the [`crate::dwallet_mpc::mpc_manager::DWalletMPCManager`].
//!
//! The service manages a lightweight computation task queue within the manager
//! to optimize resource usage.
//! Tasks are only launched when parallelization is possible on the current machine,
//! ensuring CPUs are available.
//! Before launching a task on a freed CPU core,
//! the system verifies that the computation remains relevant, 
//! considering messages received since the task was added to the queue.
//! This approach reduces the read delay from the local database
//! while maintaining efficient state synchronization.
```
There is no `///!` (note the 3 slashes).

crates/pera-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -12,7 +12,7 @@ use tokio::sync::{watch, Notify};
 use tracing::error;
 use typed_store::Map;
 
-const READ_INTERVAL_SECS: u64 = 5;
+const READ_INTERVAL_MILLIS: u64 = 100;
```
zeevmoney, line 15:
```suggestion
const READ_INTERVAL_MS: u64 = 100;
```
more common this way.

crates/pera-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -36,7 +36,7 @@ impl DWalletMPCService {
     }
 
     /// Spawns the DWallet MPC service, that
-    /// read DWallet MPC messages from the local DB every [`READ_INTERVAL_SECS`] seconds and forward them to the
+    /// read DWallet MPC messages from the local DB every [`READ_INTERVAL_MILLIS`] seconds and forward them to the
```
zeevmoney, line 39:
```suggestion
    /// read dWallet MPC messages from the local DB every
    /// [`READ_INTERVAL_MS`] seconds and forward them to the
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -29,18 +29,24 @@ use pera_types::digests::Digest;
...
-use tracing::log::{debug, warn};
-use tracing::{error, info};
+use tracing::log::debug;
+use tracing::{error, info, warn};
 use twopc_mpc::sign::Protocol;
 use typed_store::Map;
 
+/// The number of logical cores validator's need to allocate for ongoing, non-cryptographic computations.
+/// Needed to understand how many sessions computations this validator can run in parallel.
+const MACHINE_CORS_FOR_NON_COMPUTATION: usize = 3;
```
zeevmoney, line 48: Why 3?
Why is it hardcoded?
Maybe move it to the general config.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -63,12 +69,21 @@ pub struct DWalletMPCManager {
...
-    /// We need this field to skip already completed rounds, & to use the same messages,
-    /// i.e. those that have been received until the first consensus round in which a quorum have reached.
-    /// We use this first consensus round to know that all validators advance with the exact same messages.
-    ready_to_advance: HashMap<ObjectID, DWalletMPCSession>,
+    /// A map of the pending cryptographic computation sessions.
+    /// This map is needed in order to remove a session that we received a quorum of messages for
+    /// its next round, so running the current, completed round is redundant.
+    pub(crate) pending_computation_map:
+        HashMap<DWalletMPCLocalComputationMetadata, DWalletMPCSession>,
+    /// The order of the [`pending_computation_map`]. Needed to process the computations in the order they were received.
```
zeevmoney, line 77:
```suggestion
    /// The order of the [`pending_computation_map`].
    /// Needed to process the computations in the order they were received.
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -63,12 +69,21 @@ pub struct DWalletMPCManager {
...
     pub(crate) validators_data_for_network_dkg: HashMap<PartyID, ValidatorDataForNetworkDKG>,
-    /// Sessions that are ready to advance when the next [`DWalletMPCDBMessage::PerformCryptographicComputations`]
-    /// message will be received.
-    /// We need this field to skip already completed rounds, & to use the same messages,
-    /// i.e. those that have been received until the first consensus round in which a quorum have reached.
-    /// We use this first consensus round to know that all validators advance with the exact same messages.
-    ready_to_advance: HashMap<ObjectID, DWalletMPCSession>,
+    /// A map of the pending cryptographic computation sessions.
+    /// This map is needed in order to remove a session that we received a quorum of messages for
+    /// its next round, so running the current, completed round is redundant.
```
zeevmoney, line 74:
```suggestion
    /// This map is needed to remove a session that we received a quorum of messages for
    /// its next round, so running the current completed round is redundant.
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -63,12 +69,21 @@ pub struct DWalletMPCManager {
...
+    pub(crate) pending_computation_map:
+        HashMap<DWalletMPCLocalComputationMetadata, DWalletMPCSession>,
+    /// The order of the [`pending_computation_map`]. Needed to process the computations in the order they were received.
+    pending_for_computation_order: VecDeque<DWalletMPCLocalComputationMetadata>,
+    /// The number of currently running cryptographic computations - i.e. computations we called [`rayon::spawn_fifo`] for,
```
zeevmoney, line 79:
```suggestion
    /// The number of currently running cryptographic computations — i.e.,
    /// computations we called [`rayon::spawn_fifo`] for,
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -63,12 +69,21 @@ pub struct DWalletMPCManager {
...
+    /// but we didn't receive a completion message for.
+    pub(crate) currently_running_sessions_count: usize,
```
zeevmoney, line 81: why do we have this and also `active_sessions_counter`?
make the name clearer.
Maybe even move all cryptographic computation elements to a sub-struct.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -63,12 +69,21 @@ pub struct DWalletMPCManager {
...
+    /// The number of logical CPUs available for cryptographic computations on the validator's machine.
```
zeevmoney, line 82:
```suggestion
    /// The number of logical CPUs available for cryptographic 
    /// computations on the validator's machine.
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -63,12 +69,21 @@ pub struct DWalletMPCManager {
...
+    available_cores_for_cryptographic_computations: usize,
+    /// A channel sender to notify the manager that a computation has been completed.
+    /// This is needed to decrease the [`currently_running_sessions_count`] when a computation is done.
```
zeevmoney, line 85:
```suggestion
    /// This is needed to decrease the [`currently_running_sessions_count`]
    /// when a computation is done.
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -118,6 +133,16 @@ impl DWalletMPCManager {
     ) -> DwalletMPCResult<Self> {
         let weighted_threshold_access_structure =
             epoch_store.get_weighted_threshold_access_structure()?;
+        let completed_computation_channel_sender =
+            Self::listen_for_completed_computations(&epoch_store);
+        let available_cores_for_computations: usize = std::thread::available_parallelism()
+            .map_err(|e| DwalletMPCError::FailedToGetAvailableParallelism)?
```
zeevmoney, line 139: pass the error as string.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -386,32 +415,83 @@ impl DWalletMPCManager {
                             DwalletMPCNetworkKeysStatus::Ready(_)
                         ));
                 if is_ready && is_manager_ready {
+                    let session_clone = session.clone();
```
zeevmoney, line 418: why?

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -386,32 +415,83 @@ impl DWalletMPCManager {
...
+            };
+            let Some(session) = self
+                .pending_computation_map
+                .remove(&oldest_computation_metadata)
+            else {
+                return;
+            };
+            self.currently_running_sessions_count += 1;
+            if let Err(err) = self.spawn_session(&session) {
+                error!("Failed to spawn session with err: {:?}", err);
```
zeevmoney, line 463:
```suggestion
                error!("failed to spawn session with error: {:?}", err);
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -386,32 +415,83 @@ impl DWalletMPCManager {
...
-                if let Err(err) = owned_session_for_different_thread.advance(handle) {
-                    error!("Failed to advance session with error: {:?}", err);
-                }
-            });
+        while self.currently_running_sessions_count
+            < self.available_cores_for_cryptographic_computations
+        {
+            let Some(oldest_computation_metadata) = self.pending_for_computation_order.pop_front()
+            else {
+                break;
```
zeevmoney, line 453:
```suggestion
                return;
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -386,32 +415,83 @@ impl DWalletMPCManager {
...
+            .mpc_sessions
+            .get(&session_id)
+            .ok_or(DwalletMPCError::MPCSessionNotFound { session_id })?
+            .status
+            != MPCSessionStatus::Active
+        {
+            return Ok(());
         }
-        self.ready_to_advance.clear();
+        let handle = tokio::runtime::Handle::current();
```
zeevmoney, line 480:
```suggestion
        // Hook the tokio thread pool to the rayon thread pool.
        let handle = tokio::runtime::Handle::current();
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -386,32 +415,83 @@ impl DWalletMPCManager {
...
+        let session = session.clone();
+        let finished_computation_sender = self.completed_computation_channel_sender.clone();
+        rayon::spawn_fifo(move || {
+            if let Err(err) = session.advance(&handle) {
+                error!("Failed to advance session with error: {:?}", err);
```
zeevmoney, line 485:
```suggestion
                error!("failed to advance session with error: {:?}", err);
```

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -386,32 +415,83 @@ impl DWalletMPCManager {
...
+            }
+            if let Err(err) = finished_computation_sender.send(()) {
+                error!(
+                    "Failed to send finished computation message with error: {:?}",
```
zeevmoney, line 489:
```suggestion
                    "failed to send a finished computation message with error: {:?}",
```

sdk/typescript/test/e2e/dwallet-mpc.test.ts
```diff
@@ -64,6 +64,8 @@ describe('Test dWallet MPC', () => {
 	});
 
 	it('should run Presign', async () => {
+		await new Promise((r) => setTimeout(r, 2000));
```
zeevmoney, line 67: use delay() func, later also, no need to comment about the sleep time.

crates/pera-types/src/messages_dwallet_mpc.rs
```diff
@@ -85,6 +85,12 @@ pub struct DWalletMPCOutputMessage {
     pub session_info: SessionInfo,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
+pub struct DWalletMPCLocalComputationMetadata {
```
zeevmoney, line 89: doc

zeevmoney (changes requested):
see comments.

crates/pera-core/src/dwallet_mpc/mpc_manager.rs
ItayLevyOfficial: Because you don't want to include messages from future narwhal rounds in the session.

zeevmoney (approved):

PR #537, opened 2025-01-22
chore: Finish mpc session with threshold honest parties

yaelAbergel:
See #524

Create a MaliciousHandler to report malicious actors based on a quorum of validators.
New consensus message to agree on the malicious actors reported by the quorum in a failed session and re-run the round using advance.


crates/pera-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -424,7 +490,10 @@ impl DWalletMPCManager {
             "[dWallet MPC] Flagged the following parties as malicious: {:?}",
             malicious_parties_names
         );
-        self.malicious_actors.extend(malicious_parties_names);
+
+        malicious_parties_names
+            .into_iter()
+            .for_each(|party| self.malicious_handler.report_malicious_internal(party));
```
ItayLevyOfficial, line 496: it's weird that a public function is being named internal

ItayLevyOfficial (approved):
lgtm

zeevmoney (approved):

PR #547, opened 2025-01-23
chore: Verify the sign output by verifying the signature instead of by waiting for a quorum of sign outputs

ItayLevyOfficial:
This makes the sign flow significantly faster, as instead of waiting for all validators to compute the signature, we have to wait only for the fastest one. This PR does not yet cause only one validator to run the last sign step.

### DIFF
--- a/crates/pera-core/src/dwallet_mpc/mpc_outputs_verifier.rs
+++ b/crates/pera-core/src/dwallet_mpc/mpc_outputs_verifier.rs
@@ -14,7 +14,7 @@ use narwhal_types::Round;
 use pera_types::base_types::{AuthorityName, EpochId, ObjectID};
 use pera_types::committee::StakeUnit;
 use pera_types::dwallet_mpc_error::{DwalletMPCError, DwalletMPCResult};
-use pera_types::messages_dwallet_mpc::{MPCInitProtocolInfo, SessionInfo, SignSessionData};
+use pera_types::messages_dwallet_mpc::{MPCProtocolInitData, SessionInfo, SingleSignSessionData};
 use std::cmp::PartialEq;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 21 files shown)


crates/pera-types/src/messages_dwallet_mpc.rs
```diff
@@ -33,7 +33,7 @@ pub enum MPCInitProtocolInfo {
     /// the Presign first round output, and the batch session ID.
     PresignSecond(ObjectID, MPCPublicOutput, ObjectID),
     /// The first and only round of the Sign protocol.
-    Sign(SignMessageData),
+    Sign(SignSessionData),
```
zeevmoney, line 36:
```suggestion
    Sign(SingleSignSessionData),
```

crates/pera-core/src/dwallet_mpc/mpc_outputs_verifier.rs
```diff
@@ -178,6 +215,56 @@ impl DWalletMPCOutputsVerifier {
...
+                sign_session_data.network_key_version,
+            )?;
+        let protocol_public_parameters: secp256k1::class_groups::ProtocolPublicParameters =
+            bcs::from_bytes(&protocol_public_parameters)?;
+        let dwallet_public_key = secp256k1::GroupElement::new(
+            dkg_output,
+            &protocol_public_parameters.as_ref().group_public_parameters,
+        )
+        .map_err(|_| {
+            DwalletMPCError::ClassGroupsError("Failed to create public key".to_string())
```
zeevmoney, line 249: pass the real error and add the text

zeevmoney (changes requested):
overall looks good, see some comments

zeevmoney (approved):

zeevmoney (approved):

PR #548, opened 2025-01-23
fix: Prevent a single malicious validator from crashing the MPC flow

yaelAbergel:


crates/pera-core/src/dwallet_mpc/mod.rs
```diff
@@ -321,7 +334,16 @@ pub(crate) fn advance<P: AsynchronouslyAdvanceable>(
     public_input: P::PublicInput,
     private_input: P::PrivateInput,
 ) -> DwalletMPCResult<mpc::AsynchronousRoundResult<Vec<u8>, Vec<u8>, Vec<u8>>> {
-    let messages = deserialize_mpc_messages(messages)?;
+    let (messages, deserialization_malicious_parties) = deserialize_mpc_messages(messages);
+    if calculate_total_voting_weight(
+        &access_threshold.party_to_weight,
+        &deserialization_malicious_parties,
+    ) >= access_threshold.threshold as usize
```
ItayLevyOfficial, line 341: This check is not good. Instead of checking if you have a quorum of valid messages after performing the deserialization, you are checking if there are a quorum of validators that sent messages that can't get deserialized.

ItayLevyOfficial (changes requested):
This PR contains a bug, and it doesn't solve the entire issue it points to (See Scaly's comment in the issue). Please point this PR to the specific deserialization issue you are solving here, and fix the bug.

zeevmoney (approved):

zeevmoney (approved):

PR #557, opened 2025-01-26
549 malicious actors in last sign step

ItayLevyOfficial:


crates/pera-core/src/dwallet_mpc/mod.rs
ItayLevyOfficial, line 1: Needed to fix the failing build of the previous PR

PR #568, opened 2025-01-29
feat: Handle malicious behavior in the sign protocol

ItayLevyOfficial:
If any validator sends a malicious report during the sign protocol, the validators don't wait before spawning the last sign round.
This PR also includes a bug fix for the malicious handler, that marked the wrong party as malicious, and a bug fix for the computations orchestrator, that its running computations counter was out of sync.


crates/pera-core/src/dwallet_mpc/cryptographic_computations_orchestrator.rs
```diff
@@ -72,13 +72,6 @@ impl CryptographicComputationsOrchestrator {
     /// computations, the obsolete computation is removed.
     pub(crate) fn insert_ready_sessions(&mut self, sessions: Vec<DWalletMPCSession>) {
         for session in sessions.into_iter() {
-            if session.pending_quorum_for_highest_round_number > 0 {
```
ItayLevyOfficial, line 75: Needed to remove this removal as even if we receive a quorum of messages for the next cryptographic round, we are still don't know if they are valid. If some of them are invalid, maybe this validator's message will be needed

crates/pera-core/src/dwallet_mpc/malicious_handler.rs
```diff
@@ -90,13 +90,13 @@ impl MaliciousHandler {
...
 
-        let total_voting_weight = self.calculate_total_voting_weight(report);
+        let total_voting_weight = self.calculate_total_voting_weight(report.clone());
         let has_reached_quorum = total_voting_weight >= self.quorum_threshold as usize;
         let above_quorum = total_voting_weight > self.quorum_threshold as usize;
         let first_quorum_reached =
             total_voting_weight - authority_voting_weight < self.quorum_threshold as usize;
         if has_reached_quorum && first_quorum_reached {
-            self.malicious_actors.insert(authority);
+            self.malicious_actors.extend(report.malicious_actors);
```
ItayLevyOfficial, line 99: bugfix for the malicious handler - until now the reporting party was getting marked as malicious instead of the malicious party

zeevmoney (approved):

PR #579, opened 2025-02-01
fix: Agree on malicious report before completing a sign session

ItayLevyOfficial:
Until now, a sign flow was considered complete when a valid signature arrived. If the malicious actors reported in the triggering report were not agreed upon as malicious by a quorum by that time, the reporter would be marked as malicious. This behavior is problematic because a valid signature does not imply the absence of a quorum of validators agreeing on the report. This PR changes this behavior and enforces agreement on the malicious report before completing an aggregated sign session.

### DIFF
--- a/crates/pera-core/src/dwallet_mpc/mpc_outputs_verifier.rs
+++ b/crates/pera-core/src/dwallet_mpc/mpc_outputs_verifier.rs
@@ -148,39 +148,7 @@ impl DWalletMPCOutputsVerifier {
             return match Self::verify_signature(&epoch_store, sign_session_data, output) {
                 Ok(res) => {
                     session_output_data.current_result = OutputResult::AlreadyCommitted;
-                    let mpc_manager = epoch_store.get_dwallet_mpc_manager().await;
-                    let session = mpc_manager
-                        .mpc_sessions
-                        .get(&session_info.session_id)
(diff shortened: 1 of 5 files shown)


crates/pera-core/src/dwallet_mpc/mpc_outputs_verifier.rs
```diff
@@ -148,39 +148,7 @@ impl DWalletMPCOutputsVerifier {
...
                     session_output_data.current_result = OutputResult::AlreadyCommitted;
-                    let mpc_manager = epoch_store.get_dwallet_mpc_manager().await;
-                    let session = mpc_manager
-                        .mpc_sessions
-                        .get(&session_info.session_id)
-                        .ok_or(DwalletMPCError::MPCSessionNotFound {
-                            session_id: session_info.session_id,
-                        })?;
                     let mut session_malicious_actors = res.malicious_actors;
-                    if let Some(MPCSessionSpecificState::Sign(sign_state)) =
```
ItayLevyOfficial, line 159: we can remove this check from here, as the architecture changed so that every validator in a sign IA flow sends a malicious report, and the initiating report is being checked versus the agreed report

zeevmoney (approved):
💯

PR #596, opened 2025-02-05
fix: Update the committee constants to use fixed sizes for development

yaelAbergel:
Most of the files in this PR are generated Markdown files that changed when I built the Ika Move packages.  
The only files that actually changed are:  
- `crates/ika-types/src/committee.rs`  
- `crates/ika-move-packages/packages/ika_system/sources/system_v1/committee.move`


ItayLevyOfficial (approved):
lgtm

PR #635, opened 2025-02-12
chore: Refactor Class Groups key mocking 

yaelAbergel:
This PR refactors the handling of the class groups key pair and proof, which is generated for each validator and used in the network's DKG protocol.
Previously, class groups keys were mocked in multiple places, leading to overlooked issues related to validator additions to the committee and certain Sui transaction restrictions.

This update removes the redundant mocks and ensures that the class groups key is only mocked once during key generation within the swarm config creation. 

The new mocking also fixes the memory access issue that used to occur when deserializing the mocked key in a dev mode build.


crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -73,48 +63,25 @@ impl ClassGroupsKeyPairAndProof {
 pub fn generate_class_groups_keypair_and_proof_from_seed(
```
ItayLevyOfficial, line 63: update docs

ItayLevyOfficial (approved):
lgtm

zeevmoney (approved):

PR #637, opened 2025-02-12
chore: Save the Ika start configuration to a config file & use it from TS

ItayLevyOfficial:
Instead of manually updating the TypeScript each time the Ika contracts are redeployed, the Rust deployment code now writes the updated config to a JSON file, which is used by the TypeScript tests.


crates/ika-swarm-config/src/sui_client.rs
```diff
@@ -169,6 +179,14 @@ pub async fn init_ika_on_sui(
     .await?;
 
     println!("Running `init::initialize` done: system_id: {system_id}");
+    let ika_config = IkaConfig {
+        ika_package_id,
+        ika_system_package_id,
+        ika_system_obj_id: system_id,
+    };
+    let mut file = File::create("ika_config.json")?;
+    let json = serde_json::to_string_pretty(&ika_config).unwrap();
```
zeevmoney, line 188: don't unwrap()

zeevmoney: also remove the commented lines in this file please

sdk/typescript/src/dwallet-mpc/globals.ts
```diff
@@ -1,33 +1,34 @@
...
-export const IKA_SYSTEM_OBJ_ID =
-	'0x3eff62e4dfcbca5f92e5f7241041db2bfc0a0a64e15f047238805e3e9c15debe';
-export const IKA_COIN_OBJECT_PATH = `${IKA_PACKAGE_ID}::ika::IKA`;
 export const DWALLET_NETWORK_VERSION = 0;
 
 export const SUI_PACKAGE_ID = '0x2';
 
 // Mocked protocol parameters used for testing purposes in non-production environments.
 export const mockedProtocolPublicParameters = Uint8Array.from(
-	// eslint-disable-next-line no-restricted-globals
```
zeevmoney, line 20: why?

sdk/typescript/tsconfig.json
```diff
@@ -29,7 +29,7 @@
 		"resolveJsonModule": true,
 		"composite": true,
 		"module": "CommonJS",
-		"outDir": "dist/cjs",
```
zeevmoney, line 32: why?

sdk/typescript/src/dwallet-mpc/globals.ts
zeevmoney: also remove this const

zeevmoney (changes requested):
see comments

sdk/typescript/tsconfig.json
ItayLevyOfficial: Needed to change this as this folder does not exist, and compiling the TS with this configuration keeps the js files in the git repo, as you saw in the previous pr.

sdk/typescript/src/dwallet-mpc/globals.ts
ItayLevyOfficial: removed

crates/ika-swarm-config/src/sui_client.rs
ItayLevyOfficial: Which commented lines? I don't see any new commented lines in this PR.

ika_config.json
ItayLevyOfficial, line 1: doesn't need to be part of the repo

zeevmoney (approved):

PR #639, opened 2025-02-13
fix: Make the `DWalletMPCService` start with the new Move code

ItayLevyOfficial:
After fetching the new Move code, I forgot to update the Rust code structs to match the new Move structs, and this prevented the chain from starting successfully. This PR mitigates this issue, and fixes one typo.

### DIFF
--- a/crates/ika-types/src/sui/system_inner_v1.rs
+++ b/crates/ika-types/src/sui/system_inner_v1.rs
@@ -94,8 +94,8 @@ pub struct UpgradeCap {
 /// Represents a capability granting control over a specific dWallet network decryption key.
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
 pub struct DWalletNetworkDecryptionKeyCap {
-    pub id: ID,
-    pub dwallet_network_decryption_key_id: ID,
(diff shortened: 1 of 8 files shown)


crates/ika-types/src/sui/system_inner_v1.rs
```diff
@@ -104,6 +112,8 @@ pub struct SystemInnerV1 {
     pub previous_epoch_last_checkpoint_sequence_number: u64,
     pub computation_reward: Balance,
     pub authorized_protocol_cap_ids: Vec<ObjectID>,
+    pub dwallet_2pc_mpc_secp256k1_id: Option<ObjectID>,
```
zeevmoney, line 115: I don't understand this field, this is the ID of what exactly?

zeevmoney (approved):

crates/ika-types/src/sui/system_inner_v1.rs
ItayLevyOfficial: There are two shared object states objects - one is the system state, and the other is the DWalllet state. This is the ID of the DWallet state shared object.

zeevmoney (approved):

PR #647, opened 2025-02-16
chore: Save the Class Groups encryption key and proof for each validator upon joining the committee

yaelAbergel:
This PR integrates the Class Groups public key and proof into the validator metadata, ensuring it is requested when a validator joins the committee.

**Sui has some limitations that required careful design considerations:**  
1. **Object Size Restriction (≤ 250KB) vs. Class Groups Key Size (~240KB):**  
   The full encryption key and proof consist of 13 key-proof pairs, which exceed Sui's object size limit when added to the validator metadata object. To bypass this constraint, we created `ClassGroupsPublicKeyAndProof` object that dynamically store the full key.

2. **Transaction Argument Limit (≤ 16KB) vs. Single Proof Size (~17KB):**  
   A single public key-proof pair exceeds the transaction argument size limit.  
   To resolve this, each key-proof pair is now split into two parts before being processed.


crates/dwallet-classgroups-types/src/lib.rs
zeevmoney, line 1: move all structs to the top.

zeevmoney: I mean all of them, not only the one added in this PR.

crates/ika-move-packages/packages/ika_system/sources/utils/class_groups_public_key_and_proof.move
zeevmoney, line 1: add docs for all the functions, and for the structs + fields.
In the functions the important thing is to explain why we broke it down to 13.

crates/ika-types/src/sui/mod.rs
zeevmoney, line 1: Doc all public structs, alternatively reduce them to pub(crate) or pub(super) if possible.
Always give the minimum access possible.

zeevmoney (changes requested):
see comments

crates/ika-types/src/sui/mod.rs
```diff
@@ -168,3 +169,38 @@ pub struct Validator {
...
+        StructTag {
+            address: ika_system_package_address,
+            name: ident_str!("ClassGroupsPublicKeyAndProofBuilder").to_owned(),
+            module: ident_str!("class_groups_public_key_and_proof").to_owned(),
+            type_params: vec![],
+        }
+    }
+}
+
+/// Rust version of the Move ika::class_groups::ClassGroupsPublicKeyAndProof type
```
zeevmoney, line 189:
```suggestion
/// Rust version of the Move ika::class_groups::ClassGroupsPublicKeyAndProof type.
```

crates/ika-types/src/sui/mod.rs
```diff
@@ -168,3 +169,38 @@ pub struct Validator {
     pub id: ObjectID,
     pub inner: Versioned,
 }
+
+/// Rust representation of the Move ika::class_groups::ClassGroupsPublicKeyAndProofBuilder type
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct ClassGroupsPublicKeyAndProofBuilder;
+
+impl ClassGroupsPublicKeyAndProofBuilder {
+    /// Return the Move struct tag for this type
```
zeevmoney, line 178:
```suggestion
    /// Return the Move struct tag for this type.
```

crates/ika-types/src/sui/mod.rs
```diff
@@ -168,3 +169,38 @@ pub struct Validator {
     pub id: ObjectID,
     pub inner: Versioned,
 }
+
+/// Rust representation of the Move ika::class_groups::ClassGroupsPublicKeyAndProofBuilder type
```
zeevmoney, line 173:
```suggestion
/// Rust representation of the Move ika::class_groups::ClassGroupsPublicKeyAndProofBuilder type.
```

PR #661, opened 2025-02-19
660

ItayLevyOfficial:


sdk/typescript/package.json
ItayLevyOfficial, line 1: I forgot to remove those deps while removing Sui`s TS code, and they break the build

zeevmoney (approved):
see small comment

PR #664, opened 2025-02-19
feat: Reuse existing Ika smart contracts instead of redeploying on every run

ItayLevyOfficial:
Until now, every restart of the Ika network caused the re-deployment of the Ika Sui smart contracts. From now on, re-deployment will only occur if the ~/.ika folder does not exist.

The new command to start the chain (in Intellij run configuration) is: `run --package ika --bin ika -- start`

To re-deploy the contracts, run the following command before starting the network: `rm -rf ~/.ika`

This PR also updates the testing Swarm chain to start in a new thread to allow customization of its stack size.


zeevmoney (approved):
gj

PR #670, opened 2025-02-20
feat: Become candidate via CLI

yaelAbergel:
Usage: ika validator become-candidate --init-system-shared-version <init-system-shared-version> <validator-info-path>


crates/ika-sui-client/src/ika_validator_transactions.rs
yaelAbergel, line 1: This file contains duplicated code from `sui_client.rs`. Omer instructed me to keep the swarm and client transactions completely separate, despite the redundancy.

yaelAbergel:
Part of #677

PR #674, opened 2025-02-23
fix: Correct substring usage in hexToAscii for accurate hex to ASCII conversion

Himess:
###  fix: Correct Hex to ASCII Conversion in `hexToAscii` Function

This PR fixes a bug in the `hexToAscii` function where the `substring` method was incorrectly extracting characters, leading to inaccurate hex-to-ASCII conversions.

####  **Issue:**
- The previous implementation used `substring(n, 2)` inside the loop, which failed to correctly extract two-character hex pairs after the first iteration.
- This caused invalid ASCII character conversion for hex strings.

####  **Solution:**
- Updated `substring(n, n + 2)` to ensure two characters are extracted at each iteration.
- Ensured accurate parsing of hex pairs to their corresponding ASCII characters.

####  **Example:**
```javascript
hexToAscii('48656c6c6f'); // Before: "Hl" → After: "Hello"
hexToAscii('0x776f726c64'); // Before: "w¤r" → After: "world"

### DIFF
--- a/apps/explorer/src/utils/stringUtils.ts
+++ b/apps/explorer/src/utils/stringUtils.ts
@@ -9,7 +9,7 @@ export function hexToAscii(hex: string) {
 
 	var str = '';
 	for (var n = 0; n < hex.length; n += 2)
-		str += String.fromCharCode(parseInt(hex.substring(n, 2), 16));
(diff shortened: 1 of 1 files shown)


zeevmoney:
The `main` branch will be replaced soon by a complete refactor featuring `ika` code, so this code won't be used.

PR #682, opened 2025-02-27
feat: Allow users to create presigns using the TS sdk

ItayLevyOfficial:


crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -15,7 +15,7 @@ pub const START_DKG_FIRST_ROUND_EVENT_STRUCT_NAME: &IdentStr =
 pub const START_DKG_SECOND_ROUND_EVENT_STRUCT_NAME: &IdentStr =
     ident_str!("DKGSecondRoundRequestEvent");
 pub const START_PRESIGN_FIRST_ROUND_EVENT_STRUCT_NAME: &IdentStr =
-    ident_str!("StartPresignFirstRoundEvent");
```
zeevmoney, line 18: add todo to fix all the func naming after we mege with @omersadika

sdk/typescript/src/dwallet-mpc/mocks.ts
zeevmoney, line 1: move the `test`

zeevmoney (changes requested):
see comments

sdk/typescript/src/dwallet-mpc/presign.ts
```diff
@@ -0,0 +1,122 @@
...
+	c: Config,
+	sessionID: string,
+	eventType: string,
+	isEventFn: (parsedJson: any) => parsedJson is TEvent,
+): Promise<TEvent> {
+	const startTime = Date.now();
+
+	while (Date.now() - startTime <= c.timeout) {
+		// Wait for a bit before polling again, objects might not be available immediately.
+		let interval = 5_000;
```
zeevmoney, line 93:
```suggestion
		const interval = 5_000;
```

sdk/typescript/src/dwallet-mpc/presign.ts
zeevmoney, line 1: replace all `let` with `const` and add a rule, it's called `preferConst` or something like this...

zeevmoney: btw it will force you to fix it for all files, do it in this PR .

zeevmoney (changes requested):
Looks good, but fix the minor issue it in TS, then I will approve.

zeevmoney (approved):

PR #685, opened 2025-02-27
Patch 'should encrypt a secret share' issues

erik-3milabs:
This PR patches the two issues that prevented 'should encrypt a secret share' from executing:
1. The `crypto-bigint` patch is updated to point to the same commit as the `cryptography-private` repository uses, and
2. replace `OsRng` with `ChaCha20`; the former uses the `getrandom` crate which is hard to compile on `wasm` targets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1753,19 +1753,19 @@ checksum = "f46ad14479a25103f283c0f10005961cf086d8dc42205bb44c46ac563475dca6"
 [[package]]
 name = "class_groups"
 version = "0.1.0"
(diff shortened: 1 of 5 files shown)


zeevmoney:
@erik-3milabs what is this PR? why did you base it one some unknown branch?

zeevmoney (changes requested):
do not merge, talk to me first.

erik-3milabs:
@zeevmoney, I wrote this patch for @ItayLevyOfficial. @ItayLevyOfficial, please let me know which branch you'd like me to base it against.

PR #688, opened 2025-03-02
684 sign

ItayLevyOfficial:


.gitignore
```diff
@@ -75,3 +75,4 @@ docs/content/references/framework/**
 
 lcov.info
 ika_config.json
+ika_config.yaml
```
ItayLevyOfficial, line 78: needed to ignore the new config file @yaelAbergel started generating on each run

PR #689, opened 2025-03-02
feat: Incorporate cryptography-private Wasm fix

ItayLevyOfficial:
This allows running a complete dWallet creation flow without using a mock


sdk/typescript/test/e2e/dwallet-mpc.test.ts
ItayLevyOfficial, line 1: removed the tests that use a mock of the DKG first round output as they are now redundant

PR #757, opened 2025-03-17
fix: Allow deserializing the class groups key pair and proof when restarting the chain

ItayLevyOfficial:
The chain restart can still fail due to another bug in the checkpointing mechanism, this PR only mitigates the class group key deserialization issue.


zeevmoney:
fixed in #759

PR #766, opened 2025-03-20
chore: Optimize the checkpoint message size

yaelAbergel:
Optimize the checkpoint transaction size while adhering to Sui's limits.


crates/ika-core/src/sui_connector/sui_executor.rs
```diff
@@ -168,6 +170,21 @@ where
         signers_bitmap
     }
 
+    /// Break down the message to slices because of chain transaction size limits.
+    /// Limit 16 KB per Tx `pure` argument.
+    fn break_down_checkpoint_message(message: Vec<u8>) -> Vec<CallArg> {
+        let mut slices = Vec::new();
+        let messages = message.chunks(15 * 1024).collect_vec();
```
zeevmoney, line 177:
```suggestion
        // Set to 15 because the limit is up to 16 (smaller than).
        let messages = message.chunks(15 * 1024).collect_vec();
```

zeevmoney (approved):

PR #774, opened 2025-03-24
chore: Remove the duplicated implementation of the `message_digest` function 

ItayLevyOfficial:
This function uses the `k256` crate & it's being used both in the Wasm & the backend code. I moved this function & all related logic to the `shared-wasm-class-groups`, and now I use it from both the Wasm code and the backend. 

This PR also includes renaming the `class_groups_constants` crate to `shared-wasm-class-groups`, as now it contains the `message_digest` function, which is not a constant.


crates/shared-wasm-class-groups/src/message_digest.rs
```diff
@@ -0,0 +1,45 @@
...
+            &sha3::Keccak256::new_with_prefix(message).finalize_fixed(),
+        )
+        .map_err(|e| anyhow::Error::msg(format!("KECCAK256 bits2field error: {:?}", e)))?,
+
+        Hash::SHA256 => {
+            bits2field::<k256::Secp256k1>(&sha2::Sha256::new_with_prefix(message).finalize_fixed())
+                .map_err(|e| anyhow::Error::msg(format!("SHA256 bits2field error: {:?}", e)))?
+        }
+    };
+    #[allow(clippy::useless_conversion)]
```
zeevmoney, line 42: why? comment

crates/shared-wasm-class-groups/src/message_digest.rs
```diff
@@ -0,0 +1,45 @@
+use k256::ecdsa::hazmat::bits2field;
+use k256::elliptic_curve::ops::Reduce;
+use k256::{elliptic_curve, U256};
+use sha3::digest::FixedOutput;
+use sha3::Digest;
+use twopc_mpc::secp256k1;
+/// Supported hash functions for message digest.
```
zeevmoney, line 7:
```suggestion
use twopc_mpc::secp256k1;

/// Supported hash functions for message digest.
```

zeevmoney (changes requested):
see comment

crates/shared-wasm-class-groups/src/message_digest.rs
ItayLevyOfficial: this was in the original function, I removed it

zeevmoney (approved):

PR #783, opened 2025-03-26
fix: Allow full nodes to sync checkpoints from peers

ItayLevyOfficial:


crates/ika-swarm-config/src/sui_client.rs
ItayLevyOfficial, line 1: Sadika directly pushed the last pr without running `cargo fmt`

zeevmoney (approved):

PR #790, opened 2025-03-30
chore: Report Sui on rejected sessions

ItayLevyOfficial:


yaelAbergel (approved):
Looks good! Just add documentation explaining how it should work.

PR #823, opened 2025-04-03
chore: Limit the number of concurrent active MPC sessions

### LINKED ISSUE #822: load the last session to run from move all the time & run sessions only until this sequence number

ItayLevyOfficial:
This is needed to allow completing all the MPC sessions on an epoch switch in a reasonable time


crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -553,10 +582,10 @@ impl DWalletMPCManager {
         &mut self,
         session_id: &ObjectID,
         mpc_event_data: Option<MPCEventData>,
+        session_sequence_number: u64,
     ) {
         if self.mpc_sessions.contains_key(&session_id) {
-            // This should never happen, as the session ID is a Move UniqueID.
-            error!(
+            warn!(
```
ItayLevyOfficial, line 588: This can happen now due to the load_missed_events logic

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -150,9 +155,27 @@ impl DWalletMPCManager {
...
             pending_for_events_order: Default::default(),
+            last_completed_session_sequence_number: 0,
         })
     }
 
+    pub(crate) fn update_last_completed_session_sequence_number(
+        &mut self,
+        last_completed_session_sequence_number: u64,
+    ) {
+        if last_completed_session_sequence_number < self.last_completed_session_sequence_number {
```
ycscaly, line 166:
```suggestion
        if last_completed_session_sequence_number <= self.last_completed_session_sequence_number {
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -150,9 +155,27 @@ impl DWalletMPCManager {
...
+            return;
+        }
+        for i in self.last_completed_session_sequence_number
+            ..=(last_completed_session_sequence_number + MAX_ACTIVE_SESSIONS_BUFFER)
+        {
+            if let Some(session) = self.pending_sessions.remove(&i) {
+                self.mpc_sessions.insert(session.session_id, session);
+            }
+        }
```
ycscaly, line 175: It is anyways not how we discussed as far as I know: the Move should keep a `last_event_to_execute` and not be computed here. We should think if this ,akes any difference

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -509,7 +536,11 @@ impl DWalletMPCManager {
                     "received a message for an MPC session ID: `{:?}` which an event has not yet received for",
                     message.session_id
                 );
-                self.push_new_mpc_session(&message.session_id, None);
+                self.push_new_mpc_session(
```
ycscaly, line 539: We said we want to limit now what sessions can we receive message for and mark validators that send messages for inexisting sessions as malicious

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly: Where is that?

ycscaly (changes requested):

PR #837, opened 2025-04-09
fix: Compilation 

yaelAbergel:


crates/ika-core/src/dwallet_mpc/network_dkg.rs
```diff
@@ -292,15 +294,15 @@ impl DwalletMPCNetworkKeys {
...
+    pub fn get_protocol_public_parameters(
         &self,
         key_id: &ObjectID,
         key_scheme: DWalletMPCNetworkKeyScheme,
     ) -> DwalletMPCResult<Vec<u8>> {
         loop {
             let Ok(Some(result)) = self.try_get_decryption_keys(key_id) else {
                 warn!("failed to fetch the network decryption key shares for key ID: {:?}, trying again", key_id);
-                tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
+                thread::sleep(Duration::from_secs(2));
```
ItayLevyOfficial, line 305: this is bad as this will block the entire tokio thread pool

PR #844, opened 2025-04-10
fix: Increase the node Tokio thread pool size

### LINKED ISSUE #843: fix tokio node thread pool size
right now the thread pool size is 1 and this creates problems with the consensus commits creation mechanism

ItayLevyOfficial:
The current thread pool size (1) is way to small, and causes the consensus mechanism to stuck


zeevmoney (approved):

ItayLevyOfficial:
irrelevant atm

PR #869, opened 2025-04-16
feat: Advance reshare until finalized 

yaelAbergel:


crates/ika-core/src/dwallet_mpc/mod.rs
```diff
@@ -482,7 +482,13 @@ pub(crate) async fn session_input_from_event(
                             .dwallet_network_decryption_key_id,
                     )?,
                 )?,
-                None,
+                Some(bcs::to_bytes(
+                    &dwallet_mpc_manager
+                        .node_config
+                        .class_groups_key_pair_and_proof
+                        .class_groups_keypair()
```
ycscaly, line 489: Keypair is secret + public. I never mix up these two, so no way this is a keypair. Remame

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -397,7 +404,30 @@ impl DWalletMPCSession {
                     malicious_parties: vec![],
                 })
             }
-            MPCProtocolInitData::DecryptionKeyReshare(_) => todo!(),
+            MPCProtocolInitData::DecryptionKeyReshare(_) => {
+                let public_input = bcs::from_bytes(public_input)?;
+                let decryption_shares_primes = mpc_event_data
```
ycscaly, line 409: Really confusing name. There is no prime here.. only in the encryption keys and proofs
```suggestion
                let decryption_key_shares = mpc_event_data
```

ycscaly (approved):
Too easy, just rename

ItayLevyOfficial (approved):
gj, merge this pr with the public input prs

crates/ika-core/src/dwallet_mpc/mod.rs
yaelAbergel: The node config is stored locally in the validator pc. The keypair remains private, and the validator only sends the encryption key and a proof to the chain.

zeevmoney (approved):

PR #913, opened 2025-04-22
chore: Remove `dwallet_mpc_round_outputs` mutex

ItayLevyOfficial:


crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1497,13 +1516,6 @@ impl AuthorityPerEpochStore {
         session_info: SessionInfo,
         output: Vec<u8>,
     ) -> IkaResult<ConsensusCertificateResult> {
-        self.save_dwallet_mpc_output(DWalletMPCOutputMessage {
```
zeevmoney, line 1500: why was it removed? how is this related to the mutex?

crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1424,6 +1410,39 @@ impl AuthorityPerEpochStore {
             .collect()
     }
 
+    /// Filter DWalletMPCMessages from the consensus output.
+    /// Those messages will get processed when the dWallet MPC service reads
+    /// them from the DB.
+    fn filter_dwallet_mpc_outputs(
```
zeevmoney, line 1416: Why do we need this filter suddenly?

zeevmoney (changes requested):
explain what you did, there is not enough context.

zeevmoney (approved):

zeevmoney (approved):

PR #930, opened 2025-04-23
fix: Handle messages that are being received before events correctly

ItayLevyOfficial:


crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -529,6 +490,28 @@ impl DWalletMPCManager {
                 );
                 continue;
             }
+            let Some(mpc_event_data) = oldest_pending_session.mpc_event_data.clone() else {
+                // This should never happen
+                error!(
```
zeevmoney, line 495: add fields

zeevmoney (approved):

PR #931, opened 2025-04-23
bugfix: Read current committee from state object for syncing network decryption key

yaelAbergel:


crates/ika-core/src/dwallet_mpc/mod.rs
yaelAbergel, line 1: just format

crates/ika-types/src/committee.rs
yaelAbergel, line 1: format

zeevmoney (approved):

zeevmoney (approved):

PR #947, opened 2025-04-24
fix: Ignore MPC session events from different epochs

ItayLevyOfficial:


crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -106,7 +106,8 @@ impl DWalletMPCService {
             };
             for event in events {
                 match session_info_from_event(event.clone(), &epoch_store.packages_config) {
-                    Ok(Some(session_info)) => {
+                    Ok(Some(mut session_info)) => {
```
zeevmoney, line 109: add the code that ignores non current epoch sui events

zeevmoney (approved):

PR #962, opened 2025-04-28
chore: Remove redundant error log when trying to process a non-MPC event

ItayLevyOfficial:


crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1361,7 +1361,7 @@ impl AuthorityPerEpochStore {
                         Some(event)
                     }
                     Ok(None) => {
-                        error!("failed to extract session info from event");
+                        info!("Received an event that does not trigger the start of an MPC flow");
```
zeevmoney, line 1364:
```suggestion
                        warn!("Received an event that does not trigger the start of an MPC flow");
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -137,7 +137,7 @@ impl DWalletMPCService {
                         );
                     }
                     Ok(None) => {
-                        error!("Failed to extract session info from missed event");
+                        info!("Received an event that does not trigger the start of an MPC flow");
```
zeevmoney, line 140:
```suggestion
                        warn!("Received an event that does not trigger the start of an MPC flow");
```

zeevmoney (changes requested):
see comments

zeevmoney (approved):

PR #966, opened 2025-04-28
fix typos

destinyae:
check commits, fix all typos.


zeevmoney:
Thanks, but the `main` branch will be replaced soon, entirely.

PR #993, opened 2025-05-04
Change move dwallet code to be generic

omersadika:


crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -631,6 +675,12 @@ public struct ECDSASignRequestEvent has copy, drop, store {
     /// The output from the dWallet DKG process used in this session.
     dwallet_public_output: vector<u8>,
 
+    /// The elliptic curve used for the dWallet.
+    curve: u8,
```
ycscaly, line 679: Just saying it: using a `u8` limits us to `256` supported options, forever. This should be fine, but I'm wondering why not use a bigger number like `u32` which we know never, ever to be too small for our future needs. I guess maybe storage is an reason, but its probably negligible relative to other fields.

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -728,6 +780,10 @@ const EInvalidSource: u64 =13;
 const EDWalletNetworkDecryptionKeyNotActive: u64 = 14;
 const EInvalidPresign: u64 = 15;
 const ECannotAdvanceEpoch: u64 = 16;
+const EInvalidCurve: u64 = 17;
+const EInvalidSignatureAlgorithm: u64 = 18;
+const ECurvePaused: u64 = 19;
+const ESignatureAlgorithmPaused: u64 = 20;
```
ycscaly, line 786: Pause hash too?

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1087,25 +1149,11 @@ public(package) fun register_encryption_key(
 /// - **`message`**: The message that has been approved.
 public struct MessageApproval has store, drop {
     dwallet_id: ID,
+    signature_algorithm: u8,
```
ycscaly, line 1152: I know that currently there is only one curve per dWallet, but do we want to specify curve here anyways? Do we want to offer ourselves the option to support more than one curve per dWallet, or not? Maybe better not to, but we can think of what it means

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1131,24 +1179,29 @@ public(package) fun create_message_approval(
...
     dwallet_cap: &DWalletCap,
+    signature_algorithm: u8,
     hash_scheme: u8,
     message: vector<u8>
 ): MessageApproval {
-    create_message_approval(
-        dwallet_cap.dwallet_id,
+    let dwallet_id = dwallet_cap.dwallet_id;
+    let (dwallet, _) = self.get_active_dwallet_and_public_output(dwallet_id);
+    assert!(self.supported_curves_to_signature_algorithms.contains(&dwallet.curve), EInvalidCurve);
```
ycscaly, line 1191: How could a dwallet for an unsupported curve exist?

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1131,24 +1179,29 @@ public(package) fun create_message_approval(
...
+    let supported_curve_to_signature_algorithms = self.supported_curves_to_signature_algorithms[&dwallet.curve];
+    assert!(supported_curve_to_signature_algorithms.contains(&signature_algorithm), EInvalidSignatureAlgorithm);
+    assert!(self.supported_signature_algorithms_to_hash_schemes.contains(&signature_algorithm), EInvalidSignatureAlgorithm);
```
ycscaly, line 1194: I think its much better to check all of these together rather than separately. I could foresee a reason why not to use specific hash in a sepcific signature algorithm for one curve but not the other.

Let's do that you just have an array of supported combinations, and you just put there all the supported combinations and check if the current combination exists, i.e. (curve, signature_scheme, hash_scheme)

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1131,24 +1179,29 @@ public(package) fun create_message_approval(
...
+    let supported_signature_algorithm_to_hash_schemes = self.supported_signature_algorithms_to_hash_schemes[&signature_algorithm];
+    assert!(supported_signature_algorithm_to_hash_schemes.contains(&hash_scheme), EInvalidHashScheme);
+
```
ycscaly, line 1197: No pause checks?

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1625,47 +1684,126 @@ public(package) fun accept_encrypted_user_share(
...
+    let (dwallet, public_output) = self.get_active_dwallet_and_public_output(dwallet_id);
+
+    let curve = dwallet.curve;
+
+    assert!(self.supported_curves_to_signature_algorithms.contains(&curve), EInvalidCurve);
+    let supported_curve_to_signature_algorithms = self.supported_curves_to_signature_algorithms[&curve];
+    assert!(supported_curve_to_signature_algorithms.contains(&signature_algorithm), EInvalidSignatureAlgorithm);
+    assert!(self.supported_signature_algorithms_to_hash_schemes.contains(&signature_algorithm), EInvalidSignatureAlgorithm);
+
+    assert!(!self.paused_curves.contains(&curve), ECurvePaused);
```
ycscaly, line 1708: Pause hash too

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1766,51 +1901,56 @@ public(package) fun is_ecdsa_presign_valid(
...
-        ECDSAPresignState::Completed { presign } => {
+        PresignState::Completed { presign } => {
             presign
         },
         _ => abort EInvalidPresign
     };
     let presign_id = id.to_inner();
     id.delete();
-    assert!(presign_dwallet_id == message_approval_dwallet_id, EMessageApprovalMismatch);
+    assert!(presign_dwallet_id.is_none() || presign_dwallet_id.is_some_and!(|id| id == message_approval_dwallet_id), EMessageApprovalMismatch);
```
ycscaly, line 1952: But message approval dwallet id should also be an option, I think this check is wrong

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1766,51 +1901,56 @@ public(package) fun is_ecdsa_presign_valid(
...
+    assert!(signature_algorithm == message_approval_signature_algorithm, EMessageApprovalMismatch);
     assert!(presign_cap_id == cap_id, EPresignNotExist);
     assert!(presign_id == presign_cap_presign_id, EPresignNotExist);
     assert!(presign_cap_dwallet_id == presign_dwallet_id, EPresignNotExist);
```
ycscaly, line 1956: Better copy all the presign/message approval checks here too just in case: that everything is supported, that this supports keyless, nothing is paused etc

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1766,51 +1901,56 @@ public(package) fun is_ecdsa_presign_valid(
...
     is_future_sign: bool,
     ctx: &mut TxContext
 ) {
     let created_at_epoch = self.current_epoch;
-    let (dwallet, public_output) = self.get_active_dwallet_and_public_output_mut(presign_cap.dwallet_id);
 
-    assert!(dwallet.ecdsa_presigns.contains(presign_cap.presign_id), EPresignNotExist);
-    let presign = dwallet.ecdsa_presigns.remove(presign_cap.presign_id);
+    assert!(self.presigns.contains(presign_cap.presign_id), EPresignNotExist);
+    let presign = self.presigns.remove(presign_cap.presign_id);
```
ycscaly, line 1918: We need to be very careful here, if you remember we had some issues in the previous design about presigns removals etc. and the structure is changed now, and I think some of the reasons that the issues weren't serious last time was because the presigns were kept in the dwallet itself.

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
```diff
@@ -1766,51 +1901,56 @@ public(package) fun is_ecdsa_presign_valid(
 /// # Aborts
 /// - **`EExtraDataAndMessagesLenMismatch`**: If `signature_algorithm_data` and `message_approvals` have different lengths.
 /// - **`EMissingApprovalOrWrongApprovalOrder`**: If message approvals are incorrect or missing.
-fun emit_ecdsa_sign_event(
+fun emit_sign_event(
```
ycscaly, line 1904:
```suggestion
/// These function is a shared logic for both the normal and future sign flows. It checks the presign is valid and removes it, thus assuring it is never used twice. Finally it emits the sign event.
fun sign_inner(
```

crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_coordinator_inner.move
ycscaly: also, I know that's how you get the dwallet in the first place and its a duplicate check, but I rather have it in the code so if there are future changes we won't ever forget it: check that `dwallet.id` matches the `presign_dwallet_id` too

ycscaly (changes requested):
see somments

zeevmoney (approved):

zeevmoney (approved):

PR #1012, opened 2025-05-11
fix: Handle malicious behavior correctly in all flows

ItayLevyOfficial:
Until now, our malicious actor handling mechanism worked properly only when rolling back a single round after a session failure due to malicious behavior. While this sufficed in most cases, the network DKG and Reshare flows require rolling back multiple rounds. This PR correctly implements those flows by leveraging Scaly's new code. When a `ThresholdNotReached` error is received, the system now waits for additional messages (including those from previous rounds) and retries.

[This draft PR](https://github.com/dwallet-labs/dwallet-network/pull/1013) includes a passing e2e test that shows the new malicious handling mechanism actually works.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/mod.rs
+++ b/crates/ika-core/src/dwallet_mpc/mod.rs
@@ -451,7 +451,7 @@ pub(crate) fn advance_and_serialize<P: AsynchronouslyAdvanceable>(
 }
 
 struct DeserializeMPCMessagesResponse<M: DeserializeOwned + Clone> {
-    messages: Vec<HashMap<PartyID, M>>,
+    messages: HashMap<usize, HashMap<PartyID, M>>,
     malicious_parties: Vec<PartyID>,
 }
 
@@ -461,10 +461,10 @@ struct DeserializeMPCMessagesResponse<M: DeserializeOwned + Clone> {
 fn deserialize_mpc_messages<M: DeserializeOwned + Clone>(
     messages: Vec<HashMap<PartyID, MPCMessage>>,
 ) -> DeserializeMPCMessagesResponse<M> {
-    let mut deserialized_results = Vec::new();
+    let mut deserialized_results = HashMap::new();
     let mut malicious_parties = Vec::new();
 
-    for message_batch in &messages {
+    for (index, message_batch) in &messages.iter().enumerate() {
         let mut valid_messages = HashMap::new();
 
         for (party_id, message) in message_batch {
@@ -484,7 +484,7 @@ fn deserialize_mpc_messages<M: DeserializeOwned + Clone>(
         }
 
         if !valid_messages.is_empty() {
-            deserialized_results.push(valid_messages);
+            deserialized_results.insert(index + 1, valid_messages);
         }
     }
     DeserializeMPCMessagesResponse {
--- a/crates/ika-core/src/dwallet_mpc/mod.rs
+++ b/crates/ika-core/src/dwallet_mpc/mod.rs
@@ -396,7 +396,7 @@ pub(crate) fn advance_and_serialize<P: AsynchronouslyAdvanceable>(
         malicious_parties: _,
     } = deserialize_mpc_messages(messages);
 
-    let res = match P::advance(
+    let res = match P::advance_with_guaranteed_output(
         session_id,
         party_id,
         access_threshold,
--- a/crates/ika-core/src/dwallet_mpc/mod.rs
+++ b/crates/ika-core/src/dwallet_mpc/mod.rs
@@ -484,7 +484,7 @@ fn deserialize_mpc_messages<M: DeserializeOwned + Clone>(
         }
 
         if !valid_messages.is_empty() {
-            deserialized_results.insert(index + 1, valid_messages);
+            deserialized_results.insert(index, valid_messages);
         }
     }
     DeserializeMPCMessagesResponse {
--- a/crates/ika-core/src/dwallet_mpc/mod.rs
+++ b/crates/ika-core/src/dwallet_mpc/mod.rs
@@ -416,7 +416,7 @@ pub(crate) fn advance_and_serialize<P: AsynchronouslyAdvanceable>(
             return match e.into() {
                 // No threshold was reached, so we can't proceed.
                 mpc::Error::ThresholdNotReached => {
-                    todo!("Handle threshold not reached error")
+                    return Err(DwalletMPCError::TWOPCMPCThresholdNotReached)
                 }
                 _ => Err(general_error),
             };
--- a/crates/ika-core/src/dwallet_mpc/mod.rs
+++ b/crates/ika-core/src/dwallet_mpc/mod.rs
@@ -415,17 +415,8 @@ pub(crate) fn advance_and_serialize<P: AsynchronouslyAdvanceable>(
             ));
             return match e.into() {
                 // No threshold was reached, so we can't proceed.
-                mpc::Error::ThresholdNotReached { honest_subset } => {
-                    let malicious_actors = messages
-                        .last()
-                        .ok_or(general_error)?
-                        .keys()
-                        .filter(|party_id| !honest_subset.contains(*party_id))
-                        .cloned()
-                        .collect();
-                    Err(DwalletMPCError::SessionFailedWithMaliciousParties(
-                        malicious_actors,
-                    ))
+                mpc::Error::ThresholdNotReached => {
+                    todo!("Handle threshold not reached error")
                 }
                 _ => Err(general_error),
             };
--- a/crates/ika-core/src/dwallet_mpc/mod.rs
+++ b/crates/ika-core/src/dwallet_mpc/mod.rs
@@ -464,7 +464,7 @@ fn deserialize_mpc_messages<M: DeserializeOwned + Clone>(
     let mut deserialized_results = HashMap::new();
     let mut malicious_parties = Vec::new();
 
-    for (index, message_batch) in &messages.iter().enumerate() {
+    for (index, message_batch) in messages.iter().enumerate() {
         let mut valid_messages = HashMap::new();
 
         for (party_id, message) in message_batch {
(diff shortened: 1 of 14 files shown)


crates/ika-core/src/dwallet_mpc/mod.rs
```diff
@@ -396,7 +396,7 @@ pub(crate) fn advance_and_serialize<P: AsynchronouslyAdvanceable>(
         malicious_parties: _,
     } = deserialize_mpc_messages(messages);
 
-    let res = match P::advance(
+    let res = match P::advance_with_guaranteed_output(
```
ycscaly, line 399: Good

crates/ika-core/src/dwallet_mpc/mod.rs
```diff
@@ -470,10 +461,10 @@ struct DeserializeMPCMessagesResponse<M: DeserializeOwned + Clone> {
 fn deserialize_mpc_messages<M: DeserializeOwned + Clone>(
     messages: Vec<HashMap<PartyID, MPCMessage>>,
 ) -> DeserializeMPCMessagesResponse<M> {
-    let mut deserialized_results = Vec::new();
+    let mut deserialized_results = HashMap::new();
     let mut malicious_parties = Vec::new();
```
ycscaly, line 465: Its not a good idea to keep pushing into a vec as there could be repeated entries there.
```suggestion
    let mut malicious_parties = HashSet::new();
```

ycscaly: Also, there could be a problem with ordering with Vec. This is why I always call my `deduplicate_and_sort()` when handling malicious parties. So you can keep as vec but call this method in the end if you prefer

ycscaly: Please go over the code and call this method everywhere (you can do in separate pr) to assure we have no ordering issues with malicious parties reporting

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -284,6 +290,41 @@ impl DWalletMPCManager {
...
+            self.threshold_not_reached_reports.remove(&report);
+        }
+        Ok(())
+    }
+
+    fn wait_for_more_messages_and_retry(&mut self, session_id: ObjectID) {
+        if let Some(session) = self.mpc_sessions.get_mut(&session_id) {
+            session.received_more_messages_since_last_retry = false;
+            session.attempts_count += 1;
+            session.pending_quorum_for_highest_round_number -= 1;
```
ycscaly, line 325: good

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -284,6 +290,41 @@ impl DWalletMPCManager {
...
+        let committee = self.epoch_store()?.committee().clone();
+        if self
+            .threshold_not_reached_reports
+            .entry(report.clone())
+            .or_insert(StakeAggregator::new(committee))
+            .insert_generic(origin_authority, ())
+            .is_quorum_reached()
+        {
+            self.wait_for_more_messages_and_retry(report.session_id);
+            self.threshold_not_reached_reports.remove(&report);
```
ycscaly, line 316: I'm not sure about this - wouldn't it be removed after 2/3 messages come, and then the next message will recreate the report?

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -314,48 +355,14 @@ impl DWalletMPCManager {
...
-            .malicious_handler
+        self.malicious_handler
             .report_malicious_actor(report.clone(), reporting_authority)?;
         if self
             .malicious_handler
             .is_malicious_actor(&self.epoch_store()?.name)
         {
             self.recognized_self_as_malicious = true;
         }
-
```
ycscaly, line 327: Why is this removed? we still need to keep track of malicious parties

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -321,6 +311,26 @@ impl DWalletMPCSession {
         Ok(())
     }
 
+    fn report_threshold_not_reached(&self, tokio_runtime_handle: &Handle) -> DwalletMPCResult<()> {
+        let report_tx =
+            self.new_dwallet_report_threshold_not_reached(ThresholdNotReachedReport {
+                session_id: self.session_id,
+                crypto_round_number: self.pending_quorum_for_highest_round_number,
```
ycscaly, line 318: should this be `-1` to represent the round that failed?

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -640,6 +655,7 @@ impl DWalletMPCSession {
                                 .collect::<HashSet<PartyID>>(),
                         )
                         .is_ok()
+                        && self.received_more_messages_since_last_retry)
```
ycscaly, line 658: good

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -524,10 +526,23 @@ impl DWalletMPCSession {
...
+                report,
+            ),
+        )
+    }
+
     /// Stores a message in the serialized messages map.
     /// Every new message received for a session is stored.
     /// When a threshold of messages is reached, the session advances.
     pub(crate) fn store_message(&mut self, message: &DWalletMPCMessage) -> DwalletMPCResult<()> {
+        self.received_more_messages_since_last_retry = true;
```
ycscaly, line 545: good

crates/ika-types/src/dwallet_mpc_error.rs
```diff
@@ -34,8 +34,8 @@ pub enum DwalletMPCError {
     #[error("malicious parties have been detected: {0:?}")]
     MaliciousParties(Vec<PartyID>),
 
-    #[error("session failed with malicious parties: {0:?}")]
-    SessionFailedWithMaliciousParties(Vec<PartyID>),
+    #[error("two-pc MPC threshold not reached")]
+    TWOPCMPCThresholdNotReached,
```
ycscaly, line 38: good

ycscaly (changes requested):
looks good but some changes needed

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ItayLevyOfficial: exactly, but this "new" report will not reach a quorum

crates/ika-core/src/dwallet_mpc/mod.rs
ItayLevyOfficial: Added a todo and an issue

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ItayLevyOfficial: We are - I only removed the logic that restarts a round when a session fails with malicious parties, as we now do this when we reach agreement on threshold not reached.

crates/ika-core/src/dwallet_mpc/mpc_session.rs
ItayLevyOfficial: no

ItayLevyOfficial: This field was redundant actually, I removed it

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly: This is definitely not the way to do it. You are coupling here to the fact the threshold is 2/3, and its undocumented. Please instead, keep that report, and just add a flag that it has already reached quorum or something like that.

ycscaly (approved):
gg wp

crates/ika-core/src/dwallet_mpc/mod.rs
ycscaly, line 399: I don't understand how this code compiles now, sense we expect a `HashMap<>` of `messages`. I think you go the PR order wrong

ItayLevyOfficial: no, in this pr i covert the vector to a hashmap in the `deserialize_mpc_messages` function

ItayLevyOfficial: my order is fine.

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -321,6 +313,25 @@ impl DWalletMPCSession {
...
+                attempt: self.attempts_count,
+            })?;
+        let epoch_store = self.epoch_store()?.clone();
+        let consensus_adapter = self.consensus_adapter.clone();
+        tokio_runtime_handle.spawn(async move {
+            if let Err(err) = consensus_adapter
+                .submit_to_consensus(&vec![report_tx], &epoch_store)
+                .await
+            {
+                error!("failed to submit an MPC message to consensus: {:?}", err);
```
zeevmoney, line 329: don't use `{:?}` in errors, we need the feature of tracing where the error is separate.
This applies to all log severities.

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -86,6 +87,10 @@ pub(super) struct DWalletMPCSession {
     // TODO (#539): Simplify struct to only contain session related data - remove this field.
     weighted_threshold_access_structure: WeightedThresholdAccessStructure,
     pub(crate) mpc_event_data: Option<MPCEventData>,
+    pub(crate) received_more_messages_since_last_retry: bool,
```
zeevmoney, line 90: change this var to `received_more_messages_since_last_advance`
also do it, and explain why do we need it.

PR #1014, opened 2025-05-12
chore: Convert the MPC messages to a HashMap

ItayLevyOfficial:


crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -529,7 +529,9 @@ impl DWalletMPCManager {
                     let mut session_clone = session.clone();
                     session_clone
                         .serialized_full_messages
-                        .truncate(session.pending_quorum_for_highest_round_number);
+                        .retain(|round_number, _| {
+                            round_number < &session.pending_quorum_for_highest_round_number
+                        });
```
ycscaly, line 534: Ok i'm not sure about this one: we should never accept messages for a round before we had quorum, so there shouldn't be any more messages here. However, if we get a threshold not reached, it could be that some one (maliciously) submitted messages for the next round. In that case, I guess we should remove those & report them as malicious. So please add that. And then we can remove this `retain` I believe

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -576,65 +577,10 @@ impl DWalletMPCSession {
...
-                    "Received a message for two or more rounds above known round for session",
-                );
-                // Unexpected round number; rounds should grow sequentially.
-                return Err(DwalletMPCError::MaliciousParties(vec![source_party_id]));
-            }
-        }
+        self.serialized_full_messages
+            .entry(message.round_number)
+            .or_insert(HashMap::new())
+            .insert(source_party_id, message.message.clone());
```
ycscaly, line 583: much cleaner like that, isn't it?

crates/ika-core/src/dwallet_mpc/mpc_session.rs
ycscaly, line 1: had to write this here: I think a better name for `pending_quorum_for_highest_round_number` is `next_round` or `next_round_to_advance`. It should start from `1`, and be incremented every time we have quorum, like it is right now. So at the beginning, we have `next_round_to_advance == 1` so we advance and increment it. Then after we have `t` round 1 messages, we have `next_round_to_advance == 2` and we increment it and advance and so forth.

ycscaly (changes requested):
see comments

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -506,7 +507,7 @@ impl DWalletMPCSession {
             self.epoch_store()?.name,
             message,
             self.session_id.clone(),
-            self.pending_quorum_for_highest_round_number,
+            self.next_round_to_advance,
```
ycscaly, line 510:
```suggestion
            self.next_round_to_advance - 1,
```

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -217,7 +218,7 @@ impl DWalletMPCSession {
                     session_type=?base64_mpc_session_type,
                     session_id=?self.session_id,
                     validator=?self.epoch_store()?.name,
-                    crypto_round=?self.pending_quorum_for_highest_round_number,
+                    crypto_round=?self.next_round_to_advance,
```
ycscaly, line 221:
```suggestion
                    crypto_round=?self.next_round_to_advance - 1,
```
In all of these

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -75,14 +76,14 @@ pub(super) struct DWalletMPCSession {
...
     /// We need to accumulate a threshold of those before advancing the session.
     /// Vec[Round1: Map{Validator1->Message, Validator2->Message}, Round2: Map{Validator1->Message} ...]
-    pub(super) serialized_full_messages: Vec<HashMap<PartyID, MPCMessage>>,
+    pub(super) serialized_full_messages: HashMap<usize, HashMap<PartyID, MPCMessage>>,
     epoch_store: Weak<AuthorityPerEpochStore>,
     consensus_adapter: Arc<dyn SubmitToConsensus>,
     epoch_id: EpochId,
     pub(super) session_id: ObjectID,
     /// The current MPC round number of the session.
     /// Starts at 0 and increments by one each time we advance the session.
```
ycscaly, line 85:
```suggestion
    /// The next round we should advance, after receiving a quorum of messages from the current round (`next_round_to_advance - 1`).
    /// Starts at 1 and increments by one each time we advance the session (and is decremented only upon an `TWOPCMPCThresholdNotReached` Error.)
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -322,12 +322,29 @@ impl DWalletMPCManager {
...
     }
 
-    fn wait_for_more_messages_and_retry(&mut self, session_id: ObjectID) {
+    fn wait_for_more_messages_and_retry(&mut self, session_id: ObjectID) -> DwalletMPCResult<()> {
+        let epoch_store = self.epoch_store()?;
         if let Some(session) = self.mpc_sessions.get_mut(&session_id) {
             session.received_more_messages_since_last_retry = false;
             session.attempts_count += 1;
-            session.pending_quorum_for_highest_round_number -= 1;
+            // remove outputs of the previous round, as all validators should have failed
```
ycscaly, line 330:
```suggestion
            // We got an `TWOPCMPCThresholdNotReached` error, and a quorum agreement on it, so necessarily all parties that sent a message for the last executed round are malicious (as the round aborted with error, and no message was generated.)
           // Remove these messages, and mark the senders as malicious.
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -322,12 +322,29 @@ impl DWalletMPCManager {
...
+            self.malicious_handler
+                .report_malicious_actors(&party_ids_to_authority_names(
+                    &session
+                        .serialized_full_messages
+                        .get(&(session.next_round_to_advance - 1))
+                        .unwrap_or(&HashMap::new())
+                        .keys()
+                        .cloned()
+                        .collect::<Vec<PartyID>>(),
+                    &*epoch_store,
```
ycscaly, line 340: I think this code belongs to #1012, but as long as both are merged I don't care much about the order

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -75,14 +76,14 @@ pub(super) struct DWalletMPCSession {
...
-    pub(super) pending_quorum_for_highest_round_number: usize,
+    pub(super) next_round_to_advance: usize,
```
ycscaly, line 86:
```suggestion
    pub(super) current_round: usize,
```

crates/ika-core/src/dwallet_mpc/mpc_session.rs
ycscaly, line 85:
```suggestion
    /// The current MPC round number of the session.
    /// Starts at `1` and increments after each advance the session.
    /// Decremented only upon an `TWOPCMPCThresholdNotReached` Error.
```

ycscaly (changes requested):

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -327,12 +327,13 @@ impl DWalletMPCManager {
...
             session.attempts_count += 1;
-            // remove outputs of the previous round, as all validators should have failed
+            // We got an `TWOPCMPCThresholdNotReached` error, and a quorum agreement on it, so necessarily all parties that sent a message for the last executed round are malicious (as the round aborted with error, and no message was generated.)
+            // Remove these messages, and mark the senders as malicious.
             self.malicious_handler
                 .report_malicious_actors(&party_ids_to_authority_names(
                     &session
                         .serialized_full_messages
-                        .get(&(session.next_round_to_advance - 1))
+                        .get(&(session.current_round - 1))
```
ycscaly, line 336:
```suggestion
                        .get(&session.current_round)
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -341,8 +342,8 @@ impl DWalletMPCManager {
                 )?);
             session
                 .serialized_full_messages
-                .remove(&(session.next_round_to_advance - 1));
-            session.next_round_to_advance -= 1;
+                .remove(&(session.current_round - 1));
```
ycscaly, line 345:
```suggestion
                .remove(&session.current_round);
```

ycscaly (changes requested):

ycscaly (approved):

PR #1017, opened 2025-05-12
chore: Order malicious actors before submitting malicious report 

ItayLevyOfficial:
This ensures all validators agree on the same vector of malicious actors. A hash set can't be used, as it doesn't implement Hash and can't be sent over the network or used as a key in the reports map.


crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -292,8 +292,10 @@ impl DWalletMPCSession {
     fn report_malicious_actors(
         &self,
         tokio_runtime_handle: &Handle,
-        malicious_parties_ids: Vec<PartyID>,
+        mut malicious_parties_ids: Vec<PartyID>,
     ) -> DwalletMPCResult<()> {
+        // Makes sure all the validators report on the malicious actors in the same order.
+        malicious_parties_ids.sort();
```
ycscaly, line 298: No, please call `deduplicate_and_sort()`. This doesn't deduplicate

ycscaly (approved):

zeevmoney (approved):

PR #1021, opened 2025-05-13
fix: Stop fetching all Sui events

ItayLevyOfficial:
Right now, the SDK fetches the latest 50 events in the entire Sui blockchain every 0.5s to continually check for completion event for a session. This work fine on a local Sui network, but on a real network with many events this will break. We need to look for state changes in the object that is being created while starting the flow instead of looking for completion events.

To support this new design, I had to make a minor change to the dWallet DKG second-round logic and create the encrypted secret share object during the init flow move call, instead of doing so only in the response phase as was done until now.


sdk/typescript/src/dwallet-mpc/globals.ts
```diff
@@ -197,82 +206,6 @@ export async function getDWalletSecpState(c: Config): Promise<SharedObjectData>
 	};
 }
 
-export async function fetchObjectWithType<TObject>(
```
ItayLevyOfficial, line 200: This function was a redundant duplicate of getObjectWithType

PR #1022, opened 2025-05-13
chore: Initiate Rayon thread pool in the correct location with the right stack size

ItayLevyOfficial:
This PR removes both the Rayon thread pool error we used to encounter when running the chain via Swarm and the need to pass the `RUST_MIN_STACK=16777216` environment flag.


crates/ika-core/src/runtime.rs
```diff
@@ -12,8 +13,16 @@ pub struct IkaRuntimes {
 
 impl IkaRuntimes {
     pub fn new(_confg: &NodeConfig) -> Self {
+        if let Err(err) = rayon::ThreadPoolBuilder::new()
+            .panic_handler(|err| error!("Rayon thread pool task panicked: {:?}", err))
+            .stack_size(16 * 1024 * 1024)
+            .build_global()
+        {
+            error!("Failed to create rayon thread pool: {:?}", err);
```
omersadika, line 21: Change the error to be a param of error! - ?err

crates/ika/src/ika_commands.rs
```diff
@@ -333,7 +333,12 @@ async fn start(
     if ika_network_config_not_exists {
         swarm.network_config.save(&network_config_path)?;
     }
-
+    if let Err(err) = rayon::ThreadPoolBuilder::new()
```
omersadika, line 336: Why not in swarm crate?

zeevmoney (changes requested):
see comment

crates/ika/src/ika_commands.rs
ItayLevyOfficial: @omersadika I moved the initialization to the swarm crate.

@zeevmoney Why not initialize the Rayon thread pool from the swarm? It saves passing the `RUST_MIN_STACK=16777216` flag (I tested the swarm without this flag and it worked)

zeevmoney (approved):

PR #1032, opened 2025-05-15
feat: Sync consensus protocol version

yaelAbergel:
This PR ensures that validators synchronize to the same consensus protocol version for the next epoch.
When a validator first starts, it sends an `AuthorityCapabilitiesV1` message to all other validators. This message includes the validator’s current protocol version number.
Once a validator receives `AuthorityCapabilitiesV1` messages from the majority of other validators with the same protocol version, it sends an `IkaSystemTransaction` signed with its own signature. At this point, the checkpoint mechanism (copied and referred to as `ika_system_checkpoint`) is triggered to achieve quorum for the signatures and synchronize with the full node.


crates/ika-archival/src/lib.rs
```diff
@@ -97,7 +99,8 @@ const MANIFEST_FILENAME: &str = "MANIFEST";
 )]
 #[repr(u8)]
 pub enum FileType {
-    CheckpointMessage = 0,
+    IkaSystemCheckpoint = 0,
+    CheckpointMessage = 1,
 }
```
zeevmoney, line 104:
```suggestion
    IkaSystemCheckpoint = 0,
    CheckpointMessage = 1,
}
```
SystemCheckpointMessage
DWalletCoodinatorCheckpointMessage

zeevmoney (approved):

zeevmoney (approved):

PR #1036, opened 2025-05-18
Fix Epoch Advance DOS Bug with Session Completions Check

ycscaly:
I found a serious race condition bug in the session completion checks during epoch switch.

In move, we set last_session_to_complete_in_current_epoch = next_session_sequence_number (unless there are more than max_active_sessions_buffer open sessions, lets discard that case for now, even though I'm not sure its bug free either.)
number_of_completed_sessions is initialized to 0 and incremented every time a session finishes.
next_session_sequence_number  is initialized to 0 and incremented every time a new session begins.
Let's say we have just one session [0] that is now running, and we lock the last_session_to_complete_in_current_epoch variable (at =next_session_sequence_number=1).

Now, let's say that a new session 1 is created, before session 0 is done.
The network [starts any session as long as it is <= last_session_to_complete_in_current_epoch](https://github.com/dwallet-labs/dwallet-network/blob/6e1d1509826c8115904c612488131dc918d3111c/crates/ika-core/src/dwallet_mpc/mpc_manager.rs#L674).

This means we start session 1 too at this epoch. Let's say it finishes before 0 does. Now number_of_completed_sessions is at 1 as is last_session_to_complete_in_current_epoch, and all_current_epoch_sessions_completed()[ returns true as it checks that they are equal](https://github.com/dwallet-labs/dwallet-network/blob/6e1d1509826c8115904c612488131dc918d3111c/crates/ika-move-packages/packages/ika_system/sources/dwallet/dwallet_2pc_mpc_secp256k1_inner.move#L1254), before session 0 is done!
And so we can advance the epoch, and session 0 might never complete!

But even worse, if session 0 will also finish we would have number_of_completed_sessions = 2 which is bigger than last_session_to_complete_in_current_epoch = 1 and we would never be able to advance epoch resulting in a total shutdown of the network, and one that a malicious actor can cause by pushing a transactions to time it as I described earlier (assuming a relatively quite end of epoch, which is a very make-sense assumption.)

Fix: initialize next_session_sequence_number to 1 and set last_session_to_complete_in_current_epoch = next_session_sequence_number - 1


ItayLevyOfficial (approved):
lgtm, i ran an e2e test & it passes

crates/ika/src/ika_commands.rs
```diff
@@ -57,7 +57,7 @@ use tempfile::tempdir;
 use tokio::runtime::Runtime;
 use tracing::info;
 
-const DEFAULT_EPOCH_DURATION_MS: u64 = 1000 * 60 * 10;
+const DEFAULT_EPOCH_DURATION_MS: u64 = 1000 * 150;
```
zeevmoney, line 60: revert

ycscaly: its @ItayLevyOfficial's snapshot commit, idk how to fix this

zeevmoney (approved):

PR #1046, opened 2025-05-20
log the broken checkpoint messages nicely

ItayLevyOfficial:
nicely means without all the content


ItayLevyOfficial:
the bug has been resolved

PR #1099, opened 2025-06-01
fix: Assure the Sui notifier executes one transaction at a time

ItayLevyOfficial:
This is crucial, as executing two transactions simultaneously will lock the notifier's Coin object until the end of the epoch, rendering it non-functional.

This PR also includes a bug fix in the consensus delay mechanism I discovered while working on this pr.


crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -887,23 +888,28 @@ impl DWalletMPCSession {
...
-        // Safe to unwrap as this function is only called after the MPC event is received.
-        match &self.mpc_event_data.clone().unwrap().init_protocol_data {
-            MPCProtocolInitData::Sign(_) => {
+        if self.agreed_mpc_protocol.is_none() {
+            return Ok(ReadyToAdvanceCheckResult {
+                is_ready: false,
+                malicious_parties: vec![],
+            });
+        }
+        match &self.agreed_mpc_protocol.clone().unwrap() {
```
zeevmoney: do not unwrap

zeevmoney (approved):

PR #1109, opened 2025-06-03
chore: Full hash on MPC message/output 

yaelAbergel:


ItayLevyOfficial (commented):
why have you added the session info to every message?

ItayLevyOfficial:
what i meant in the issue was to hash the session info when you already have it (in the output message), not to add it to every message

ItayLevyOfficial (approved):
lgtm

PR #1112, opened 2025-06-03
fix: Accept messages for the current crypto round

ItayLevyOfficial:


crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -846,7 +846,7 @@ impl DWalletMPCSession {
             .epoch_store()?
             .authority_name_to_party_id(&message.authority)?;
         // We should only receive outputs of previous rounds.
-        if message.round_number >= self.current_round {
+        if message.round_number > self.current_round {
```
zeevmoney, line 849: also add to the warn message:
                recieved_message_round_number=?message.round_number,

PR #1130, opened 2025-06-05
Improve move code

omersadika:


crates/ika-config/src/initiation.rs
```diff
@@ -97,6 +98,8 @@ impl InitiationParameters {
 
     fn default_chain_start_timestamp_ms() -> u64 {
         std::time::SystemTime::now()
+            .checked_sub(Duration::from_secs(24 * 60 * 60)) // subtract 24 hours
+            .unwrap()
```
ItayLevyOfficial: default chain start time is 24 hours ago? weird.

zeevmoney: Why start the chain in the past?

PR #1148, opened 2025-06-11
fix: Stop wiping outputs after receiving a late session event

ItayLevyOfficial:


crates/ika-core/src/consensus_handler.rs
```diff
@@ -424,9 +424,6 @@ impl<C: DWalletCheckpointServiceNotify + Send + Sync> ConsensusHandler<C> {
             .epoch_store
             .get_dwallet_mpc_outputs_verifier_write()
             .await;
-        for event in self.epoch_store.tables()?.get_all_dwallet_mpc_events()? {
```
ItayLevyOfficial, line 427: This call is not needed, as when receiving an output for a non-existing session its entry gets created automatically

yaelAbergel (approved):

PR #1154, opened 2025-06-11
chore: Use epoch first event as Sui event fetching cursor

ItayLevyOfficial:
Until now users had to fetch all the events from Sui since Sui's start. This PR fixes this issue, and allows validators to sync from the first TX of the current epoch.

### DIFF
--- a/crates/ika-core/src/sui_connector/sui_syncer.rs
+++ b/crates/ika-core/src/sui_connector/sui_syncer.rs
@@ -351,9 +351,7 @@ where
             // as it is unexpected to change often.
             if loop_index % 10 == 0 {
                 debug!("Querying epoch start cursor from Sui");
-                let system_inner = match sui_client.must_get_system_inner_object().await {
-                    SystemInner::V1(system_inner) => system_inner,
-                };
+                let SystemInner::V1(system_inner) = sui_client.must_get_system_inner_object().await;
                 let Ok(epoch_start_tx_digest) = system_inner.epoch_start_tx_digest.try_into()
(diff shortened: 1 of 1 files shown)


crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -343,8 +343,27 @@ where
...
+            if loop_index % 10 == 0 {
+                info!("Querying epoch start cursor from Sui");
+                let system_inner = match sui_client.must_get_system_inner_object().await {
+                    SystemInner::V1(system_inner) => system_inner,
+                };
+                let start_epoch_event = EventID::from((
+                    system_inner
+                        .epoch_start_tx_digest
+                        .try_into()
+                        .expect("start epoch TX digest in wrong length"),
```
zeevmoney: remove except, continue the loop, retry later...

zeevmoney (changes requested):
see comment

crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -343,8 +343,27 @@ where
                 last_synced_sui_checkpoints_metric.set(latest_checkpoint_sequence_number as i64);
             }
         });
-
+        let mut start_epoch_cursor: Option<EventID> = None;
+        let mut loop_index = 0;
         loop {
+            loop_index += 1;
+            if loop_index % 10 == 0 {
```
zeevmoney, line 352: add doc

PR #1159, opened 2025-06-12
fix: Fetch missed events from Sui every 5 minutes

ItayLevyOfficial:


crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -174,7 +174,13 @@ impl DWalletMPCService {
     /// The service automatically terminates when an epoch switch occurs.
     pub async fn spawn(&mut self) {
         self.load_missed_events().await;
+        let mut loop_index = 0;
         loop {
+            loop_index += 1;
+            // Load events from Sui every 5 minutes.
+            if loop_index % 3_000 == 0 {
+                self.load_missed_events().await;
+            }
```
yaelAbergel:
```suggestion
        let mut loop_index = 0;
        loop {
            // Load events from Sui every 5 minutes.
            if loop_index % 3_000 == 0 {
                self.load_missed_events().await;
            }
            loop_index += 1;
```

PR #1163, opened 2025-06-12
chore: Read dWallet MPC events from channel

yaelAbergel:


zeevmoney (approved):

crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -364,6 +357,8 @@ where
                     cursor = start_epoch_cursor;
                 }
             }
+            loop_index += 1;
```
ItayLevyOfficial, line 360: the start epoch cursor is suspicious imo - it may not work when fetching events from a different module

PR #1169, opened 2025-06-15
chore: MPC bench tests 

yaelAbergel:
Add a test to run N full flows in parallel and time each MPC session. To achieve this, I adapted the TypeScript code and split some methods into separate functions.


ItayLevyOfficial (changes requested):
Can you please explain in the pr title/description what's the purpose of this PR? Or link it to an issue?

sdk/typescript/src/dwallet-mpc/encrypt-user-share.ts
```diff
@@ -202,13 +202,30 @@ async function registerEncryptionKey(
...
+				await delay(5_000); // Wait 5 seconds before retrying
+				continue;
+			}
+			throw error; // If we've exceeded timeout, throw the error
+		}
+	}
+
+	if (!res) {
+		throw new Error(`Failed to get transaction response within ${conf.timeout / (60 * 1000)} minutes`);
+	}
```
ItayLevyOfficial, line 228: This is a bug. If a user cannot register an encryption key, for any reason, we should tell him what the error is - not try again & again & drain his Sui balance for those failing txs.

ItayLevyOfficial (changes requested):

PR #1172, opened 2025-06-16
chore: Replace all Move string error codes with numbers

### LINKED ISSUE #1171: replace all string error codes with numbers

ItayLevyOfficial:
the string error codes are impossible to debug through the explorer


zeevmoney:
Not needed eventually, @omersadika did it already.

PR #1175, opened 2025-06-17
chore: Cache the protocol public parameters

ItayLevyOfficial:
This will save us the time it takes to reconstruct it from the network key


sdk/typescript/src/dwallet-mpc/globals.ts
```diff
@@ -321,7 +327,7 @@ export async function getNetworkDecryptionKeyID(c: Config): Promise<string> {
 	return decryptionKeyID;
 }
 
-export function cacheNetworkKey(key_id: string, epoch: number, networkKey: Uint8Array) {
+export async function cachePublicParameters(key_id: string, epoch: number, networkKey: Uint8Array) {
```
ycscaly, line 330: @omersadika this is the only thing I'm not sure of, will this work in a browser?

ycscaly (approved):
I approve, but @omersadika should say if its fine to cache it this way

zeevmoney (approved):

PR #1210, opened 2025-06-24
fix: Handle reading of MPC network key reconfiguration output

yaelAbergel:
If a validator started running after the first network MPC key reconfiguration output had been written, it would consistently fail to read the key. As a result, it could never start an MPC session, because the session would remain permanently pending on the network key.


ycscaly (approved):

ItayLevyOfficial (approved):
lgtm

zeevmoney:
This one was eventually replaced by #1211

PR #1211, opened 2025-06-24
fix: Handle reading of MPC network key reconfiguration output with Move update

yaelAbergel:


omersadika (approved):

crates/ika-sui-client/src/lib.rs
```diff
@@ -1041,29 +1041,43 @@ impl SuiClientInner for SuiSdkClient {
             .read_table_vec_as_raw_bytes(key.network_dkg_public_output.contents.id)
             .await?;
 
-        let current_reconfiguration_public_output = if key.reconfiguration_public_outputs.size == 0
+        // Note that if we try to read the reconfiguration public output during the first epoch,
+        // where we only had NetworkDKG, this function will error.
+        // In this case, the validator will be stuck in a loop where it can't process events
+        // until the epoch is switched, since it will be endlessly waiting for the network key.
```
ItayLevyOfficial: So what happens in the first epoch? How can the validator fetch the network dkg output from the chain?

ItayLevyOfficial: I can't see where the function errors exactly also..

crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -270,7 +270,9 @@ where
                         continue 'sync_network_keys;
                     }
                     Err(err) => {
-                        warn!(
+                        // DO NOT CHANGE THIS TO WARNING!
+                        // THIS IS A CRITICAL ERROR IN SOME CASES!
```
ItayLevyOfficial, line 274: Those comments above an error seems strange for me

yaelAbergel: @zeevmoney

crates/ika-sui-client/src/lib.rs
```diff
@@ -1041,29 +1041,43 @@ impl SuiClientInner for SuiSdkClient {
...
+                is_first: true,
+            });
+        let awaiting_first_reconfiguration_to_complete = key.state
+            == (DWalletNetworkEncryptionKeyState::AwaitingNetworkReconfiguration {
+                is_first: true,
+            });
+        let no_reconfiguration_key_data = key.reconfiguration_public_outputs.size == 0;
+        let mut current_reconfiguration_public_output = vec![];
+
+        if no_reconfiguration_key_data
```
ycscaly, line 1059: Better to do `let x = if {vec![]} else {...}`, no reason to `mut` here

ycscaly (approved):

crates/ika-core/src/sui_connector/sui_syncer.rs
zeevmoney: It's needed.

crates/ika-sui-client/src/lib.rs
zeevmoney: I changed it, it was unreadable.

zeevmoney (approved):

PR #1213, opened 2025-06-25
fix: System Checkpoint Sync

yaelAbergel:


crates/ika-network/src/state_sync/mod.rs
```diff
@@ -178,8 +178,10 @@ struct PeerStateSyncInfo {
     chain_identifier: ChainIdentifier,
     /// Indicates if this Peer is on the same chain as us.
     on_same_chain_as_us: bool,
-    /// Highest checkpoint sequence number we know of for this Peer.
+    /// Highest dwallet checkpoint sequence number we know of for this Peer.
     height: Option<DWalletCheckpointSequenceNumber>,
```
omersadika: Rename to dwallet_checkpoint_height

ycscaly (approved):

PR #1224, opened 2025-06-29
1223 fix deserialization fix

ItayLevyOfficial:


zeevmoney (approved):

crates/ika-sui-client/src/lib.rs
```diff
@@ -373,25 +373,20 @@ where
                     .members
                     .iter()
                     .map(|m| {
-                        let validator = validators
-                            .iter()
-                            .find(|v| v.id == m.validator_id)
-                            .unwrap();
+                        let validator = validators.iter().find(|v| v.id == m.validator_id).unwrap();
```
zeevmoney: remove this unwarp()_

zeevmoney (changes requested):
see comments

crates/ika-sui-client/src/lib.rs
zeevmoney: return a descriptive error.

PR #1233, opened 2025-06-30
chore: Remove redundant class groups keys fetching 

ItayLevyOfficial:
Until now, we fetched the class group keys of all validators every time we retrieved a new network key, without any clear reason. This PR addresses the issue by fetching the keys only once per epoch and reusing them thereafter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -541,7 +541,7 @@ ika-protocol-config = { path = "crates/ika-protocol-config" }
 ika-protocol-config-macros = { path = "crates/ika-protocol-config-macros" }
 #ika-sdk = { path = "crates/ika-sdk" }
 ika-swarm = { path = "crates/ika-swarm" }
-ika-swarm-config = { path = "crates/ika-swarm-config", features = ["mock-class-groups"] }
+ika-swarm-config = { path = "crates/ika-swarm-config" }
 ika-telemetry = { path = "crates/ika-telemetry" }
 ika-types = { path = "crates/ika-types" }
 ika-move-packages = { path = "crates/ika-move-packages" }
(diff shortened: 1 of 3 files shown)


zeevmoney (approved):

zeevmoney (approved):

crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -192,10 +187,11 @@ where
     /// Sync the DwalletMPC network keys from the Sui client to the local store.
     async fn sync_dwallet_network_keys(
         sui_client: Arc<SuiClient<C>>,
-        network_keys_sender: Sender<Arc<HashMap<ObjectID, NetworkDecryptionKeyPublicData>>>,
+        network_keys_sender: Sender<Arc<HashMap<ObjectID, DWalletNetworkDecryptionKeyData>>>,
     ) {
-        // (Key Obj ID, Epoch)
-        let mut network_keys_cache: HashSet<(ObjectID, u64)> = HashSet::new();
+        // Cache for network keys by (ObjectID, Epoch) to avoid redundant fetching.
+        let mut network_keys_cache: HashMap<(ObjectID, u64), DWalletNetworkDecryptionKeyData> =
```
yaelAbergel: There is no need to save an extra copy of the key

crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -225,67 +221,30 @@ where
...
             }
-            let active_bls_committee = system_inner.get_ika_active_committee();
-            let active_committee = system_inner.read_bls_committee(&active_bls_committee);
-            let current_keys = system_inner.dwallet_2pc_mpc_coordinator_network_encryption_keys();
-            let should_fetch_keys = current_keys.iter().any(|key| {
-                !network_keys_cache
-                    .contains(&(key.dwallet_network_decryption_key_id, system_inner.epoch()))
+            let should_fetch_keys = network_encryption_keys.values().any(|key| {
+                match network_keys_cache.get(&(key.id, system_inner.epoch())) {
+                    Some(cached_key) => cached_key.state != key.state,
```
yaelAbergel: The state mismatch doesn't have to affect the key fetch, some states indicate the next epoch not the current one. As long as the cached key is in the current epoch and the state is different then `WaitingNetworkDKG` we can read it.

PR #1234, opened 2025-06-30
Derive a One-Time Use, MPC Protocol and Round Specific, Deterministic Random Generator for `advance()`

ycscaly:
This is an important security fix, as we might run `advance()` twice, and it is unsafe for us to send two messages for the same round and the same session. So by using a deterministic rng, we assure the same message will be created in that case.

Another important reason is for debugging - given only the `private_seed`, we can deterministically reconstruct the entire MPC protocol after a bug occurred, so we can debug it line-by-line. The validators just need to save this seed, and export it to us in the case some bug happened (or debug it themselves locally).


crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -143,19 +150,31 @@ impl DWalletMPCManager {
...
         };
         let dwallet_network_keys = DwalletMPCNetworkKeys::new(validator_private_data);
+
+        // Derive the private part of the deterministic seed for this session from the class-groups decryption key of the validator.
+        // This seed is derived using Keccak256, a one-way hash function, which means the key can never be recovered from that value.
+        //
+        // However, this seed is the secret material from which the randomness in all MPC protocols will be generated. As such,
+        // it *must be kept private*, otherwise all of this validator's secrets in the MPC protocol will be leaked!
+        let serialized_decryption_key = bcs::to_bytes(&class_groups_decryption_key)?;
+        let private_seed = sha3::Keccak256::new_with_prefix(&serialized_decryption_key)
```
ycscaly: @omersadika a much better idea is to use the validator's seed directly, and not a hash on the class-groups keys.

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -438,6 +449,23 @@ impl DWalletMPCSession {
...
+        transcript.append_u64(b"$ pid $", self.party_id as u64);
+        transcript.append_uint(b"$ sid $", &session_identifier);
+        transcript.append_u64(b"$ current round $", self.current_round as u64);
+        transcript.append_u64(b"$ attempts count $", self.attempts_count as u64);
+        transcript.append_u64(b"$ epoch $", self.epoch_id as u64);
+
+        let mut seed: [u8; 32] = [0; 32];
+        transcript.challenge_bytes(b"seed", &mut seed);
+
+        let rng = ChaCha20Rng::from(ChaCha20Core::from_seed(seed));
```
ItayLevyOfficial: extract to a separate function, ideally in the cryptography private repository.

ItayLevyOfficial: What's the point of adding all those fields to the private seed anyway?

ItayLevyOfficial (approved):
lgtm

crates/ika-core/src/dwallet_mpc/mpc_session.rs
ycscaly: Can't be in cryptography, because it depends on network-specific information like the epoch etc.
We need this to make sure the rng is unique, and we never use the same one for different sessions

crates/dwallet-rng/src/lib.rs
```diff
@@ -1,11 +1,17 @@
...
 use merlin::Transcript;
+use proof::TranscriptProtocol;
 use rand_chacha::rand_core::{RngCore, SeedableRng};
 use rand_chacha::ChaCha20Rng;
 use serde::{Deserialize, Serialize};
 
+/// The Root Seed for this validator, used to deterministically derive purpose-specific child seeds
+/// for all cryptographically-secure random generation operations.
+///
+/// SECURITY NOTICE: *MUST KEEP PRIVATE*.
```
zeevmoney:
```suggestion
/// SECURITY NOTICE: *MUST BE KEPT PRIVATE*.
```

omersadika (approved):

crates/dwallet-rng/src/lib.rs
```diff
@@ -64,11 +69,65 @@ impl RootSeed {
         seed
     }
 
+    /// Derive a seed deterministically for advancing an MPC round.
+    ///
+    /// We don't use the root seed directly, as it would be used for other purposes.
```
zeevmoney:
```suggestion
    /// We don't use the root seed directly, as it may be used for other purposes.
```

yaelAbergel (approved):

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -144,21 +149,24 @@ impl DWalletMPCManager {
...
         let mpc_computations_orchestrator = CryptographicComputationsOrchestrator::try_new()?;
         let party_id = epoch_store.authority_name_to_party_id(&epoch_store.name)?;
+        let root_seed = node_config
+            .root_seed
+            .clone()
+            // Since only a validator executes the DWalletMPCManager, we can unwrap
+            // the `root_seed`.
+            .expect("Root seed must be present")
+            .root_seed()
+            .clone();
```
zeevmoney:
```suggestion
        let root_seed = node_config
            .root_seed
            .clone()
            .ok_or(DwalletMPCError::MissingRootSeed)?
            .root_seed()
            .clone();
```
This function returns result, it's very easy to avoid except()
This error is not in the enum, needs to be added

crates/dwallet-rng/src/lib.rs
```diff
@@ -64,11 +69,65 @@ impl RootSeed {
...
+    ///
+    /// We don't use the root seed directly, as it may be used for other purposes.
+    /// Instead, we derive a seed from it using a distinct hard-coded label.
+    fn mpc_round_seed(
+        &self,
+        session_identifier: CommitmentSizedNumber,
+        party_id: u64,
+        current_round: u64,
+        attempts_count: u64,
+        epoch_id: u64,
```
dolevmu, line 82: I think these last two maybe can cause a party to send to different messages in the same session same round if it does state sync. Why do we need those?

dolevmu (commented):
See my comment!

PR #1236, opened 2025-06-30
fix: Make imported DWallet verification flow use session ID

ItayLevyOfficial:


crates/ika-types/src/messages_dwallet_mpc.rs
```diff
@@ -613,8 +613,7 @@ pub struct DWalletSessionEvent<E: DWalletSessionEventTrait> {
     pub epoch: u64,
     pub session_object_id: ObjectID,
     pub session_type: SessionType,
-    // DO NOT MAKE THIS PUBLIC! ONLY CALL `session_identifier_digest`
```
ItayLevyOfficial, line 616: @omersadika This is needed as you are using this field in the centralized part of the imported dwallet flow - Do you have any alternative in mind?

PR #1239, opened 2025-06-30
fix: Remove redundant dynamic fields fetching logic while fetching class groups keys

ItayLevyOfficial:
Until now all the validator's object dynamic fields has been fetched from Sui while fetching the validators, without any clear reason. This pr removes this redundant logic

### DIFF
--- a/crates/ika-sui-client/src/lib.rs
+++ b/crates/ika-sui-client/src/lib.rs
@@ -334,7 +334,7 @@ where
 
                 let validators = self
                     .inner
-                    .get_validators_from_object_table(validator_ids)
+                    .get_validators(validator_ids)
                     .await
                     .map_err(|e| {
                         IkaError::SuiClientInternalError(format!(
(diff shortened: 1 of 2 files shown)


crates/ika/src/ika_commands.rs
```diff
@@ -23,7 +23,7 @@ use tokio::runtime::Runtime;
 use tracing::info;
 
 // 24 Hours.
-const DEFAULT_EPOCH_DURATION_MS: u64 = 1000 * 60 * 60 * 24;
```
zeevmoney: STOP CHANGING IT, USE FLAGS

crates/ika-sui-client/src/lib.rs
```diff
@@ -1312,49 +1303,11 @@ impl SuiClientInner for SuiSdkClient {
 
     async fn get_validators_from_object_table(
```
omersadika: Rename to get_validators

zeevmoney (approved):

PR #1246, opened 2025-07-08
Refactor Event Handling

ycscaly:
This PR moves both code & responsibility that is related to event handling, from the MPC Service into the MPC Manager (implemented in `mpc_event.rs`)

Events are handled in a batch, and before we handle any event in a batch we try to fetch the network keys & next committee, for which events might me be pending.
Then, we only handle an event and try to create its `mpc_event_data` from it if we know its going to succeed (i.e. all its requirement are met, otherwise it is queued).

It adds an explicit security check that events come from the right SUI module.

It fixes event deserialization, clearly separating between the cases of a pulled or pushed event, and not trying and retrying upon failure (which is bad practice!)
Service reads events as `DBSuiEvent`, just collects them, notes if they are pulled or pushed, and parsing is done in `handle_dwallet_db_event()`
Then we also only check the event type_ once, after which we already have an enum when we create the `mpc_event_data` so we use that.

It also actually orders the sessions that are ready to advance, which wasn't the case previously. The order is as such:
System events first, and FIFO between the system events.
User events are after all system events, and the order between them is by the (lowest) sequence number.

It also removes a lot of redundant async from sync functions.


crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -59,6 +52,7 @@ impl DWalletMPCService {
         let dwallet_mpc_manager = DWalletMPCManager::must_create_dwallet_mpc_manager(
             consensus_adapter.clone(),
             epoch_store.clone(),
+            network_keys_receiver,
```
ycscaly, line 55: mpc manager now receives network keys, which makes sense because it handles event and some event require the network key

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -254,7 +138,6 @@ impl DWalletMPCService {
                 // This signifies a bug, we can't proceed before we fix it.
                 break;
             }
-            self.update_network_keys().await;
```
ycscaly, line 250: happens in mpc manager inside `handle_dwallet_db_events()`

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -380,61 +264,4 @@ impl DWalletMPCService {
             session.status = MPCSessionStatus::Finished;
         }
     }
-
-    /// Read events from perpetual tables, remove them, and store in the current epoch tables.
-    fn receive_new_sui_events(&mut self) -> IkaResult<Vec<DWalletMPCEvent>> {
```
ycscaly, line 374: Moved to `mpc_event.rs`

crates/ika-core/src/dwallet_mpc/mod.rs
```diff
@@ -116,15 +117,17 @@ pub(crate) fn party_ids_to_authority_names(
...
 /// fetch the event's the dynamic field directly from Sui.
-/// This function first tried to deserialize the event as a [`DWalletSessionEvent`], and if it fails,
-/// it tries to deserialize it as a [`Field<ID, DWalletSessionEvent<T>>`].
-fn deserialize_event_or_dynamic_field<T: DeserializeOwned + DWalletSessionEventTrait>(
+fn deserialize_event_contents<T: DeserializeOwned + DWalletSessionEventTrait>(
     event_contents: &[u8],
+    pulled: bool,
 ) -> Result<DWalletSessionEvent<T>, bcs::Error> {
-    bcs::from_bytes::<DWalletSessionEvent<T>>(event_contents).or_else(|_| {
+    // TODO(Scaly): unit test
```
ycscaly: @ycscaly I need to do this

crates/ika-core/src/dwallet_mpc/mod.rs
```diff
@@ -116,15 +117,17 @@ pub(crate) fn party_ids_to_authority_names(
 
 /// The type of the event is different when we receive an emitted event and when we
 /// fetch the event's the dynamic field directly from Sui.
-/// This function first tried to deserialize the event as a [`DWalletSessionEvent`], and if it fails,
-/// it tries to deserialize it as a [`Field<ID, DWalletSessionEvent<T>>`].
-fn deserialize_event_or_dynamic_field<T: DeserializeOwned + DWalletSessionEventTrait>(
+fn deserialize_event_contents<T: DeserializeOwned + DWalletSessionEventTrait>(
```
ycscaly: The old logic was extremely error prone. You never do this, try to do something, hope it works, and retry a different way if it fails. You need to know why and how things work and fail, and do things you know will work. Here, there were two ways we got events, one is that we pulled uncompleted events from the state, the other is that emitted events were pushed to us - so I added a `bool` in the event structs upon creation and propogated here, and now we know for sure how to deserialize it.

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,418 @@
...
+            }
+        }
+
+        for event in events {
+            self.handle_sui_event(event, epoch_store);
+        }
+    }
+
+    fn handle_sui_event(&mut self, event: DBSuiEvent, epoch_store: &AuthorityPerEpochStore) {
+        if event.type_.address != *epoch_store.packages_config.ika_system_package_id
```
ycscaly: This kind of check was implicit later in the code (when comparing the type of the event with the packages config), but its an exteremly important security check and it must be highlighted and explicit. Otherwise, later someone might remove some code that wasn't commented to have security effects, and a vulnerability could be inserted unknowingly.

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,421 @@
...
+            {
+                // We don't yet have the data for this network encryption key, so we add it to the queue.
+                debug!(
+                    session_request=?event.session_request,
+                    session_type=?event.session_request.session_type,
+                    network_encryption_key_id=?network_encryption_key_id,
+                    "Adding event to pending for the network key"
+                );
+
+                // TODO: need to check it doesn't exist first
```
ycscaly: @ycscaly

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,421 @@
...
+            && self.next_active_committee.is_none()
+        {
+            // We don't have the next active committee yet, so we have to add this event to the pending queue until it arrives.
+            debug!(
+                session_request=?event.session_request,
+                session_type=?event.session_request.session_type,
+                "Adding event to pending for the next epoch active committee"
+            );
+
+            // TODO: need to check it doesn't exist first
```
ycscaly: @ycscaly

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -203,49 +215,6 @@ impl DWalletMPCManager {
...
 
-    /// Handle an MPC event.
-    ///
-    /// This function might be called more than once for a given session,
-    /// as we periodically check for uncompleted events.
-    /// A new MPC session is only created once, the first time the event was received.
-    /// If the event already exists in `self.mpc_sessions`, we do not add it.
-    /// If there is no session info, and we've got it in this call, we update that field
-    /// in the open session.
-    pub(crate) async fn handle_dwallet_db_event(&mut self, event: DWalletMPCEvent) {
```
ycscaly, line 227: moved to `mpc_event.rs`

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -548,43 +427,7 @@ impl DWalletMPCManager {
...
     /// If no local CPUs are available, computations will execute as CPUs are freed.
     pub(crate) async fn perform_cryptographic_computation(&mut self) {
-        for event in self
-            .events_pending_for_network_key
-            .drain(..)
-            .collect::<Vec<_>>()
-            .into_iter()
-        {
-            self.handle_dwallet_db_event(event.clone()).await;
-        }
```
ycscaly, line 572: I mean this is really funny, we shouldn't just try and retry and retry until we succeed. There should be a succeess criterion we know how to check, and we should know if something should fail or succeed in advance of doing it, and not just do random things and hope for the best

ycscaly: Also, this event handling logic has nothing to do with `perform_cryptographic_computation`

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -594,12 +437,18 @@ impl DWalletMPCManager {
...
+                .pop_front()
+                .unwrap();
             // Safe to unwarp since the session was ready to compute.
             let live_session = self
                 .mpc_sessions
                 .get(&oldest_pending_session.session_identifier)
                 .unwrap();
+
+            // TODO(Scaly): What does it even mean for a session to be non-active, if its already ready for advance?
+            // TODO - if its finished, it shouldn't be here, and also failed what??
```
ycscaly: @ItayLevyOfficial @yaelAbergel @zeevmoney ?

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -772,12 +622,129 @@ impl DWalletMPCManager {
         self.mpc_sessions.insert(*session_identifier, new_session);
     }
 
-    pub(super) async fn must_get_next_active_committee(&self) -> Committee {
-        self.next_epoch_committee_receiver
-            .clone()
-            .wait_for(|committee| committee.epoch == self.epoch_id + 1)
```
ycscaly, line 791: I don't like this - we `must wait`? and block? why, we can simply get it when it comes, and do other stuff in the meantime.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -772,12 +622,130 @@ impl DWalletMPCManager {
...
+                vec![]
+            }
+        }
+    }
+
+    /// Insert `session` into `self.ordered_sessions_pending_for_computation`, keeping order:
+    /// System sessions come first, and user sessions are sorted by their sequence number.
+    ///
+    /// Note: at this point, `mpc_event_data` must be set!
+    pub(crate) fn insert_session_into_ordered_pending_for_computation_queue(
```
ycscaly, line 706: @ycscaly I should unit-test this

zeevmoney (changes requested):
Write down some description for why this refactor is needed.

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -81,10 +81,70 @@ pub enum PublicInput {
 #[derive(Clone)]
 pub struct MPCEventData {
     pub private_input: MPCPrivateInput,
-    pub init_protocol_data: MPCProtocolInitData,
-    pub(crate) decryption_shares: HashMap<PartyID, <AsyncProtocol as Protocol>::DecryptionKeyShare>,
+    pub request_input: MPCRequestInput,
+    pub(crate) decryption_key_shares:
```
ycscaly, line 85: If something is optional, it should be `Option`, not an empty `HashMap`, this is dangerous.

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
```diff
@@ -96,188 +73,133 @@ pub(crate) async fn session_input_from_event(
...
                 PublicInput::NetworkEncryptionKeyDkg(network_dkg_public_input(
-                    &dwallet_mpc_manager
-                        .epoch_store()?
-                        .get_weighted_threshold_access_structure()?,
-                    dwallet_mpc_manager
-                        .validators_class_groups_public_keys_and_proofs
-                        .clone(),
+                    &epoch_store.get_weighted_threshold_access_structure()?,
+                    validators_class_groups_public_keys_and_proofs,
+                    // TODO: make generic
```
ycscaly, line 85: issue number?

ycscaly:
```suggestion
                    // Todo (#473): Support generic network key scheme
```

crates/ika-core/src/dwallet_mpc/mpc_session/session_info.rs
ycscaly, line 1: moved to `event_handling.rs`

crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -1617,20 +1617,20 @@ impl AuthorityPerEpochStore {
...
             "Creating session output checkpoint transaction"
         );
         let (is_rejected, output) = match bcs::from_bytes(&output)? {
             MPCSessionPublicOutput::CompletedSuccessfully(output) => (false, output),
             MPCSessionPublicOutput::SessionFailed => (true, vec![]),
         };
-        match &session_info.mpc_round {
-            MPCProtocolInitData::DKGFirst(event_data) => {
+        match &session_request.request_input {
+            MPCRequestInput::DKGFirst(event_data) => {
```
yaelAbergel: It might also be nice if you renamed all occurrences of `event_data`/`init_event_data` within MPCRequestInput to use a consistent name, such as `session_req_event`.

crates/ika-core/src/dwallet_mpc/mpc_protocols/network_dkg.rs
yaelAbergel, line 1: Ideally, key storage should be moved to a separate file so this one focuses solely on network DKG logic.

crates/ika-core/src/dwallet_mpc/mpc_protocols/network_dkg.rs
```diff
@@ -191,13 +191,29 @@ impl DwalletMPCNetworkKeys {
...
+    }
+
     /// Retrieves the protocol public parameters for the specified key ID.
     pub fn get_protocol_public_parameters(
         &self,
         key_id: &ObjectID,
     ) -> DwalletMPCResult<twopc_mpc::secp256k1::class_groups::ProtocolPublicParameters> {
         let Some(result) = self.network_encryption_keys.get(key_id) else {
-            warn!(
+            debug!(
```
yaelAbergel: Why did you lower the log level?

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
```diff
@@ -14,53 +13,41 @@ use commitment::CommitmentSizedNumber;
...
-        {
-            let deserialized_event: DWalletSessionEvent<
-                DWalletImportedKeyVerificationRequestEvent,
-            > = deserialize_event_or_dynamic_field(&event.contents)?;
-
-            let protocol_public_parameters = dwallet_mpc_manager.get_protocol_public_parameters(
+    let session_id = CommitmentSizedNumber::from_le_slice(
+        event.session_request.session_identifier.to_vec().as_slice(),
+    );
+    match event.session_request.request_input {
```
yaelAbergel, line 41: :)

crates/ika-core/src/dwallet_mpc/cryptographic_computations_orchestrator.rs
```diff
@@ -146,19 +146,19 @@ impl CryptographicComputationsOrchestrator {
         let handle = Handle::current();
         let mut session = session.clone();
         // Safe to unwrap here (event must exist before this).
-        let mpc_event_data = session.mpc_event_data.clone().unwrap().init_protocol_data;
+        let mpc_event_data = session.mpc_event_data.clone().unwrap().request_input;
```
yaelAbergel:
```suggestion
```

You can use only `request_input`

crates/ika-core/src/dwallet_mpc/mod.rs
```diff
@@ -114,19 +115,6 @@ pub(crate) fn party_ids_to_authority_names(
...
-fn deserialize_event_or_dynamic_field<T: DeserializeOwned + DWalletSessionEventTrait>(
-    event_contents: &[u8],
-) -> Result<DWalletSessionEvent<T>, bcs::Error> {
-    bcs::from_bytes::<DWalletSessionEvent<T>>(event_contents).or_else(|_| {
-        bcs::from_bytes::<Field<ID, DWalletSessionEvent<T>>>(event_contents)
-            .map(|field| field.value)
-    })
-}
-
 // TODO (#683): Parse the network key version from the network key object ID
```
yaelAbergel: Please delete this code as well

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,501 @@
...
+    /// check for uncompleted events - in which case the event will be ignored.
+    ///
+    /// A new MPC session is only created once at the first time the event was received
+    /// (per-epoch, if it was uncompleted in the previous epoch it will be created again for the next one.)
+    ///
+    /// If the event already exists in `self.mpc_sessions`, we do not add it.
+    ///
+    /// If there is no session info, and we've got it in this call,
+    /// we update that field in the open session.
+    pub(crate) fn handle_dwallet_db_events(
```
yaelAbergel: Consider renaming  `handle_dwallet_events_batch`

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -772,12 +622,130 @@ impl DWalletMPCManager {
...
+            }
+            Err(err) => {
+                error!(?err, "failed to check next epoch committee receiver");
+            }
+        }
+
+        false
+    }
+
+    pub(crate) fn update_network_keys(&mut self) -> Vec<ObjectID> {
```
yaelAbergel:
```suggestion
    pub(crate) fn maybe_update_network_keys(&mut self) -> Vec<ObjectID> {
```

crates/ika-core/src/dwallet_mpc/mod.rs
zeevmoney, line 1: remove

```
warning: unused imports: `DWalletSessionEventTrait` and `DWalletSessionEvent`
 --> crates/ika-core/src/dwallet_mpc/mod.rs:6:39
  |
6 | ...::{DWalletSessionEvent, DWalletSessionEventTrai...
  |       ^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `serde::de::DeserializeOwned`
 --> crates/ika-core/src/dwallet_mpc/mod.rs:9:5
  |
9 | use serde::de::DeserializeOwned;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused import: `sui_types::dynamic_field::Field`
  --> crates/ika-core/src/dwallet_mpc/mod.rs:15:5
   |
15 | use sui_types::dynamic_field::Fie...
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: unused import: `sui_types::id::ID`
  --> crates/ika-core/src/dwallet_mpc/mod.rs:16:5
   |
16 | use sui_types::id::ID;
   |     ^^^^^^^^^^^^^^^^^
  ```

File: crates/ika-core/src/dwallet_mpc/mpc_manager.rs [hunk 23]
```diff
@@ -66,17 +66,20 @@ pub(crate) struct DWalletMPCManager {
     /// is part of the [`AuthorityPerEpochStore`].
     pub(crate) malicious_handler: MaliciousHandler,
     /// The order of the sessions that are ready to get computed.
-    pub(crate) pending_for_computation_order: VecDeque<DWalletMPCSession>,
+    pub(crate) ordered_sessions_pending_for_computation: VecDeque<DWalletMPCSession>,
     /// The order of the sessions that have received quorum for their current round, but we have not
     /// yet received an event for from Sui.
-    pub(crate) pending_for_events_order: VecDeque<DWalletMPCSession>,
+    pub(crate) sessions_pending_for_events: VecDeque<DWalletMPCSession>,
```
yaelAbergel, line 76: If they are not ordered as the name suggest, you should also update the documentation above

zeevmoney (changes requested):
This PR doesn't pass clippy

```bash
cargo fmt && cargo clippy --all-features && cargo check && cargo build
```

Please run all of these, fix the errors.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -66,17 +66,20 @@ pub(crate) struct DWalletMPCManager {
     /// is part of the [`AuthorityPerEpochStore`].
     pub(crate) malicious_handler: MaliciousHandler,
     /// The order of the sessions that are ready to get computed.
-    pub(crate) pending_for_computation_order: VecDeque<DWalletMPCSession>,
+    pub(crate) ordered_sessions_pending_for_computation: VecDeque<DWalletMPCSession>,
```
yaelAbergel, line 73: Maybe add the order rules from the PR description to the documentation above

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,535 @@
...
+use tracing::{debug, error, warn};
+
+impl DWalletMPCManager {
+    /// Handle a batch of MPC events.
+    ///
+    /// This function might be called more than once for a given session, as we periodically
+    /// check for uncompleted events - in which case the event will be ignored.
+    ///
+    /// A new MPC session is only created once at the first time the event was received
+    /// (per-epoch, if it was uncompleted in the previous epoch it will be created again for the next one.)
```
zeevmoney:
```suggestion
    /// A new MPC session is only created once at the first time the event was received
    /// (per-epoch, if it was uncompleted in the previous epoch,
    /// it will be created again for the next one.)
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -641,7 +490,8 @@ impl DWalletMPCManager {
                 error!(
                     session_identifier=?oldest_pending_session.session_identifier,
                     last_session_to_complete_in_current_epoch=?self.last_session_to_complete_in_current_epoch,
-                    mpc_protocol=?mpc_event_data.init_protocol_data,
+                    session_type=?mpc_event_data.session_type,
+                    mpc_protocol=?mpc_event_data.request_input,
```
yaelAbergel, line 493:
```suggestion
                    session_type=?mpc_event_data.session_type,
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,535 @@
...
+    ///
+    /// If the event already exists in `self.mpc_sessions`, we do not add it.
+    ///
+    /// If there is no session info, and we've got it in this call,
```
zeevmoney:
```suggestion
    /// If there is no `session_request`, and we've got it in this call,
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,535 @@
...
+    /// we update that field in the open session.
+    pub(crate) fn handle_dwallet_db_events(
+        &mut self,
+        events: Vec<DBSuiEvent>,
+        epoch_store: &AuthorityPerEpochStore,
+    ) {
+        // We only update `next_active_committee` here, so once we update it there is no longer going to be any pending events for it in this epoch.
```
zeevmoney:
```suggestion
        // We only update `next_active_committee` in this block. Once it's set,
        // there will no longer be any pending events targeting it for this epoch.
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,535 @@
...
+        if self.next_active_committee.is_none() {
+            let got_next_active_committee = self.try_receiving_next_active_committee();
+            if got_next_active_committee {
+                // `..` stands for `RangeFull`, and calling `drain(..)` will drain (i.e. mutate) the vector thus removing all events from it.
```
zeevmoney, line 58:
```suggestion
```
this is obvious

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,535 @@
...
+                    .drain(..)
+                    .collect();
+
+                for event in events_pending_for_next_active_committee {
+                    self.handle_mpc_event(event, epoch_store);
+                }
+            }
+        }
+
+        // First try to update the network keys.
```
zeevmoney:
```suggestion
        // First, try to update the network keys.
```

crates/ika-core/src/dwallet_mpc/mod.rs
ycscaly: done

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,535 @@
...
+        let newly_updated_network_keys_ids = self.update_network_keys();
+
+        // Then handle events for which we just received the public data for.
+        // Since events are only added to the `events_pending_for_network_key` queue in this function,
+        // we know once we got the network key data no more will be pending for that key,
+        // so its safe to only handle these at the time of update, as it will remain an empty queue afterward.
```
zeevmoney:
```suggestion
	// Now handle events for which we've just received the corresponding public data.
	// Since events are only queued in `events_pending_for_network_key` within this function,
	// receiving the network key ensures no further events will be pending for that key.
	// Therefore, it's safe to process them now, as the queue will remain empty afterward.
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,535 @@
...
+        for key_id in newly_updated_network_keys_ids {
+            let events_pending_for_newly_updated_network_key = self
+                .events_pending_for_network_key
+                .remove(&key_id)
+                .unwrap_or_default();
+
+            for event in events_pending_for_newly_updated_network_key {
+                // We know this won't fail on missing network key, but it could be waiting for the next committee,
```
zeevmoney:
```suggestion
                // We know this won't fail on a missing network key,
                // but it could be waiting for the next committee,
                // in which case it would be added to that queue.
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+        self.handle_mpc_event(event, epoch_store)
+    }
+
+    /// Handle an MPC event.
+    ///
+    /// This function might be called more than once for a given session, as we periodically
+    /// check for uncompleted events.
+    ///
+    /// A new MPC session is only created once at the first time the event was received
+    /// (per-epoch, if it was uncompleted in the previous epoch it will be created again for the next one.)
```
zeevmoney:
```suggestion
    /// (per-epoch, if it was uncompleted in the previous epoch, it will be created again for the next one.)
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+    ///
+    /// If the event already exists in `self.mpc_sessions`, we do not add it.
+    ///
+    /// If there is no session info, and we've got it in this call,
```
zeevmoney:
```suggestion
    /// If there is no `session_request`, and we've got it in this call,
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+    /// we update that field in the open session.
+    fn handle_mpc_event(&mut self, event: DWalletMPCEvent, epoch_store: &AuthorityPerEpochStore) {
+        let session_identifier = event.session_request.session_identifier;
+
+        // Avoid instantiation of completed events by checking they belong to the current epoch.
+        // We only pull uncompleted events, so we skip the check for those, but pushed events might be completed.
```
zeevmoney:
```suggestion
        // Avoid instantiation of completed events by checking they belong to the current epoch.
        // We only pull uncompleted events, so we skip the check for those,
        // but pushed events might be completed.
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+            if let Some(network_encryption_key_id) = event
+                .session_request
+                .request_input
+                .get_network_encryption_key_id()
+            {
+                if !self
+                    .network_keys
+                    .key_public_data_exists(&network_encryption_key_id)
+                {
+                    // We don't yet have the data for this network encryption key, so we add it to the queue.
```
zeevmoney:
```suggestion
                    // We don't yet have the data for this network encryption key, 
                    // so we add it to the queue.
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+
+                    return;
+                }
+            }
+        }
+
+        if event.session_request.requires_next_active_committee
+            && self.next_active_committee.is_none()
+        {
+            // We don't have the next active committee yet, so we have to add this event to the pending queue until it arrives.
```
zeevmoney:
```suggestion
            // We don't have the next active committee yet, 
            // so we have to add this event to the pending queue until it arrives.
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+        let mpc_event_data = match MPCEventData::try_new(
+            event.clone(),
+            epoch_store,
+            &self.network_keys,
+            self.next_active_committee.clone(),
+            self.validators_class_groups_public_keys_and_proofs.clone(),
+        ) {
+            Ok(mpc_event_data) => mpc_event_data,
+            Err(e) => {
+                error!(e=?e, event=?event, "failed to handle dWallet MPC event with error");
```
zeevmoney:
```suggestion
                error!(error=?e, event=?event, "failed to handle dWallet MPC event with error");
```
keep it `error` everywhere, easier to filter it

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+
+                return;
+            }
+        };
+
+        self.dwallet_mpc_metrics
+            .add_received_event_start(&mpc_event_data.request_input);
+
+        if let Some(session) = self.mpc_sessions.get_mut(&session_identifier) {
+            if session.mpc_event_data.is_none() {
```
zeevmoney: why do you check it again, you just did it on line 229? the code is sync, there is no way that someone could push this data between the calls

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+                session.mpc_event_data = Some(mpc_event_data.clone());
+
+                // It could be that this session was pending for computation, but was missing the event.
```
zeevmoney:
```suggestion
                // It could be that this session was pending
                // for computation but was missing the event.
                // In that case, move it to the right queue.
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+                        ready_to_advance_session_copy,
+                    );
+                }
+            }
+        } else {
+            self.new_mpc_session(&session_identifier, Some(mpc_event_data));
+        }
+    }
+
+    /// Parses a SUI event into a dWallet MPC event.
```
zeevmoney:
```suggestion
    /// Parses a Sui event into a dWallet MPC event.
```
SUI is the token.

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+        };
+
+        Ok(event)
+    }
+}
+
+impl DWalletMPCService {
+    /// Proactively pull uncompleted events from the Sui network.
+    /// We do that to assure we don't miss any events.
+    /// These events might be from a different Epoch, not necessarily the current one.
```
zeevmoney:
```suggestion
    /// Proactively pull uncompleted events from the Sui network.
    /// We do that to ensure we don't miss any events.
    /// These events might be from a different Epoch, not necessarily the current one
```

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,533 @@
...
+                        );
+                    }
+                    return events;
+                }
+                Err(err) => {
+                    error!(
+                        ?err,
+                        current_epoch=?self.epoch_store.epoch(),
+                         "Failed to load missed events from Sui"
+                    );
```
zeevmoney:
```suggestion
                    error!(
                        error=?err,
                        current_epoch=?self.epoch_store.epoch(),
                         "failed to load missed events from Sui"
                    );
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -280,25 +162,26 @@ impl DWalletMPCService {
...
                     continue;
                 }
             };
 
-            for (round, messages) in mpc_msgs_iter {
+            // Sort the MPC messages by round in ascending order.
+            mpc_messages.sort_by(|(round, _), (other_round, _)| round.cmp(other_round));
+
+            for (round, messages) in mpc_messages {
+                // Since we sorted, this assures this variable will be the last read in this batch when we are done iterating.
```
zeevmoney:
```suggestion
                // Since we sorted, this assures this variable will be the 
                // last read in this batch when we are done iterating.
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -165,6 +171,9 @@ impl DWalletMPCManager {
         };
         let dwallet_network_keys = DwalletMPCNetworkKeys::new(validator_private_data);
 
+        // Re-initialize the malicious handler every epoch. This is done intentionally:
+        // we want to "forget" the malicious actors from the previous epoch and start from scratch.
```
zeevmoney:
```suggestion
        // Re-initialize the malicious handler every epoch. This is done intentionally:
        // We want to "forget" the malicious actors from the previous epoch and start from scratch.
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -361,15 +329,27 @@ impl DWalletMPCManager {
...
         }
-        self.pending_for_computation_order
-            .extend(ready_sessions_response.ready_sessions);
-        self.pending_for_events_order
+
+        // Note that because the Ika consensus isn't in sync with the Sui consensus, it might be that a session has gotten quorum of messages whilst
+        // the current validator haven't received the event from which its public input can be generated (and therefore cannot advance it yet).
+        //
+        // Because of this reason, we place these on two separate queues. Note that in either cases, we must use the copy at this point in time,
+        // so that we will advance it with exactly the same messages as those who already have their event data ready.
```
zeevmoney:
```suggestion
        // Since the Ika and Sui consensuses are not in sync,
        // a session might reach quorum before a validator has received
        // the event needed to generate its public input (and therefore cannot advance it yet).
        // To handle this, we use two separate queues. We also snapshot
        // the current messages so that when we do advance,
        // we use exactly the same inputs as peers who already have the data.
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -361,15 +329,27 @@ impl DWalletMPCManager {
     /// Advance all the MPC sessions that either received enough messages
     /// or perform the first step of the flow.
     /// We parallelize the advances with `Rayon` to speed up the process.
```
zeevmoney: This doc is invalid, this function doesn't advance the sessions.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
zeevmoney: Non Active = Finished or Failed.
This logic might have changed already, if it's incorrect we need to handle it in another PR.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
zeevmoney: Why should we remove the protocol?

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -772,12 +622,130 @@ impl DWalletMPCManager {
...
+        }
+
+        false
+    }
+
+    pub(crate) fn update_network_keys(&mut self) -> Vec<ObjectID> {
+        match self.network_keys_receiver.has_changed() {
+            Ok(has_changed) => {
+                let access_structure = &self.weighted_threshold_access_structure;
+                if has_changed {
```
zeevmoney:
```suggestion
                if has_changed {
                    let access_structure = &self.weighted_threshold_access_structure;
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -772,12 +622,130 @@ impl DWalletMPCManager {
...
+                            Ok(key) => {
+                                info!(key_id=?key_id, "Updating (decrypting new shares) network key for key_id");
+                                if let Err(e) = self
+                                    .network_keys
+                                    .update_network_key(
+                                        *key_id,
+                                        &key,
+                                        &self.weighted_threshold_access_structure,
+                                    ) {
+                                    error!(?e, key_id=?key_id, "failed to update network key");
```
zeevmoney:
```suggestion
                                    error!(error=?e, key_id=?key_id, "failed to update the network key");
```

crates/ika-types/src/messages_dwallet_mpc.rs
```diff
@@ -255,22 +288,23 @@ impl Debug for MPCProtocolInitData {
 pub struct DBSuiEvent {
     pub type_: StructTag,
     pub contents: Vec<u8>,
+    // Set when the event was pulled from the state of the object, and unset when it was pushed as an event.
```
zeevmoney:
```suggestion
    // True when the event was pulled from the state of the object,
    // and False when it was pushed as an event.
```

crates/ika-types/src/messages_dwallet_mpc.rs
```diff
@@ -255,22 +288,23 @@ impl Debug for MPCProtocolInitData {
...
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct DWalletMPCEvent {
-    // TODO: remove event - do all parsing beforehand.
-    pub event: DBSuiEvent,
-    pub session_info: SessionInfo,
-    pub override_epoch_check: bool,
+    pub session_request: MPCSessionRequest,
+    // Set when the event was pulled from the state of the object, and unset when it was pushed as an event.
```
zeevmoney:
```suggestion
    // True when the event was pulled from the state of the object,
    // and False when it was pushed as an event.
```
Set and Unset sounds like you are manipulating it later.

crates/ika-core/src/dwallet_mpc/mpc_protocols/network_dkg.rs
ycscaly: open issue?

crates/ika-types/src/dwallet_mpc_error.rs
```diff
@@ -136,6 +133,12 @@ pub enum DwalletMPCError {
     #[error("the first MPC step should not not receive any messages from the other parties")]
     MessageForFirstMPCStep,
 
+    #[error("no event data when spawning session: BUG")]
```
zeevmoney: Add the session ID, or some identifier.

crates/ika-types/src/dwallet_mpc_error.rs
```diff
@@ -136,6 +133,12 @@ pub enum DwalletMPCError {
...
+    MissingEventData,
+
+    #[error("no next active committee for an event that required it: BUG")]
```
zeevmoney: also here, add some identifier.

zeevmoney (changes requested):
See comments

crates/ika-core/src/authority/authority_per_epoch_store.rs
ycscaly: done

crates/ika-core/src/dwallet_mpc/mpc_event.rs
ycscaly: It wasn't to me

crates/ika-core/src/dwallet_mpc/mpc_event.rs
ycscaly: true, done

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly: done

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly: done

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly: Yeah, in my next PR I will attend to this

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly: we shouldn't

crates/ika-types/src/dwallet_mpc_error.rs
ycscaly: done

crates/ika-core/src/dwallet_mpc/mpc_protocols/network_dkg.rs
ycscaly: nice catch, fixed

crates/ika-core/src/dwallet_mpc/mpc_protocols/network_dkg.rs
yaelAbergel: #1249

yaelAbergel (approved):

zeevmoney (approved):

crates/ika-core/src/dwallet_mpc/mpc_event.rs
```diff
@@ -0,0 +1,548 @@
...
+                    if let IkaError::EpochEnded(_) = err {
+                        return vec![];
+                    };
+                    tokio::time::sleep(Duration::from_secs(2)).await;
+                }
+            }
+        }
+    }
+
+    /// Read events from perpetual tables, remove them, and store in the current epoch tables.
```
ItayLevyOfficial, line 426: misleading documentation

PR #1253, opened 2025-07-09
fix: get reporters of 

Iamknownasfesal:
## Description

Fixes function `get_reporters_of` since it uses itself instead of `validator_set`

## High Notes

Please use lint to fix lint of the file.


zeevmoney (approved):
Thank you for your contribution!

PR #1257, opened 2025-07-12
Refactor Cryptography Computation

ycscaly:
This PR changes a lot of things, you can say it re-writes the `mpc_manager`, `mpc_session` and `cryptorgraphic_computations_orchestrator`. 

Messages are now being stored by consensus round in the session, and are gathered at the end of every consensus round during perform cryptographic computation. There are no longer any queues, and no session copies; there is no need for the `MPCSession` object in order to advance, just the inputs and the messages, which form a computation request. Sessions are ordered every time we want to perform computation, and the ready sessions are advanced.

It also removes a lot of the internal state of a session, which is not needed: we can check if a session is ready by the messages every time, and we can see how much to delay it and if a threshold was reached and so forth from the consensus rounds of the messages, and from saving only the consensus rounds at which threshold not reached occurred - no going back in mpc rounds needed, and the current mpc round is only advanced after we receive a message for that round from oursevles from the consensus, ensuring perfect synchronization. 

Also, only the MPC service handles consensus, it bridges between the MPC logic and the consensus logic, so that the `mpc_manager` and `mpc_session` files do not even know what an `epoch_store` is, they don't broadcast the results, the MPC service does.


crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly, line 1: please `view_file` and read it in its entirety, not the diff

crates/ika-core/src/dwallet_mpc/mpc_session.rs
ycscaly, line 1: please view_file and read it in its entirety, not the diff

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -430,4 +479,271 @@ impl DWalletMPCService {
             session.status = MPCSessionStatus::Finished;
         }
     }
+
+    async fn handle_and_broadcast_computation_results(
```
omersadika: rename to handle_and_submit_to_consensus_computation_results

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -430,4 +479,271 @@ impl DWalletMPCService {
...
+                                            ?session_identifier,
+                                            validator=?validator_name,
+                                            error=?e,
+                                            ?public_output,
+                                            "failed to serialize an MPC output",
+                                        );
+                                    }
+                                }
+                            }
+                            Err(DwalletMPCError::TWOPCMPCThresholdNotReached) => {
```
omersadika, line 606: I don't think it's a good idea to handle this case as an error, better to return an enum

crates/ika-core/src/dwallet_mpc/crytographic_computation.rs
```diff
@@ -0,0 +1,676 @@
...
+/// A unique key for a computation request.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub(crate) struct ComputationId {
+    pub(crate) session_identifier: SessionIdentifier,
+    /// The consensus round at which this computation executed, if it is synced with the consensus.
+    /// The first MPC round will be `None`, since it isn't synced - it is launched when the event is received from Sui.
+    /// All other MPC rounds will set this to `Some()` with the value being the last consensus round from which we gathered messages to advance.
+    pub(crate) consensus_round: Option<u64>,
+    pub(crate) mpc_round: usize,
+    pub(crate) attempt_number: usize,
```
omersadika: I don't think we should use usize for mpc_round or attempt_number.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -635,27 +317,142 @@ impl DWalletMPCManager {
...
+        ComputationId,
+        DwalletMPCResult<
+            mpc::AsynchronousRoundResult<
+                MPCMessage,
+                MPCPrivateOutput,
+                SerializedWrappedMPCPublicOutput,
+            >,
+        >,
+    > {
+        let computation_requests: Vec<_> = self
```
omersadika: I would split it to vars with meaning full names, e.g. mpc_sessions_with_event_data, sorted_mpc_sessions_with_event_data and computation_requests

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -635,27 +317,142 @@ impl DWalletMPCManager {
...
+                        SessionType::System => true,
+                    }
+                } else {
+                    // Cannot advance sessions without MPC event data
+                    false
+                }
+            })
+            .sorted_by(|(_, session), (_, other_session)| {
+                // Safe to `unwrap`, we filtered sessions without `mpc_event_data`.
+                let session_mpc_event_data = session.mpc_event_data.as_ref().unwrap();
```
omersadika: I will be cleaner to add mpc_event_data to the tuple so then you don't have to keep unwrap it

omersadika: As part of filter (maybe should be filter_map instead)

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -732,61 +529,56 @@ impl DWalletMPCManager {
         }
     }
 
-    /// Insert `session` into `self.ordered_sessions_pending_for_computation`, keeping order:
-    /// System sessions come first, and user sessions are sorted by their sequence number.
-    ///
-    /// Note: at this point, `mpc_event_data` must be set!
-    pub(crate) fn insert_session_into_ordered_pending_for_computation_queue(
+    pub(crate) fn handle_dwallet_db_output(
```
omersadika, line 546: Rename to handle_consensus_round_output

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -1,1340 +1,294 @@
...
     /// The current MPC round number of the session.
-    /// Starts at `1` and increments after each advance of the session.
+    /// Starts at `1` and increments after each successful advance of the session.
     /// In round `1` We start the flow, without messages, from the event trigger.
-    /// Decremented only upon an `TWOPCMPCThresholdNotReached` Error.
-    pub(super) current_round: usize,
-    pub(crate) party_id: PartyID,
-    // TODO (#539): Simplify struct to only contain session related data - remove this field.
-    weighted_threshold_access_structure: WeightedThresholdAccessStructure,
+    pub(super) current_mpc_round: usize,
```
omersadika: Again here, I don't like the uszie

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -1,1340 +1,294 @@
...
+        let mut messages_for_advance: HashMap<usize, HashMap<PartyID, MPCMessage>> = HashMap::new();
+
+        let sorted_messages_by_consensus_round = self
+            .messages_by_consensus_round
+            .clone()
+            .into_iter()
+            .sorted_by(|(first_consensus_round, _), (second_consensus_round, _)| {
+                first_consensus_round.cmp(second_consensus_round)
+            });
+        for (consensus_round, consensus_round_messages) in sorted_messages_by_consensus_round {
```
omersadika: Try to break that to functions, too many nested loops

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -1,1340 +1,294 @@
...
-                    validator=?self.epoch_store()?.name,
-                    session_identifier=?self.session_identifier,
-                    crypto_round=?self.current_round,
-                    weighted_parties=?self.weighted_threshold_access_structure,
-                    ?serialized_messages_skeleton,
-                    "no decryption key shares for a session that requires them (reconfiguration)"
-                    );
+            // Check if we have the threshold of messages for the previous round
+            // to advance to the next round.
+            let is_quorum_reached = if let Some(previous_round_messages) =
```
omersadika: Move to a func

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -1,1340 +1,294 @@
...
+    /// - a quorum of messages from the previous round `current_mpc_round - 1` must exist (except for the first round, which is always ready to advance requiring no messages as input.)
+    /// - a minimum number of consensus rounds that was required to delay the execution (in order to allow more messages to come in before advancing)
+    ///   has passed since the first consensus round where we got a quorum for this round.
+    /// - this quorum must be "fresh", in the sense we never tried to advance with it before.
+    ///   There is only one case in which we attempt to advance the same round twice: when we get a threshold not reached error.
+    ///   Therefore, if such an error occurred for a consensus round, we don't stop the search,
+    ///   and wait for at least one new message to come in a subsequent consensus round before returning the messages to advance with.
+    ///
+    /// Duplicate messages are ignored - the first message a party has sent for an MPC round is always used.
+    pub(crate) fn build_messages_to_advance(
```
omersadika: I feel like this function should not be from MPCSession but to receive the info you need from MPCSession and build it

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -1,1340 +1,294 @@
...
-                        is_ready: false,
-                        malicious_parties: vec![],
-                    })
-                }
+    /// - **NetworkDkg protocol**: requires delay for the third round
+    ///   using `network_dkg_third_round_delay` config.
+    /// - **DecryptionKeyReconfiguration protocol**: requires delay for the third round
+    ///   using `decryption_key_reconfiguration_third_round_delay` config.
+    /// - **Other protocols**: No delay required, always ready to advance
+    fn consensus_rounds_delay_for_mpc_round(&self) -> usize {
```
omersadika: Also here, I think this func should get mpc_event_data and not assume it's there

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -94,6 +96,28 @@ pub(crate) fn sign_session_public_input(
     )
 }
 
+pub(crate) fn update_expected_decrypters_metrics(
```
omersadika, line 109: document this func, why it's good metrics

omersadika (changes requested):

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -238,390 +211,99 @@ impl DWalletMPCManager {
...
-                        .unwrap_or(&HashMap::new())
-                        .keys()
-                        .cloned()
-                        .collect::<Vec<PartyID>>(),
-                    &epoch_store,
-                )?);
+        consensus_round: u64,
+        messages: Vec<DWalletMPCMessage>,
+    ) {
+        for (_, session) in self.mpc_sessions.iter_mut() {
```
omersadika, line 220: The session can be created before by an event. You should only add rounds of messages or session with rounds

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -222,60 +262,64 @@ impl DWalletMPCService {
...
+        for (
+            mpc_messages_iteration_result,
+            mpc_outputs_iteration_result,
+            verified_dwallet_checkpoint_messages_iteration_result,
+        ) in zipped_consensus_rounds_iter
         {
-            let Ok((mpc_messages_round, mpc_messages)) = mpc_messages else {
+            let Ok((mpc_messages_consensus_round, mpc_messages)) = mpc_messages_iteration_result
+            else {
                 error!("Failed to load DWallet MPC messages from the local DB");
```
yaelAbergel:
```suggestion
                error!("failed to load DWallet MPC messages from the local DB");
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -222,60 +262,64 @@ impl DWalletMPCService {
...
+
                 return false;
             };
-            let Ok((mpc_outputs_round, mpc_outputs)) = mpc_outputs else {
+            let Ok((mpc_outputs_consensus_round, mpc_outputs)) = mpc_outputs_iteration_result
+            else {
                 error!("Failed to load DWallet MPC outputs from the local DB");
```
yaelAbergel:
```suggestion
                error!("failed to load DWallet MPC outputs from the local DB");
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -222,60 +262,64 @@ impl DWalletMPCService {
...
+                    != verified_dwallet_checkpoint_messages_consensus_round
             {
                 error!(
-                        ?mpc_messages_round,
-                        ?mpc_outputs_round,
-                        ?verified_dwallet_checkpoint_messages_round,
+                        ?mpc_messages_consensus_round,
+                        ?mpc_outputs_consensus_round,
+                        ?verified_dwallet_checkpoint_messages_consensus_round,
                         "The consensus rounds of MPC messages, MPC outputs and checkpoint messages do not match"
```
yaelAbergel:
```suggestion
                        "the consensus rounds of MPC messages, MPC outputs and checkpoint messages do not match"
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -222,60 +262,64 @@ impl DWalletMPCService {
...
             }
 
-            let consensus_round = mpc_messages_round;
+            let consensus_round = mpc_messages_consensus_round;
 
             if self.last_read_consensus_round >= Some(consensus_round) {
                 error!(
                     consensus_round,
                     last_read_consensus_round=?self.last_read_consensus_round,
                     "Consensus round must be in a ascending order, should never happen"
```
yaelAbergel:
```suggestion
                    "consensus round must be in a ascending order, should never happen"
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -430,4 +479,271 @@ impl DWalletMPCService {
...
+            if let Some(session) = self
+                .dwallet_mpc_manager
+                .mpc_sessions
+                .get(&session_identifier)
+            {
+                if session.status == MPCSessionStatus::Active {
+                    if let Some(mpc_event_data) = session.mpc_event_data.clone() {
+                        match computation_result {
+                            Ok(AsynchronousRoundResult::Advance {
+                                malicious_parties: _,
```
yaelAbergel, line 514: Do not ignore malicious, add a report

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -430,4 +479,271 @@ impl DWalletMPCService {
...
+                                        &malicious_parties,
+                                        &committee,
+                                    );
+
+                                    error!(
+                                        ?session_identifier,
+                                            validator=?validator_name,
+                                            ?malicious_parties,
+                                            ?malicious_authorities,
+                                        "Malicious parties detected upon MPC session finalize",
```
yaelAbergel:
```suggestion
                                        "malicious parties detected upon MPC session finalize",
```

yaelAbergel: Also I think this log could be `warn` level. If so the first comment should be ignored.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -98,16 +104,19 @@ pub(crate) fn advance_and_serialize<P: AsynchronouslyAdvanceable>(
 struct DeserializeMPCMessagesResponse<M: DeserializeOwned + Clone> {
     /// round -> {party -> message}
     messages: HashMap<usize, HashMap<PartyID, M>>,
-    #[allow(dead_code)]
     malicious_parties: Vec<PartyID>,
 }
 
 /// Deserializes the messages received from other parties for the next advancement.
 /// Any value that fails to deserialize is considered to be sent by a malicious party.
```
dolevmu, line 241: 👍

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -98,16 +104,19 @@ pub(crate) fn advance_and_serialize<P: AsynchronouslyAdvanceable>(
...
+///
+/// Note that deserialization of a message depends on the type of the message which is only known once the event data comes,
+/// and so we can only handle this here. Malicious messages are ignored, and we assure that we still have quorum, otherwise returning a `ThresholdNotReached` error.
+fn deserialize_mpc_messages_and_check_quorum<M: DeserializeOwned + Clone>(
     messages: &HashMap<usize, HashMap<PartyID, MPCMessage>>,
-) -> DeserializeMPCMessagesResponse<M> {
+    access_structure: &WeightedThresholdAccessStructure,
+) -> DwalletMPCResult<DeserializeMPCMessagesResponse<M>> {
     let mut deserialized_results = HashMap::new();
     let mut malicious_parties = Vec::new();
```
dolevmu, line 253: Is index == round number within the session id?
And message_batch == the set of all messages from all parties received in this round?
Do we go over the index,message_batch sorted by the index (=round number?)

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -37,10 +38,9 @@ pub(crate) type SignPublicInput =
...
 /// Note: this is only an optimization: if we don't have at least `t` online decrypters out of the `expected_decrypters` subset, the Sign protocol still completes successfully, just slower.
 fn generate_expected_decrypters(
-    epoch_store: &AuthorityPerEpochStore,
+    access_structure: &WeightedThresholdAccessStructure,
     session_identifier: SessionIdentifier,
 ) -> DwalletMPCResult<HashSet<PartyID>> {
-    let access_structure = epoch_store.get_weighted_threshold_access_structure()?;
     let total_weight = access_structure.total_weight();
     let expected_decrypters_weight =
         access_structure.threshold + (total_weight as f64 * 0.10).floor() as Weight;
```
dolevmu, line 53: I don't like that 10% hard-coded like this, even just add a global constant value and use it maybe.
If we can make it 5%, then sign could maybe be 2x faster becuse we only wait to see what 5% we don't need instead of 10% for instance.
On the other hand, if 10% is too small optimization could fail.
I'd rather keep the option to change this constant, in version update or even dynamically every epoch or whatever, and set it to this constant for now.

dolevmu: Also I see we pick them randomly, maybe it is okay, but we are not really "expecting them" they are chosen pseudorandomly, so the naming is misleading a bit. Originally we wanted them to really be the expected ones based on some metrics as you suggested to do in the future if we see it is necessary.

crates/ika-core/src/dwallet_mpc/crytographic_computation/orchestrator.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// Tracks tasks that have been spawned with [`rayon::spawn_fifo`] but haven't completed yet.
+    /// Used to prevent exceeding available CPU cores.
+    currently_running_cryptographic_computations: HashSet<ComputationId>,
+
+    /// The list of completed cryptographic computations in the current epoch.
+    completed_cryptographic_computations: HashSet<ComputationId>,
+
+    /// The root seed of this validator, used for deriving the per-round seed for advancing this session.
+    /// SECURITY NOTICE: *MUST KEEP PRIVATE*.
+    root_seed: RootSeed,
```
dolevmu, line 81: So this is the seed from which we also derive the psuedorandomness for deterministic answers to every round of every mpc session every party, that we should answer the same message in state sync? (and everything else).

crates/ika-core/src/dwallet_mpc/crytographic_computation/orchestrator.rs
```diff
@@ -0,0 +1,267 @@
...
+
+        if self
+            .currently_running_cryptographic_computations
+            .contains(&computation_id)
+            || self
+                .completed_cryptographic_computations
+                .contains(&computation_id)
+        {
+            // Don't run a task that we already spawned.
+            return true;
```
dolevmu, line 194: Can this happen? That we try twice to run the same computation task? Maybe log this too with info! or debug! ?

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
ycscaly: this will be moved to the cryptography side

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
ycscaly: Its debateable, I think we definitely want to be immediately alerted if a malicious player exists

crates/ika-core/src/dwallet_mpc/crytographic_computation/orchestrator.rs
```diff
@@ -0,0 +1,267 @@
...
+
+        let handle = Handle::current();
+
+        dwallet_mpc_metrics.add_advance_call(
+            &computation_request.request_input,
+            &computation_id.mpc_round.to_string(),
+        );
+
+        let computation_channel_sender = self.completed_computation_sender.clone();
+        let root_seed = self.root_seed.clone();
```
dolevmu, line 205: I don't like this practice, `self.root_seed` is the most critical asset a validator has, we don't want copies of it all over the place, 128bits of random copied accross all cpu cores and all tasks. If we can derive the seed for current sid and round already, it might be much better. We already put much effort to avoid side channel and cache attacks so its best to avoid copying this value directly and in parallel.

crates/ika-core/src/dwallet_mpc/mpc_session.rs
ycscaly: I also thought of that

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -113,9 +144,9 @@ impl DWalletMPCService {
         loop {
             let mut events = vec![];
 
-            // Load events from Sui every 30 seconds (300 * READ_INTERVAL_MS=100ms = 30,000ms = 30s).
+            // Load events from Sui every 30 seconds (1500 * READ_INTERVAL_MS=20ms = 30,000ms = 30s).
```
dolevmu, line 146: This is why I like macros. BTW, how does this 30sec wait from SUI affect our latency? Couldn't this be a problem in some applications?

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -113,9 +144,9 @@ impl DWalletMPCService {
...
             // Note: when we spawn, `loop_index == 0`, so we fetch uncompleted events on spawn.
-            if loop_index % 300 == 0 {
+            if loop_index % 1500 == 0 {
```
dolevmu, line 148: The "1500" constant can be computed from the READ_INTERVAL_MS constant and the "30 seconds" constant for loading events from SUI. Those two sound to me like the ones you care about so I would macro them and compute this one based on them and macro it too.

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -130,12 +161,11 @@ impl DWalletMPCService {
                 }
                 Ok(false) => (),
             };
-            tokio::time::sleep(Duration::from_millis(READ_INTERVAL_MS)).await;
 
             if self.dwallet_mpc_manager.recognized_self_as_malicious {
                 error!(
                     authority=?self.epoch_store.name,
-                    "the node has identified itself as malicious and is no longer participating in MPC protocols"
+                    "the node has identified itself as malicious, breaking from MPC service loop"
```
dolevmu, line 167: @ycscaly can we reach here in the case we talked about that the output we voted for was wrong and a "fork" happened, and we may want to regret somehow?

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -238,390 +211,99 @@ impl DWalletMPCManager {
...
-        {
-            return Ok(());
-        }
-        self.malicious_handler
-            .report_malicious_actor(report.clone(), reporting_authority)?;
-        let epoch_store = self.epoch_store()?;
-        if self.malicious_handler.is_malicious_actor(&epoch_store.name) {
-            self.recognized_self_as_malicious = true;
+    /// Handles a message by forwarding it to the relevant MPC session.
+    /// If the session does not exist, punish the sender.
```
yaelAbergel: We can receive a message before an event

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -238,390 +211,99 @@ impl DWalletMPCManager {
...
-                authority=?epoch_store.name,
-                reporting_authority=?reporting_authority,
-                malicious_actors=?report.malicious_actors,
-                session_identifier=?report.session_identifier,
-                "node recognized itself as malicious"
+                session_identifier=?session_identifier,
+                sender_authority=?sender_authority,
+                receiver_authority=?self.validator_name,
+                mpc_round_number=?mpc_round_number,
+                "Got a message for an authority without party ID",
```
yaelAbergel:
```suggestion
                "got a message for an authority without party ID",
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
dolevmu:
```suggestion
                    "Consensus round must be in ascending order, should never happen"
```

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -287,15 +331,17 @@ impl DWalletMPCService {
...
+
                             info!(
+                                authority=?self.epoch_store.name,
                                 ?output_digest,
                                 consensus_round,
                                 ?session_identifier,
                                 "MPC output is verified and reached quorum"
                             );
                         }
                         OutputVerificationStatus::Malicious => {
```
dolevmu, line 342: If we have GOD, this should never be reached, I suppose that's why its warn!

dolevmu: How can we have an output with not enough votes? Shouldn't we wait forever until we get enough of them? Does it mean we have contradicting votes so that we will never reach the threshold with none of them? @ycscaly is this the point you want to jump in and allow regret to recover?

This should also be a warn! I think...

dolevmu: Already commit also I think should be a warn!, we shouldn't commit the same output twice right?

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -635,27 +321,151 @@ impl DWalletMPCManager {
...
+    /// if they come before the last session to complete in the current epoch (at the current time).
+    ///
+    /// System sessions are always advanced before any user session,
+    /// and both system and user sessions are ordered internally by their sequence numbers.
+    ///
+    /// The messages to advance with are built on the spot, assuming they satisfy required conditions.
+    /// They are put on a `ComputationRequest` and forwarded to the `orchestrator` for execution.
+    pub(crate) async fn perform_cryptographic_computation(
+        &mut self,
+    ) -> HashMap<
```
yaelAbergel, line 362: Explain the return value in docs please

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -1,1340 +1,166 @@
...
-        if message.round_number > self.current_round {
-            warn!(
-                session_id=?message.session_identifier,
-                from_authority=?message.authority,
-                receiving_authority=?self.epoch_store()?.name,
-                recieved_message_round_number=?message.round_number,
-                "Received a message for a future round",
+
+        if sender_party_id == self.party_id && self.current_mpc_round <= message.round_number {
+            // Received a message from ourselves from the consensus, so it's safe to advance the round.
```
yaelAbergel, line 100: What if there are threshold messages and no message from ourselves yet?

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
ycscaly: documentation was outdated

crates/ika-core/src/dwallet_mpc/mpc_session.rs
ycscaly: then we should wait until we finish executing it, send it, and receive it before running a subsequent round. We should run every round and send a message for every round.

yaelAbergel (approved):

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
dolevmu: Understanding which parties were malicious during some round is detected internally in crypto-private I guess, but I still don't understand how this is empty: what about the malicious parties that sent a message that doesn't deserialize (@ycscaly  you said you will move it to your logic as well in future PR but not in this one)? What about parties that are recognized as malicious in an MPCOutput of another session that ended before this consensus round? Should be passed from network code as well, crypto-private only handles one sid at a time and can't see that.

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
dolevmu: If I remember correctly, it is handeled as an error in crypto-private as well

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
dolevmu: To be consistent, every detection of a malicious behaviour should probably be logged with same level.

crates/ika-core/src/dwallet_mpc/mpc_session/mpc_event_data.rs
```diff
@@ -0,0 +1,113 @@
...
+impl PartialOrd<Self> for MPCEventData {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for MPCEventData {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // System sessions have a higher priority than user session.
+        // Both system and user sessions are sorted by their sequence number between themselves.
```
dolevmu, line 101: In general I understand, but I must point out.
Once you have rules like that, you may have MEV and sandwich attacks, I might see an mpc event that is user type that I want to stall or make something happen before it, so I can use system type. Not sure what system session types are, is it just Net. DKG and reconfig, maybe also end epoch (so I can try to close epoch and cancel a user MPC output / message), maybe even DApps that do ECDSA MPC without client share (should those get priority? Can it cause issues?)

crates/ika-core/src/dwallet_mpc/mpc_session/mpc_event_data.rs
```diff
@@ -0,0 +1,113 @@
...
+        match (self.session_type, other.session_type) {
+            (SessionType::User, SessionType::User) => self
+                .session_sequence_number
```
dolevmu, line 104: Also, what is the session_sequence_number? Looks like a counter maybe? Is it how SUI sorts its event and you take it from there? Maybe it means we are also exposed to SUI MEV attacks? Can't we sort by session id?

PR #1285, opened 2025-07-17
fix: Replace class groups key on epoch switch and enforce setting if missing in next committee

yaelAbergel:
## Description 

This PR addresses two key issues related to class groups key handling and next epoch validator enforcement:
Class Groups Key Optional 
Previously, validators could set a `class_groups_pubkey_and_proof` without size constraints. This could lead to large keys that exceed , potentially causing failures in critical functions like `advance_epoch`.
➤ Fix: Hold all related class groups keys as option.

Next Epoch Field Enforcement
Validators were allowed to set next epoch fields (e.g., network/consensus/protocol keys) even when they are part of the next committee, potentially unsafe.
➤ Fix: Setting next epoch-related fields is now restricted only to validators who will not participate in the upcoming committee.

## Test plan 

Tested with new commands from branch `more-validator-commands`

---

### DIFF
--- a/crates/ika-move-packages/packages/ika_system/sources/staking/validator_info.move
+++ b/crates/ika-move-packages/packages/ika_system/sources/staking/validator_info.move
@@ -231,7 +231,17 @@ public(package) fun set_next_epoch_consensus_pubkey_bytes(
     self.validate();
 }
 
-/// Sets class groups public key and proof for next epoch.
+/// Sets the class groups public key and proof for the next epoch.
+/// 
+/// - If `next_epoch_class_groups_pubkey_and_proof_bytes` is already set, 
+///   this function returns its stored value and replaces it with the new value.
+/// - If it is not set, but `previous_class_groups_pubkey_and_proof_bytes` is set, 
+///   this function returns the value from the previous field and replaces it with the new value.
+/// - If neither is set, the new value is simply stored, and the function returns `None`.
+/// 
+/// The validator must drop the returned value if it is not `None`.
+/// 
+/// Using `Option` for the MPC data helps avoid latency issues due to large data sizes.
 public(package) fun set_next_epoch_class_groups_pubkey_and_proof_bytes(
     self: &mut ValidatorInfo,
     class_groups_pubkey_and_proof: TableVec<vector<u8>>,
--- a/crates/ika-types/src/sui/staking.rs
+++ b/crates/ika-types/src/sui/staking.rs
@@ -240,7 +240,8 @@ pub enum PoolState {
 
 impl StakingPool {
     pub fn verified_validator_info(&self) -> &VerifiedValidatorInfo {
+        // Todo (#1298): Remove unwrap.
         self.verified_validator_info
-            .get_or_init(|| self.validator_info.verify().unwrap()) // Why unwrap? this could fail
+            .get_or_init(|| self.validator_info.verify().unwrap())
     }
 }


crates/ika-move-packages/packages/ika_system/sources/staking/validator_info.move
```diff
@@ -277,13 +291,19 @@ public(package) fun rotate_next_epoch_info(self: &mut ValidatorInfo) {
     };
 
     if (self.next_epoch_class_groups_pubkey_and_proof_bytes.is_some()) {
```
omersadika: Maybe let's add validation here for previous_class_groups_pubkey_and_proof_bytes anyway (is none) for the case we have an issue

crates/ika-move-packages/packages/ika_system/sources/system/validator_set.move
```diff
@@ -437,6 +440,10 @@ public(package) fun set_validator_name(
     cap: &ValidatorOperationCap,
 ) {
     let validator_id = cap.validator_id();
+    let is_next_committee = self
```
omersadika: We need that only for set next, so no for name

crates/ika-move-packages/packages/ika_system/sources/system/validator_set.move
```diff
@@ -452,6 +459,10 @@ public(package) fun set_validator_metadata(
     metadata: ValidatorMetadata,
 ) {
     let validator_id = cap.validator_id();
+    let is_next_committee = self
```
omersadika: We need that only for set next, so no for metadata

crates/ika-move-packages/packages/ika_system/sources/staking/validator_info.move
```diff
@@ -78,16 +78,19 @@ public struct ValidatorInfo has store {
     consensus_pubkey_bytes: vector<u8>,
     /// The validator's Class Groups public key and its associated proof.
     /// This key is used for the network DKG process and for resharing the network MPC key
-    class_groups_pubkey_and_proof_bytes: TableVec<vector<u8>>,
+    /// Must always contain value 
+    class_groups_pubkey_and_proof_bytes: Option<TableVec<vector<u8>>>,
```
omersadika, line 82: You need to fix also the rust code to get option

crates/ika-move-packages/packages/ika_system/sources/staking/validator_info.move
```diff
@@ -78,16 +78,19 @@ public struct ValidatorInfo has store {
...
     /// Next epoch configurations - only take effect in the next epoch
     /// If none, current value will stay unchanged.
     next_epoch_protocol_pubkey_bytes: Option<vector<u8>>,
     next_epoch_network_pubkey_bytes: Option<vector<u8>>,
     next_epoch_consensus_pubkey_bytes: Option<vector<u8>>,
-    next_epoch_class_groups_pubkey_and_proof_bytes: Option<ClassGroupsPublicKeyAndProof>,
+    next_epoch_class_groups_pubkey_and_proof_bytes: Option<TableVec<vector<u8>>>,
```
omersadika, line 88: Need to update the rust code to take this if some for reconfiguration

omersadika (changes requested):

crates/ika-types/src/sui/staking.rs
```diff
@@ -230,6 +241,6 @@ pub enum PoolState {
 impl StakingPool {
     pub fn verified_validator_info(&self) -> &VerifiedValidatorInfo {
         self.verified_validator_info
-            .get_or_init(|| self.validator_info.verify().unwrap())
+            .get_or_init(|| self.validator_info.verify().unwrap()) // Why unwrap? this could fail
```
yaelAbergel: I added this comment here to highlight that using `verified_validator_info` anywhere can potentially crash the node!

PR #1327, opened 2025-07-24
chore: Verify validator keys

yaelAbergel:


ycscaly (approved):

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -916,4 +920,58 @@ impl DWalletMPCService {
         }
         slices
     }
+
+    pub fn verify_validator_keys(
```
ItayLevyOfficial, line 924: why are you not verifying the `config.protocol_key_pair`?

ItayLevyOfficial (approved):
lgtm

PR #1333, opened 2025-07-27
chore: Ignore network keys from different epochs

ItayLevyOfficial:
## Description 

Until now the network keys syncer tried to update the dwallet MPC manager with every network key it fetched, without verifying the key's epoch matched the epoch the validator run. This PR mitigates this issue by verifying the key's epoch first, and replacing the error log with a warn log.

## Test plan 

I tried to reproduce the error log locally without success, but there are good reasons to believe it may occur on a real-world network, where communication delays are more variable.

After introducing my change and running tests on it however, I saw my warn log is being printed multiple times right after an epoch switch.

I verified the network keeps working & switch epochs successfully after this change.

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/mpc_manager.rs
+++ b/crates/ika-core/src/dwallet_mpc/mpc_manager.rs
@@ -553,11 +553,12 @@ impl DWalletMPCManager {
                         match res {
                             Ok(key) => {
                                 if key.epoch != self.epoch_id {
-                                    warn!(
+                                    info!(
                                         key_id=?key_id,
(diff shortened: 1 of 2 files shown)


yaelAbergel (changes requested):
/

yaelAbergel (approved):

ycscaly (approved):

PR #1341, opened 2025-07-29
chore: Remove rounding from average & variance metrics

ItayLevyOfficial:
## Description 

Until now, I used integers for the average round completion duration and variance metrics. This required rounding, which introduced trailing errors that accumulated over time. This PR mitigates the issue by switching to floating-point values for these metrics.

## Test plan 

I added three more test cases for the new-variance calculation function, with values that has been verified using online variance calculators. 

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/ika-core/src/sui_connector/mod.rs
+++ b/crates/ika-core/src/sui_connector/mod.rs
@@ -269,7 +269,7 @@ mod tests {
     use super::*;
     use ika_sui_client::retry_with_max_elapsed_time;
     use std::time::Duration;
-    use tracing::error;
+    use tracing::warn;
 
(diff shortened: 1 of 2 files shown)


crates/ika-core/src/sui_connector/mod.rs
ItayLevyOfficial, line 1: needed to get the tests to compile

ycscaly (approved):

PR #1346, opened 2025-07-30
test: Add an e2e test of the join committee flow

ItayLevyOfficial:
## Description 

This PR introduces a TypeScript test that deploys an Ika network to a Kubernetes cluster. It adds X validators, waits for an epoch switch, then kills N validators, and verifies that the network continues to function properly.

The validator "kill" step is used to confirm that the newly added validators are operational. In the test, I deployed a network with four nodes, added three validators to the next committee, and killed two of the original validators. The fact that the network continued to function correctly after this transition confirms that the new validators were indeed operational. Out of the new seven-validator committee, at least five need to be functioning to reach quorum.

## Test plan 

I deployed a network with four nodes, added three validators to the next committee, and killed two of the original validators. The fact that the network continued to function correctly after this transition confirms that the new validators were indeed operational. Out of the new seven-validator committee, at least five need to be functioning to reach quorum.

My e2e test runs a full flow after adding and removing validators to ensure the network functions properly.

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
+++ b/contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
@@ -660,8 +660,6 @@ public fun commit_upgrade(self: &mut DWalletCoordinator, upgrade_package_approve
 /// to migrate changes in the `coordinator_inner` object if needed.
 /// This function can be called immediately after the upgrade is committed.
 public fun try_migrate_by_cap(self: &mut DWalletCoordinator, _: &VerifiedProtocolCap) {
-    assert!(self.version < VERSION, EInvalidMigration);
-    assert!(self.new_package_id.is_some(), EInvalidMigration);
     self.try_migrate_impl();
 }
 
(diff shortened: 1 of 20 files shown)


sdk/typescript/src/dwallet-mpc/globals.ts
```diff
@@ -162,6 +163,21 @@ export function isSystemInner(obj: any): obj is SystemInner {
 	);
 }
 
+export async function getSystemInner(conf: Config): Promise<SystemInner> {
```
ItayLevyOfficial, line 167: needed to continually check for an epoch switch in the end to end test

sdk/typescript/test/chaos-test/pods.ts
```diff
@@ -51,6 +51,12 @@ export async function createValidatorPod(
 					command: ['/opt/ika/bin/ika-node', '--config-path', '/opt/ika/config/validator.yaml'],
 					name: 'ika-node',
 					image: process.env.DOCKER_TAG,
+					resources: {
```
ItayLevyOfficial, line 54: I started working with the Devnet k8s cluster while working on this PR, so I needed to specify the hardware requirements for the pods. This wasn’t necessary until now, as I had been using my local Minikube.

ycscaly (approved):

sdk/typescript/test/chaos-test/create-ika-genesis-mac.sh
```diff
@@ -198,64 +195,6 @@ echo "Ika dWallet 2PC MPC Package ID: $IKA_DWALLET_2PC_MPC_PACKAGE_ID"
 # Request Tokens and Create Validator.yaml (Max 5 Parallel + Retry)
 ############################
 
-request_and_generate_yaml() {
```
ItayLevyOfficial, line 201: moved to shared.sh

ItayLevyOfficial: accidentally not been removed in the previous pr

PR #1353, opened 2025-07-31
fix: Make `party_id_to_authority_name` not panic on zero input

ItayLevyOfficial:
## Description 

I added some unit tests to the `party_id_to_authority_name` function, and while working on them I find that bug & fixed it.

## Test plan 

I ran the unit tests I introduced in this PR.

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/ika-types/src/committee.rs
+++ b/crates/ika-types/src/committee.rs
@@ -89,10 +89,10 @@ impl Committee {
     /// of the Ika System object.
     pub fn new_for_testing_with_normalized_voting_power(
         epoch: EpochId,
-        mut voting_weights: BTreeMap<AuthorityName, StakeUnit>,
+        mut voting_weights: Vec<(AuthorityName, StakeUnit)>,
     ) -> Self {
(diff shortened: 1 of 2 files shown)


crates/ika-types/src/committee.rs
```diff
@@ -89,10 +89,10 @@ impl Committee {
     /// of the Ika System object.
     pub fn new_for_testing_with_normalized_voting_power(
         epoch: EpochId,
-        mut voting_weights: BTreeMap<AuthorityName, StakeUnit>,
+        mut voting_weights: Vec<(AuthorityName, StakeUnit)>,
```
ItayLevyOfficial, line 92: did this change to simplify testing and avoid changing the order of the validators.

ycscaly (approved):

PR #1362, opened 2025-08-03
test: Add a TS test that creates a network key and runs a full flow with it

ItayLevyOfficial:
## Description 

Until now, some of our SDK functions always used the first network key found in the list of network keys on Sui, without considering the scenario where multiple network keys exist. This PR fixes that issue by allowing the SDK functions to accept a specific network key ID to work with. It also adds a test that creates a network key and uses it to run a full flow.

## Test plan 

I ran the new test I wrote.

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:


sdk/typescript/test/e2e/dwallet-mpc.test.ts
```diff
@@ -71,35 +72,6 @@ export async function createConf(): Promise<Config> {
 	};
 }
 
-export async function runSignFullFlow(conf: Config) {
```
ItayLevyOfficial, line 74: moved down

PR #1385, opened 2025-08-11
Test: Add an integration test in which one N parties behave maliciously

ItayLevyOfficial:
## Description 

Add an integration test in which N parties send an invalid message for MPC round no. 1. Those parties are expected to be marked by all the validators as malicious, and the flow completes successfully.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/malicious_behavior.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/malicious_behavior.rs
@@ -9,12 +9,16 @@ use tracing::info;
 
 #[tokio::test]
 #[cfg(test)]
-async fn test_malicious_behavior() {
+async fn test_some_malicious_validators_flows_succeed() {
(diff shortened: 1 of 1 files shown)


crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -190,6 +190,10 @@ impl DWalletMPCService {
         }
     }
 
+    pub(crate) fn dwallet_mpc_manager(&self) -> &DWalletMPCManager {
```
ItayLevyOfficial, line 193: preferred to use a getter for testing so it will be clear this field can't change from outside of this module

PR #1405, opened 2025-08-12
Upgrade move contracts to v2

omersadika:


contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -479,11 +480,33 @@ public fun request_sign(
     payment_sui: &mut Coin<SUI>,
     ctx: &mut TxContext,
 ) {
+    let _ = self
```
Iamknownasfesal, line 483: Instead of this if the return is a droppable value you can just call the function.

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -502,11 +525,33 @@ public fun request_imported_key_sign(
...
```
Iamknownasfesal, line 528: same here

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -571,6 +616,26 @@ public fun request_sign_with_partial_user_signature(
...
```
Iamknownasfesal, line 619: same here

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -592,6 +657,26 @@ public fun request_imported_key_sign_with_partial_user_signature(
...
```
Iamknownasfesal, line 660: same here

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -4875,14 +4848,162 @@ public(package) fun get_network_encryption_key_supported_curves(
...
+/// ### Aborts
+/// - `EDWalletInactive`: If the dWallet is not in the `Active` state
+///
+/// ### Active State Requirements
+/// A dWallet is considered active when:
+/// - DKG process has completed successfully, OR
+/// - Imported key verification has completed successfully
+/// - User has accepted their encrypted key share
+/// - Public output is available for cryptographic operations
+public fun validate_active_and_get_public_output(self: &DWallet): &vector<u8> {
```
Iamknownasfesal, line 4941: Instead of covering everything we can do:

```
match (&self.state) {
    DWalletState::Active {
        public_output,
    } => public_output,
    _ => abort EDWalletInactive,
}
```

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
omersadika: It's the preferred way to do it in rust and I think it's good to explicitly indicate that we are dropping the returned value

contracts/dwallet_utils/sources/dwallet_utils.move
```diff
@@ -0,0 +1,64 @@
...
+
+
+//     //let signature = vector[35,226,152,163,30,13,80,43,85,163,108,93,168,242,130,26,185,126,237,20,27,183,34,72,1,48,104,253,112,105,52,243,15,20,90,55,183,115,127,71,111,255,79,4,126,169,90,43,153,34,205,62,174,154,143,233,27,29,234,171,183,119,221,248];
+//     // std::debug::print(&signature.length());
+
+
+//     // let msg = vector[104,101,108,108,111,32,119,111,114,108,100];
+//     // let verified = ecdsa_k1::secp256k1_verify(&signature, &public_key, &msg, 0);
+//     // assert!(verified, 1);
+// }
```
ItayLevyOfficial, line 64:
```suggestion
```

contracts/dwallet_utils/sources/dwallet_utils.move
```diff
@@ -0,0 +1,64 @@
...
+            };
+            i = i + 1;
+        };
+        option::some(public_key_bytes)
+    } else {
+        option::none()
+    }
+}
+
+public fun non_recoverable_ecdsa_k1_signature(dwallet: &DWallet, sign_id: ID): vector<u8> {
```
ItayLevyOfficial, line 29: What does non recoverable mean here? In which sense is it non recoverable?

Iamknownasfesal: It is a signature type where you cannot recover the public key from the signature, versus where you can

recoverable
(r, s, v)

non recoverable

(r, s)

---

r – the x-coordinate of the ephemeral point used in signing
s – the signature proof scalar
v – the recovery ID(that lets you recover the public key from r, s, and the message hash)

ItayLevyOfficial: cool, thanks! @Iamknownasfesal

ItayLevyOfficial (approved):
lgtm

PR #1438, opened 2025-08-18
cover all cases

Iamknownasfesal:
since we need a lot more ram in our nodejs process, i suggest you guys to test it with this: `NODE_OPTIONS=--max-old-space-size=8192 pnpm vitest run`, i know it is crazy... let me know if you guys see any optimizations can be done

### DIFF
--- a/sdk/typescript/test/imported-dwallet/imported-dwallet-sign.test.ts
+++ b/sdk/typescript/test/imported-dwallet/imported-dwallet-sign.test.ts
@@ -143,7 +143,5 @@ describe('Imported DWallet Signing', () => {
 			userShareEncryptionKeys,
 			testName,
 		);
-
-		expect(true).toBe(true);
 	});
 });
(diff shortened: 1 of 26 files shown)


sdk/typescript/test/imported-dwallet/imported-dwallet-sign.test.ts
```diff
@@ -17,140 +17,133 @@ import {
...
+			presignObject,
+			message,
+			Hash.KECCAK256,
+			SignatureAlgorithm.ECDSA,
+			encryptedUserSecretKeyShare,
+			userShareEncryptionKeys,
+			testName,
+		);
+
+		expect(true).toBe(true);
```
ycscaly: what's this?

Iamknownasfesal: It's a shallow verification if the test is passed or not, i needed it, removing now

PR #1456, opened 2025-08-21
fix: Stop ignoring threshold not reached messages

ItayLevyOfficial:
## Description 

The current chain's code had a bug that prevented it from handling threshold not reached scenario correctly. This PR fixes this issue.

## Test plan 

I ran the threshold not reached integration test with it & it functioned as expected.

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:


crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -257,10 +257,11 @@ impl DWalletMPCManager {
                     sender_authority=?sender_authority,
                     receiver_authority=?self.validator_name,
                     serialized_message=?message.message,
-                    "got a threshold not reached message, ignoring",
+                    "got a threshold not reached message",
                 );
 
-                return;
+                let serialized_mpc_round_number = &message.message[1..=8];
```
ycscaly, line 264:
```suggestion
                // TODO: temporary, remove this
                let serialized_mpc_round_number = &message.message[1..=8];
```

PR #1457, opened 2025-08-21
chore: Sessions decoupling

yaelAbergel:


crates/ika-core/src/dwallet_mpc/crytographic_computation.rs
```diff
@@ -18,6 +18,6 @@ const MPC_SIGN_SECOND_ROUND: u64 = 2;
 pub(crate) struct ComputationId {
     pub(crate) session_identifier: SessionIdentifier,
     pub(crate) consensus_round: u64,
-    pub(crate) mpc_round: u64,
+    pub(crate) current_round: u64,
```
ycscaly: Leave it as `mpc_round`, just make it an `Option`

crates/ika-types/src/messages_dwallet_mpc.rs
```diff
@@ -237,7 +240,7 @@ pub type AsyncProtocol = twopc_mpc::secp256k1::class_groups::AsyncProtocol;
 pub struct DWalletSessionEvent<E: DWalletSessionEventTrait> {
     pub epoch: u64,
     pub session_object_id: ObjectID,
-    pub session_type: SessionType,
+    pub session_type: SessionSource,
```
ycscaly: if you rename the struct, rename the variable

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -28,53 +30,99 @@ pub(crate) struct DWalletMPCSessionOutput {
...
+        /// Starts at `1` and increments after each successful advance of the session.
+        /// In round `1` We start the flow, without messages, from the event trigger.
+        current_mpc_round: u64,
+
+        /// All the messages that have been received for this session from each party, by consensus round and then by MPC round.
+        /// Used to build the input of messages to advance each round of the session.
+        messages_by_consensus_round: HashMap<u64, HashMap<PartyID, MPCMessage>>,
+    },
+    Native,
+}
```
ycscaly: yeah, that's the way

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -150,10 +211,11 @@ impl DWalletMPCSession {
         }
     }
 
+    // Todo (#1452): Remove.
     /// Records a threshold not reached error that we got when advancing
     /// this session with messages up to `consensus_round`.
     pub(crate) fn record_threshold_not_reached(&mut self, consensus_round: u64) {
```
ycscaly: you can delete this in this PR already I think

ycscaly (commented):
In the right direction

crates/ika-types/src/messages_dwallet_mpc.rs
omersadika: Why to be inconsistent with the move struct name?

crates/ika-core/src/dwallet_mpc/mpc_session.rs
omersadika: I don't think Native is a good name, maybe something like NetworkConsensus?

ycscaly: I think its a much better name then `NetworkConsensus` - these are computations that cannot be made in a smart contract, as they require native code to be executed i.e. native

crates/ika-core/src/dwallet_mpc/crytographic_computation/orchestrator.rs
```diff
@@ -120,13 +120,13 @@ impl CryptographicComputationsOrchestrator {
...
             let session_identifier = computation_update.computation_id.session_identifier;
-            let mpc_round = computation_update.computation_id.mpc_round;
+            let mpc_round = computation_update.computation_id.current_round;
             let attempt_number = computation_update.computation_id.attempt_number;
             let elapsed_ms = computation_update.elapsed_ms;
 
             debug!(
                 session_identifier=?computation_update.computation_id.session_identifier,
-                mpc_round=?computation_update.computation_id.mpc_round,
+                current_round=?computation_update.computation_id.current_round,
```
ycscaly: so I don't want it even printed for native computations. I.e., if `None`, the field shouldn't even be logged - there is no meaning for `round` in a native computation

crates/ika-core/src/dwallet_mpc/crytographic_computation/orchestrator.rs
```diff
@@ -227,13 +227,15 @@ impl CryptographicComputationsOrchestrator {
         let protocol_metadata: DWalletSessionRequestMetricData =
             (&computation_request.protocol_cryptographic_data).into();
 
-        dwallet_mpc_metrics
-            .add_advance_call(&protocol_metadata, &computation_id.mpc_round.to_string());
+        dwallet_mpc_metrics.add_advance_call(
+            &protocol_metadata,
+            &computation_id.current_round.to_string(),
```
ycscaly: same in metrics, no label for native

crates/ika-core/src/dwallet_mpc/crytographic_computation/request.rs
```diff
@@ -70,7 +70,7 @@ impl Request {
         // and keep private!
         let mut rng = root_seed.mpc_round_rng(
             session_id,
-            computation_id.mpc_round,
+            computation_id.current_round,
```
ycscaly:
```suggestion
            computation_id.current_round.unwrap_or(1),
```

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -28,53 +30,99 @@ pub(crate) struct DWalletMPCSessionOutput {
...
-    pub(super) status: MPCSessionStatus,
+    pub(super) status: SessionStatus,
 
-    /// The current MPC round number of the session.
-    /// Starts at `1` and increments after each successful advance of the session.
-    /// In round `1` We start the flow, without messages, from the event trigger.
-    pub(super) current_mpc_round: u64,
+    pub(super) session_type: SessionType,
 
+    // Todo (#1452): Remove threshold not reached.
```
ycscaly: delete it now

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -28,53 +30,99 @@ pub(crate) struct DWalletMPCSessionOutput {
...
+    Failed,
+}
+
+#[derive(Clone, Debug)]
+pub enum SessionType {
+    MPC {
+        /// The current MPC round number of the session.
+        /// Starts at `1` and increments after each successful advance of the session.
+        /// In round `1` We start the flow, without messages, from the event trigger.
+        current_mpc_round: u64,
```
ycscaly:
```suggestion
        current_round: u64,
```
Don't stutter (its inside `MPC` variant)

ItayLevyOfficial:
CI failing

crates/ika-core/src/dwallet_mpc/crytographic_computation/protocol_cryptographic_data.rs
```diff
@@ -465,21 +387,444 @@ impl ProtocolCryptographicData {
...
+            ProtocolCryptographicData::ImportedKeyVerification {
+                advance_request, ..
+            } => Some(advance_request.mpc_round_number),
+            ProtocolCryptographicData::EncryptedShareVerification { .. }
+            | ProtocolCryptographicData::PartialSignatureVerification { .. }
+            | ProtocolCryptographicData::MakeDWalletUserSecretKeySharesPublic { .. } => None,
+        }
+    }
+
+    pub(crate) fn compute_mpc(
```
ycscaly: move to the mpc computations module

crates/ika-core/src/dwallet_mpc/crytographic_computation/protocol_cryptographic_data.rs
```diff
@@ -465,21 +387,444 @@ impl ProtocolCryptographicData {
...
+            _ => {
+                error!(
+                    session_type=?protocol_metadata,
+                    session_identifier=?session_identifier,
+                    "Invalid session type for mpc computation");
+                Err(DwalletMPCError::InvalidSessionType)
+            }
+        }
+    }
+    pub(crate) fn compute_native(
```
ycscaly: move to the native computations module

crates/ika-core/src/dwallet_mpc/crytographic_computation/request.rs
```diff
@@ -55,374 +34,25 @@ impl Request {
...
-                let result = Party::<PresignParty>::advance_with_guaranteed_output(
-                    session_id,
-                    self.party_id,
-                    &self.access_structure,
-                    advance_request,
-                    None,
-                    &public_input,
-                    &mut rng,
-                )?;
+        let Some(mpc_round) = computation_id.mpc_round else {
```
ycscaly: `if let {} else {}`

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -183,15 +183,15 @@ impl DWalletMPCManager {
...
         for (_, session) in self.mpc_sessions.iter_mut() {
-            let SessionType::MPC {
+            let ComputationType::MPC {
                 messages_by_consensus_round,
                 ..
-            } = &mut session.session_type
+            } = &mut session.computation_type
             else {
                 error!(
                     should_never_happen=true,
```
ycscaly: I'm not sure it shouldn't happen, a malicious validator could cause this right?

ycscaly: in which case we shouldn't `return`

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -172,35 +159,34 @@ impl DWalletSession {
...
-                message_mpc_round=?mpc_round_number,
-                current_mpc_round,
-                new_mpc_round,
-                mpc_protocol,
-                "Advancing current MPC round",
-            );
-
-            *current_mpc_round = new_mpc_round;
-        }
+        // if sender_party_id == self.party_id && *current_mpc_round <= mpc_round_number {
```
ycscaly: don't comment out, delete, also delete the `sessions.mpc_round`

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
yaelAbergel: We can delete this code because we use `entry().or_default()` we we insert a new message: 

```
let consensus_round_messages_map = messages_by_consensus_round
            .entry(consensus_round)
            .or_default();
```

yaelAbergel: Also it was a bug

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -118,65 +151,32 @@ impl DWalletMPCSession {
...
+        } = &mut self.computation_type
+        else {
             error!(
                 should_never_happen=true,
                 session_identifier=?self.session_identifier,
-                "tried to record threshold not reached for a non-active MPC session"
+                computation_type=?self.computation_type,
+                "tried to add a message to a non-MPC session"
             );
             return;
```
ycscaly: You just moved the bug here, a malicious validator can send a message for a native session, you won't report it as malicious, and you have `should_never_happen=true,` which is wrong

crates/ika-core/src/dwallet_mpc/integration_tests/threshold_not_reached.rs
ycscaly, line 1: @yaelAbergel @ItayLevyOfficial why delete? we should still test it

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -1,8 +1,570 @@
...
+
+                ProtocolCryptographicData::NetworkEncryptionKeyReconfiguration {
+                    data: data.clone(),
+                    public_input: public_input.clone(),
+                    advance_request,
+                    decryption_key_shares: decryption_key_shares.clone(),
+                }
+            }
+            _ => {
+                return Err(DwalletMPCError::InvalidSessionType);
```
ycscaly: why session type ? its protocol type right?

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -28,58 +31,96 @@ pub(crate) struct DWalletMPCSessionOutput {
...
+    },
+    WaitingForSessionRequest,
+    ComputationCompleted,
+    Completed,
+    Failed,
+}
 
-    outputs_by_consensus_round: HashMap<u64, HashMap<PartyID, DWalletMPCSessionOutput>>,
+#[derive(Clone, Debug)]
+pub enum ComputationType {
```
ycscaly:
```suggestion
pub enum SessionComputationType {
```

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -320,46 +276,62 @@ impl DWalletMPCManager {
...
+                        messages_by_consensus_round: HashMap::new(),
+                    },
                 );
                 // Safe to `unwrap()`: we just created the session.
                 self.mpc_sessions.get_mut(&session_identifier).unwrap()
             }
         };
 
-        session.add_message(consensus_round, mpc_round_number, sender_party_id, message);
+        let SessionComputationType::MPC { .. } = &mut session.computation_type else {
```
ycscaly: I think a better way to do this is `if matches!(..)`

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -320,46 +276,62 @@ impl DWalletMPCManager {
...
+        status: SessionStatus,
+        session_type: SessionComputationType,
     ) {
         info!(
             status=?status,
             "Received start MPC flow request for session identifier {:?}",
             session_identifier,
         );
-        let with_request_data = request.is_some();
+        let with_request_data = matches!(status, SessionStatus::Active { .. });
```
ycscaly:
```suggestion
        let active = matches!(status, SessionStatus::Active { .. });
```

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -204,7 +214,7 @@ impl DWalletMPCSession {
             // Received an output from ourselves from the consensus, so it's safe to mark the session as computation completed.
             info!(
                 authority=?self.validator_name,
-                current_mpc_round=self.current_mpc_round,
+                // computation_type=self.computation_type, // yael
```
ycscaly: @yaelAbergel

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -474,10 +494,61 @@ impl DWalletMPCManager {
...
+                            "Tried to update an existing session with a different computation type; \
+                             marking senders as malicious."
+                        );
+
+                        if !malicious_parties.is_empty() {
+                            // Convert HashSet -> Vec for the API (if that’s what it expects)
+                            let malicious_vec: Vec<_> = malicious_parties.iter().cloned().collect();
+                            self.record_malicious_actors(&malicious_vec);
+                        }
+                        return;
```
ycscaly: its not good to return here, it allows other validators to DOS a protocol. We should instead switch the session type, and drop the messages

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -474,10 +494,61 @@ impl DWalletMPCManager {
...
+                        error!(
+                            should_never_happen=true,
+                            session_identifier=?session_identifier,
+                            existing_computation_type=?session.computation_type,
+                            new_computation_type=?new_type,
+                            "tried to update an existing session with a different computation type"
+                        );
+                        return;
+                    }
+                }
```
ycscaly: I don't really think we need this check, since we don't get messages from other validators here, the only case we can have is getting two events with different types for the same session ID, but in that case we don't really have much to do but ignore the second event

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -497,56 +498,40 @@ impl DWalletMPCManager {
...
-                        error!(
-                            should_never_happen=true,
-                            session_identifier=?session_identifier,
-                            existing_computation_type=?session.computation_type,
-                            new_computation_type=?new_type,
-                            "tried to update an existing session with a different computation type"
-                        );
-                        return;
+                    session.computation_type = new_type;
+                    if !malicious_parties.is_empty() {
```
ycscaly: has to be non-empty if we're here

ycscaly (approved):

PR #1468, opened 2025-08-26
chore: Update `cryptography-private` version

ItayLevyOfficial:
## Description 

Update to new version of cryptography-private crate

The new version introduces breaking changes to DWallet DKG outputs. 
Version wrapping and conversion have been added to maintain backward 
compatibility.

I also updated the DWallet DKG second round to use the DWallet ID as the 
unique identifier instead of the session ID. The session ID changes 
between the first and second rounds, which caused proof verification 
to fail.

## Test plan 

I ran the chain and ran a full sign TS flow against it.

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,7 +61,7 @@ jobs:
         run: cargo fmt -- --check
       - name: Fmt Check - Wasm
         run: cargo fmt -- --check
-        working-directory: sdk/mpc-wasm
+        working-directory: sdk/ika-wasm
   clippy:
     if: github.event_name == 'push'
     name: Clippy
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,5 @@ ika_config.json
 
 # TypeDoc
 sdk/typescript/docs
+move_pkg_lock_root.lock
+node-compile-cache
--- a/crates/dwallet-mpc-centralized-party/src/lib.rs
+++ b/crates/dwallet-mpc-centralized-party/src/lib.rs
@@ -10,9 +10,10 @@ use anyhow::{Context, anyhow};
 use class_groups::dkg::Secp256k1Party;
 use class_groups::setup::get_setup_parameters_secp256k1;
 use class_groups::{
-    CiphertextSpaceGroupElement, DecryptionKey, EncryptionKey,
-    SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS, SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
-    Secp256k1DecryptionKey,
+    CiphertextSpaceGroupElement, DEFAULT_COMPUTATIONAL_SECURITY_PARAMETER, DecryptionKey,
+    EncryptionKey, SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+    SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS, SECP256K1_SCALAR_LIMBS, Secp256k1DecryptionKey,
+    setup::DeriveFromPlaintextPublicParameters,
 };
 use dwallet_mpc_types::dwallet_mpc::{
     DKGDecentralizedPartyOutputSecp256k1, DKGDecentralizedPartyVersionedOutputSecp256k1,
@@ -493,16 +494,36 @@ fn protocol_public_parameters_by_key_scheme(
                     let encryption_scheme_public_parameters = network_dkg_public_output
                         .default_encryption_scheme_public_parameters::<secp256k1::GroupElement>(
                     )?;
+
+                    let setup_parameters =
+                        class_groups::setup::SetupParameters::SetupParameters::<
+                            SECP256K1_SCALAR_LIMBS,
+                            SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                            SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                            group::secp256k1::scalar::PublicParameters,
+                        >::derive_from_plaintext_parameters::<group::secp256k1::Scalar>(
+                            group::secp256k1::scalar::PublicParameters::default(),
+                            DEFAULT_COMPUTATIONAL_SECURITY_PARAMETER,
+                        )?;
+
+                    let neutral_group_value =
+                        group::secp256k1::GroupElement::neutral_from_public_parameters(
+                            &group::secp256k1::group_element::PublicParameters::default(),
+                        )
+                        .map_err(twopc_mpc::Error::from)?
+                        .value();
+                    let neutral_ciphertext_value = ::class_groups::CiphertextSpaceGroupElement::neutral_from_public_parameters(&setup_parameters.ciphertext_space_public_parameters())?.value();
+
                     let protocol_public_parameters = ProtocolPublicParameters::new::<
                         { secp256k1::SCALAR_LIMBS },
                         { SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS },
                         { SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS },
                         secp256k1::GroupElement,
                     >(
-                        Default::default(),
-                        Default::default(),
-                        Default::default(),
-                        Default::default(),
+                        neutral_group_value,
+                        neutral_group_value,
+                        neutral_ciphertext_value,
+                        neutral_ciphertext_value,
                         encryption_scheme_public_parameters.clone(),
                     );
                     Ok(protocol_public_parameters)
--- a/crates/dwallet-mpc-centralized-party/src/lib.rs
+++ b/crates/dwallet-mpc-centralized-party/src/lib.rs
@@ -347,6 +347,7 @@ pub fn advance_centralized_sign_party(
         bcs::from_bytes(&centralized_party_secret_key_share)?;
     let VersionedDwalletUserSecretShare::V1(centralized_party_secret_key_share) =
         centralized_party_secret_key_share;
+    // TODO (#1478): Use From to convert the decentralized DKG output to centralized public output.
     let decentralized_output = match decentralized_dkg_output {
         DKGDecentralizedPartyVersionedOutput::<
             { group::secp256k1::SCALAR_LIMBS },
--- a/crates/dwallet-mpc-centralized-party/src/lib.rs
+++ b/crates/dwallet-mpc-centralized-party/src/lib.rs
@@ -170,7 +170,7 @@ pub fn create_dkg_output_v1(
     protocol_pp: Vec<u8>,
     decentralized_first_round_public_output: SerializedWrappedMPCPublicOutput,
 ) -> anyhow::Result<CentralizedDKGWasmResult> {
-    let public_parameters: ProtocolPublicParameters = bcs::from_bytes(&protocol_pp)?;
+    let protocol_public_parameters: ProtocolPublicParameters = bcs::from_bytes(&protocol_pp)?;
     let decentralized_first_round_public_output =
         bcs::from_bytes(&decentralized_first_round_public_output)?;
     match decentralized_first_round_public_output {
@@ -194,7 +194,7 @@ pub fn create_dkg_output_v1(
                 second_second_part,
                 first_first_part,
                 second_first_part,
-                public_parameters
+                protocol_public_parameters
                     .encryption_scheme_public_parameters
                     .clone(),
             );
--- a/crates/dwallet-mpc-centralized-party/src/lib.rs
+++ b/crates/dwallet-mpc-centralized-party/src/lib.rs
@@ -15,6 +15,7 @@ use class_groups::{
     Secp256k1DecryptionKey,
 };
 use dwallet_mpc_types::dwallet_mpc::{
+    DKGDecentralizedPartyOutputSecp256k1, DKGDecentralizedPartyVersionedOutputSecp256k1,
     DWalletMPCNetworkKeyScheme, SerializedWrappedMPCPublicOutput,
     VersionedCentralizedDKGPublicOutput, VersionedDwalletDKGFirstRoundPublicOutput,
     VersionedDwalletDKGSecondRoundPublicOutput, VersionedDwalletUserSecretShare,
@@ -22,7 +23,7 @@ use dwallet_mpc_types::dwallet_mpc::{
     VersionedImportedDwalletOutgoingMessage, VersionedNetworkDkgOutput, VersionedPresignOutput,
     VersionedPublicKeyShareAndProof, VersionedSignOutput, VersionedUserSignedMessage,
 };
-use group::{CyclicGroupElement, GroupElement, OsCsRng, Samplable, secp256k1};
+use group::{CyclicGroupElement, GroupElement, HashType, OsCsRng, Samplable, secp256k1};
 use homomorphic_encryption::{
     AdditivelyHomomorphicDecryptionKey, AdditivelyHomomorphicEncryptionKey,
     GroupsPublicParametersAccessors,
@@ -35,12 +36,17 @@ use twopc_mpc::secp256k1::SCALAR_LIMBS;
 use class_groups::encryption_key::public_parameters::Instantiate;
 use commitment::CommitmentSizedNumber;
 use message_digest::message_digest::message_digest;
-use twopc_mpc::class_groups::{DKGCentralizedPartyOutput, DKGDecentralizedPartyOutput};
+use twopc_mpc::class_groups::{
+    DKGCentralizedPartyOutput, DKGCentralizedPartyVersionedOutput, DKGDecentralizedPartyOutput,
+    DKGDecentralizedPartyVersionedOutput,
+};
 use twopc_mpc::dkg::Protocol;
-use twopc_mpc::secp256k1::class_groups::ProtocolPublicParameters;
-use twopc_mpc::sign::verify_signature;
+use twopc_mpc::ecdsa::sign::verify_signature;
+use twopc_mpc::secp256k1::class_groups::{
+    FUNDAMENTAL_DISCRIMINANT_LIMBS, NON_FUNDAMENTAL_DISCRIMINANT_LIMBS, ProtocolPublicParameters,
+};
 
-type AsyncProtocol = twopc_mpc::secp256k1::class_groups::AsyncProtocol;
+type AsyncProtocol = twopc_mpc::secp256k1::class_groups::AsyncECDSAProtocol;
 type DKGCentralizedParty = <AsyncProtocol as twopc_mpc::dkg::Protocol>::DKGCentralizedPartyRound;
 pub type SignCentralizedParty = <AsyncProtocol as twopc_mpc::sign::Protocol>::SignCentralizedParty;
 
@@ -75,6 +81,8 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
     Ok(bcs::to_bytes(&public_parameters)?)
 }
 
+pub type DWalletDKGFirstParty = twopc_mpc::secp256k1::class_groups::EncryptionOfSecretKeyShareParty;
+
 /// Executes the second phase of the DKG protocol, part of a three-phase DKG flow.
 ///
 /// This function is invoked by the centralized party to produce:
@@ -97,26 +105,107 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
 /// # Errors
 /// Return an error if decoding or advancing the protocol fails.
 /// This is okay since a malicious blockchain can always block a client.
-pub fn create_dkg_output(
+pub fn create_dkg_output_v2(
+    protocol_pp: Vec<u8>,
+    session_id: Vec<u8>,
+) -> anyhow::Result<CentralizedDKGWasmResult> {
+    let public_parameters: ProtocolPublicParameters = bcs::from_bytes(&protocol_pp)?;
+    let session_identifier = CommitmentSizedNumber::from_le_slice(&session_id);
+    let round_result = DKGCentralizedParty::advance(
+        (),
+        &(),
+        &(public_parameters, session_identifier).into(),
+        &mut OsCsRng,
+    )
+    .context("advance() failed on the DKGCentralizedParty")?;
+
+    // Centralized Public Key Share and Proof.
+    let public_key_share_and_proof =
+        VersionedPublicKeyShareAndProof::V1(bcs::to_bytes(&round_result.outgoing_message)?);
+
+    let public_key_share_and_proof = bcs::to_bytes(&public_key_share_and_proof)?;
+    let centralized_output = round_result.public_output;
+
+    // Public Output:
+    let public_output = bcs::to_bytes(&VersionedCentralizedDKGPublicOutput::V2(bcs::to_bytes(
+        &centralized_output,
+    )?))?;
+    // Centralized Secret Key Share.
+    // Warning:
+    // The secret (private)
+    // key share returned from this function should never be sent
+    // and should always be kept private.
+    let centralized_secret_output =
+        VersionedDwalletUserSecretShare::V1(bcs::to_bytes(&round_result.private_output)?);
+    let centralized_secret_output = bcs::to_bytes(&centralized_secret_output)?;
+    Ok(CentralizedDKGWasmResult {
+        public_output,
+        public_key_share_and_proof,
+        centralized_secret_output,
+    })
+}
+/// Executes the second phase of the DKG protocol, part of a three-phase DKG flow.
+///
+/// This function is invoked by the centralized party to produce:
+/// - A public key share and its proof.
+/// - Centralized DKG output required for further protocol steps.
+/// # Warning
+/// The secret (private) key returned from this function should never be sent
+/// and should always be kept private.
+///
+/// # Parameters
+/// — `decentralized_first_round_output`:
+///    Serialized output of the decentralized party from the first DKG round.
+/// — `session_id`: Unique hexadecimal string identifying the session.
+///
+/// # Returns
+/// A tuple containing:
+/// - Serialized public key share and proof.
+/// - Serialized centralized DKG output.
+///
+/// # Errors
+/// Return an error if decoding or advancing the protocol fails.
+/// This is okay since a malicious blockchain can always block a client.
+pub fn create_dkg_output_v1(
     protocol_pp: Vec<u8>,
     decentralized_first_round_public_output: SerializedWrappedMPCPublicOutput,
-    session_identifier: Vec<u8>,
 ) -> anyhow::Result<CentralizedDKGWasmResult> {
     let public_parameters: ProtocolPublicParameters = bcs::from_bytes(&protocol_pp)?;
     let decentralized_first_round_public_output =
         bcs::from_bytes(&decentralized_first_round_public_output)?;
     match decentralized_first_round_public_output {
         VersionedDwalletDKGFirstRoundPublicOutput::V1(decentralized_first_round_public_output) => {
-            let (decentralized_first_round_public_output, _): <<AsyncProtocol as Protocol>::EncryptionOfSecretKeyShareRoundParty as Party>::PublicOutput =
-                bcs::from_bytes(&decentralized_first_round_public_output)
+            let (output, session_identifier) =
+                bcs::from_bytes::<(Vec<u8>, _)>(&decentralized_first_round_public_output)?;
+            let [first_part, second_part]: <DWalletDKGFirstParty as Party>::PublicOutput =
+                bcs::from_bytes(&output)
                     .context("failed to deserialize decentralized first round DKG output")?;
-
-            let session_identifier = CommitmentSizedNumber::from_le_slice(&session_identifier);
-
+            let (first_first_part, first_second_part) = first_part.into();
+            let (second_first_part, second_second_part) = second_part.into();
+            // This is a temporary hack to keep working with the existing 2-round dWallet DKG mechanism.
+            // TODO (#1470): Use one network round in the dWallet DKG flow.
+            let protocol_pp_with_decentralized_dkg_output = ProtocolPublicParameters::new::<
+                { group::secp256k1::SCALAR_LIMBS },
+                SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                group::secp256k1::GroupElement,
+            >(
+                first_second_part,
+                second_second_part,
+                first_first_part,
+                second_first_part,
+                public_parameters
+                    .encryption_scheme_public_parameters
+                    .clone(),
+            );
             let round_result = DKGCentralizedParty::advance(
-                decentralized_first_round_public_output,
+                (),
                 &(),
-                &(public_parameters, session_identifier).into(),
+                &(
+                    protocol_pp_with_decentralized_dkg_output,
+                    session_identifier,
+                )
+                    .into(),
                 &mut OsCsRng,
             )
             .context("advance() failed on the DKGCentralizedParty")?;
@@ -126,11 +215,26 @@ pub fn create_dkg_output(
                 VersionedPublicKeyShareAndProof::V1(bcs::to_bytes(&round_result.outgoing_message)?);
 
             let public_key_share_and_proof = bcs::to_bytes(&public_key_share_and_proof)?;
+            // TODO(#1470): Use one network round in the dWallet DKG flow.
+            // This is a temporary hack to keep working with the existing 2-round dWallet DKG mechanism.
+            let centralized_output = match round_result.public_output {
+                DKGCentralizedPartyVersionedOutput::<
+                    { group::secp256k1::SCALAR_LIMBS },
+                    group::secp256k1::GroupElement,
+                >::UniversalPublicDKGOutput {
+                    output: dkg_output,
+                    ..
+                } => dkg_output,
+                DKGCentralizedPartyVersionedOutput::<
+                    { group::secp256k1::SCALAR_LIMBS },
+                    group::secp256k1::GroupElement,
+                >::TargetedPublicDKGOutput(output) => output,
+            };
 
             // Public Output:
             // centralized_public_key_share + public_key + decentralized_party_public_key_share
             let public_output = bcs::to_bytes(&VersionedCentralizedDKGPublicOutput::V1(
-                bcs::to_bytes(&round_result.public_output)?,
+                bcs::to_bytes(&centralized_output)?,
             ))?;
             // Centralized Secret Key Share.
             // Warning:
@@ -153,8 +257,28 @@ pub fn public_key_from_dwallet_output_inner(dwallet_output: Vec<u8>) -> anyhow::
     let dkg_output: VersionedDwalletDKGSecondRoundPublicOutput = bcs::from_bytes(&dwallet_output)?;
     match dkg_output {
         VersionedDwalletDKGSecondRoundPublicOutput::V1(dkg_output) => {
+            let output: DKGDecentralizedPartyOutputSecp256k1 = bcs::from_bytes(&dkg_output)?;
+            Ok(bcs::to_bytes(&output.public_key)?)
+        }
+        VersionedDwalletDKGSecondRoundPublicOutput::V2(dkg_output) => {
             let dkg_output: DKGDecentralizedOutput = bcs::from_bytes(&dkg_output)?;
-            let public_key = dkg_output.public_key;
+            let public_key = match dkg_output {
+                DKGDecentralizedPartyVersionedOutput::<
+                    { group::secp256k1::SCALAR_LIMBS },
+                    SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                    SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                    group::secp256k1::GroupElement,
+                >::UniversalPublicDKGOutput {
+                    output: dkg_output,
+                    ..
+                } => dkg_output.public_key,
+                DKGDecentralizedPartyVersionedOutput::<
+                    { group::secp256k1::SCALAR_LIMBS },
+                    SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                    SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+                    group::secp256k1::GroupElement,
+                >::TargetedPublicDKGOutput(output) => output.public_key,
+            };
             Ok(bcs::to_bytes(&public_key)?)
         }
     }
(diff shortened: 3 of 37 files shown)


crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -83,18 +83,12 @@ pub(crate) fn sign_session_public_input(
...
-                    .map_err(|e| DwalletMPCError::SignatureVerificationFailed(e.to_string()))?,
-            )
-            .map_err(|e| DwalletMPCError::SignatureVerificationFailed(e.to_string()))?,
-        )?,
+        message,
         presign,
         message_centralized_signature,
         decryption_pp,
         expected_decrypters,
+        hash_scheme,
```
ycscaly, line 92: here we need to be careful, we need to make sure the enums convert well because there is one here/in the move, and one in the cryptography. Also now we need to delete the `message_digest` crate

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/dwallet_dkg.rs
```diff
@@ -96,6 +107,23 @@ impl DWalletDKGSecondPartyPublicInputGenerator for DWalletDKGSecondParty {
             VersionedCentralizedDKGPublicOutput::V1(first_round_output) => {
```
ycscaly: No but, you should definitely add a `V2` here too - the outputs of this party and the old one are different.
If you want to avoid adding a V2 here, we should have the exact same structure, which I think adds an `SID` there too

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -399,6 +399,10 @@ fn instantiate_dwallet_mpc_network_encryption_key_public_data_from_dkg_public_ou
                     { NON_FUNDAMENTAL_DISCRIMINANT_LIMBS },
                     secp256k1::GroupElement,
                 >(
+                    Default::default(),
+                    Default::default(),
+                    Default::default(),
+                    Default::default(),
```
ycscaly: I think a safer option would be to have `neutral_from_public_parameters().value()` here

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -209,9 +206,9 @@ pub(crate) fn verify_partial_signature(
             let partial: <AsyncProtocol as twopc_mpc::sign::Protocol>::SignMessage =
                 bcs::from_bytes(&partially_signed_message)?;
 
-            twopc_mpc::sign::decentralized_party::signature_partial_decryption_round::Party::verify_encryption_of_signature_parts_prehash_class_groups(
+            twopc_mpc::ecdsa::sign::decentralized_party::signature_partial_decryption_round::Party::verify_encryption_of_signature_parts_prehash_class_groups(
                 protocol_public_parameters,
-                dkg_output,
+                dkg_output.into(),
```
ycscaly: This is safe only if `dkg_output` was network version V1, in V2 you don't need it

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -107,16 +113,34 @@ pub fn create_dkg_output(
...
                 &(),
-                &(public_parameters, session_identifier).into(),
+                &(
+                    protocol_pp_with_decentralized_dkg_output,
+                    session_identifier,
+                )
+                    .into(),
                 &mut OsCsRng,
             )
             .context("advance() failed on the DKGCentralizedParty")?;
```
ycscaly, line 212: there's another ugly part you are missing for both the user and network: when you get back the public output, it is already in the `VersionedOutput::UniversalPublicDKGOutput`. But this was a hack, it isn't really the case, and instead we should have it be as the old V1 `output` field inside it.

Later, when V1 (network version) wallets are being deserialized, they should be deserialized as `Output` and not `VersionedOutput`, and then you use `.into()` to transform them into a `VersionedOutput::TargetedPublicDKGOutput`

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ItayLevyOfficial: Are you sure we need to delete the message digest crate? It's still being used before calling the `verify_partial_signature` function

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ItayLevyOfficial: Should I change it or no then? to what exactly?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ItayLevyOfficial: The code doesn't compile without this - what should I change to get it to compile?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ItayLevyOfficial: scaly said I should do this now as it may cause the bug

ItayLevyOfficial: encryption_scheme.cipher_text public params

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ItayLevyOfficial: wdym by being careful? what should I change? the message_digest crate is being used by the verify_secp_signature_inner function - are you sure it need to be removed? What should I use there instead?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/dwallet_dkg.rs
ItayLevyOfficial: Added v2

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -153,7 +153,7 @@ pub fn create_dkg_output(
 
             // Public Output:
             // centralized_public_key_share + public_key + decentralized_party_public_key_share
-            let public_output = bcs::to_bytes(&VersionedCentralizedDKGPublicOutput::V1(
+            let public_output = bcs::to_bytes(&VersionedCentralizedDKGPublicOutput::V2(
```
ycscaly: No, so - V2 should only use when we don't hack. When we hack, keep it V1 as its the same struct

ycscaly: also, you shouldn't save `round_result.public_output`, you should take the `output` field from the `Universal..` variant and save it here

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -205,17 +205,23 @@ pub fn centralized_and_decentralized_parties_dkg_output_match_inner(
 ) -> anyhow::Result<bool> {
     let versioned_centralized_dkg_output =
         bcs::from_bytes::<VersionedCentralizedDKGPublicOutput>(centralized_dkg_output)?;
-    let VersionedCentralizedDKGPublicOutput::V1(centralized_dkg_output) =
-        versioned_centralized_dkg_output;
+    let VersionedCentralizedDKGPublicOutput::V2(centralized_dkg_output) =
+        versioned_centralized_dkg_output
+    else {
+        return Err(anyhow!("Only V2 centralized DKG output is supported"));
+    };
```
ycscaly: Not true, both are supported, and you should transition between them like I said using `into()`

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -205,17 +205,23 @@ pub fn centralized_and_decentralized_parties_dkg_output_match_inner(
...
 
     let versioned_decentralized_dkg_output =
         bcs::from_bytes::<VersionedCentralizedDKGPublicOutput>(decentralized_dkg_output)?;
-    let VersionedCentralizedDKGPublicOutput::V1(decentralized_dkg_output) =
-        versioned_decentralized_dkg_output;
+    let VersionedCentralizedDKGPublicOutput::V2(decentralized_dkg_output) =
+        versioned_decentralized_dkg_output
+    else {
+        return Err(anyhow!("Only V2 decentralized DKG output is supported"));
+    };
```
ycscaly: same

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -403,7 +409,7 @@ fn protocol_public_parameters_by_key_scheme(
                     let encryption_scheme_public_parameters = network_dkg_public_output
                         .default_encryption_scheme_public_parameters::<secp256k1::GroupElement>(
                     )?;
-                    Ok(ProtocolPublicParameters::new::<
+                    let ppp = ProtocolPublicParameters::new::<
```
ycscaly:
```suggestion
                    let protocol_public_parameters = ProtocolPublicParameters::new::<
```

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -410,10 +412,26 @@ impl ProtocolCryptographicData {
...
+                            output: decentralized_output,
+                            ..
+                        } = decentralized_output
+                        else {
+                            return Err(DwalletMPCError::InvalidHashScheme);
+                        };
+                        let public_output_value =
+                            bcs::to_bytes(&VersionedDwalletDKGSecondRoundPublicOutput::V1(
+                                bcs::to_bytes(&decentralized_output).unwrap(),
+                            ))?;
```
ycscaly, line 469: good

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/dwallet_dkg.rs
ycscaly: its not in place of v1, you should support both

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: interesting, this will may fix the bug

.github/workflows/ci.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
         run: cargo fmt -- --check
       - name: Fmt Check - Wasm
         run: cargo fmt -- --check
-        working-directory: sdk/mpc-wasm
+        working-directory: sdk/ika-wasm
```
ItayLevyOfficial, line 64: needed as the current ci was broken

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -102,21 +112,43 @@ pub fn create_dkg_output(
     decentralized_first_round_public_output: SerializedWrappedMPCPublicOutput,
     session_identifier: Vec<u8>,
 ) -> anyhow::Result<CentralizedDKGWasmResult> {
+    web_sys::console::log_1(&"Hello, world!".into());
```
ycscaly: delete prints

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -102,21 +112,43 @@ pub fn create_dkg_output(
...
+            >(
+                first_second_part,
+                second_second_part,
+                first_first_part,
+                second_first_part,
+                public_parameters
+                    .encryption_scheme_public_parameters
+                    .clone(),
+            );
             let session_identifier = CommitmentSizedNumber::from_le_slice(&session_identifier);
```
ycscaly: where is it coming from?

ycscaly: we should take it to be the same like the first round

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -102,21 +112,43 @@ pub fn create_dkg_output(
     decentralized_first_round_public_output: SerializedWrappedMPCPublicOutput,
     session_identifier: Vec<u8>,
 ) -> anyhow::Result<CentralizedDKGWasmResult> {
```
ycscaly, line 173: Now, this function should only be called in the first epoch, after which we shouldn't have a two-stage DKG process

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -153,8 +200,28 @@ pub fn public_key_from_dwallet_output_inner(dwallet_output: Vec<u8>) -> anyhow::
     let dkg_output: VersionedDwalletDKGSecondRoundPublicOutput = bcs::from_bytes(&dwallet_output)?;
     match dkg_output {
         VersionedDwalletDKGSecondRoundPublicOutput::V1(dkg_output) => {
+            let output: SpecificDKGDecentralizedPartyOutput = bcs::from_bytes(&dkg_output)?;
+            Ok(bcs::to_bytes(&output.public_key)?)
+        }
+        VersionedDwalletDKGSecondRoundPublicOutput::V2(dkg_output) => {
             let dkg_output: DKGDecentralizedOutput = bcs::from_bytes(&dkg_output)?;
-            let public_key = dkg_output.public_key;
+            let public_key = match dkg_output {
```
ycscaly, line 266:
```suggestion
            let public_output = match dkg_output {
```

ycscaly: or dkg_output

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -169,33 +236,29 @@ pub fn centralized_and_decentralized_parties_dkg_output_match_inner(
...
-            group::secp256k1::GroupElement,
-        >,
-    >(decentralized_dkg_output.as_slice())?;
+        bcs::from_bytes::<VersionedDwalletDKGSecondRoundPublicOutput>(decentralized_dkg_output)?;
+    let decentralized_dkg_output = match versioned_decentralized_dkg_output {
+        VersionedDwalletDKGSecondRoundPublicOutput::V1(output) => {
+            bcs::from_bytes::<SpecificDKGDecentralizedPartyOutput>(output.as_slice())?.into()
+        }
+        VersionedDwalletDKGSecondRoundPublicOutput::V2(output) => {
+            bcs::from_bytes::<SpecificDKGDecentralizedPartyVersionedOutput>(output.as_slice())?
```
ycscaly: what is `specific`?

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -169,33 +236,29 @@ pub fn centralized_and_decentralized_parties_dkg_output_match_inner(
...
+    };
 
-    let does_centralized_and_decentralized_parties_dkg_output_match = centralized_dkg_output
-        .public_key_share
-        == decentralized_dkg_output.centralized_party_public_key_share
-        && centralized_dkg_output.decentralized_party_public_key_share
-            == decentralized_dkg_output.public_key_share
-        && centralized_dkg_output.public_key == decentralized_dkg_output.public_key;
+    let does_centralized_and_decentralized_parties_dkg_output_match =
+        decentralized_dkg_output == centralized_dkg_output;
```
ycscaly, line 319: good

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -213,53 +276,69 @@ pub fn advance_centralized_sign_party(
...
-            )
-            .context("advance() failed on the SignCentralizedParty")?;
-
-            let signed_message =
-                VersionedUserSignedMessage::V1(bcs::to_bytes(&round_result.outgoing_message)?);
-            let signed_message = bcs::to_bytes(&signed_message)?;
-            Ok(signed_message)
+    let decentralized_dkg_output = match bcs::from_bytes(&decentralized_party_dkg_public_output)? {
+        VersionedDwalletDKGSecondRoundPublicOutput::V1(output) => {
+            bcs::from_bytes::<SpecificDKGDecentralizedPartyOutput>(output.as_slice())?.into()
```
ycscaly: good `into()`

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -213,53 +276,69 @@ pub fn advance_centralized_sign_party(
...
+            output: dkg_output,
+            ..
+        } => dkg_output,
+        DKGDecentralizedPartyVersionedOutput::<
+            { group::secp256k1::SCALAR_LIMBS },
+            SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+            SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+            group::secp256k1::GroupElement,
+        >::TargetedPublicDKGOutput(output) => output,
+    };
```
ycscaly, line 368: No, if its `UniversalPublicDKGOutput` you must have it stay as `UniversalPublicDKGOutput` for the centralized party. I can create a `From<>`

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: be careful because the enum variants might differ

ycscaly: yes you should remove that crate

ItayLevyOfficial: And what should I use instead to verify the final signature?

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: without generics - which name would make more sense for you?

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: Not sure I understand. I need to use in the following lines the 
```
public_key_share: decentralized_output.centralized_party_public_key_share,
        public_key: decentralized_output.public_key,
        decentralized_party_public_key_share: decentralized_output.public_key_share,
    };
```
How will I be able to use those fields without decomposing the DKG output?

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: Renamed to dwallet_id

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: Ok, I renamed it to create_dkg_v1 & created a separate create_dkg_v2 wasm function

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: Are you sure? The name of the inner variable I am taking here is `public_key`

ItayLevyOfficial:
> 2. I am missing a conversion of the dkg output in the network in the case of the hack flow, into `V1` and taking only the `output` like you did in the user

https://github.com/dwallet-labs/dwallet-network/blob/c6e18d9b510faf062a985cae659d347954b1c533/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs#L443

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: I think you can just remove the `Specific`

ycscaly: or add Secp256k1 at the end

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -83,8 +83,100 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
...
+    let protocol_pp_with_decentralized_dkg_output = ProtocolPublicParameters::new::<
+        { group::secp256k1::SCALAR_LIMBS },
+        SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+        SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+        group::secp256k1::GroupElement,
+    >(
+        Default::default(),
+        Default::default(),
+        Default::default(),
+        Default::default(),
```
ycscaly: No, but this has to come from the reconfiguration public output. Add a `todo` here

ycscaly: we can do it in the next PR just add a todo

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -83,8 +83,100 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
...
+        public_parameters
+            .encryption_scheme_public_parameters
+            .clone(),
+    );
+    let session_identifier = CommitmentSizedNumber::from_le_slice(&dwallet_id);
```
ycscaly: why? this is not the session id...

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -140,7 +230,7 @@ pub fn create_dkg_output(
                     .encryption_scheme_public_parameters
                     .clone(),
             );
-            let session_identifier = CommitmentSizedNumber::from_le_slice(&session_identifier);
+            let session_identifier = CommitmentSizedNumber::from_le_slice(&dwallet_id);
```
ycscaly: Again this is not the session ID, you must re-use the one from the first round

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: you need to take it from the event

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: I told you multiple times that this needs to be the same as the second's round session ID (This has been changed by you in your new version), and making the first & second dwallet dkg rounds use the same session ID is hard and require changes in the Move code.

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: I told you multiple times that this needs to be the same as the second's round session ID (This has been changed by you in your new version), and making the first & second dwallet dkg rounds use the same session ID is hard and require changes in the Move code.

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: It doesn’t require any changes at all, thats the way it was up until now too

ycscaly: It was just saved in the public output of the first round and taken from there. You can do it again.

.gitignore
```diff
@@ -88,4 +88,6 @@ ika_config.json
 .cursor
 
 # TypeDoc
-sdk/typescript/docs
\ No newline at end of file
+sdk/typescript/docs
+move_pkg_lock_root.lock
```
ItayLevyOfficial, line 92: node cache files that are being auto generated when using the new TS sdk on ubuntu

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: Here is where you should use the From<> impl

crates/ika-core/src/dwallet_mpc/crytographic_computation/native_computations/make_dwallet_user_secret_key_shares_public.rs
```diff
@@ -22,9 +22,9 @@ pub fn verify_secret_share(
         VersionedDwalletDKGSecondRoundPublicOutput::V1(output) => {
             bcs::from_bytes::<SpecificDKGDecentralizedPartyOutput>(output.as_slice())?.into()
         }
-        VersionedDwalletDKGSecondRoundPublicOutput::V2(output) => {
-            bcs::from_bytes::<SpecificDKGDecentralizedPartyVersionedOutput>(output.as_slice())?
-        }
+        VersionedDwalletDKGSecondRoundPublicOutput::V2(output) => bcs::from_bytes::<
+            SpecificDKGDecentralizedPartyVersionedOutputSecp256k1,
```
ycscaly:
```suggestion
           DKGDecentralizedPartyVersionedOutputSecp256k1,
```

sdk/ika-wasm/src/lib.rs
```diff
@@ -3,26 +3,38 @@
...
     decentralized_first_round_public_output: Vec<u8>,
+) -> Result<JsValue, JsError> {
+    let dkg_centralized_result =
+        &create_dkg_output_v1(protocol_pp, decentralized_first_round_public_output)
+            .map_err(|e| JsError::new(&e.to_string()))?;
+    serde_wasm_bindgen::to_value(&(
+        dkg_centralized_result.public_key_share_and_proof.clone(),
+        dkg_centralized_result.public_output.clone(),
+        dkg_centralized_result.centralized_secret_output.clone(),
+    ))
```
ycscaly, line 27: why are we taking these fields seperately?

sdk/ika-wasm/src/lib.rs
```diff
@@ -3,26 +3,38 @@
...
+    .map_err(|e| JsError::new(&e.to_string()))
+}
+
+#[wasm_bindgen]
+pub fn create_dkg_centralized_output_v2(
```
ycscaly, line 32: also logic here is different as there are different fields

sdk/ika-wasm/src/lib.rs
ItayLevyOfficial: This is how it has been done until now, you see those changes as new because of the indentation. returning a list of multiple values is the only way to return multiple values from a Wasm function, and those values need to be passed separately to the Move code

sdk/ika-wasm/src/lib.rs
ItayLevyOfficial: ok..? What do you want me to do about it?

sdk/ika-wasm/src/lib.rs
ycscaly: yes, but why was it done like this until now? why was the output parts returned, and not just the output itself as bytes? this needs to change, espsecially now when we have the new dkg output version

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -75,6 +81,103 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
...
+        group::secp256k1::GroupElement,
+    >(
+        Default::default(),
+        Default::default(),
+        Default::default(),
+        Default::default(),
+        public_parameters
+            .encryption_scheme_public_parameters
+            .clone(),
+    );
```
ycscaly:
```suggestion
```

ycscaly: no patch needed for v2, you will simply get the correct protocol public parameters in this function already, as they are global for an encryption key, taken from the reconfiguration public output

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -75,6 +81,103 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
...
+            ..
+        } => dkg_output,
+        DKGCentralizedPartyVersionedOutput::<
+            { group::secp256k1::SCALAR_LIMBS },
+            group::secp256k1::GroupElement,
+        >::TargetedPublicDKGOutput(output) => output,
+    };
+
+    // Public Output:
+    // centralized_public_key_share + public_key + decentralized_party_public_key_share
```
ycscaly:
```suggestion
```

ycscaly: not true anymore

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -75,6 +81,103 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
...
+    )
+    .context("advance() failed on the DKGCentralizedParty")?;
+
+    // Centralized Public Key Share and Proof.
+    let public_key_share_and_proof =
+        VersionedPublicKeyShareAndProof::V1(bcs::to_bytes(&round_result.outgoing_message)?);
+
+    let public_key_share_and_proof = bcs::to_bytes(&public_key_share_and_proof)?;
+    // TODO(#1470): Use one network round in the dWallet DKG flow.
+    // This is a temporary hack to keep working with the existing 2-round dWallet DKG mechanism.
```
ycscaly:
```suggestion
```

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -75,6 +81,103 @@ pub fn network_dkg_public_output_to_protocol_pp_inner(
...
+            group::secp256k1::GroupElement,
+        >::UniversalPublicDKGOutput {
+            output: dkg_output,
+            ..
+        } => dkg_output,
+        DKGCentralizedPartyVersionedOutput::<
+            { group::secp256k1::SCALAR_LIMBS },
+            group::secp256k1::GroupElement,
+        >::TargetedPublicDKGOutput(output) => output,
+    };
```
ycscaly:
```suggestion
    let centralized_output = round_result.public_output;;
```

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: good it was fixed

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: you are right, you only take the public key, but why? it shouldn't be like that.

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -169,33 +327,29 @@ pub fn centralized_and_decentralized_parties_dkg_output_match_inner(
 ) -> anyhow::Result<bool> {
     let versioned_centralized_dkg_output =
         bcs::from_bytes::<VersionedCentralizedDKGPublicOutput>(centralized_dkg_output)?;
```
ycscaly, line 296: where does `centralized_dkg_output` come from? you only output the public key above?

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: Add todo with link to issue

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -355,12 +525,19 @@ fn protocol_public_parameters_by_key_scheme(
...
                         secp256k1::GroupElement,
-                    >(encryption_scheme_public_parameters))
+                    >(
+                        Default::default(),
+                        Default::default(),
+                        Default::default(),
+                        Default::default(),
+                        encryption_scheme_public_parameters.clone(),
+                    );
+                    Ok(protocol_public_parameters)
```
ycscaly, line 529: where is the `v2` case? todo + issue

crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -26,6 +27,20 @@ pub enum NetworkDecryptionKeyPublicOutputType {
     Reconfiguration,
 }
 
+pub type SpecificDKGDecentralizedPartyOutput = DKGDecentralizedPartyOutput<
```
ycscaly:
```suggestion
pub type DKGDecentralizedPartyOutputSecp256k1 = DKGDecentralizedPartyOutput<
```

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -372,8 +384,16 @@ impl ProtocolCryptographicData {
...
+                first_round_output,
                 ..
             } => {
+                let session_id = match bcs::from_bytes(&first_round_output)? {
+                    VersionedDwalletDKGFirstRoundPublicOutput::V1(output) => {
+                        let (_, session_id) =
+                            bcs::from_bytes::<(Vec<u8>, CommitmentSizedNumber)>(&output)?;
+                        session_id
+                    }
+                };
```
ycscaly, line 396: isn't this code duplicated? also where is the patch for the protocol public parameters here? was it done already?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -389,12 +409,31 @@ impl ProtocolCryptographicData {
...
+                            },
+                            group::secp256k1::GroupElement,
+                        >::TargetedPublicDKGOutput(output) => output,
+                    };
                     verify_encrypted_share(
                         &data.encrypted_centralized_secret_share_and_proof,
                         &bcs::to_bytes(&VersionedDwalletDKGSecondRoundPublicOutput::V1(
-                            public_output_value.clone(),
+                            bcs::to_bytes(&decentralized_output)?,
                         ))?,
```
ycscaly, line 441: This should only be in the `v1` patchy case, add a todo here to check protocol config before

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -389,12 +409,31 @@ impl ProtocolCryptographicData {
...
+                        } => dkg_output,
+                        DKGDecentralizedPartyVersionedOutput::<
+                            { group::secp256k1::SCALAR_LIMBS },
+                            { twopc_mpc::secp256k1::class_groups::FUNDAMENTAL_DISCRIMINANT_LIMBS },
+                            {
+                                twopc_mpc::secp256k1::class_groups::NON_FUNDAMENTAL_DISCRIMINANT_LIMBS
+                            },
+                            group::secp256k1::GroupElement,
+                        >::TargetedPublicDKGOutput(output) => output,
+                    };
```
ycscaly, line 434: This should only be in the `v1` patchy case, add a todo here to check protocol config before

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -410,10 +449,19 @@ impl ProtocolCryptographicData {
...
+                                output, ..
+                            } => output,
+                            DKGDecentralizedPartyVersionedOutputSecp256k1::TargetedPublicDKGOutput (
+                                output
+                            ) => output,
+                        };
+                        let public_output_value =
+                            bcs::to_bytes(&VersionedDwalletDKGSecondRoundPublicOutput::V1(
+                                bcs::to_bytes(&decentralized_output).unwrap(),
+                            ))?;
```
ycscaly, line 469: This should only be in the `v1` patchy case, add a todo here to check protocol config before

ycscaly: link issue

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/dwallet_dkg.rs
```diff
@@ -74,28 +77,61 @@ impl DWalletDKGFirstPartyPublicInputGenerator for DWalletDKGFirstParty {
...
     ) -> DwalletMPCResult<<DWalletDKGSecondParty as mpc::Party>::PublicInput> {
-        let first_round_output_buf: VersionedCentralizedDKGPublicOutput =
+        let first_round_output_buf: VersionedDwalletDKGFirstRoundPublicOutput =
             bcs::from_bytes(first_round_output_buf).map_err(DwalletMPCError::BcsError)?;
 
         let centralized_party_public_key_share: VersionedPublicKeyShareAndProof =
             bcs::from_bytes(centralized_party_public_key_share_buf)
                 .map_err(DwalletMPCError::BcsError)?;
 
         match first_round_output_buf {
```
ycscaly, line 111: add todo to add `v2` case

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
ycscaly, line 397: This should only be in the v1 patchy case, add a todo here to check protocol config before

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ycscaly: you didn't do this, and you resolved the comment. Why?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -132,13 +133,19 @@ pub(crate) fn instantiate_dwallet_mpc_network_encryption_key_public_data_from_re
...
             let protocol_public_parameters = ProtocolPublicParameters::new::<
                 { secp256k1::SCALAR_LIMBS },
                 { FUNDAMENTAL_DISCRIMINANT_LIMBS },
                 { NON_FUNDAMENTAL_DISCRIMINANT_LIMBS },
                 secp256k1::GroupElement,
             >(
+                Default::default(),
+                Default::default(),
+                Default::default(),
+                Default::default(),
```
ycscaly: use neutral here too, but why is this code duplicated anyways, can you explain me the two different flows?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: what do you mean?

ycscaly: I now export a `verify()` function in the `ecdsa::VerifyingKey` trait, use that one

sdk/ika-wasm/src/lib.rs
ycscaly: sorry, this is my mistake, I was confused, its fine

sdk/ika-wasm/src/lib.rs
ycscaly: you did it with the `v2` function

sdk/typescript/src/client/wasm-loader.ts
```diff
@@ -65,17 +65,12 @@ export async function generate_secp_cg_keypair_from_seed(
 	return wasm.generate_secp_cg_keypair_from_seed(seed);
 }
 
-export async function create_dkg_centralized_output(
+export async function create_dkg_centralized_output_v1(
```
ycscaly, line 69: add a `v2` function too, and we should call it from the `JS` accordingly, so add a TODO + issue link here

ycscaly (changes requested):
see comments

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: Not sure I understand your question. The centralized dkg output is being returned from the `create_dkg_centralized_output_v1` function, when creating the dwallet.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
ItayLevyOfficial: What is neutral? Can you please tell me exactly what do you want me to use?

About the duplication - we have two separate flows to create the network key data - one from the first network DKG output, and one from a reconfiguration output.

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -110,30 +110,11 @@ pub fn create_dkg_output_v2(
...
     let session_identifier = CommitmentSizedNumber::from_le_slice(&session_id);
     let round_result = DKGCentralizedParty::advance(
         (),
         &(),
-        &(
-            protocol_pp_with_decentralized_dkg_output,
-            session_identifier,
-        )
-            .into(),
+        &(public_parameters, session_identifier).into(),
```
ycscaly:
```suggestion
        &(protocol_public_parameters, session_identifier).into(),
```

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -110,30 +110,11 @@ pub fn create_dkg_output_v2(
     session_id: Vec<u8>,
 ) -> anyhow::Result<CentralizedDKGWasmResult> {
     let public_parameters: ProtocolPublicParameters = bcs::from_bytes(&protocol_pp)?;
```
ycscaly:
```suggestion
    let protocol_public_parameters: ProtocolPublicParameters = bcs::from_bytes(&protocol_pp)?;
```

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: the entire purpose of this wasm function is to allow users to get the public key of their dwallet

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ItayLevyOfficial: You only added it in your latest version, I have added this task to the version bump issue & I'll implement it there.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ItayLevyOfficial: I think I did & then undid it. Anyway, is it "something you think is safer" or a must? Maybe change the default implementation to use this?

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: I see

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ycscaly: just do it please

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: In `V2` you should have the `VersionedOutput`

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
ycscaly: it is neutral because we just want a valid group element there. I told you exactly what to do, take it from the public parameters

ycscaly (approved):

PR #1477, opened 2025-09-01
chore: Add V1 to V2 reconfiguration flow

ItayLevyOfficial:
## Description 

Add chain support for V1 to V2 MPC flow & performs it automatically on old keys upon a protocol config version upgrade

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:


crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -589,6 +623,59 @@ impl ProtocolCryptographicData {
...
+                            .network_encryption_key_version
+                        {
+                            Some(2) => {
+                                bcs::to_bytes(&VersionedDecryptionKeyReconfigurationOutput::V2(
+                                    public_output_value,
+                                ))?
+                            }
+                            _ => bcs::to_bytes(&VersionedDecryptionKeyReconfigurationOutput::V1(
+                                public_output_value,
+                            ))?,
```
ycscaly: I'm confused, you added a `protocol_config` based match on the public output, but not on the advance. Do you plan to add this in a future PR? if so, issue + todo

crates/ika-core/src/request_protocol_data.rs
```diff
@@ -67,6 +67,10 @@ pub struct NetworkEncryptionKeyDkgData {
 #[display("Network Encryption Key Reconfiguration")]
 pub struct NetworkEncryptionKeyReconfigurationData {}
 
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, derive_more::Display)]
+#[display("Network Encryption Key V1 to V2 Reconfiguration")]
+pub struct NetworkEncryptionKeyV1ToV2ReconfigurationData {}
```
ycscaly, line 72: Please add an issue to remove v1-to-v2 code, as it is just a temporary transitional code, and link it with a TODO everywhere a v1-to-v2 specific struct/code exists

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
```diff
@@ -127,8 +133,30 @@ pub(crate) fn session_input_from_request(
             let next_active_committee = next_active_committee.ok_or(
                 DwalletMPCError::MissingNextActiveCommittee(session_id.to_be_bytes().to_vec()),
             )?;
-
-            Ok((
+            let last_reconfiguration_version = network_keys
+                .get_latest_reconfiguration_version(dwallet_network_encryption_key_id)?;
+            if last_reconfiguration_version == Some(1)
+                && protocol_config.network_encryption_key_version == Some(2)
```
ycscaly: This is a good check, but you are missing the case of the v2 reconfiguration flow - please add a TODO here to add it + link issue.

Also, I think its better to do it within a `match protocol_config.network_encryption_key_version`, then you'll have two supported cases, `Some(1)`, `Some(2)`, and in `Some(2)` you can have another `match` on the `last_reconfiguration_version`. It's cleaner this way

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
```diff
@@ -127,8 +133,30 @@ pub(crate) fn session_input_from_request(
             let next_active_committee = next_active_committee.ok_or(
                 DwalletMPCError::MissingNextActiveCommittee(session_id.to_be_bytes().to_vec()),
             )?;
-
-            Ok((
+            let last_reconfiguration_version = network_keys
+                .get_latest_reconfiguration_version(dwallet_network_encryption_key_id)?;
```
ycscaly: So this is a bit weird, you are returning an `Option<usize>` from `get_latest_reconfiguration_version()` instead of `usize`, because maybe we don't have the network key yet, but you are completely ignoring the `None` case here -  which should mean we are not ready for reconfiguration yet, as we don't have the network key.

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -166,6 +167,7 @@ impl DWalletMPCService {
                 0,
                 DWalletMPCMetrics::new(&Registry::new()),
                 sui_data_receivers.clone(),
+                ProtocolConfig::get_for_min_version(),
```
ycscaly, line 170: what is this?

crates/ika-core/src/dwallet_mpc/crytographic_computation/orchestrator.rs
```diff
@@ -256,11 +262,16 @@ impl CryptographicComputationsOrchestrator {
 
         let computation_channel_sender = self.completed_computation_sender.clone();
         let root_seed = self.root_seed.clone();
+        let protocol_config = self.protocol_config.clone();
```
ycscaly, line 265:
```suggestion
        let protocol_config = self.protocol_config.clone();
        
```

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -37,6 +38,59 @@ pub(crate) trait ReconfigurationPartyPublicInputGenerator: Party {
...
+        reconfiguration_public_output: VersionedDecryptionKeyReconfigurationOutput,
+    ) -> DwalletMPCResult<<ReconfigurationV1toV2Secp256k1Party as mpc::Party>::PublicInput>;
+}
+
+impl ReconfigurationV1ToV2PartyPublicInputGenerator for ReconfigurationV1toV2Secp256k1Party {
+    fn generate_public_input(
+        current_committee: &Committee,
+        upcoming_committee: Committee,
+        network_dkg_public_output: VersionedNetworkDkgOutput,
+        reconfiguration_public_output: VersionedDecryptionKeyReconfigurationOutput,
```
ycscaly: not required anymore, I support the case of doing v1-to-v2 just after a v1 dkg (no v1 reconfiguration yet). Please change to support both cases.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -37,6 +38,59 @@ pub(crate) trait ReconfigurationPartyPublicInputGenerator: Party {
...
+        let public_input: <ReconfigurationV1toV2Secp256k1Party as Party>::PublicInput =
+            <ReconfigurationV1toV2Secp256k1Party as Party>::PublicInput::new(
+                &current_access_structure,
+                upcoming_access_structure,
+                current_encryption_keys_per_crt_prime_and_proofs.clone(),
+                upcoming_encryption_keys_per_crt_prime_and_proofs.clone(),
+                current_tangible_party_id_to_upcoming(current_committee, upcoming_committee)
+                    .clone(),
+                bcs::from_bytes(&network_dkg_public_output)?,
+                bcs::from_bytes(&reconfiguration_public_output)?,
```
ycscaly: did you test this? its the wrong input. We should pass the decryption key share public parameters + network dkg v1 output.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -271,6 +291,18 @@ impl DwalletMPCNetworkKeys {
             .network_dkg_output
             .clone())
     }
+
+    pub fn get_last_reconfiguration_output(
```
ycscaly, line 295: do we have a guarantee that this belongs to the current epoch (meaning, it was the reconfiguration that completed in the previous epoch?)

ycscaly (changes requested):

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
ItayLevyOfficial: dummy protocol config for tests

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
ItayLevyOfficial: https://github.com/dwallet-labs/dwallet-network/blob/75a1483f14849ba0e31cb455495915f181fb5873/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs#L688

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ItayLevyOfficial: https://github.com/dwallet-labs/dwallet-network/blob/75b214657845b7a5d97d0a1d013d2797e6aeacaa/crates/ika-core/src/dwallet_mpc/mpc_manager.rs#L513

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -234,7 +236,7 @@ impl ProtocolCryptographicData {
                 data,
                 dwallet_network_encryption_key_id,
             } => match public_input {
-                PublicInput::NetworkEncryptionKeyReconfiguration(public_input) => {
+                PublicInput::NetworkEncryptionKeyV1Reconfiguration(public_input) => {
```
ycscaly:
```suggestion
                PublicInput::NetworkEncryptionKeyReconfigurationV1(public_input) => {
```

ycscaly: be consistent

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -230,22 +230,22 @@ impl DwalletMPCNetworkKeys {
...
-        &self,
-        key_id: &ObjectID,
-    ) -> DwalletMPCResult<Option<usize>> {
-        let versioned_output = self
+    pub fn get_network_key_version(&self, key_id: &ObjectID) -> DwalletMPCResult<usize> {
+        let key_data = self
             .network_encryption_keys
             .get(key_id)
-            .ok_or(DwalletMPCError::WaitingForNetworkKey(*key_id))?
+            .ok_or(DwalletMPCError::WaitingForNetworkKey(*key_id))?;
```
ycscaly, line 237: but will the event be ignored if we reach this? or will we wait for the network key and retry?

crates/ika-core/src/dwallet_session_request.rs
```diff
@@ -233,6 +233,14 @@ impl From<&ProtocolCryptographicData> for DWalletSessionRequestMetricData {
                 hash_scheme: None,
                 signature_algorithm: None,
             },
+            ProtocolCryptographicData::NetworkEncryptionKeyV2Reconfiguration { data, .. } => {
+                DWalletSessionRequestMetricData {
+                    name: data.to_string(),
+                    curve: None,
```
ycscaly, line 240: create an issue to set these fields and link todo here

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ItayLevyOfficial: because of [this check](https://github.com/dwallet-labs/dwallet-network/blob/ef10dc4f44d272d5e035463839377b03b0879bfd/crates/ika-core/src/dwallet_mpc/mpc_session.rs#L409), this should never happen

ycscaly: In that case, this is the wrong error code. You should use one that signifies there is a bug, and also log it as `should_never_happen`

PR #1486, opened 2025-09-02
chore: Add support for more curves when parsing V2 network key data

ItayLevyOfficial:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:


crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -57,11 +57,18 @@ pub struct NetworkEncryptionKeyPublicData {
...
 
     /// The public output of the `NetworkDKG` process (the first and only one).
     /// On first instance it will be equal to `latest_public_output`.
     pub network_dkg_output: VersionedNetworkDkgOutput,
+    pub secp256r1_protocol_public_parameters:
+        Option<twopc_mpc::secp256r1::class_groups::ProtocolPublicParameters>,
+    pub ristretto_protocol_public_parameters:
+        Option<twopc_mpc::ristretto::class_groups::ProtocolPublicParameters>,
+    pub curve25519_protocol_public_parameters:
+        Option<twopc_mpc::curve25519::class_groups::ProtocolPublicParameters>,
```
ycscaly: I don't like this, maybe we should have a `NetworkEncryptionKeyPublicDataV2` struct that holds these as non-option.

crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -57,11 +57,18 @@ pub struct NetworkEncryptionKeyPublicData {
     pub decryption_key_share_public_parameters:
```
ycscaly: Also add for each curve

crates/dwallet-mpc-types/src/dwallet_mpc.rs
ItayLevyOfficial: So should I create it or no?

ycscaly: yes if that makes sense

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -532,6 +532,11 @@ fn protocol_public_parameters_by_key_scheme(
                     // function generic or have an enum over `ProtocolPublicParameters`.
                     todo!()
                 }
+                DWalletMPCNetworkKeyScheme::Secp256r1 => {
+                    // To add support here, we need to either make this
+                    // function generic or have an enum over `ProtocolPublicParameters`.
+                    todo!()
+                }
```
ycscaly, line 539: Remove both this and the `Ristretto` cases from here, they will never be supported in `V1`. We will support them in the `V2` network dkg

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -602,10 +603,24 @@ impl ProtocolCryptographicData {
                         malicious_parties,
                         private_output,
                     } => {
-                        let signature: ECDSASecp256k1Signature =
-                            bcs::from_bytes(&public_output_value)?;
+                        let public_output_value = match data.curve {
+                            DWalletMPCNetworkKeyScheme::Secp256k1 => {
+                                let signature: ECDSASecp256k1Signature =
```
ycscaly, line 610: Either add support for TaprootSignature if its schnorr, or link it to an issue to add support for schnorr

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -602,10 +603,24 @@ impl ProtocolCryptographicData {
...
+                                signature.to_bytes().to_vec()
+                            }
+                            DWalletMPCNetworkKeyScheme::Secp256r1 => {
+                                let signature: ECDSASecp256r1Signature =
+                                    bcs::from_bytes(&public_output_value)?;
+                                signature.to_bytes().to_vec()
+                            }
+                            _ => {
+                                return Err(DwalletMPCError::InvalidDWalletProtocolType);
+                            }
```
ycscaly, line 621: Either add support for EdDSASignature if its schnorr for curve25519, or link it to an issue to add support for schnorr.

Same for SchnorrkelSubstrateSignature for schnorr Ristretto

ycscaly: I think itll be easy for you to just add them here

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: Actually, in the current design there is no meaning for this `DWalletMPCNetworkKeyScheme` enum anymore.
In `V1`, there is only secp256k1, and in `V2`, its the same network key for all curves.

The curve comes from the dwallet exclusively in our current design.

So maybe the best thing to do, is to remove this enum entirely.

crates/dwallet-mpc-types/src/dwallet_mpc.rs
ycscaly: Is this going to be done in the current PR?

ycscaly (changes requested):

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: will be done in a separate pr

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
ItayLevyOfficial: this is problematic right now because of the false network key scheme enum, it will be simpler to do that after removing it

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: just open an issue for it

crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -41,6 +41,24 @@ pub type DKGDecentralizedPartyVersionedOutputSecp256k1 = DKGDecentralizedPartyVe
...
+
+    pub ristretto_decryption_key_share_public_parameters:
+        class_groups::RistrettoDecryptionKeySharePublicParameters,
+
+    pub curve25519_decryption_key_share_public_parameters:
+        class_groups::Curve25519DecryptionKeySharePublicParameters,
+    pub ristretto_protocol_public_parameters:
+        twopc_mpc::ristretto::class_groups::ProtocolPublicParameters,
+    pub curve25519_protocol_public_parameters:
+        twopc_mpc::curve25519::class_groups::ProtocolPublicParameters,
```
ycscaly: 1. consistent ordering (of curves)
2. where is k1?

crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -54,14 +72,14 @@ pub struct NetworkEncryptionKeyPublicData {
...
         class_groups::Secp256k1DecryptionKeySharePublicParameters,
-
-    pub protocol_public_parameters: twopc_mpc::secp256k1::class_groups::ProtocolPublicParameters,
-
     /// The public output of the `NetworkDKG` process (the first and only one).
     /// On first instance it will be equal to `latest_public_output`.
     pub network_dkg_output: VersionedNetworkDkgOutput,
+    pub secp256k1_protocol_public_parameters:
+        twopc_mpc::secp256k1::class_groups::ProtocolPublicParameters,
+    pub v2_data: Option<V2NetworkKeyData>,
```
ycscaly: No, the whole idea is for this not to be optional at all, meaning, we should have either two different structs or an enum with v1 and v2, and the fields are non-optional; once we upgrade the network I want it to be v2 and we can remove the old version. But for the v1 to v2 epoch, it needs to stay v1.

You can do an enum with accessor functions if its easier, but I think the best thing is to have two different types so we can remove the old version later, as this struct isn't sent anywhere its just local

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: https://github.com/dwallet-labs/ika/issues/1493

ycscaly (approved):

PR #1499, opened 2025-09-04
feat: add metrics proxy and some missing infra

zeevmoney:


ItayLevyOfficial:
We already have a CI for ika-node docker image build and push, [here](https://github.com/dwallet-labs/dwallet-network/blob/9c52a1f3d90424340e084ba5aaa7fd45065243d4/.github/workflows/docker-ci.yaml#L1)

ItayLevyOfficial:
> **In order to enable the github actions in this PR you must push them to `main` branch, since all of them are `workflow_dispatch` which will be visible only in `main`.**

This is not true. You can make the workflow run on push, run it once, then remove the push trigger, and then you'll be able to manually trigger the workflow dispatch workflow without merging it to main.

PR #1513, opened 2025-09-08
chore: Add getters for network encryption key by curve

yaelAbergel:


crates/dwallet-mpc-types/src/dwallet_mpc.rs
yaelAbergel, line 1: Moved implementations to the end of the file

crates/dwallet-mpc-types/src/dwallet_mpc.rs
```diff
@@ -42,161 +42,41 @@ pub type DKGDecentralizedPartyVersionedOutputSecp256k1 = DKGDecentralizedPartyVe
...
-    fn network_dkg_output(&self) -> &VersionedNetworkDkgOutput {
-        &self.network_dkg_output
-    }
-
-    fn state(&self) -> &NetworkDecryptionKeyPublicOutputType {
-        &self.state
-    }
-
-    fn latest_network_reconfiguration_public_output(
+    ) -> Option<twopc_mpc::secp256r1::class_groups::ProtocolPublicParameters>;
```
ycscaly: Return `Result<>` not option, this should be a bug if `None`, so we should return an error instead.

PR #1515, opened 2025-09-09
fix: Stop crashing on first V2 reconfiguration

ItayLevyOfficial:
## Description 

Previously, the first reconfiguration of a V2 network key would crash because the public input generation function expected a reconfiguration output that didn’t exist. This PR fixes the issue by handling that case correctly.

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
+++ b/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
@@ -106,15 +106,17 @@ impl ReconfigurationV2PartyPublicInputGenerator for ReconfigurationV2Secp256k1Pa
                                 upcoming_encryption_keys_per_crt_prime_and_proofs.clone(),
                                 current_tangible_party_id_to_upcoming(current_committee, upcoming_committee)
                                     .clone(),
-                                bcs::from_bytes::<Secp256k1Party::PublicOutput>(&network_dkg_public_output)?.into(),
+                                bcs::from_bytes(&network_dkg_public_output)?,
                                 bcs::from_bytes(&latest_reconfiguration_public_output)?,
                             )
                                 .map_err(DwalletMPCError::from)?;
-                        Ok(public_input);
+                        Ok(public_input)
                     }
                 }
             }
             VersionedNetworkDkgOutput::V2(network_dkg_public_output) => {
+                let public_output: <twopc_mpc::decentralized_party::dkg::Party as mpc::Party>::PublicOutput =
+                    bcs::from_bytes(&network_dkg_public_output)?;
                 let public_input: <ReconfigurationV2Secp256k1Party as Party>::PublicInput =
                     <twopc_mpc::decentralized_party::reconfiguration::Party as Party>::PublicInput::new_from_dkg_output(
                         &current_access_structure,
@@ -123,7 +125,7 @@ impl ReconfigurationV2PartyPublicInputGenerator for ReconfigurationV2Secp256k1Pa
                         upcoming_encryption_keys_per_crt_prime_and_proofs.clone(),
                         current_tangible_party_id_to_upcoming(current_committee, upcoming_committee)
                             .clone(),
-                        bcs::from_bytes(&network_dkg_public_output)?,
+                        public_output.into(),
                     )
                         .map_err(DwalletMPCError::from)?;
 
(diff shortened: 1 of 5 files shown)


crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -72,31 +72,51 @@ impl ReconfigurationV2PartyPublicInputGenerator for ReconfigurationV2Secp256k1Pa
...
+                                current_encryption_keys_per_crt_prime_and_proofs.clone(),
+                                upcoming_encryption_keys_per_crt_prime_and_proofs.clone(),
+                                current_tangible_party_id_to_upcoming(current_committee, upcoming_committee)
+                                    .clone(),
+                                bcs::from_bytes(&network_dkg_public_output)?,
+                            )
+                                .map_err(DwalletMPCError::from)?;
+
+                        Ok(public_input)
+                    }
```
ycscaly: No, this case is not supported.

We have the reconfiguration v1 to v2 protocol because there is already a network key that has v1 reconfiguration output, but no network key has v1 dkg output only.

And if a new key will be generated after the update, it will be with the v2 output.

So this options does not exist, and an error should be returned, and the session rejected.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -72,31 +72,51 @@ impl ReconfigurationV2PartyPublicInputGenerator for ReconfigurationV2Secp256k1Pa
...
+                    Some(latest_reconfiguration_public_output) => {
+                        let VersionedDecryptionKeyReconfigurationOutput::V2(
+                            latest_reconfiguration_public_output,
+                        ) = latest_reconfiguration_public_output
+                        else {
+                            return Err(DwalletMPCError::InternalError(
+                                "V2 Reconfiguration only supports latest reconfiguration public output of version V2"
```
ycscaly:
```suggestion
                                "The Reconfiguration v2 protocol can only be executed after a v1-to-v2 protocol, or after another reconfiguration v2 protocol."
```

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -72,31 +72,51 @@ impl ReconfigurationV2PartyPublicInputGenerator for ReconfigurationV2Secp256k1Pa
...
+                            )
+                                .map_err(DwalletMPCError::from)?;
+                        Ok(public_input)
+                    }
+                }
             }
             VersionedNetworkDkgOutput::V2(network_dkg_public_output) => {
+                let public_output: <twopc_mpc::decentralized_party::dkg::Party as mpc::Party>::PublicOutput =
+                    bcs::from_bytes(&network_dkg_public_output)?;
                 let public_input: <ReconfigurationV2Secp256k1Party as Party>::PublicInput =
```
ycscaly: what do you mean by `ReconfigurationV2Secp256k1Party`? there is only one party, its not curve specific. It should just be `ReconfigurationV2Party`. Same for v1 to v2 and dkg v2

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -105,7 +125,7 @@ impl ReconfigurationV2PartyPublicInputGenerator for ReconfigurationV2Secp256k1Pa
                         upcoming_encryption_keys_per_crt_prime_and_proofs.clone(),
                         current_tangible_party_id_to_upcoming(current_committee, upcoming_committee)
                             .clone(),
-                        bcs::from_bytes(&network_dkg_public_output)?,
+                        public_output.into(),
```
ycscaly:
```suggestion
                        public_output,
```

ycscaly: No `.into()` needed. Maybe you are in an old version, use the latest from `main`.

ycscaly (changes requested):

ycscaly (approved):

PR #1524, opened 2025-09-14
1518 full version bump

ItayLevyOfficial:
## Description 

## Test plan 

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/ika-protocol-config/src/lib.rs
+++ b/crates/ika-protocol-config/src/lib.rs
@@ -16,7 +16,7 @@ use tracing::{info, warn};
 
 /// The minimum and maximum protocol versions supported by this build.
 const MIN_PROTOCOL_VERSION: u64 = 1;
-const MAX_PROTOCOL_VERSION: u64 = 1;
+const MAX_PROTOCOL_VERSION: u64 = 2;
 
 // Record history of protocol version allocations here:
 //
@@ -487,6 +487,9 @@ impl ProtocolConfig {
                 //
                 //     // Remove a constant (ensure that it is never accessed during this version).
                 //     existing_constant: None,
+                2 => {
+                    cfg.network_encryption_key_version = Some(2);
+                }
                 _ => panic!("unsupported version {version:?}"),
             }
         }
--- a/sdk/typescript/src/client/ika-client.ts
+++ b/sdk/typescript/src/client/ika-client.ts
@@ -678,7 +678,7 @@ export class IkaClient {
 		}
 
 		const protocolPublicParameters = await networkDkgPublicOutputToProtocolPublicParameters(
-			await this.#readTableVecAsRawBytes(networkEncryptionKeyPublicOutputID),
+			await this.readTableVecAsRawBytes(networkEncryptionKeyPublicOutputID),
 		);
 
 		// Cache the parameters by encryption key ID
@@ -939,7 +939,7 @@ export class IkaClient {
 	 * @throws {NetworkError} If network requests fail
 	 * @private
 	 */
-	async #readTableVecAsRawBytes(tableID: string): Promise<Uint8Array> {
+	async readTableVecAsRawBytes(tableID: string): Promise<Uint8Array> {
 		try {
 			let cursor: string | null = null;
 			const allTableRows: { objectId: string }[] = [];
--- a/sdk/typescript/test/helpers/test-utils.ts
+++ b/sdk/typescript/test/helpers/test-utils.ts
@@ -368,7 +368,7 @@ export function delay(seconds: number): Promise<void> {
 	return new Promise((resolve) => setTimeout(resolve, seconds * 1000));
 }
 
-export async function runSignFullFlow(
+export async function runSignFullFlowWithV1Dwallet(
 	ikaClient: IkaClient,
 	suiClient: SuiClient,
 	testName: string,
--- a/sdk/typescript/test/helpers/test-utils.ts
+++ b/sdk/typescript/test/helpers/test-utils.ts
@@ -23,7 +23,7 @@ import {
 	ZeroTrustDWallet,
 } from '../../src/client/types.js';
 import { UserShareEncryptionKeys } from '../../src/client/user-share-encryption-keys.js';
-import { createCompleteDWallet, testPresign, testSign } from './dwallet-test-helpers';
+import { createCompleteDWallet, createCompleteDWalletV2, testPresign, testSign } from './dwallet-test-helpers';
 
 // Store random seeds per test to ensure deterministic behavior within each test
 const testSeeds = new Map<string, Uint8Array>();
@@ -379,6 +379,59 @@ export async function runSignFullFlowWithV1Dwallet(
 		userShareEncryptionKeys,
 		signerAddress,
 	} = await createCompleteDWallet(ikaClient, suiClient, testName);
+
+	// Step 2: Create presign
+	const presignRequestEvent = await testPresign(
+		ikaClient,
+		suiClient,
+		activeDWallet,
+		SignatureAlgorithm.ECDSA,
+		signerAddress,
+		testName,
+	);
+
+	expect(presignRequestEvent).toBeDefined();
+	expect(presignRequestEvent.event_data.presign_id).toBeDefined();
+
+	// Step 3: Wait for presign to complete
+	const presignObject = await retryUntil(
+		() =>
+			ikaClient.getPresignInParticularState(presignRequestEvent.event_data.presign_id, 'Completed'),
+		(presign) => presign !== null,
+		30,
+		2000,
+	);
+
+	expect(presignObject).toBeDefined();
+	expect(presignObject.state.$kind).toBe('Completed');
+
+	// Step 4: Sign a message
+	const message = createTestMessage(testName);
+	await testSign(
+		ikaClient,
+		suiClient,
+		activeDWallet as ZeroTrustDWallet,
+		userShareEncryptionKeys,
+		presignObject,
+		encryptedUserSecretKeyShare,
+		message,
+		Hash.KECCAK256,
+		SignatureAlgorithm.ECDSA,
+		testName,
+	);
+}
+
+export async function runSignFullFlowWithV2Dwallet(
+	ikaClient: IkaClient,
+	suiClient: SuiClient,
+	testName: string,
+) {
+	const {
+		dWallet: activeDWallet,
+		encryptedUserSecretKeyShare,
+		userShareEncryptionKeys,
+		signerAddress,
+	} = await createCompleteDWalletV2(ikaClient, suiClient, testName);
 
 	// Step 2: Create presign
 	const presignRequestEvent = await testPresign(
--- a/sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
+++ b/sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
@@ -9,7 +9,6 @@ describe('system tests', () => {
 	it('run a full flow test of upgrading the network key version and the move code', async () => {
 		require('dotenv').config({ path: `${TEST_ROOT_DIR}/.env` });
 
-		const startCommitteeSize = Number(process.env.VALIDATOR_NUM);
 		// ------------ Create Ika Genesis ------------
 		const createIkaGenesisPath = `${TEST_ROOT_DIR}/create-ika-genesis-mac.sh`;
 		await execa({
@@ -23,7 +22,6 @@ describe('system tests', () => {
 		);
 
 		console.log(`Ika genesis created, deploying ika network`);
-		const addValidatorScriptPath = `${TEST_ROOT_DIR}/add-validators-to-next-committee.sh`;
 		console.log('Validators added to the next committee, deploying ika network');
 		await deployIkaNetwork();
 
--- a/sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
+++ b/sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
@@ -7,7 +7,7 @@ import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
 import {
 	createTestIkaClient,
-	createTestSuiClient, delay,
+	createTestSuiClient, delay, runSignFullFlowWithV1Dwallet,
 	waitForEpochSwitch,
 } from '../../helpers/test-utils';
 import { deployIkaNetwork, NAMESPACE_NAME, TEST_ROOT_DIR } from '../globals';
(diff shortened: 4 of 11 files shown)


crates/ika-protocol-config/src/lib.rs
ItayLevyOfficial, line 1: code needed to make a validator vote to upgrade the protocol version & create v2 network keys

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,104 @@
...
+		console.log('Ika network deployed, waiting for epoch switch');
+		const suiClient = createTestSuiClient();
+		const ikaClient = createTestIkaClient(suiClient);
+		await ikaClient.initialize();
+		await waitForEpochSwitch(ikaClient);
+		console.log('Epoch switched, verifying the network key version is V1');
+		const networkKey = await ikaClient.getConfiguredNetworkEncryptionKey();
+		let networkKeyBytes = await ikaClient.readTableVecAsRawBytes(networkKey.publicOutputID);
+		const networkKeyVersion = network_key_version(networkKeyBytes);
+		expect(networkKeyVersion).toBe(1);
```
ycscaly, line 71: Already here, with the old binaries, create a dWallet

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,104 @@
...
+		// Restart each validator pod one by one to pick up the docker tag change
+		for (let i = 0; i < Number(process.env.VALIDATOR_NUM); i++) {
+			await killValidatorPod(kc, NAMESPACE_NAME, i + 1);
+			await delay(1);
+			await createValidatorPod(kc, NAMESPACE_NAME, i + 1);
+		}
+		console.log(
+			'All validators upgraded, waiting for two epoch switches (the protocol version may have been changed after the epoch middle)',
+		);
+		await waitForEpochSwitch(ikaClient);
```
ycscaly: after the upgrade happened, but befoe the reconfiguration happened - sign with that dwallet
also after the reconfiguration but before epoch switch sign with
and also after the epoch switched - sign with it

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,104 @@
...
+		await waitForEpochSwitch(ikaClient);
+		console.log('Two epochs switched, verifying the network key version is V2');
```
ycscaly: here too sign with that dwallet

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,104 @@
...
+			twopc_mpc_contracts_path,
+			ikaClient,
+			protocolCapID,
+		);
+		await delay(5); // wait for the upgrade to be fully processed
+		await migrateCoordinator(suiClient, signer, ikaClient, protocolCapID, upgradedPackageID);
+		await delay(5); // wait for the migration to be fully processed
+		console.log('Move contracts upgraded to V2, verifying v2 dWallet full flow works');
+		await runSignFullFlowWithV2Dwallet(ikaClient, suiClient, `v2-dwallet-sign-full-flow-test`);
+		console.log('V2 dWallet full flow works, test completed successfully');
```
ycscaly, line 170: also create here an imported key and sign with it

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
ItayLevyOfficial: Signing after a reconfiguration but before an epoch switch is tricky — I can detect the reconfig, but how can I prevent the network from switching epochs? This test needs to run with a short epoch duration to complete in a reasonable time, which makes it problematic.

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,136 @@
...
+		await ikaClient.initialize();
+		await waitForEpochSwitch(ikaClient);
+		console.log('Epoch switched, verifying the network key version is V1');
+		const networkKey = await ikaClient.getConfiguredNetworkEncryptionKey();
+		let networkKeyBytes = await ikaClient.readTableVecAsRawBytes(networkKey.publicOutputID);
+		const networkKeyVersion = network_key_version(networkKeyBytes);
+		expect(networkKeyVersion).toBe(1);
+		console.log('Network key version is V1, creating a dWallet with it');
+		const dwallet = await createCompleteDWallet(ikaClient, suiClient, 'create-complete-dwallet');
+		console.log('DWallet created successfully, upgrading the validators docker image');
```
ycscaly: lets also run a full-flow with this dWallet here just for completeness

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,136 @@
...
+			stderr: ['pipe', 'inherit'],
+			cwd: TEST_ROOT_DIR,
+		})`${createIkaGenesisPath}`;
+
+		await fs.copyFile(
+			`${TEST_ROOT_DIR}/${process.env.SUBDOMAIN}/publisher/ika_config.json`,
+			path.resolve(process.cwd(), '../../ika_config.json'),
+		);
+		console.log(`Ika genesis created, deploying ika network`);
+		await deployIkaNetwork();
```
ycscaly, line 61: which contracts are deployed here? where can I see that?

ycscaly: or is it deployed from the outside, and how? Better deploy it as a part of this test

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
ycscaly: I think it can be ok, just increase the epoch time a bit, it can be a few minutes and the whole test can run in 20 minutes that's fine

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,136 @@
...
+		);
+		await runSignFullFlowTestWithImportedDwallet(
+			`imported-dwallet-sign-full-flow-test`,
+			ikaClient,
+			suiClient,
+		);
+		console.log(
+			'Imported dWallet full flow works, creating a new v2 dWallet and verifying it works',
+		);
+		await runSignFullFlowWithV2Dwallet(ikaClient, suiClient, `v2-dwallet-sign-full-flow-test`);
```
ycscaly: also here sign with the old v1 dwallet, after the contact was upgraded

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
```diff
@@ -0,0 +1,136 @@
...
+		console.log('V2 dWallet full flow works, test completed successfully');
```
ycscaly: also we should do all the curves, all the signatures here. But not in this pr, open issue and add a todo, we will do this after the multiple curves/sig code is merged

sdk/typescript/test/helpers/test-utils.ts
```diff
@@ -421,6 +485,59 @@ export async function runSignFullFlow(
...
+) {
+	const {
+		dWallet: activeDWallet,
+		encryptedUserSecretKeyShare,
+		userShareEncryptionKeys,
+		signerAddress,
+	} = await createCompleteDWalletV2(ikaClient, suiClient, testName);
+
+	// Step 2: Create presign
+	const presignRequestEvent = await testPresign(
```
ycscaly, line 506: please assert that the dwallet version is 2

sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts
ItayLevyOfficial: The contracts are being deployed [here](https://github.com/dwallet-labs/dwallet-network/blob/b48a1fb71d2efe04627b7cc8bc4e0e68145435c7/sdk/typescript/test/system-tests/upgrade-network-key/upgrade-network-key.test.ts#L47), from within the `create-ika-genesis` script

.github/workflows/docker-ci.yaml
ItayLevyOfficial, line 1: fixes needed for the docker ci

crates/ika-node/src/lib.rs
```diff
@@ -1161,9 +1161,7 @@ impl IkaNode {
                     cur_epoch_store.get_chain_identifier().chain(),
                     self.config
                         .supported_protocol_versions
-                        .expect("Supported versions should be populated")
-                        // no need to send digests of versions less than the current version
-                        .truncate_below(config.version),
+                        .expect("Supported versions should be populated"),
```
ItayLevyOfficial: a bugfix - until now a validator always voted to the protocol config version it currently uses, this change make it vote for the last version it support

crates/ika-core/src/sui_connector/sui_event_into_request.rs
ItayLevyOfficial, line 1: those changes are needed as the event's package id can be either the old one or the upgraded one

sdk/typescript/src/client/ika-client.ts
```diff
@@ -681,15 +681,14 @@ export class IkaClient {
...
-		const protocolPublicParameters =
-			networkEncryptionKey.epoch === Number(objects.coordinatorInner.current_epoch)
-				? await networkDkgPublicOutputToProtocolPublicParameters(
-						await this.#readTableVecAsRawBytes(networkEncryptionKeyPublicOutputID),
-					)
-				: await reconfigurationPublicOutputToProtocolPublicParameters(
-						await this.#readTableVecAsRawBytes(networkEncryptionKey.reconfigurationOutputID!),
-						await this.#readTableVecAsRawBytes(networkEncryptionKeyPublicOutputID),
-					);
+		const protocolPublicParameters = !networkEncryptionKey.reconfigurationOutputID
```
ItayLevyOfficial, line 744: a bugfix

sdk/typescript/test/helpers/dwallet-test-helpers.ts
```diff
@@ -192,7 +196,9 @@ export async function createCompleteDWalletV2(
 	await delay(5); // Wait for 5 seconds to ensure the DWallet is created
 
 	// Step 2: Register encryption key
-	await registerTestEncryptionKey(ikaClient, suiClient, userShareEncryptionKeys, testName);
+	if (registerEncryptionKey) {
+		await registerTestEncryptionKey(ikaClient, suiClient, userShareEncryptionKeys, testName);
```
ItayLevyOfficial, line 200: trying to register an encryption key for a sui address more than once results in a bug

crates/ika-core/src/sui_connector/sui_event_into_request.rs
```diff
@@ -9,26 +9,35 @@ use crate::request_protocol_data::{
...
+    if (event_type.address != *packages_config.packages.ika_dwallet_2pc_mpc_package_id
+        && (packages_config
+            .packages
+            .ika_dwallet_2pc_mpc_package_id_v2
+            .is_none()
+            || event_type.address
+                != *packages_config
+                    .packages
+                    .ika_dwallet_2pc_mpc_package_id_v2
+                    .unwrap()))
```
ycscaly, line 39:
```suggestion
        && Some(event_type.address)
                != *packages_config
                    .packages
                    .ika_dwallet_2pc_mpc_package_id_v2))
```

crates/ika-core/src/sui_connector/sui_event_into_request.rs
```diff
@@ -40,73 +49,102 @@ pub fn sui_event_into_session_request(
             "received an event from a wrong SUI module - rejecting!"
         ));
     }
-
-    let session_request = if event_type
-        == DWalletSessionEvent::<DWalletImportedKeyVerificationRequestEvent>::type_(packages_config)
+    if !event_type
+        .to_string()
+        .contains(&DWALLET_SESSION_EVENT_STRUCT_NAME.to_string())
```
ycscaly, line 53: explain please. contains is scary, too general

sdk/typescript/src/client/ika-client.ts
```diff
@@ -937,10 +936,10 @@ export class IkaClient {
...
 						}),
 					)
-				).find(
-					(reconfigOutput) =>
-						Number(reconfigOutput.name) === Number(objects.coordinatorInner.current_epoch) - 1,
-				);
+				)
+					.sort((a, b) => Number(a.name) - Number(b.name))
+					// The last reconfiguration has not necessarily been completed, so we take the second to last
+					.at(-2);
```
ycscaly, line 1002: but it could have been completed, in which case you take the non-latest one (?)

sdk/typescript/test/helpers/dwallet-test-helpers.ts
ycscaly: A better thing to do is to check if its already registered, and only register if its not

ycscaly: @Iamknownasfesal

crates/ika-core/src/sui_connector/sui_event_into_request.rs
ItayLevyOfficial: Agreed, but as the event package id behaves strangely after the upgrade (sometimes the outer `DWalletSessionEvent` has a different package id then the inner event), this is the cleanest solution imo (I also discussed it with Omer). The `DWALLET_SESSION_EVENT_STRUCT_NAME` wraps every session start event, and only those events, so if the full event type does not contain it - it can be safely ignored.

sdk/typescript/test/helpers/dwallet-test-helpers.ts
ItayLevyOfficial: this is only a test util so it's reasonable to keep it this way imo

sdk/typescript/src/client/ika-client.ts
ItayLevyOfficial: ye, shouldn't be a problem afaik

crates/ika-core/src/sui_connector/sui_event_into_request.rs
ycscaly: what does strangely means? We should understand why it is like that

sdk/typescript/src/client/ika-client.ts
ycscaly: why not a problem? looks like it is

ItayLevyOfficial: is using the previous epoch reconfiguration output to generate the protocol public parameters a problem?

crates/ika-core/src/sui_connector/sui_event_into_request.rs
ItayLevyOfficial: as I said, sometimes the outer `DWalletSessionEvent` has a different package id then the inner event. I don't know why it happens, maybe Omer knows

PR #1533, opened 2025-09-16
Improve move dkg and global presign

omersadika:


contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -300,11 +331,12 @@ public fun request_dwallet_dkg(
     encryption_key_address: address,
     user_public_output: vector<u8>,
     singer_public_key: vector<u8>,
+    sign_during_dkg_request: Option<SignDuringDKGRequest>,
     session_identifier: SessionIdentifier,
     payment_ika: &mut Coin<IKA>,
     payment_sui: &mut Coin<SUI>,
     ctx: &mut TxContext,
-): DWalletCap {
+): (DWalletCap, Option<ID>) {
```
Iamknownasfesal, line 339: Should we do like this or make a variant where we return dwallet cap and id directly instead of option, keeping dWalletCap return only one return and not a tuple

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -315,6 +347,36 @@ public fun request_dwallet_dkg(
...
+    curve: u32,
+    centralized_public_key_share_and_proof: vector<u8>,
+    user_public_output: vector<u8>,
+    public_user_secret_key_share: vector<u8>,
+    sign_during_dkg_request: Option<SignDuringDKGRequest>,
+    session_identifier: SessionIdentifier,
+    payment_ika: &mut Coin<IKA>,
+    payment_sui: &mut Coin<SUI>,
+    ctx: &mut TxContext,
+): (DWalletCap, Option<ID>) {
```
Iamknownasfesal, line 370: Should we do like this or make a variant where we return dwallet cap and id directly instead of option, keeping dWalletCap return only one return and not a tuple

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -1018,6 +1037,36 @@ public struct RejectedDWalletDKGSecondRoundEvent has copy, drop, store {
...
+    sign_id: ID,
+    presign_id: ID,
+    presign: vector<u8>,
+    signature_algorithm: u32,
+    hash_scheme: u32,
+    message: vector<u8>,
+    message_centralized_signature: vector<u8>,
+}
+
+public enum UserSecretKeySharEvent has copy, drop, store {
```
Iamknownasfesal: UserSecretKeyShareEvent

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -2576,23 +2685,130 @@ public(package) fun request_dwallet_dkg(
...
+    assert!(encryption_key_curve == curve, EMismatchCurve);
     self.support_config.validate_curve(curve);
 
-    let pricing_value = self
-        .pricing_and_fee_manager
-        .get_pricing_value_for_protocol(curve, option::none(), DWALLET_DKG_PROTOCOL_FLAG);
+    let id = object::new(ctx);
+    let encrypted_user_secret_key_share_id = id.to_inner();
+
+    let user_secret_key_share = UserSecretKeySharEvent::Encrypted {
```
Iamknownasfesal: UserSecretKeyShareEvent

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -4697,6 +4991,26 @@ public(package) fun set_paused_curves_and_signature_algorithms(
...
+/// ### Parameters
+/// - **`curve_to_signature_algorithms_for_dkg`**: Map of curves to signature algorithms for dWallets that are created via DKG.
+/// - **`curve_to_signature_algorithms_for_imported_key`**: Map of curves to signature algorithms for dWallets that are created via imported key.
+public(package) fun set_global_presign_config(
+    self: &mut DWalletCoordinatorInner,
+    curve_to_signature_algorithms_for_dkg: VecMap<u32, vector<u32>>,
+    curve_to_signature_algorithms_for_imported_key: VecMap<u32, vector<u32>>,
+    _: &VerifiedProtocolCap,
+) {
+    if(self.extra_fields.contains(b"global_presign_config")) {
```
Iamknownasfesal: should we have this as a constant?

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
ycscaly: agreed

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -2576,23 +2685,130 @@ public(package) fun request_dwallet_dkg(
...
+        state: EncryptedUserSecretKeyShareState::AwaitingNetworkVerification,
+    };
+
+    let dwallet = self.get_dwallet_mut(dwallet_cap.dwallet_id);
+
+    dwallet
+        .encrypted_user_secret_key_shares
+        .add(encrypted_user_secret_key_share_id, encrypted_user_share);
+    
+    (dwallet_cap, sign_id)
```
ycscaly, line 2754: what is this

ycscaly: you moved it from the inner function so it wouldn't be used for the public user share? but why

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -3896,6 +4173,21 @@ public(package) fun request_future_sign(
 
     let signature_algorithm = presign_obj.signature_algorithm;
 
+    
+    let global_presign_config = self.global_presign_config();
+    let is_global_presign = if(is_imported_key_dwallet) {
+        global_presign_config.is_global_presign_for_imported_key(curve, signature_algorithm)
+    } else {
+        global_presign_config.is_global_presign_for_dkg(curve, signature_algorithm)
```
ycscaly, line 4205: need also the version of the dwallet

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
omersadika: I think it will make ptbs more complex, it's easy to deal with tuples with ptbs

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: For the public user share we don't really need the encrypted_centralized_secret_share_and_proof so I won't do any of the this

contracts/ika_dwallet_2pc_mpc/sources/sessions_manager.move
```diff
@@ -231,19 +231,27 @@ public(package) fun lock_last_user_initiated_session_to_complete_in_current_epoc
...
-    identifier_preimage: vector<u8>,
+    bytes: vector<u8>,
     ctx: &mut TxContext,
 ): SessionIdentifier {
     assert!(
-        identifier_preimage.length() == SESSION_IDENTIFIER_LENGTH,
+        bytes.length() == SESSION_IDENTIFIER_LENGTH,
         ESessionIdentifierInvalidLength,
     );
+    let mut hasher = ctx.sender().to_bytes();
```
ycscaly:
```suggestion
    let mut bytes_to_hash = ctx.sender().to_bytes();
```

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -2527,6 +2573,74 @@ fun validate_approve_message(
...
+
+    let presign_id = id.to_inner();
+
+    let dwallet_network_encryption_key_id = dynamic_field::remove(&mut id, b"dwallet_network_encryption_key_id");
+    id.delete();
+
+    // Sanity checks: check that the IDs of the capability and presign match, and that they point to this dWallet.
+    assert!(presign_cap_id == cap_id, EPresignNotExist);
+    assert!(presign_id == presign_cap_presign_id, EPresignNotExist);
+    assert!(presign_cap_dwallet_id == presign_dwallet_id, EPresignNotExist);
```
ycscaly:
```suggestion
    assert!(presign_dwallet_id.is_none(), EOnlyGlobalPresignAllowed);
```

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -3285,12 +3550,20 @@ public(package) fun request_presign(
...
     let dwallet_network_encryption_key_id = dwallet.dwallet_network_encryption_key_id;
 
+    let is_imported_key_dwallet = dwallet.is_imported_key_dwallet;
+
+    let global_presign_config = self.global_presign_config();
+
+    let is_global_presign = if(is_imported_key_dwallet) {
+        global_presign_config.is_global_presign_for_imported_key(curve, signature_algorithm)
+    } else {
+        global_presign_config.is_global_presign_for_dkg(curve, signature_algorithm)
```
ycscaly, line 3572: this should also take in the dkg output as input, and check it is v2

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -3285,12 +3550,20 @@ public(package) fun request_presign(
...
-        .validate_signature_algorithm_not_allowed_global_presign(signature_algorithm);
 
     let dwallet_network_encryption_key_id = dwallet.dwallet_network_encryption_key_id;
 
+    let is_imported_key_dwallet = dwallet.is_imported_key_dwallet;
+
+    let global_presign_config = self.global_presign_config();
+
+    let is_global_presign = if(is_imported_key_dwallet) {
+        global_presign_config.is_global_presign_for_imported_key(curve, signature_algorithm)
```
ycscaly, line 3570: I think this is overcomplicating, just `if(is_imported_key_dwallet) { false }`

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -3399,9 +3672,13 @@ public(package) fun request_global_presign(
 
     self.support_config.validate_curve_and_signature_algorithm(curve, signature_algorithm);
     self.validate_network_encryption_key_supports_curve(dwallet_network_encryption_key_id, curve);
-    self.support_config.validate_signature_algorithm_allowed_global_presign(signature_algorithm);
+    let global_presign_config = self.global_presign_config();
+    let is_global_presign = global_presign_config.is_global_presign_for_dkg(curve, signature_algorithm);
```
ycscaly: why `for_dkg` here?

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -4697,6 +5006,26 @@ public(package) fun set_paused_curves_and_signature_algorithms(
     self.support_config.set_paused(paused_curves, paused_signature_algorithms, paused_hash_schemes);
 }
 
+/// Sets the global presign config.
+///
+/// ### Parameters
+/// - **`curve_to_signature_algorithms_for_dkg`**: Map of curves to signature algorithms for dWallets that are created via DKG.
+/// - **`curve_to_signature_algorithms_for_imported_key`**: Map of curves to signature algorithms for dWallets that are created via imported key.
+public(package) fun set_global_presign_config(
```
ycscaly, line 5025: why is this a new function? it should have been existed no?

contracts/ika_dwallet_2pc_mpc/sources/support_config.move
```diff
@@ -34,9 +34,26 @@ public struct SupportConfig has store {
...
+/// If a presign is not global, it can only be used by the dWallet it was created for.
+public struct GlobalPresignConfig has store {
+    /// Map of curves to signature algorithms for dWallets that are created via DKG.
+    /// This means for this curve and this signature algorithm, it is only allowed to use global presign.
+    /// e.g. secp256k1 -> [(ecdsa -> [sha256, keccak256]), (schnorr -> [sha256])]
+    curve_to_signature_algorithms_for_dkg: VecMap<u32, vector<u32>>,
+    /// Map of curves to signature algorithms for dWallets that are created via imported key.
+    /// This means for this curve and this signature algorithm, it is only allowed to use global presign.
+    /// e.g. secp256k1 -> [(ecdsa -> [sha256, keccak256]), (schnorr -> [sha256])]
+    curve_to_signature_algorithms_for_imported_key: VecMap<u32, vector<u32>>,
```
ycscaly, line 54: remove

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: We already have that before - assert!(presign_cap_dwallet_id.is_none(), EOnlyGlobalPresignAllowed);

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: Because we support global for imported for schnorr

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: Why?

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: Why?

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: Here is before we know the dwallet, so I'm just checking if that signature_algorithm should be supported, we can add also "or" for_imported, but i assume if you have for_imported you must have for_dkg

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: We don't have that, and I recreated all the global presign staff, and the old value is deprecated

contracts/ika_dwallet_2pc_mpc/sources/support_config.move
omersadika: ?

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -1018,6 +1040,36 @@ public struct RejectedDWalletDKGSecondRoundEvent has copy, drop, store {
...
+    sign_id: ID,
+    presign_id: ID,
+    presign: vector<u8>,
+    signature_algorithm: u32,
+    hash_scheme: u32,
+    message: vector<u8>,
+    message_centralized_signature: vector<u8>,
+}
+
+public enum UserSecretKeyShareEvent has copy, drop, store {
```
ItayLevyOfficial:
```suggestion
public enum UserSecretKeyShareType has copy, drop, store {
```

This event is never emitted on its own, it is always encompassed within another event

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -771,32 +833,34 @@ public fun commit_upgrade(self: &mut DWalletCoordinator, upgrade_package_approve
...
+    self.try_migrate_impl(ctx);
 }
 
 /// Try to migrate the coordinator object to the new package id.
 ///
 /// This function sets the new package id and version and can be modified in future versions
 /// to migrate changes in the `coordinator_inner` object if needed.
 /// Call this function after the migration epoch is reached.
-public fun try_migrate(self: &mut DWalletCoordinator) {
+public fun try_migrate(self: &mut DWalletCoordinator, ctx: &mut TxContext) {
```
ItayLevyOfficial, line 845:
```suggestion
public fun try_migrate(self: &mut DWalletCoordinator) {
```

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -771,32 +833,34 @@ public fun commit_upgrade(self: &mut DWalletCoordinator, upgrade_package_approve
...
-    self.try_migrate_impl();
+    self.try_migrate_impl(ctx);
 }
 
 /// Migrate the coordinator object to the new package id.
 ///
 /// This function sets the new package id and version and can be modified in future versions
 /// to migrate changes in the `coordinator_inner` object if needed.
-fun try_migrate_impl(self: &mut DWalletCoordinator) {
+fun try_migrate_impl(self: &mut DWalletCoordinator, _ctx: &mut TxContext) {
```
ItayLevyOfficial, line 854:
```suggestion
fun try_migrate_impl(self: &mut DWalletCoordinator) {
```

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
```diff
@@ -771,32 +833,34 @@ public fun commit_upgrade(self: &mut DWalletCoordinator, upgrade_package_approve
...
 /// Try to migrate the coordinator object to the new package id.
 ///
 /// This function sets the new package id and version and can be modified in future versions
 /// to migrate changes in the `coordinator_inner` object if needed.
 /// Call this function after the migration epoch is reached.
-public fun try_migrate(self: &mut DWalletCoordinator) {
+public fun try_migrate(self: &mut DWalletCoordinator, ctx: &mut TxContext) {
     assert!(self.migration_epoch.is_some_and!(|e| self.inner_without_version_check().epoch() >= *e), EInvalidMigration);
-    self.try_migrate_impl();
+    self.try_migrate_impl(ctx);
```
ItayLevyOfficial, line 847:
```suggestion
    self.try_migrate_impl();
```

contracts/ika_dwallet_2pc_mpc/sources/coordinator.move
Iamknownasfesal: ctx might be needed here later on, that's why it needs to be added to the signature of the public function(because you cannot change args later on and need to introduce new functions)

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -2527,6 +2573,74 @@ fun validate_approve_message(
...
+        _ => abort EInvalidPresign,
+    };
+
+    let presign_id = id.to_inner();
+
+    let dwallet_network_encryption_key_id = dynamic_field::remove(&mut id, b"dwallet_network_encryption_key_id");
+    id.delete();
+
+    // Sanity checks: check that the IDs of the capability and presign match, and that they point to this dWallet.
+    assert!(presign_cap_id == cap_id, EPresignNotExist);
```
ItayLevyOfficial: The error code is incorrect, and having the same error code for multiple different errors makes debugging harder

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -2576,23 +2690,136 @@ public(package) fun request_dwallet_dkg(
...
     ctx: &mut TxContext,
-): DWalletCap {
+): (DWalletCap, Option<ID>) {
+    let encryption_key = self.encryption_keys.borrow(encryption_key_address);
+    let encryption_key_curve = encryption_key.curve;
+    let encryption_key_id = encryption_key.id.to_inner();
+    let encryption_key = encryption_key.encryption_key;
+    let created_at_epoch: u64 = self.current_epoch;
+
+    assert!(encryption_key_curve == curve, EMismatchCurve);
```
ItayLevyOfficial, line 2709: Are you sure they have to be the same? The encryption key in this context is the key the user secret share is encrypted to, no?

contracts/ika_dwallet_2pc_mpc/sources/support_config.move
```diff
@@ -130,24 +147,44 @@ public(package) fun validate_curve_and_signature_algorithm_and_hash_scheme(
...
+/// - `signature_algorithm`: Signature algorithm to check
+/// ### Returns
+/// True if only global presign is allowed for the dWallet that is created via imported key, false otherwise
+public(package) fun is_global_presign_for_imported_key(
+    self: & GlobalPresignConfig,
+    curve: u32,
     signature_algorithm: u32,
-) {
+): bool {
     assert!(
```
ItayLevyOfficial: Why do you throw an error in this case, but return false in the other?

contracts/ika_dwallet_2pc_mpc/sources/sessions_manager.move
ItayLevyOfficial: This change requires changing the `sessionIdentifierDigest` function in the TS sdk & the `session_identifier_digest` in the Rust code - please do that before merging, otherwise this pr will break the code

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
omersadika: the curve  of the encryption key must be the same as the the dwallet curve

contracts/ika_dwallet_2pc_mpc/sources/support_config.move
omersadika: Fixed

contracts/ika_dwallet_2pc_mpc/sources/coordinator_inner.move
```diff
@@ -1018,6 +1042,36 @@ public struct RejectedDWalletDKGSecondRoundEvent has copy, drop, store {
...
+    sign_id: ID,
+    presign_id: ID,
+    presign: vector<u8>,
+    signature_algorithm: u32,
+    hash_scheme: u32,
+    message: vector<u8>,
+    message_centralized_signature: vector<u8>,
+}
+
+public enum UserSecretKeyShareEventType has copy, drop, store {
```
omersadika, line 1055: @ycscaly enum for user share

crates/ika-protocol-config/src/lib.rs
```diff
@@ -503,9 +503,6 @@ impl ProtocolConfig {
                 //
                 //     // Remove a constant (ensure that it is never accessed during this version).
                 //     existing_constant: None,
-                2 => {
```
omersadika, line 506: Duplicated variant in match

PR #1535, opened 2025-09-18
fix: Use the latest network key reconfiguration output instead of the initial DKG output

ItayLevyOfficial:
## Description

Until now, the TS SDK always used the network's first DKG output to derive the protocol's public parameters. This is problematic because the network key upgrade requires users to use the latest reconfiguration output to derive them. This PR fixes the issue by using the most recent reconfiguration output when it is available.

## Test Plan

I created a dWallet before and after a successful reconfiguration, verifying that the post-reconfiguration flow uses the latest reconfiguration key with the debugger.

---

## Release notes

- [ ] Protocol: 
- [ ] Indexer: 
- [ ] JSON-RPC: 
- [ ] GraphQL: 
- [ ] CLI: 
- [ ] Rust SDK:
- [ ] REST API:

### DIFF
--- a/crates/dwallet-mpc-centralized-party/src/lib.rs
+++ b/crates/dwallet-mpc-centralized-party/src/lib.rs
@@ -15,14 +15,24 @@ use class_groups::{
     SECP256K1_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS, SECP256K1_SCALAR_LIMBS, Secp256k1DecryptionKey,
     setup::DeriveFromPlaintextPublicParameters,
 };
-use dwallet_mpc_types::dwallet_mpc::{DKGDecentralizedPartyOutputSecp256k1, DKGDecentralizedPartyVersionedOutputSecp256k1, DWalletCurve, NetworkDecryptionKeyPublicOutputType, NetworkEncryptionKeyPublicDataV1, NetworkEncryptionKeyPublicDataV2, ReconfigurationParty, SerializedWrappedMPCPublicOutput, VersionedCentralizedDKGPublicOutput, VersionedDecryptionKeyReconfigurationOutput, VersionedDwalletDKGFirstRoundPublicOutput, VersionedDwalletDKGSecondRoundPublicOutput, VersionedDwalletUserSecretShare, VersionedEncryptedUserShare, VersionedImportedDWalletPublicOutput, VersionedImportedDwalletOutgoingMessage, VersionedNetworkDkgOutput, VersionedNetworkEncryptionKeyPublicData, VersionedPresignOutput, VersionedPublicKeyShareAndProof, VersionedSignOutput, VersionedUserSignedMessage};
-use group::{CyclicGroupElement, GroupElement, HashType, OsCsRng, Samplable, secp256k1};
+use dwallet_mpc_types::dwallet_mpc::{
+    DKGDecentralizedPartyOutputSecp256k1, DKGDecentralizedPartyVersionedOutputSecp256k1,
+    DWalletCurve, NetworkDecryptionKeyPublicOutputType, NetworkEncryptionKeyPublicDataV1,
+    NetworkEncryptionKeyPublicDataV2, ReconfigurationParty, SerializedWrappedMPCPublicOutput,
+    VersionedCentralizedDKGPublicOutput, VersionedDecryptionKeyReconfigurationOutput,
+    VersionedDwalletDKGFirstRoundPublicOutput, VersionedDwalletDKGSecondRoundPublicOutput,
+    VersionedDwalletUserSecretShare, VersionedEncryptedUserShare,
+    VersionedImportedDWalletPublicOutput, VersionedImportedDwalletOutgoingMessage,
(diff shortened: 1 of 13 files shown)


sdk/typescript/src/client/cryptography.ts
```diff
@@ -363,6 +364,28 @@ export async function networkDkgPublicOutputToProtocolPublicParameters(
...
+ * @param threshold - The threshold value
+ * @returns The protocol public parameters
+ */
+export async function reconfigurationPublicOutputToProtocolPublicParameters(
+	reconfiguration_dkg_public_output: Uint8Array,
+	committee_size: number,
+	threshold: number,
+): Promise<Uint8Array> {
+	return Uint8Array.from(
+		await reconfiguration_public_output_to_protocol_pp(
```
Iamknownasfesal, line 379: you cannot do this, you need to add it into wasm_loader instead and initialize wasm module first and then run it, check wasm loader file for it

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -558,6 +575,76 @@ fn protocol_public_parameters(
...
+                neutral_group_value,
+                neutral_ciphertext_value,
+                neutral_ciphertext_value,
+                decryption_key_share_public_parameters
+                    .encryption_scheme_public_parameters
+                    .clone(),
+            );
+
+            Ok(protocol_public_parameters)
+        }
```
ycscaly: I don't like this.. we should upgrade the TS SDK to use the latest cryptography code only after the reconfiguration v1 to v2 is done (until then use the old TS + WASM), and then assert is it is V2.

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -558,6 +575,76 @@ fn protocol_public_parameters(
...
+    quorum_threshold: usize,
+) -> anyhow::Result<ProtocolPublicParameters> {
+    let reconfiguration_dkg_public_output: VersionedDecryptionKeyReconfigurationOutput =
+        bcs::from_bytes(&reconfiguration_dkg_public_output)?;
+
+    // Every member has a voting power of 1 in the current Move code.
+    let access_structure = WeightedThresholdAccessStructure::new(
+        quorum_threshold as Weight,
+        (1..=committee_size).map(|i| (i as PartyID, 1)).collect(),
+    )?;
```
ycscaly: I don't like this either, there's a reason why we never needed the user to know the committee structure so far. Delete

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -558,6 +575,76 @@ fn protocol_public_parameters(
...
+        }
+        VersionedDecryptionKeyReconfigurationOutput::V2(public_output_bytes) => {
+            let public_output: <twopc_mpc::decentralized_party::reconfiguration::Party as mpc::Party>::PublicOutput =
+                bcs::from_bytes(public_output_bytes)?;
+            let secp256k1_protocol_public_parameters =
+                twopc_mpc::decentralized_party::reconfiguration::PublicOutput::secp256k1_protocol_public_parameters(
+                    &public_output,
+                )?;
+
+            Ok(secp256k1_protocol_public_parameters)
```
ycscaly: We have functions that are in Yael's latest PRs for the centralized party that creates the public input for protocols generically.

I know we want to cache things in the user side, we still should, but maybe we should do that generically too - or just cache for all the protocols.

ycscaly (changes requested):

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: So what do you want me to do here? Create separate v1 and v2 functions?

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: The access structure is required to generate the protocol pp, please suggest an alternative. If I'll simply delete, the code will not compile.

crates/dwallet-mpc-centralized-party/src/lib.rs
ItayLevyOfficial: In the code you commented on I am generating the protocol public parameters, not public input.

Iamknownasfesal (approved):

crates/dwallet-mpc-centralized-party/src/lib.rs
```diff
@@ -558,6 +573,65 @@ fn protocol_public_parameters(
...
+                secp256k1::GroupElement,
+            >(
+                neutral_group_value,
+                neutral_group_value,
+                neutral_ciphertext_value,
+                neutral_ciphertext_value,
+                encryption_scheme_public_parameters,
+            );
+
+            Ok(protocol_public_parameters)
```
ycscaly:
```suggestion
            protocol_public_parameters(versioned_network_dkg_output)
```

ycscaly (approved):

PR #1540, opened 2025-09-28
fix: Presign session rejection 

yaelAbergel:


ItayLevyOfficial (changes requested):
not needed, a fix for the presign has already been merged to main

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
ycscaly, line 1: all of this should be reverted

ycscaly: I don't want specialized logic for this case.

ycscaly: we should use the same logic as session rejecting when advancing flow

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -388,7 +399,7 @@ impl DWalletMPCManager {
     ///
     /// If there is no `session_request`, and we've got it in this call,
     /// we update that field in the open session.
-    fn handle_mpc_request(&mut self, request: DWalletSessionRequest) {
+    fn handle_mpc_request(&mut self, request: DWalletSessionRequest) -> DwalletMPCResult<()> {
```
ycscaly:
```suggestion
    fn handle_mpc_request(&mut self, request: DWalletSessionRequest) {
```

ycscaly: I don't want to return an error here, instead we should handle the errornous case like a session that was created and failed

crates/ika-core/src/dwallet_mpc/mpc_session.rs
```diff
@@ -482,7 +493,7 @@ impl DWalletMPCManager {
             Ok((public_input, private_input)) => (public_input, private_input),
             Err(e) => {
                 error!(should_never_happen=true, error=?e, ?request, "create session input from dWallet request with error");
-                return;
+                return Err(e);
```
ycscaly: Instead of doing this, create the session and mark it as completed/rejected as if it has advanced and failed. We also want to save it to the db in the same way as advance failed, to avoid doing this again.
We want to make sure this is handled later on in the mpc service in the same way, and that we reject the session

PR #1645, opened 2026-02-19
Fix rust tests

omersadika:


crates/ika-core/src/sui_connector/sui_event_into_request.rs
```diff
@@ -354,22 +354,26 @@ mod tests {
...
+            5,
+            vec![42; 32],
+            DWalletDKGFirstRoundRequestEvent {
+                dwallet_id: ObjectID::random(),
+                dwallet_cap_id: ObjectID::random(),
+                dwallet_network_encryption_key_id: ObjectID::random(),
+                curve: 0,
+            },
+        );
+        let contents = bcs::to_bytes(&event).expect("should serialize pushed event");
```
ycscaly, line 376: why? i used a pre-defined serialized value

PR #1646, opened 2026-02-19
feat(sdk): migrate TypeScript SDK to @mysten/sui v2.4.0 client APIs

arbuthnot-eth:
## Summary

Migrate the `@ika.xyz/sdk` TypeScript SDK from `@mysten/sui` v1 (`^1.44.0`) to v2 (`^2.4.0`), updating all client APIs, type signatures, and BCS decoding paths to align with the new Sui SDK surface.

### Breaking Changes

- **Client type**: `SuiClient` → `SuiClientTypes.TransportMethods` (supports both gRPC and JSON-RPC transports)
- **Consumers must upgrade** `@mysten/sui` to `^2.4.0`
- **Version bump**: `0.2.7` → `0.3.0` (minor, via changeset)

### Source Changes (`src/client/`)

- **`ika-client.ts`** — rewrite all Sui RPC calls to v2 API surface:
  - `getObject` → `getObjects` (batched, with `include: { objectBcs: true }`)
  - `getOwnedObjects` → `listOwnedObjects` (with structured `type` filter)
  - `getDynamicFields` → `listDynamicFields`
  - `devInspectTransactionBlock` → `simulateTransaction` (with `include: { commandResults: true }`)
  - v2 error handling: `obj instanceof Error` checks on batched responses
- **`types.ts`** — update `IkaClientOptions.suiClient` type from `SuiClient` to `SuiClientTypes.TransportMethods`
- **`utils.ts`** — `objResToBcs()` now reads `obj.objectBcs` (raw `Uint8Array`) instead of base64 decoding from `obj.data.bcs.bcsBytes`; `fetchAllDynamicFields()` updated for v2 pagination shape

### Test Changes

- Migrate all test helpers from `SuiClient` to `SuiGrpcClient` (v2 gRPC transport)
- Update `executeTestTransactionWithKeypair` to handle v2 transaction result shape (`$kind` discriminated union)
- Update `getObjectWithType` to use v2 `getObject` with `include: { json: true }`
- Add `cwd`-based fallback paths in `findIkaConfigFile()` for flexible test runner execution
- Stabilize vitest config: `pool: 'forks'` for memory isolation, conditional test exclusion based on env vars

### Versioning

- Added `.changeset/migrate-sui-v2.md` for minor bump → `0.3.0`

## Test Plan

- [x] `pnpm build` — passes cleanly (no type errors)
- [x] Unit tests — 95/95 pass across 6 test files (utils, errors, network-config, cryptography, encryption keys, index exports)
- [ ] Integration tests — require a live Ika network (`RUN_INTEGRATION_TESTS=true` + `ika_config.json`); to be validated in CI
- [ ] System tests — require full node cluster (`RUN_SYSTEM_TESTS=true` + `test/system-tests/.env`); to be validated in CI

### DIFF
--- a/sdk/typescript/package.json
+++ b/sdk/typescript/package.json
@@ -1,5 +1,5 @@
 {
-	"name": "@dika.sui/sdk",
+	"name": "@imbibe/dika",
 	"author": "brando.sui (https://github.com/arbuthnot-eth)",
 	"description": "Dika typescript sdk",
 	"repository": {


arbuthnot-eth:
Hey @Iamknownasfesal, appreciate the feedback — you were right about the BCS.

Dug into the actual `@mysten/sui` v2 source and found the root cause of the "non compatible bcs" error. The previous migration used `objectBcs` which returns the **full Object BCS envelope** (type + owner + data + storage rebate — the whole `bcs.Object` serialization). The generated `MoveStruct.parse()` calls expect only the **raw MoveObject struct bytes**, which is what `content` provides. Swapped all `include: { objectBcs: true }` → `include: { content: true }` and `objResToBcs` now reads `obj.content`.

Also restored the BCS-based `package_id` extraction you had — parsing `DWalletCoordinator` and `System` objects via their generated BCS types to get the actual `package_id` field, rather than splitting it from `obj.type`. The type string keeps the original defining package after a Sui package upgrade, but the on-chain `package_id` field reflects the current upgraded package, which is what the move call targets in `coordinator.ts`/`system.ts` need.

Kept the DF improvements from the migration (paginated `fetchAllDynamicFields`, version-sorted DF selection via `#selectHighestVersionDynamicField`) since those are strictly better than the original single-page fetch + "take last" approach.

Summary of `50e1088`:
- `objectBcs` → `content` for correct BCS parsing (this was the breaking bug)
- Restored `CoordinatorModule.DWalletCoordinator.parse()` / `SystemModule.System.parse()` for `package_id`
- Paginated DF fetching + highest-version DF selection preserved
- `#assertObjectType()` defensive check before parsing
- Unit tests updated, all 95 pass + `tsc --noEmit` clean

Still need a live Ika network to run the `test/v2` integration suite — happy to work through that if you can point me at the right environment or CI setup.

Iamknownasfesal:
> i dont currently possess the capability to test with real live ika validator nodes actually running the MPC protocols

You can run the tests with running localnet and running the tests via vitest, https://docs.ika.xyz/docs/sdk/setup-localnet

PR #1664, opened 2026-03-09
Add NOA signed checkpoint infrastructure

omersadika:
## Summary
- Add Network Owned Address (NOA) signed checkpoint infrastructure for submitting MPC session results to Sui without BLS quorum signatures
- Implement NOA checkpoint building, signing pipeline integration, deduplication, and state sync
- Add conditional BLS/NOA routing controlled by protocol config flags (`bls_checkpoints`, `noa_checkpoints`)
- Add `hash_scheme` to `NetworkOwnedAddressSign` consensus output for deterministic checkpoint signing
- Change `session_sequence_number` to `Option<u64>` in `DWalletSessionRequest` for sessions that don't need sequence numbers

## Test plan
- [ ] Verify `cargo build --release -p ika-types -p ika-core -p ika-node` compiles
- [ ] Verify `cargo test --release -p ika-types` passes
- [ ] Verify Move contracts build: `cd contracts/ika_dwallet_2pc_mpc && sui move build` and `cd contracts/ika_system && sui move build`
- [ ] Verify BLS-only mode works when `noa_checkpoints` is disabled
- [ ] Verify NOA checkpoint flow end-to-end when enabled

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
@@ -54,8 +54,8 @@ use ika_types::messages_dwallet_mpc::{
 use ika_types::messages_system_checkpoints::SystemCheckpointMessageKind;
 use ika_types::noa_checkpoint;
 use ika_types::noa_checkpoint::{
-    CounterpartyChainKind, NOACheckpointKindName, NOACheckpointTxObservation, SuiChainContext,
-    SuiChainObservation,
+    CounterpartyChainKind, NOACheckpointKindName, NOACheckpointResolution,
+    NOACheckpointTxObservation, SuiChainContext, SuiChainObservation,
 };
 use ika_types::sui::EpochStartSystem;
 use ika_types::sui::{EpochStartSystemTrait, EpochStartValidatorInfoTrait};
@@ -654,8 +654,6 @@ impl DWalletMPCService {
         &mut self,
         agreed_status: &crate::dwallet_mpc::mpc_manager::AgreedStatusUpdate,
     ) {
-        use ika_types::noa_checkpoint::NOACheckpointResolution;
-
         for tx_ref in &agreed_status.newly_finalized_tx_refs {
             let resolution = NOACheckpointResolution::Finalized(tx_ref.clone());
             self.route_resolution(resolution, tx_ref.kind_name);
@@ -1972,7 +1970,7 @@ impl DWalletMPCService {
             } => {
                 let tx = DWalletCheckpointMessageKind::RespondDWalletImportedKeyVerificationOutput(
                     DWalletImportedKeyVerificationOutput {
-                        dwallet_id: dwallet_id.to_vec().clone(),
+                        dwallet_id: dwallet_id.to_vec(),
                         public_output: output,
                         encrypted_user_secret_key_share_id: encrypted_user_secret_key_share_id
                             .to_vec(),
--- a/crates/ika-core/src/noa_checkpoints/chain_submitter.rs
+++ b/crates/ika-core/src/noa_checkpoints/chain_submitter.rs
@@ -35,7 +35,8 @@ pub trait NOAChainSubmitter<K: NOACheckpointKind>: Send + Sync + 'static {
 
 /// No-op chain submitter that logs operations and always reports execution success.
 /// Used as a placeholder until actual chain submission is implemented.
-// TODO: delete it once we have a real implementation
+// TODO(noa-checkpoints): Remove LogOnlyChainSubmitter once SuiChainSubmitter is implemented
+// and wired into the NOA checkpoint pipeline.
 pub struct LogOnlyChainSubmitter;
 
 #[async_trait]
--- a/crates/ika-core/src/noa_checkpoints/local_store.rs
+++ b/crates/ika-core/src/noa_checkpoints/local_store.rs
@@ -7,6 +7,7 @@ use ika_types::noa_checkpoint::{
     CertifiedNOACheckpointMessage, CounterpartyChain, NOACheckpointKind, NOACheckpointMessage,
     NOACheckpointTxRef, NOACheckpointTxStatus,
 };
+use tracing::{error, warn};
 
 // === NOACheckpointLocalStore ===
 
@@ -174,21 +175,29 @@ impl<K: NOACheckpointKind> NOACheckpointLocalStore<K> {
     /// Record that a transaction has been submitted to the chain.
     pub fn mark_submitted(&mut self, tx_ref: &NOACheckpointTxRef, chain_tx_id: Vec<u8>) {
         let Some(entry) = self.entries.get_mut(&tx_ref.sequence_number) else {
+            warn!(?tx_ref, "mark_submitted: unknown tx_ref, ignoring");
             return;
         };
         let Some(tx_state) = entry.transactions.get_mut(tx_ref.tx_index as usize) else {
+            warn!(?tx_ref, "mark_submitted: unknown tx_index, ignoring");
             return;
         };
-        debug_assert!(
-            matches!(
-                tx_state.finalization_status,
-                None | Some(NOACheckpointTxStatus::RetryPending)
-                    | Some(NOACheckpointTxStatus::SubmitFailed)
-            ),
-            "mark_submitted: unexpected current status {:?} for tx_ref {:?}",
+        if !matches!(
             tx_state.finalization_status,
-            tx_ref,
-        );
+            None | Some(NOACheckpointTxStatus::RetryPending)
+                | Some(NOACheckpointTxStatus::SubmitFailed)
+        ) {
+            error!(
+                current_status = ?tx_state.finalization_status,
+                ?tx_ref,
+                should_never_happen = true,
+                "mark_submitted: unexpected status transition",
+            );
+            panic!(
+                "mark_submitted: unexpected status {:?} for tx_ref {:?}",
+                tx_state.finalization_status, tx_ref
+            );
+        }
         tx_state.finalization_status = Some(NOACheckpointTxStatus::Pending);
         tx_state.chain_tx_id = Some(chain_tx_id);
     }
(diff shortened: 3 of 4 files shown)


crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -700,6 +725,12 @@ fn verified_dwallet_checkpoint_messages_table_default_config() -> DBOptions {
         .optimize_for_large_values_no_scan(1 << 10)
 }
 
+fn verified_system_checkpoint_messages_table_default_config() -> DBOptions {
```
ycscaly, line 728: @omersadika what is this

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -468,12 +569,24 @@ impl DWalletMPCService {
                 .collect()
         };
 
+        // FIXME(noa-checkpoints): Without a real SuiChainObservation, the entire NOA
```
ycscaly, line 572: @omersadika ?

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -502,6 +617,86 @@ impl DWalletMPCService {
...
+        }
+    }
+
+    /// Dispatch NOA checkpoint resolutions directly to the appropriate handler
+    /// based on the quorum results from the current consensus round.
+    fn dispatch_noa_resolutions(
+        &mut self,
+        agreed_status: &crate::dwallet_mpc::mpc_manager::AgreedStatusUpdate,
+    ) {
+        use ika_types::noa_checkpoint::NOACheckpointResolution;
```
ycscaly: @omersadika remove imports in functions

crates/ika-core/src/authority/authority_per_epoch_store.rs
omersadika: Rocks db config for the table

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
omersadika: For later PR

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -502,6 +617,84 @@ impl DWalletMPCService {
...
+    /// Each handler's `add_signature` silently ignores outputs for tx bytes it doesn't
+    /// own (returns `None`), so broadcasting is correct. The `debug!` log in `add_signature`
+    /// is the only side-effect of sending to the wrong handler.
+    async fn handle_noa_sign_outputs(&mut self) {
+        while let Ok(output) = self.network_owned_address_sign_output_receiver.try_recv() {
+            if let Some(ref mut handler) = self.dwallet_checkpoint_handler {
+                handler.handle_sign_output(output.clone()).await;
+            }
+            if let Some(ref mut handler) = self.system_checkpoint_handler {
+                handler.handle_sign_output(output).await;
```
ycscaly, line 682: why can't we know based on `output` which handler it should use?

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -723,6 +916,32 @@ impl DWalletMPCService {
                 }
             };
 
+            let verified_system_checkpoint_messages = self
+                .epoch_store
+                .next_verified_system_checkpoint_message(self.last_read_consensus_round);
```
ycscaly, line 921: where was this code before? we had system checkpoints too

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -970,45 +1207,105 @@ impl DWalletMPCService {
...
-                        .epoch_store
-                        .insert_pending_dwallet_checkpoint(pending_checkpoint)
+                for (chain, checkpoint_messages) in &mut messages_by_chain {
+                    if checkpoint_messages.is_empty() {
+                        continue;
+                    }
+
+                    // BLS checkpoint path (Sui only for now).
+                    if *chain == CounterpartyChainKind::Sui
+                        && self.protocol_config.bls_checkpoints()
```
ycscaly, line 1217: you already did this, and saved in sui_checkpoint_messages. dont do twice

And self.protocol_config.bls_checkpoints() should be checked outside, not inside the loop

crates/ika-core/src/noa_checkpoints/handler.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub async fn handle_sign_output(&mut self, sign_output: NetworkOwnedAddressSignOutput) {
+        let certified = match self.store.add_signature(
+            &sign_output.message,
+            sign_output.signature,
+            sign_output.curve,
+            sign_output.signature_algorithm,
+        ) {
+            Some(c) => c,
+            None => return,
+        };
```
ycscaly, line 117: You unified the MPC signature accumlation and BLS ones here?
but there was already code that was accumulating it generically in mpc_manager for outputs of a sign session, im confused

crates/ika-core/src/noa_checkpoints/handler.rs
```diff
@@ -0,0 +1,373 @@
...
+                self.initiate_tx_retry_with_context(&tx_ref, &context)
+                    .into_iter()
+                    .collect()
+            }
+        }
+    }
+
+    /// Poll chain for submitted tx status.
+    /// Returns observations to send through consensus.
+    pub async fn poll_chain_status(&mut self) -> Vec<NOACheckpointTxObservation> {
```
ycscaly, line 160: this name is misleading as it doesn't actually do the polling right? it just checks the polled status in the store. maybe `handle_chain_status_changes`

crates/ika-node/src/lib.rs
```diff
@@ -934,17 +964,65 @@ impl IkaNode {
...
+                dyn NOAChainSubmitter<noa_checkpoint::SuiDWalletCheckpoint>,
+            > = Arc::new(LogOnlyChainSubmitter);
+            let system_chain_submitter: Arc<
+                dyn NOAChainSubmitter<noa_checkpoint::SuiSystemCheckpoint>,
+            > = Arc::new(LogOnlyChainSubmitter);
+
+            let dwallet_handler = NOACheckpointHandler::<noa_checkpoint::SuiDWalletCheckpoint>::new(
+                dwallet_chain_submitter,
+                epoch_store.epoch(),
+                vec![],
```
ycscaly, line 1010: @omersadika i dont think it is a good idea to pass it as an empty vec for placeholder. Consider having it as a placeholder option

crates/ika-node/src/lib.rs
```diff
@@ -507,6 +518,16 @@ impl IkaNode {
         ) = watch::channel((0, 0));
         let (uncompleted_requests_sender, uncompleted_requests_receiver) =
             watch::channel((Vec::new(), 0));
+        // Separate flags for each NOA checkpoint kind to avoid race between handlers.
+        let noa_dwallet_finalized = Arc::new(std::sync::atomic::AtomicBool::new(true));
+        let noa_system_finalized = Arc::new(std::sync::atomic::AtomicBool::new(true));
```
ycscaly, line 523: @omersadika why is this atomic bool?

"It is a bit weird. Every other condition in the epoch gate is derived from on-chain coordinator state that the SuiSyncer already reads. The       
  AtomicBool breaks that pattern — it's out-of-band shared mutable state between two independent tasks.
                                                                                                                                                    
  The reason it exists: NOA checkpoint finalization is determined locally through consensus quorum (2f+1 validators observe the tx executed
  on-chain). That result lives in the NOACheckpointHandler's in-memory store, which is owned by DWalletMPCService. The SuiSyncer runs in a          
  completely different task and has no access to that state. The AtomicBool is the bridge.

  But it has issues:

  1. Different pattern from everything else — every other gate condition is read from the Sui coordinator object in one place. This one is a
  side-channel.
  2. Staleness window — the code even acknowledges it in a comment (lib.rs:982: "There is a brief window where the flag could be stale").
  3. Implicit coupling — two independent tasks communicating through shared atomic state rather than an explicit message/channel.

  Alternatives that would stay consistent with the existing pattern:
  - Write NOA checkpoint finalization status to the Sui coordinator object, so the syncer reads it the same way as everything else. But that's
  somewhat circular — you're submitting checkpoints TO Sui and then reading back from Sui whether they landed.
  - Use a watch channel instead of AtomicBool — the service already uses watch channels for other cross-task signaling (e.g.,
  last_session_to_complete_in_current_epoch_sender). That would at least be a consistent mechanism.

  It reads like a pragmatic shortcut to avoid the heavier integration work."

crates/ika-core/src/noa_checkpoints/handler.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// Process new checkpoint messages. Returns sign requests to submit.
+    pub fn handle_new_checkpoint(
+        &mut self,
+        messages: Vec<K::MessageKind>,
+        chain_context: <K::Counterparty as CounterpartyChain>::Context,
+    ) -> Vec<NetworkOwnedAddressSignRequest> {
+        let seq = self.next_sequence_number;
+        self.next_sequence_number += 1;
+
+        let message_groups = K::split_messages(&messages);
```
ycscaly, line 60: I'm not sure this mechanism is fully thought on:
"Right, that's worth questioning. Each tx gets its own independent MPC signature over its own subset of messages. They're not "joined"             
  cryptographically — they'd be separate Sui transactions, each submitted independently.
                                                                                                                                                    
  The on-chain Move contract would need to handle reassembly. Each tx includes (epoch, sequence_number, tx_index, messages) in its bytes, so the
  contract knows which checkpoint and which part it's receiving. It would:                                                                          
                                                            
  1. Accept each partial tx independently
  2. Track which tx_indexes for a given sequence_number have arrived
  3. Consider the checkpoint complete when all parts are in

  **But this raises questions that aren't answered yet:
  - How does the contract know the total number of parts to expect? That's not included in build_tx_bytes.
  - Ordering guarantees — do parts need to arrive in order?**
  - What if some parts finalize and others need retry with different context? Now you have a half-applied checkpoint on-chain.

  Since split_messages is currently a no-op (always one group), none of this is fleshed out. The multi-tx plumbing in the handler/store is ready,
  but the on-chain side and the actual splitting logic are both TODOs. It might be premature scaffolding — or it might be necessary groundwork if
  checkpoints are expected to exceed Sui's tx size limit in practice."

PR #1672, opened 2026-03-18
Fix NOA checkpoint issues, split consensus messages, and fix clippy

omersadika:
## Summary
- **NOA checkpoint infrastructure**: Rename types, consolidate pipeline into direct handler, add chain context agreement, per-tx retry, failure voting, and integration tests
- **Split `InternalSessionsStatusUpdate`**: Break bundled consensus message into 4 independent types (`InternalSessionsStatusUpdate`, `ConsensusGlobalPresignRequest`, `ConsensusNetworkKeyData`, `ConsensusNOAObservation`) to prevent unbounded Vec growth
- **Fix all clippy warnings**: Delete dead code, collapse nested ifs, remove unused parameters, replace unbounded channels with bounded (capacity 1024), add type aliases for complex types, fix module inception

## Test plan
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes clean
- [ ] `cargo test --release -p ika-core dwallet_mpc::integration_tests -- --test-threads=1` — all 39 tests pass
- [ ] `cargo build --release` succeeds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,4 +91,4 @@ ika_config.json
 # TypeDoc
 sdk/typescript/docs
 move_pkg_lock_root.lock
-node-compile-cachedebug_output.txt
+node-compile-cache


crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -902,32 +1006,61 @@ impl AuthorityEpochTables {
...
+                .unwrap_or(0);
+            let mut batch = table.batch();
+            batch.delete_batch(table, [&key])?;
+            if current_size > 0 {
+                batch.insert_batch(
+                    &self.internal_presign_pool_sizes,
+                    [(&size_key, &(current_size - 1))],
+                )?;
+            }
+            batch.write()?;
```
ycscaly, line 1055: @omersadika im not sure this is a good idea

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -590,20 +599,47 @@ impl DWalletMPCService {
...
-            new_key_data.clone(),
-            sui_chain_observation.clone(),
-            self.buffered_noa_observations.clone(),
-        );
+        // Build a batch of consensus transactions.
+        let mut transactions = Vec::new();
+
+        // Slimmed status update (idle + chain observation) when those changed.
+        if idle_status_changed || observation_changed {
+            let status_update = InternalSessionsStatusUpdate::new(
```
ycscaly: So now this `InternalSessionsStatusUpdate` is a pretty bad name, how about `SyncedStateStatusUpdate` or something like that? @omersadika

crates/ika-core/src/authority/authority_per_epoch_store.rs
omersadika: The code was:

 // This shouldn't happen, but handle it gracefully
table.remove(&key)?;

It's the same thing but with batch and more cleaner

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
omersadika: Maybe we can just split sui_chain_observation and then it will be just sync idle status or something like that. What do you think?

PR #1703, opened 2026-05-01
Add ika-solana-signer crate (Rust signer for Solana via Ika dWallets)

Iamknownasfesal:
## Summary
Adds `crates/ika-solana-signer` — a Rust signer that produces Ed25519 signatures for Solana via a pre-provisioned Ika dWallet — and lifts the read-side dWallet helpers it needs out of the CLI into `ika-sui-client`. Verified end-to-end against live Sui + Ika testnet: presign + sign + ed25519 verify under the dWallet's pubkey.

Companion to https://github.com/solana-foundation/solana-keychain (TS package `@solana/keychain-ika`, separate effort) — a downstream crate can adapt `IkaSigner` to the upstream `SolanaSigner` trait without depending on this workspace's git/patch graph.

## Public surface
```rust
pub struct IkaSigner { /* ... */ }

impl IkaSigner {
    pub async fn create(cfg: IkaSignerConfig) -> Result<Self, IkaSignerError>;
    pub fn pubkey(&self) -> [u8; 32];
    pub async fn sign_message(&self, msg: &[u8]) -> Result<[u8; 64], IkaSignerError>;
    pub async fn is_available(&self) -> bool;
}

pub enum SecretShareSource {
    Bytes(Vec<u8>),
    OnChainEncrypted { decryption_key: Vec<u8>, encryption_key_address: SuiAddress },
}

pub enum PresignMode {
    PerSignGlobal,                  // fresh global presign per call (stateless)
    SingleProvided(ObjectID),       // single-shot caller-supplied verified cap
}
```

## Commits
1. `Lift read-side dWallet signer helpers into ika-sui-client` — moves `fetch_dwallet_metadata`, `fetch_presign_output`, `find_sign_session_id`, `poll_sign_session`, `fetch_encrypted_share_for_dwallet`, `get_network_key_info_for`, `find_ika_coin`, etc. into a new `ika_sui_client::dwallet_signer` module so both the CLI and this new crate consume them via the same code path.
2. `Add ika-solana-signer crate` — the signer itself + tx-builder helpers (`request_sign_tx_with_signer`, `request_global_presign_tx_with_signer`, `request_imported_key_sign_tx_with_signer`) that take lower-level inputs than the CLI's `WalletContext`-based wrappers.
3. `Fix sign-session lookup + completion detection in dwallet_signer` — three real correctness bugs (see below) caught by the new integration test.
4. `Route ika-solana-signer move calls through V2 package + add testnet test` — V2 package routing + the live integration test.

## Bugs fixed (caught by the new testnet integration test)
1. **V1 → V2 package routing.** Tx builders always called through `ika_dwallet_2pc_mpc_package_id` (V1). The on-chain coordinator on testnet is at `version == 2`; V1 calls abort inside `coordinator::inner_mut` with `EWrongInnerVersion` because that function asserts `self.version == VERSION` and V1's `VERSION` is `1`. Now resolves V2 from `IkaNetworkConfig` and falls back to V1 only when V2 is unset.
2. **Wrong sign-session ID.** `find_sign_session_id` returned the outer `session_object_id` from `SignRequestEvent`, which is the transient SessionManager wrapper record (cleaned up almost immediately after dispatch). The actual `SignSession` object — the one that flips to `Completed { signature }` / `NetworkRejected` — is `event_data.sign_id`. Polling the wrong ID timed out at 300s with "Object not found" once the wrapper was cleaned. (This also affected the existing CLI sign path on networks that GC sessions quickly.)
3. **Wrong enum-variant detection.** `poll_sign_session` looked for `state.variant`, but Sui's parsedJson renders Move enums as `{ "type": "<EnumStructTag>", "fields": { ...active variant inner fields... } }` with no explicit variant tag. Active variant is inferred from which inner fields are present. For `SignState`, only `Completed` is unambiguously detectable (via `state.fields.signature`); `Requested` and `NetworkRejected` both render as empty `fields`. Left a `TODO` to detect rejection via the validator-emitted rejection event in a follow-up.

The TS SDK didn't hit any of these because it routes V2 automatically and decodes events / `SignSession` via BCS (where the variant tag is explicit).

## Verification
- `cargo build --release --all-targets` — clean.
- `cargo clippy --release --all-targets -p ika-solana-signer -p ika-sui-client --no-deps` — clean.
- `cargo test --release -p ika-solana-signer --test testnet_integration` against live Sui + Ika testnet — passes in ~85s (presign + sign + ed25519 verify). Test is gated on `IKA_DWALLET_ID` so it auto-skips for contributors without a provisioned dWallet.

## Out of scope / follow-ups
- The on-chain encrypted share flow (`SecretShareSource::OnChainEncrypted`) is wired up but not exercised by the integration test (the test uses raw `Bytes`). Worth a follow-up that DKGs a fresh dWallet via the SDK and signs through the encrypted-share path.
- An `examples/` devnet binary that submits a real Solana tx end-to-end. Not strictly needed to merge; useful smoke test for downstream consumers.
- `NetworkRejected` detection via the validator's rejection event (currently indistinguishable from `Requested` via the SignSession object alone — see bug 3 above).
- Validator-tx dry-run / gas estimation against testnet currently goes through the `[TEMP]` skip in #1702. The proper fix is bumping the workspace Sui pin from `testnet-v1.67.1` to `testnet-v1.71.0`, which is a separate refactor (RustCrypto pre-release pinning needs work).

## Test plan
- [x] `cargo build --release --all-targets`
- [x] `cargo clippy --release --all-targets -p ika-solana-signer -p ika-sui-client --no-deps`
- [x] `cargo test --release -p ika-solana-signer --test testnet_integration` against live testnet (full presign + sign + verify cycle, ~85s)
- [ ] CI runs the existing `crates/ika/tests/integration/test_all_combinations.sh` to confirm the helper-lift didn't break the CLI Sign path


Iamknownasfesal:
Closing as duplicate — superseded by the existing stacked pair #1700 (refactor: lift helpers → main) + #1701 (this crate, stacked on #1700). I opened this without realizing the stack existed; #1700/#1701 both rebased forward onto the new mainnet-v1.70.2 main, all the bug fixes from this PR already live on the same `feat/ika-solana-signer` branch and will land via #1701.

PR #1707, opened 2026-05-12
Bump cryptography-private to 9d35fa76 (latest main)

ycscaly:
## Summary

Bumps all 7 `dwalletlabs-com/cryptography-private` workspace deps from `babbb483` → `9d35fa76` and ports ika to the new MPC/curve/PVSS surface. Single-pin migration (no dual-pin backcompat). **Local-network only — wire-format incompatible with mainnet** (see "Mainnet incompatibility" below).

The work is captured in two plan docs that document the as-built decisions:
- `docs/plan-bump-crypto-private-to-main.md` — 12-phase port of the upstream API surface
- `docs/plan-update-crypto-latest.md` — followup plan to restore mainnet wire compatibility

## What changed

**Crypto pin**
- All 7 cryptography-private deps at git rev `9d35fa76` (`mpc`, `proof`, `class_groups`, `commitment`, `twopc_mpc`, `group`, `homomorphic_encryption`).
- Walked the upstream rc cascade across all signed/AEAD primitives — final `[patch.crates-io]` set pins `crypto-bigint` to ycscaly fork `8aabcee5` plus a chain of `rc.*` versions for sha2/hmac/digest/sha3/keccak/crypto-common/aead/cipher/aes/ctr/aes-gcm/chacha20/chacha20poly1305/poly1305/polyval/universal-hash/ghash/pkcs8/der/sec1/spki/elliptic-curve/ecdsa to match upstream.

**Protocol changes** (Phase 4a–g, 5, 6, 9 in plan)
- Network DKG is now 7 rounds (was 4) — upstream activated the threshold-encryption-to-sharing sub-protocol. Integration-test assertions bumped accordingly (`malicious_behavior`, `missing_network_key`, `message_before_event`, `threshold_not_reached`).
- New `SignData<SignMessage, VerifiedSignData>` enum (`Unverified` / `Verified` / `ToBeEmulated`) wired end-to-end.
- NOA emulator deleted (≈700 lines from `network_owned_address_sign_dkg_emulation.rs`); the upstream `dkg::Protocol::threshold_dkg_output` is now used directly (per-curve concrete builders).
- `<P, D>` two-generic pattern with `presign::Protocol<DKGProtocol = D>` bound in sign code paths.
- `verify_partial_signature` returns `Result<P::VerifiedSignData>` so the verified state is threaded through.
- `VersionedDwalletDKGPublicOutput::V2 { public_key_bytes, dkg_output }` wrapper for NOA-DKG outputs so the downstream sign path can decode them uniformly.

**PVSS HPKE validator keys** (originally scoped out by the bump plan; bundled here so local network actually works)
- 3 new per-curve PVSS HPKE keypair types — `Secp256k1PvssEncryptionKeyAndProof`, `Secp256r1PvssEncryptionKeyAndProof`, `RistrettoPvssEncryptionKeyAndProof`. New combined struct `ValidatorEncryptionKeysAndProofs { class_groups, secp256k1_pvss, secp256r1_pvss, ristretto_pvss }`.
- `Committee` and `NetworkMetadata` gain 3 sibling `HashMap<AuthorityName, *PvssEncryptionKeyAndProof>` fields (per the user's "3 sibling fields" choice).
- Per-curve PVSS keypairs derived from existing `RootSeed` via 3 domain-separated child RNGs — never rotates (per user's "Derived from RootSeed" choice).
- Validator-side: `ClassGroupsKeyPairAndProof::from_seed` now also generates 3 PVSS keypairs via upstream's `class_groups::publicly_verifiable_secret_sharing::small_prime::encryption::generate_and_prove_encryption_keypair`.
- Wired through `network_dkg_v2_public_input` and all 3 reconfiguration `PublicInput` constructors (UPCOMING committee's PVSS keys per upstream's `participating_parties_access_structure: upcoming_access_structure`).

**WASM**
- `sdk/ika-wasm/Cargo.lock` updated to align with the upstream rc pin set (`crypto-primes 0.7.0 → 0.7.0-pre.3`, `crypto-bigint 0.7.3 → 0.7.0-rc.9` so the patched git rev applies).

## ⚠️ Mainnet wire-format incompatibility

Per the user's design choice for this branch's local-network scope, the Move-side `class_groups_public_key_and_proof` byte field is **overloaded** to carry the combined `ValidatorEncryptionKeysAndProofs` (class groups + 3 PVSS) instead of the bare `ClassGroupsEncryptionKeyAndProof` mainnet expects. Validators on either side of this serialization boundary **cannot decode each other's records**. The followup `docs/plan-update-crypto-latest.md` covers the dual-pin / `_backward_compatible` path that restores mainnet compatibility.

The incompatibility is loudly documented in:
- `ValidatorEncryptionKeysAndProofs` rustdoc
- Every serialization site (`validator_initialization_config.rs`, `validator_commands.rs`)
- Both deserialization sites (`epoch_start_system.rs`, `sui_syncer.rs`)
- The self-check at `dwallet_mpc_service.rs:2194`

## Test plan

Rust workspace
- [x] `cargo build --release` — clean
- [x] `cargo clippy --all-targets --all-features` — clean
- [x] `cargo test --release -p ika-core --lib` — passes (long-running; the 5 tests with bumped DKG-round-count assertions all pass cleanly in isolation; they only flake under heavy CPU contention from concurrent builds)

SDK / WASM
- [x] `sdk/ika-wasm` builds fresh
- [x] `sdk/typescript` builds, BCS bindings regenerated, ts-codegen happy
- [x] Local-network bring-up works (sui 1.67.1 + ika): network DKG completes round 7, `Network key data has been agreed upon` reached, presigns fill pools, internal MPC outputs reach quorum, dwallet DKG outputs are accepted by the chain (`status=Success`)
- [x] SDK integration tests partially run on this branch (`global-presign` 5/5 passed, `dwallet-creation - Ed25519 zero trust` passed), then hit a chain epoch-advance deadlock — see below

## Pre-existing chain epoch-advance deadlock (NOT caused by this PR)

After ~17 minutes of sustained SDK load (multiple test files, ~12+ dwallet DKGs), the chain enters a state where it cannot advance epochs:

```
last_user_initiated_session_to_complete_in_current_epoch: 7   (the cap)
locked: true
user_completed: 12       (already past the cap!)
user_next: 29
sys_completed: 6 == sys_started: 6
```

Move's `all_current_epoch_sessions_completed()` in `contracts/ika_dwallet_2pc_mpc/sources/sessions_manager.move:613` requires **strict equality** `user_sessions_completed_count == last_user_initiated_session_to_complete_in_current_epoch`. Once `completed` overshoots `cap`, the equality never holds again, and `update_last_user_initiated_session_to_complete_in_current_epoch` is gated on `!locked` (line 630) — so the cap can't bump itself.

Validators correctly cap their work at `session_seq ≤ 7` (`mpc_manager.rs:1217`), so any new dwallet DKG with `seq > 7` sits unprocessed, SDK times out waiting for state transitions, all subsequent dwallet-creation tests fail.

**Verified: this is pre-existing on `dev`.** Running the same SDK sweep against fresh dev localnet (sui 1.67.1, identical epoch cadence) reproduces the deadlock — in fact dev hits it harder: `cap=1, locked=true, user_completed=10`, and 8/8 SDK test files failed (dev's `global-presign` failed too, while this PR's branch passed it 5/5).

So the SDK test failures are not a regression introduced by this PR. They expose a pre-existing protocol issue in the Move-side epoch-advance condition that needs a separate fix (likely `>=` instead of `==`, or removing the lock gate on the cap bump).


crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -263,12 +263,18 @@ pub(crate) fn advance_network_dkg_v2(
     class_groups_decryption_key: ClassGroupsDecryptionKey,
     rng: &mut ChaCha20Rng,
 ) -> DwalletMPCResult<GuaranteedOutputDeliveryRoundResult> {
+    // Phase 4 of crypto bump: dkg::Party's PrivateInput is now a struct wrapping the
+    // CRT-decryption-key array (see decentralized_party/dkg.rs:52); previously it was the
+    // bare array. Wrap our existing class_groups_decryption_key value into the struct.
```
ycscaly: Remove useless comment

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -377,35 +393,128 @@ fn compute_network_owned_address_dkg_and_public_key(
...
-        DWalletCurve::Secp256k1 => bcs::to_bytes(secp256k1_protocol_public_parameters)?,
-        DWalletCurve::Secp256r1 => bcs::to_bytes(secp256r1_protocol_public_parameters)?,
-        DWalletCurve::Ristretto => bcs::to_bytes(ristretto_protocol_public_parameters)?,
-        DWalletCurve::Curve25519 => bcs::to_bytes(curve25519_protocol_public_parameters)?,
+    // Phase 4f of crypto bump: ika's NOA emulator was deleted. Use upstream's native
+    // `<DKGProtocol as dkg::Protocol>::threshold_dkg_output(pp, session_id)` directly to
+    // produce the decentralized-party DKG output for threshold (no centralized party) mode.
+    // The session id is derived from network_key_id + curve (curve-specific, sig-algo
+    // independent) per the surviving session_id helper.
+    use twopc_mpc::dkg::Protocol as DkgProtocol;
```
ycscaly: The rules say that you are not alllowed to import inside a function

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -377,35 +393,128 @@ fn compute_network_owned_address_dkg_and_public_key(
...
+
+    let _ = (party_id, access_structure);
```
ycscaly: What is this???????? So sloppy@@@!!!!

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -25,23 +39,62 @@ pub type SingleEncryptionKeyAndProof = (
...
+/// PVSS HPKE secret decryption key, secp256k1 plaintext space.
+///
+/// Sized by the secp256k1 fundamental discriminant; private to the validator
+/// — never sent on the wire. Mirrored by `Secp256r1PvssDecryptionKey` and
+/// `RistrettoPvssDecryptionKey` for the other two PVSS curves.
+pub type Secp256k1PvssDecryptionKey = Uint<{ SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS }>;
+/// PVSS HPKE secret decryption key, secp256r1 plaintext space.
+pub type Secp256r1PvssDecryptionKey = Uint<{ SECP256R1_FUNDAMENTAL_DISCRIMINANT_LIMBS }>;
+/// PVSS HPKE secret decryption key, ristretto plaintext space.
+pub type RistrettoPvssDecryptionKey = Uint<{ RISTRETTO_FUNDAMENTAL_DISCRIMINANT_LIMBS }>;
```
ycscaly: There are already these types exported in class_groups, use them..

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -25,23 +39,62 @@ pub type SingleEncryptionKeyAndProof = (
...
     encryption_key_and_proof: ClassGroupsEncryptionKeyAndProof,
+
+    secp256k1_pvss_decryption_key: Secp256k1PvssDecryptionKey,
+    secp256k1_pvss_encryption_key_and_proof: Secp256k1PvssEncryptionKeyAndProof,
+
+    secp256r1_pvss_decryption_key: Secp256r1PvssDecryptionKey,
+    secp256r1_pvss_encryption_key_and_proof: Secp256r1PvssEncryptionKeyAndProof,
+
+    ristretto_pvss_decryption_key: RistrettoPvssDecryptionKey,
+    ristretto_pvss_encryption_key_and_proof: RistrettoPvssEncryptionKeyAndProof,
```
ycscaly: I dont want you to add these as members to this struct, as it will break serialization of existing keys; the better approach is to add a new struct with these fields, that will also take the (unchanged)ClassGroupsKeyAndProof as a member. This will ease our backward compat

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -363,47 +381,42 @@ pub(crate) struct AllCurvesNetworkOwnedAddressDkgData {
...
+    D: twopc_mpc::dkg::Protocol,
+{
+    let session_identifier =
+        network_owned_address_sign_dkg_session_identifier(network_key_id, curve);
+    let session_id = CommitmentSizedNumber::from_le_slice(&session_identifier.into_bytes());
+
+    let out = D::threshold_dkg_output(protocol_public_parameters, session_id).map_err(|e| {
+        DwalletMPCError::InternalError(format!("threshold_dkg_output {curve:?}: {e}"))
+    })?;
+    let raw_bytes = bcs::to_bytes(&out)?;
```
ycscaly: Member names! You are breaking convention! And use shadowing!

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -363,47 +381,42 @@ pub(crate) struct AllCurvesNetworkOwnedAddressDkgData {
...
+    let public_key = public_key_from_decentralized_dkg_output_by_curve_v2(curve, &raw_bytes)
+        .map_err(|e| {
+            DwalletMPCError::InternalError(format!("public_key extract {curve:?}: {e}"))
+        })?;
+    // Wrap in VersionedDwalletDKGPublicOutput::V2 so downstream sign-input
+    // construction (which decodes via VersionedDwalletDKGPublicOutput) can
+    // round-trip through the standard user-driven sign path.
+    let dkg_output = bcs::to_bytes(&VersionedDwalletDKGPublicOutput::V2 {
+        public_key_bytes: public_key.clone(),
+        dkg_output: raw_bytes,
```
ycscaly: This is so ugly, when you use a variable to instantiate a member of struct, name it like the member! Write in claude.md

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
```diff
@@ -51,6 +51,17 @@ impl ReconfigurationPartyPublicInputGenerator for ReconfigurationParty {
...
+        // Per-curve PVSS HPKE encryption keys + proofs. Upstream's
+        // `new_from_dkg_output` / `new_from_reconfiguration_output` accept a
+        // single set of PVSS HashMaps keyed by `PartyID`; their internal use
+        // (`participating_parties_access_structure: upcoming_access_structure`
+        // in `2pc-mpc/src/decentralized_party/reconfiguration.rs:401, 689`)
+        // shows they correspond to the UPCOMING committee — the dealers send
+        // ciphertexts encrypted under each upcoming participating party's PVSS
+        // public key.
+        let upcoming_validators_keys =
+            crate::dwallet_mpc::get_validators_encryption_keys_by_party_id(&upcoming_committee)?;
```
ycscaly: Why, in the name of god, would you use such non-descriptive 'encryption_keys' field name and function names for the HPKE encryption keys where there are also the class group encryption keys? SPECIFY WHICH TYPE OF ENCRYPTION KEYS IT IS. DO IT EVERYWHERE

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -343,181 +344,444 @@ impl SignPublicInputByProtocol {
...
+                    )))
+                })?,
+            };
+            presign_value = bcs::from_bytes(&presign_bytes).map_err(|e| {
+                DwalletMPCError::BcsError(bcs::Error::Custom(format!(
+                    "Failed to deserialize presign V2: {e}"
+                )))
+            })?;
+        }
+    };
```
ycscaly, line 455: Why dont you use the decode function here?

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
```diff
@@ -270,51 +265,23 @@ pub(crate) fn session_input_from_request(
...
 
-            // Get the stored network-owned-address sign DKG output for the specific curve
-            // (contains both centralized and decentralized DKG).
-            // This is pre-computed during network key construction.
+            // Phase 4f of crypto bump: ika no longer pre-emulates the centralized party's
+            // partial signature off-Rayon. The stored output is now just the decentralized
+            // DKG output bytes (network_dkg.rs writes it via `threshold_dkg_output`). Pass
+            // empty `message_centralized_signature` to signal NOA → SignPublicInputByProtocol
+            // dispatches `SignData::ToBeEmulated` so the upstream protocol emulates inside
+            // its Rayon-scheduled advance.
```
ycscaly: Remove these kind of useless comments, so what if we no longer do something, you dont need to document what we dont do

crates/ika-core/src/dwallet_mpc/mod.rs
```diff
@@ -79,18 +82,68 @@ pub(crate) fn authority_name_to_party_id_from_committee(
...
+/// PVSS curve in the future requires touching only this struct + the
+/// extractor below, not every helper signature on the call path.
+#[derive(Clone, Debug)]
+pub(crate) struct ValidatorsEncryptionKeysByPartyId {
+    pub class_groups: HashMap<PartyID, ClassGroupsEncryptionKeyAndProof>,
+    pub secp256k1_pvss: HashMap<PartyID, Secp256k1PvssEncryptionKeyAndProof>,
+    pub secp256r1_pvss: HashMap<PartyID, Secp256r1PvssEncryptionKeyAndProof>,
+    pub ristretto_pvss: HashMap<PartyID, RistrettoPvssEncryptionKeyAndProof>,
+}
+
```
ycscaly, line 102: Im not sure why we need this new struct

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
ycscaly: You should keep the doc that says why we pass empty message from user however, as that is detailing what we are doing now.

ycscaly (commented):
Also write in claude.md to never write comments saying which part of the plan it is

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ycscaly: Done — module doc/comments rewritten as part of the NOA-file rename (compute_noa_dkg<D> moved into `network_owned_address_sign_dkg.rs`). 396af2647e

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ycscaly: Confirmed clean — `grep -nE '^use |fn ' network_dkg.rs` shows all `use` declarations at file top; no imports inside any function body. 396af2647e

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ycscaly: Cleaned up. 396af2647e

crates/dwallet-classgroups-types/src/lib.rs
ycscaly: Checked — `class_groups` exports `Secp256k1DecryptionKey = DecryptionKey<...>` (a wrapper struct for the full FHE decryption-key API), not the raw `Uint<FUNDAMENTAL_DISCRIMINANT_LIMBS>` that `generate_and_prove_encryption_keypair` returns (line 122 of upstream `small_prime/encryption.rs`). No upstream `*PvssDecryptionKey` alias for the raw return type exists, so the local aliases stay with a clarifying comment. 396af2647e

crates/dwallet-classgroups-types/src/lib.rs
ycscaly: Done — `ClassGroupsKeyPairAndProof` reverted to its pre-PR 2-field shape (wire format preserved for existing keys); new `ClassGroupsAndPvssKeyPairAndProof` wrapper holds it as a `class_groups` member + the 6 PVSS HPKE fields. `from_seed` lives on the wrapper and calls `ClassGroupsKeyPairAndProof::from_seed` first. 5 call sites updated. 396af2647e

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ycscaly: Done — `compute_noa_dkg<D>` now uses field-name-matching locals (`dkg_output`, `public_key`) and shadowing for the transform chain (`let dkg_output = D::threshold_dkg_output(...)?; let dkg_output = bcs::to_bytes(&dkg_output)?;`). Lives in `network_owned_address_sign_dkg.rs` post-move. 396af2647e

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
ycscaly: Done — same fix as the sibling comment, plus added a CLAUDE.md rule under Rust code style ("When initializing a struct with locals, name the local like the field; use struct-init shorthand or shadowing"). 396af2647e

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs
ycscaly: Done — `extract_encryption_keys_from_committee` → `extract_class_groups_encryption_keys_from_committee` (function body operates only on class-groups); local var `upcoming_validators_keys` → `upcoming_validators_pvss_hpke_keys_by_party_id`. The existing `*_per_crt_prime_and_proofs` variables already disambiguate as class-groups via their suffix and were left alone. The bundle function `get_validators_encryption_keys_by_party_id` also got renamed to `get_validator_mpc_keys_by_party_id` (see the mod.rs:102 thread). 396af2647e

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: Done — `build_secp256k1_ecdsa_sign_public_input` refactored (~120 → ~50 lines). Extracted a new `decode_ecdsa_dkg<D>` helper for the DKG-only decode; V2 path now uses the existing `decode_ecdsa_dkg_and_presign<D, P>` + `decode_ecdsa_sign_data<P>` like the other four builders. V1-only raw-presign branch stays inline (no other curve has V1 presign history; the helper would have to be secp256k1-shaped to absorb it). 396af2647e

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
ycscaly: Done — `mpc_session/input.rs:268-273` comment surgery: dropped the "we no longer pre-emulate" cruft, kept only the WHY-empty-message rationale per your sibling comment. 396af2647e

crates/ika-core/src/dwallet_mpc/mod.rs
ycscaly: Renamed `ValidatorsEncryptionKeysByPartyId` → `ValidatorMpcKeysByPartyId` (and the function `get_validators_encryption_keys_by_party_id` → `get_validator_mpc_keys_by_party_id`). Docstring rewritten to lead with the real argument: one parameter through the network-DKG / reconfiguration call chain instead of four separate `HashMap` arguments, and adding a fifth PVSS curve becomes a one-field change rather than threading a new `HashMap` through six call sites. 396af2647e

crates/ika-core/src/dwallet_mpc/mpc_session/input.rs
ycscaly: Kept — the surviving sentence reads: "Pass an empty `message_centralized_signature` so `SignPublicInputByProtocol` dispatches `SignData::ToBeEmulated` — the upstream sign protocol then emulates the centralized party's partial signature inside its Rayon-scheduled advance." 396af2647e

ycscaly:
## Review feedback addressed — 396af2647e

Replies posted on all 12 inline threads. Summary of changes:

**Architectural / breakage-preserving**
- **Struct split** (lib.rs:80): `ClassGroupsKeyPairAndProof` reverted to its pre-PR 2-field shape so existing serialized keys round-trip; new `ClassGroupsAndPvssKeyPairAndProof` wrapper holds it as a `class_groups` member + the 6 PVSS HPKE fields. `from_seed` lives on the wrapper; 5 call sites updated.
- **NOA module rename**: `network_owned_address_sign_dkg_emulation.rs` → `network_owned_address_sign_dkg.rs` (no longer an emulator); `compute_noa_dkg<D>` moved into it.

**Naming / clarity**
- **HPKE disambiguation**: `extract_encryption_keys_from_committee` → `extract_class_groups_encryption_keys_from_committee`; `upcoming_validators_keys` → `upcoming_validators_pvss_hpke_keys_by_party_id`.
- **Bundle struct**: `ValidatorsEncryptionKeysByPartyId` → `ValidatorMpcKeysByPartyId` (and matching fn rename), docstring rewritten with the real argument (one parameter through call chain instead of four `HashMap`s; adding a 5th PVSS curve is a one-field change).
- **Member-naming convention**: `compute_noa_dkg<D>` now uses field-name-matching locals + shadowing (`let dkg_output = ...; let dkg_output = bcs::to_bytes(&dkg_output)?;`).

**Refactor**
- `build_secp256k1_ecdsa_sign_public_input` ~120 → ~50 lines. Extracted `decode_ecdsa_dkg<D>`; V2 path uses the existing `decode_ecdsa_dkg_and_presign<D, P>` + `decode_ecdsa_sign_data<P>` like the other 4 builders. V1-only raw-presign branch stays inline (no other curve has V1 history).

**Comments cleanup**
- Dropped all "Phase X of crypto bump" plan-tag comments (6 sites). Kept the technical content.
- `mpc_session/input.rs`: surgery per the dual review — dropped "we used to X" cruft, kept the WHY-empty-message rationale.
- Added two CLAUDE.md rules: never reference plan/phase names in comments, and name struct-init locals after the field.

**New tests** (both pass in release mode)
- `validator_encryption_keys_and_proofs_round_trips_through_bcs` — BCS round-trip through the wire-format-overloaded payload.
- `class_groups_and_pvss_key_pair_from_seed_is_deterministic` — same seed → byte-identical struct.

**Magic-number lift**
- `utils::EXPECTED_NETWORK_DKG_ROUND_COUNT = 7`; 5 integration-test sites now derive their assertion from it.

**Re item 12** (`There are already these types exported in class_groups`): checked — `class_groups` exports `Secp256k1DecryptionKey = DecryptionKey<...>` (wrapper struct for the full FHE decryption-key API), not the raw `Uint<FUNDAMENTAL_DISCRIMINANT_LIMBS>` that `generate_and_prove_encryption_keypair` returns. Kept local aliases with a clarifying comment.

**Verification**
- `cargo build --release` clean
- `cargo clippy --release --all-targets --all-features` clean (13 pre-existing warnings in `ika/src/dwallet_commands.rs` only)
- `cargo test --release -p dwallet-classgroups-types` — 2/2 pass (BCS round-trip + determinism, 41.8s)
- `cargo test --release -p ika-core --lib -- --test-threads=1 dwallet_mpc::integration_tests::{malicious_behavior,missing_network_key,message_before_event,threshold_not_reached}` — 5/5 pass (12m 47s)

PR #1709, opened 2026-05-18
sui_executor: cache gas ObjectRef from prev tx effects to avoid versi…

ycscaly:


Iamknownasfesal:
i think this PR will not be needed or can be reshaped once balance accumulators are open in sui mainnet, you will be able to pay gas through the accumulator and not via coins, thus not needing to manage gas coins or sponsorship

PR #1714, opened 2026-05-20
feat: Fast Schnorr (VSS Schnorr) support

ycscaly:
## Fast Schnorr (VSS) support, integrated onto the upgrade-test branch

**Base is now `feat/ika-upgrade-test`** (was `dev`). This branch is the VSS
crypto work merged with the cross-binary upgrade-test branch, so the diff here
is just the Fast Schnorr delta — not the upgrade harness or #1721, which live
in the base.

### What's in the VSS delta
- **Crypto**: `SchnorrkelSubstrate` + the VSS signature algorithms (TaprootVSS,
  EdDSAVSS, SchnorrkelSubstrateVSS), the `b2a01b3c` cryptography-private pin,
  the internal VSS presign pools, `fast_schnorr_supported` at protocol v4.
- **Wire format (integrated, not either-or)**: on-chain stays the bare
  mainnet-v1.1.8 `ClassGroupsEncryptionKeyAndProof` shape (so the v118 upgrade
  rehearsal boots against a literal 1.1.8 binary); the off-chain v1 bundle
  carries the **5th** Fast Schnorr VSS HPKE key, threaded end-to-end through the
  producer (`ValidatorMPCSecrets`), `OffChainCommitteeBundles`, the assembler
  (required / fail-closed), `Committee::new`, and the shape-tolerant
  `decode_validator_encryption_keys` (extended to 5 fields).

### Integration choices
- Crypto/VSS comes from fast-schnorr; the off-chain / consensus-message /
  per-epoch-store subsystem stays the upgrade-test branch's evolved design
  (handoff-cert convergence, unified consensus messages, the v3→v4 fixes). The
  off-chain consumer files are byte-identical to the base.

### Verified
- `ika-node` + `ika-upgrade-test` + `ika-test-cluster` + all test targets build
  clean; `ika-validator`/`ika-notifier`/`ika` binaries build.
- `ika-protocol-config` + `dwallet-classgroups-types` unit tests green
  (snapshots regenerated for the carried-over `grace_rounds` fields).
- Bidirectional loss review: no upgrade-test logic dropped (harness 0-diff,
  restorations present); no fast-schnorr VSS logic dropped (VSS key flows
  off-chain → committee, pools/tables/verification all present).

### Not yet validated
The build + unit tests prove structure and wiring, **not** VSS signing
end-to-end (key derivation → off-chain propagation → committee verification →
sign). That needs the VSS sign suite + the v118 rehearsal.


crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -131,6 +133,14 @@ pub struct ClassGroupsAndPvssKeyPairAndProof {
 
     ristretto_pvss_decryption_key: RistrettoPvssDecryptionKey,
     ristretto_pvss_encryption_key_and_proof: RistrettoPvssEncryptionKeyAndProof,
+
+    /// Fast Schnorr (VSS) HPKE encryption public key (curve25519) + UC proof. The
+    /// matching secret key is NOT stored here (it would force `curve25519::Scalar`
```
ycscaly: WHAT ARE YOU TALKING ABOUT?!? OF COURSE YOU CANT STORE SECRET KEYS OF OTHER VALIDATORS?!???

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -131,6 +133,14 @@ pub struct ClassGroupsAndPvssKeyPairAndProof {
...
+    /// serializability onto this `Serialize` struct); instead it is re-derived
+    /// deterministically from the same `RootSeed` where the validator's private
+    /// MPC key data is assembled (see `vss_schnorr_hpke_secret_key_from_seed`).
+    /// One curve25519 keypair serves all VSS signing curves.
+    vss_schnorr_hpke_public_key_and_proof: VssSchnorrHpkeEncryptionKeyAndProof,
```
ycscaly: Dont say in member names or struct types that it is schnorr specific; it is just VSS HPKE keys.

crates/dwallet-rng/src/lib.rs
```diff
@@ -151,6 +151,28 @@ impl RootSeed {
         ChaCha20Rng::from_seed(self.pvss_ristretto_decryption_key_seed())
     }
 
+    fn vss_schnorr_hpke_encryption_key_seed(&self) -> [u8; Self::SEED_LENGTH] {
+        let mut transcript = Transcript::new(b"Fast Schnorr VSS HPKE Encryption Key Seed");
```
ycscaly: Its a bit weird to say the seed is for the encryption key as its public; you can say its for the secret key and the proof

crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -773,6 +808,16 @@ pub struct AuthorityEpochTables {
     #[default_options_override_fn = "internal_presign_pool_table_default_config"]
     internal_presign_pool_schnorrkel_substrate:
         DBMap<(ObjectID, u64), (SessionIdentifier, Vec<Vec<u8>>)>,
```
ycscaly: Write here what the map structure is (types of keys, values and meaning)

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -192,14 +202,36 @@ impl ProtocolCryptographicData {
...
+                    // presign nonce data, instead of the AHE decryption key shares.
+                    let reconfiguration_public_output = decryption_key_shares
+                        .get_network_encryption_key_public_data(dwallet_network_encryption_key_id)?
+                        .latest_network_reconfiguration_public_output()
+                        .ok_or_else(|| {
+                            DwalletMPCError::InvalidInput(
+                                "Fast Schnorr (VSS) sign requires a reconfigured network key"
+                                    .to_string(),
+                            )
+                        })?;
```
ycscaly: Not true, you can use dkg output if no reconfiguration happened.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/presign.rs
```diff
@@ -141,20 +173,125 @@ impl PresignAdvanceRequestByProtocol {
...
+                curve25519::GroupElement::new(*encryption_key_value, &public_parameters).map_err(
+                    |e| {
+                        DwalletMPCError::InvalidInput(format!(
+                            "invalid VSS HPKE encryption key for party {party_id}: {e:?}"
+                        ))
+                    },
+                )?;
+            Ok((*party_id, (encryption_key, proof.clone())))
+        })
+        .collect::<DwalletMPCResult<HashMap<PartyID, _>>>()?;
```
ycscaly: NO! A single malicious validator can dos the network if you do it this way1!!! There is handle_messages_asynchronous for this, but in any case it would be better to put this logic as part of the cryptography

ycscaly: We can move this entire function there, open a PR against main, put it in the right place with non-ika doc and well strcutured

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -322,6 +394,19 @@ impl DWalletDKGAndSignAdvanceRequestByProtocol {
...
+            // Fast Schnorr (VSS) does not support the combined DKG-and-sign fast
+            // path (deferred by decision; the upstream combined VSS party exists
+            // but is not wired here). Such requests are rejected at construction
+            // (`dwallet_dkg_and_sign_protocol_data`), so this is unreachable; kept
+            // exhaustive with a clear error as defense-in-depth.
+            DWalletSignatureAlgorithm::TaprootVSS
+            | DWalletSignatureAlgorithm::EdDSAVSS
+            | DWalletSignatureAlgorithm::SchnorrkelSubstrateVSS => {
+                return Err(DwalletMPCError::InvalidInput(format!(
+                    "Fast Schnorr (VSS) algorithm {protocol} is not supported for \
```
ycscaly: Are we sure it isnt supported? Im pretty sure it is

crates/ika-core/src/authority/authority_per_epoch_store.rs
```diff
@@ -319,6 +320,23 @@ pub trait AuthorityPerEpochStoreTrait: Sync + Send + 'static {
     /// Checks if a presign has already been used.
     fn is_presign_used(&self, presign_session_id: SessionIdentifier) -> IkaResult<bool>;
 
+    /// Persists a Fast Schnorr (VSS) presign's private output (`bcs(Vec<PrivatePresignOutput>)`)
+    /// keyed by the presign session id, so the later sign can recover the nonce
+    /// shares. Self-prunes at epoch rotation. VSS sessions only.
+    fn store_presign_private_output(
+        &self,
+        presign_session_id: CommitmentSizedNumber,
+        private_output: Vec<u8>,
```
ycscaly, line 347: Cannot be one private output per session; either you keep the blending index in the keys here and take it as argument, or it has to be a vector of private outputs right?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -410,6 +531,438 @@ impl SignPublicInputByProtocol {
...
+        DwalletMPCError::BcsError(bcs::Error::Custom(format!(
+            "failed to deserialize VSS presign output: {e}"
+        )))
+    })?;
+    let VersionedPresignOutput::V2(inner) = versioned else {
+        return Err(DwalletMPCError::InvalidInput(
+            "Fast Schnorr (VSS) presign must be V2".to_string(),
+        ));
+    };
+    let decode = |bytes: &[u8]| -> DwalletMPCResult<commitment::CommitmentSizedNumber> {
```
ycscaly: No functions as variables

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -410,6 +531,438 @@ impl SignPublicInputByProtocol {
 // `message_centralized_signature` selects `SignData::ToBeEmulated` (NOA path);
 // otherwise `SignData::Unverified(deserialized_sign_message)` (user-driven path).
 
+/// Recovers the presign `session_id` from a public VSS presign — the key under
+/// which this presign's private nonce output was persisted at presign-finalize.
+/// Deserializes the curve-specific public `schnorr::vss::Presign`.
+pub(crate) fn vss_public_presign_session_id(
```
ycscaly: I am not sure why we need this; we can keep the session id saved seperately inside the db perhaps as key right? Why would we need to do this at fetch time?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -410,6 +531,438 @@ impl SignPublicInputByProtocol {
...
+        }
+    };
+    decode(&inner)
+}
+
+/// Deserializes the latest network Reconfiguration V3 public output, which holds
+/// the per-curve secret-key polynomial commitments + masked parts + randomizer
+/// dealings the Fast Schnorr (VSS) sign needs. VSS sign requires a reconfigured
+/// network key — the DKG-only output does not expose these accessors.
+fn vss_reconfiguration_public_output(
```
ycscaly: I think its a bit too much cryptography for here, and also there should be the network dkg option as well, and I'm not sure what your point is. We can think of moving much of the cryptogrpahy you did in this pr onto crypto private lib

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -410,6 +531,438 @@ impl SignPublicInputByProtocol {
...
+        <twopc_mpc::decentralized_party::reconfiguration::Party as Party>::PublicOutput,
+    >(reconfig_bytes)
+    .map_err(|e| {
+        DwalletMPCError::BcsError(bcs::Error::Custom(format!(
+            "failed to deserialize reconfiguration V3 public output: {e}"
+        )))
+    })?;
+
+    let validator_keys = ClassGroupsAndPvssKeyPairAndProof::from_seed(root_seed);
+    let (secret_key_share_first_part, secret_key_share_second_part) = reconfiguration_public_output
```
ycscaly: Could be dkg output as well...

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -410,6 +531,438 @@ impl SignPublicInputByProtocol {
...
+        session_id,
+        presign_blending_index,
+    )?;
+    let entry = private_presign_outputs
+        .into_iter()
+        .find(|output| {
+            output.session_id == session_id
+                && output.presign_blending_index == presign_blending_index
+        })
+        .ok_or_else(|| {
```
ycscaly: All of these weird computations would have been solved had you keyed the presign by the blending index as well. I do remember we had issues with that and thought to skip that in ahe blending schnorr presigns, so we need to think carefully which is better

crates/ika-core/src/dwallet_mpc/crytographic_computation/protocol_cryptographic_data.rs
```diff
@@ -73,6 +75,25 @@ pub(crate) enum ProtocolCryptographicData {
...
 
+    /// Fast Schnorr (VSS) sign. Serves both external (`ProtocolData::Sign`) and NOA
+    /// (`ProtocolData::NetworkOwnedAddressSign`) VSS signs: VSS has no AHE decrypters,
+    /// so the AHE NOA-vs-external compute split is unnecessary — NOA output is still
+    /// routed to its channel by the request's session type. Unlike AHE `Sign`, the
+    /// private input is not the class-groups decryption key shares but the validator's
+    /// VSS secret-key Shamir shares (recovered at compute time from
+    /// `reconfiguration_public_output` + the validator's PVSS keys via `root_seed`)
+    /// joined with the persisted presign nonce data (`presign_private_output`). The
+    /// non-serializable 8-field VSS `PrivateInput` is assembled in `compute_mpc`.
```
ycscaly: Im not sure why this conflating is good, why not seperate?

ycscaly (commented):
See comments

crates/dwallet-classgroups-types/src/lib.rs
ycscaly: Fixed in 805189d8d0. The justification was actually false — `curve25519::Scalar` *is* `Serialize`. The real issue is that a struct holding private decryption keys shouldn't be `Serialize` at all: dropped the derive, renamed the struct to `ValidatorMPCSecrets`, and added a SECRET. doc spelling out it must never be serialized/persisted/transmitted (publish via the public-only extractor).

crates/dwallet-classgroups-types/src/lib.rs
ycscaly: Done in 805189d8d0 — dropped "schnorr" from all VSS HPKE identifiers (VssSchnorrHpke… → VssHpke…, vss_schnorr_hpke… → vss_hpke…) across the codebase. They are plain VSS HPKE keys; protocol-name prose ("Fast Schnorr (VSS)") is left as-is.

crates/dwallet-rng/src/lib.rs
ycscaly: Renamed to vss_hpke_secret_key_seed / _rng in 8b06ceb4a7; the doc now says the secret key is sampled from this RNG and the public encryption key + UC proof are derived from it. Transcript label updated too.

crates/ika-core/src/authority/authority_per_epoch_store.rs
ycscaly: Documented the VSS presign-pool maps (key types, value, meaning) in bab9957204.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
ycscaly: Right — fixed in 29f7f00718. VSS sign now uses the reconfiguration output if the key has reconfigured, else the network DKG output (no longer hard-requires reconfiguration). Needed the DKG outputs polynomial_commitments accessors, added in cryptography-private PR #534.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/presign.rs
ycscaly: Fixed. Moved parse+verify into cryptography-private as parse_and_uc_verify_encryption_keys (PR #534), which parses + UC-verifies with per-party exclusion (a malformed/unprovable key drops only that party, never aborts). ika side wired in 1484978be1. The downstream dealing already uses handle_invalid_messages_async.

ycscaly: Moved the whole function to cryptography-private (mpc::hybrid_public_key_encryption::parse_and_uc_verify_encryption_keys, PR #534, non-ika docs); the ika function is deleted and call sites inlined (1484978be1).

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: You were right — it is supported. The crypto DKGSignDecentralizedParty exists for VSS, its PrivateInput is identical to the standalone VSS sign party, and VSS presigns are dWallet-independent, so there is no protocol constraint. Wired the combined VSS DKG-and-sign fast path in cd6aa4b363 (the "not supported" comments were the actual error).

crates/ika-core/src/authority/authority_per_epoch_store.rs
ycscaly: Reworked to key by (session_id, blending_index) — one PrivatePresignOutput per row, not a vector per session. The AHE presign storage + single-use guard were keyed the same way first (PR #1716, merged), then extended to the VSS store in the merge bf44cb4e0a.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: The presign identity is no longer reconstructed with an ad-hoc parse at fetch — it is parsed once into (session_id, blending_index) via vss_public_presign_identity and used for a direct point-get of the store (now keyed by that pair). bf44cb4e0a.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: Removed in bf44cb4e0a — the decode closure is gone; identity recovery is the plain fn vss_public_presign_identity (parses the presign once, returns (session_id, blending_index)).

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: Reworked in 29f7f00718. The secret-key-share recovery now works off either the reconfiguration output or the network DKG output (whichever is current) via a small ika dispatch; the misleading "DKG-only output does not expose these accessors" claim is gone (it does — same derive_*/polynomial_commitments, the latter added in PR #534).

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: Yes — fixed in 29f7f00718; sources from the DKG output when no reconfiguration has happened.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
ycscaly: Keyed the presign store AND the single-use guard by (session_id, blending_index) — for AHE (PR #1716) and VSS (bf44cb4e0a) — which removes the .find() and the session_id reconstruction. The AHE-blending concern turned out to be about keying the presign *pool* by blending index (sequential pop over a shrinking vector), which is a separate structure from the private-output store; the store keys cleanly.

crates/ika-core/src/dwallet_mpc/crytographic_computation/protocol_cryptographic_data.rs
ycscaly: Split into two variants in cb2d115a04: SignVSS (external ProtocolData::Sign) and a new NetworkOwnedAddressSignVSS, mirroring the AHE Sign/NetworkOwnedAddressSign split. The NOA data is no longer coerced to SignData at construction; the coercion happens at the compute site like the AHE NOA arms.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/presign.rs
ycscaly: Moved the whole function to cryptography-private (mpc::hybrid_public_key_encryption::parse_and_uc_verify_encryption_keys, PR #534, with non-ika docs and a unit test); the ika function is deleted and the call sites inlined (1484978be1).

ycscaly: Moved the whole function to cryptography-private (mpc::hybrid_public_key_encryption::parse_and_uc_verify_encryption_keys, PR #534, with non-ika docs and a unit test); the ika function is deleted and the call sites inlined (1484978be1).

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -109,18 +111,26 @@ impl ClassGroupsKeyPairAndProof {
...
-pub struct ClassGroupsAndPvssKeyPairAndProof {
+/// Composition: the existing [`ClassGroupsKeyPairAndProof`] as a member PLUS the three
+/// per-curve PVSS HPKE keypairs (private decryption key + public encryption key +
+/// UC-secure proof), introduced at the `cryptography-private @ 9d35fa76` bump for
+/// upstream's threshold-encryption-to-sharing sub-protocol.
+///
+/// Generated deterministically from the validator's [`RootSeed`]; the seed must be
+/// cryptographically secure and kept confidential.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ValidatorMPCSecrets {
```
ycscaly, line 119: We talked about it, that it is not a good idea to put both public (enc, proof) values and private (dec) values into a single struct. Bad practice. This struct should have had only the secrets, and you should have created another struct for the enc keys and proofs.

crates/dwallet-classgroups-types/src/lib.rs
```diff
@@ -210,17 +229,39 @@ impl ClassGroupsAndPvssKeyPairAndProof {
             >(&ristretto_setup, &mut ristretto_rng)
             .unwrap();
 
-        ClassGroupsAndPvssKeyPairAndProof {
+        // Fast Schnorr (VSS) HPKE keypair: a single curve25519 keypair (not class
+        // groups, not per-curve) used as the known-order threshold-encryption-to-
+        // sharing transport for the VSS Schnorr presign. The secret is discarded
+        // here and re-derived from the same `RootSeed` where it's needed.
+        let (_vss_hpke_secret, vss_hpke_public, vss_hpke_proof) =
```
ycscaly: Why do you ignore the secret? Put it in the strcut.

crates/dwallet-mpc-types/src/mpc_protocol_configuration.rs
```diff
@@ -114,6 +121,14 @@ lazy_static! {
     };
 
     /// Global presign supported curves to signature algorithms for imported keys
+    ///
+    /// VSS (Fast Schnorr) variants are intentionally absent: the decentralized
+    /// party's secret key share must be Shamir-shared by the network, which is
+    /// impossible for a user-imported secret. NOTE: omission from this map alone
```
ycscaly: This is not the reason. The reason is that vss presigns live only for a single epoch, and thus cannot be guranteed to be usable in a general sense, which contradicts the current Move design. We will allow dwallets to use fast schnorr later on, when we have users talk to the network directly

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -192,14 +202,34 @@ impl ProtocolCryptographicData {
...
+                    // Fast Schnorr (VSS): carry the secret-key-share source (the
+                    // reconfiguration output if the key reconfigured, else the network
+                    // DKG output) for secret-key-share recovery at compute time, plus
+                    // the persisted presign nonce data, instead of the AHE decryption
+                    // key shares.
+                    let secret_key_shares_source = VssSecretKeySharesVersionedSource::select(
+                        decryption_key_shares.get_network_encryption_key_public_data(
+                            dwallet_network_encryption_key_id,
+                        )?,
+                    );
```
ycscaly: Why is this so complicated?

We already have a mechanism to compute the decryption key shares out of dkg/reconfig, and we store them somewhere that is accessible here, why not follow the same pattern for the shamir secrets of each network key id?

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/presign.rs
```diff
@@ -275,6 +363,77 @@ impl PresignPublicInputByProtocol {
...
+                let _ = dwallet_dkg_output;
+                let protocol_public_parameters =
+                    network_encryption_key_public_data.secp256k1_protocol_public_parameters();
+                let party_encryption_keys =
+                    parse_and_uc_verify_encryption_keys(&validator_mpc_keys_by_party_id.vss_hpke)
+                        .map_err(|e| {
+                        DwalletMPCError::InvalidInput(format!(
+                            "failed to parse/verify VSS HPKE encryption keys: {e:?}"
+                        ))
+                    })?;
```
ycscaly: Most certainly not!!!!! Are you crazy? You will re-verify the proofs for each public input? This is more expensive then the entire presign protocol, is redundant, AND is computed out of Rayon!?!?!??!

The whole idea is that you only verify it once, and then keep only the verified result. I stressed that out and you messed up again.

When we get the vadliator keys and proofs, we dont even need to save them, we just call this there, and then dont even save the proofs, just the result of that verification function.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -410,6 +591,586 @@ impl SignPublicInputByProtocol {
...
+pub(crate) fn vss_public_presign_identity(
+    signature_algorithm: DWalletSignatureAlgorithm,
+    presign: &SerializedWrappedMPCPublicOutput,
+) -> DwalletMPCResult<(commitment::CommitmentSizedNumber, u16)> {
+    let versioned: VersionedPresignOutput = bcs::from_bytes(presign).map_err(|e| {
+        DwalletMPCError::BcsError(bcs::Error::Custom(format!(
+            "failed to deserialize VSS presign output: {e}"
+        )))
+    })?;
+    let VersionedPresignOutput::V2(inner) = versioned else {
```
ycscaly: V2?? It must be V3 are you kidding me? The vss presign struct is a different structure than the ahe one, and v2 was for ahe????

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
```diff
@@ -410,6 +591,586 @@ impl SignPublicInputByProtocol {
...
+    ) -> (
+        &Vec<group::Value<group::ristretto::GroupElement>>,
+        &Vec<group::Value<group::ristretto::GroupElement>>,
+    ) {
+        match self {
+            Self::Reconfiguration(output) => output.ristretto_polynomial_commitments(),
+            Self::NetworkDkg(output) => output.ristretto_polynomial_commitments(),
+        }
+    }
+}
```
ycscaly: All of this should have been computed when we got the network key data directly, you can move that code there, follow and model the path of computing decryption key shares, its the most efficient and make-sense design.

ycscaly: And I mean ALL OF IT, including the public data you parse here as well.

crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -359,6 +359,10 @@ where
             .iter()
             .filter_map(|(n, v)| v.ristretto_pvss.clone().map(|k| (*n, k)))
             .collect();
+        let vss_hpke_public_keys_and_proofs: HashMap<_, _> = decoded_per_validator
```
ycscaly: Maybe this would be the spot to verify the proofs? And then save only the return value?

crates/ika-protocol-config/src/lib.rs
```diff
@@ -17,11 +17,13 @@ use tracing::{info, warn};
 
 /// The minimum and maximum protocol versions supported by this build.
 const MIN_PROTOCOL_VERSION: u64 = 3;
-const MAX_PROTOCOL_VERSION: u64 = 4;
+const MAX_PROTOCOL_VERSION: u64 = 5;
```
ycscaly: What? Why? It should be kept to 4

crates/ika-types/src/committee.rs
```diff
@@ -609,26 +665,43 @@ pub struct DecodedValidatorEncryptionKeys {
...
+        });
+    }
+    // v4 bundle (4 fields, no VSS HPKE key).
+    if let Ok(bundle) = bcs::from_bytes::<LegacyValidatorEncryptionKeysAndProofsV4>(bytes) {
+        return Some(DecodedValidatorEncryptionKeys {
+            class_groups: bundle.class_groups,
+            secp256k1_pvss: Some(bundle.secp256k1_pvss),
+            secp256r1_pvss: Some(bundle.secp256r1_pvss),
+            ristretto_pvss: Some(bundle.ristretto_pvss),
+            vss_hpke_public_key_and_proof: None,
```
ycscaly: What is this weird code here? What is the backward compataibilitty for exactly? I think you are wrong

docs/plan-fast-schnorr.md
ycscaly, line 1: Delete

sdk/typescript/src/client/hash-signature-validation.ts
ycscaly, line 1: Why is this file changed as part of this pr? I think it should be reverted unless there is something i miss?

sdk/typescript/src/client/ika-transaction.ts
ycscaly, line 1: again revert

sdk/typescript/test/unit/imported-key-vss-gates.test.ts
ycscaly, line 1: Revert

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
```diff
@@ -53,6 +131,22 @@ pub struct ValidatorPrivateDecryptionKeyData {
...
 
+    /// Validator-private PVSS HPKE secret decryption keys (secp256k1 +
+    /// ristretto). Used at network-key ingestion to pre-derive the Fast
+    /// Schnorr (VSS) Shamir shares; paired with `validator_pvss_publics_for_vss`
+    /// (the matching public encryption keys). `None` if Fast Schnorr is not
+    /// configured on this node (e.g. integration tests that don't exercise
+    /// VSS). The VSS HPKE curve25519 **secret** key isn't here — it's needed
+    /// only at the presign hot path and is cached on
+    /// `CryptographicComputationsOrchestrator`.
+    pub validator_pvss_secrets_for_vss: Option<ValidatorPvssSecretsForVss>,
```
ycscaly: Our secrets should never be optional, we know we upgraded the binary..

crates/ika-core/src/sui_connector/sui_syncer.rs
```diff
@@ -411,53 +413,45 @@ where
...
+                ) {
+                    Ok(k) => Some((*name, k)),
+                    Err(e) => {
+                        warn!(
+                            authority = ?name,
+                            error = ?e,
+                            "Failed to decode mainnet-v1.1.8 ClassGroupsEncryptionKeyAndProof from Move-side mpc_data"
+                        );
+                        None
+                    }
```
ycscaly: why are there such diffs in this file now? we dont actually need to change everything against dev no?

Issue #575, opened 2025-01-30
Improve error handling


zeevmoney:

zeevmoney:
In the: `crates/pera-core/src/dwallet_mpc/encrypt_user_share.rs` file, there are server `DwalletMPCError`, for all of them add the error from the closure as string.

Dr-Maxwell:
Could you please provide a more detailed description of the desired error handling improvements? It would be helpful to understand the specific scenarios where these error strings are needed

PR #1721, opened 2026-05-25
Off-chain validator metadata + EndOfPublishV2

omersadika:
## Summary

Re-lands the off-chain validator-metadata pipeline on top of `dev` (full clean-slate rebuild from the original branch), adds the EndOfPublishV2 consensus message that bundles the handoff signature with the EndOfPublish vote, and closes the final propagation gap (P2P blob fetch) so chain reads are eliminated for `mpc_data` / network-key / reconfig blobs in v4.

Gated entirely by `off_chain_validator_metadata` (and the related `network_encryption_key_version` / `reconfiguration_message_version` bumps); v3 stays on the original chain-driven flow.

### Pipeline (all v4-gated)

- **Producer side**: `MpcDataAnnouncementSender` broadcasts each validator's mpc_data digest via consensus + `EpochMpcDataReadySignal` per epoch + `NetworkKeyDKGReadySignal` per key. Validator's own blob is mirrored into both the perpetual store and the in-memory cache backing the local Anemo `GetMpcDataBlob` server.
- **Consensus side**: per-epoch `validator_mpc_data_announcements` table; `EpochMpcDataReadySignal` quorum freezes the snapshot into `frozen_validator_mpc_data_input_set` (idempotent on first quorum).
- **P2P side**: new `PeerBlobFetcher` task pulls peer mpc_data blobs over Anemo, hash-verifies, writes to perpetual + in-memory cache so this validator can also serve other peers without restart.
- **Handoff path**: pure attestation/signature/aggregate helpers; per-epoch record + per-validator perpetual `CertifiedHandoffAttestation`s; Anemo endpoint for joiners to bootstrap from a cert.
- **Joiner path**: late-binding `JoinerPubkeyProvider` (next-epoch committee) + `SubmitMpcDataAnnouncement` relay RPC.
- **Sui syncer**: prefers off-chain class-groups assembly; skips chain blob reads (`get_network_encryption_key_with_full_data_by_epoch`, `get_mpc_data_from_validators_pool`) in v4 — overlay fills bytes from local producer cache + P2P fetches.

### EndOfPublishV2

New consensus message variant `EndOfPublishV2 { authority, handoff_signature }` that bundles the validator's signed handoff attestation with its EndOfPublish vote at exactly the consensus point EOP fires. Eliminates the V1 race where a separate `HandoffSignature` consensus tx could arrive at peers out of order with EndOfPublish and produce divergent aggregator state across the committee. Producer emits V2 instead of V1 + separate handoff when `off_chain_validator_metadata` is on; consumer side splits the bundled message back into its two parts and routes each through the existing v1 paths.

### Protocol version

Bumps `MAX_PROTOCOL_VERSION` to 4. v4 activates: `internal_presign_sessions`, `off_chain_validator_metadata`, `consensus_skip_gced_blocks_in_direct_finalization`, `network_encryption_key_version = 3`, `reconfiguration_message_version = 3`.

## Test plan

- [x] Acceptance gate: `cargo test --release -p ika-core test_network_dkg_full_flow` passes (181s).
- [x] V2 unit tests: BCS round-trip + V1/V2 key distinctness pass.
- [x] Protocol-config snapshot tests pass.
- [x] 10-epoch churn cluster test passes (~21 min; 14 aggregate certs across the cluster, all 10 epoch transitions advance within timeout).
- [x] New cluster test `off_chain_metadata_v4_does_not_read_blobs_from_chain` passes (256s) — chain blob reads delta = 0 across an epoch transition under v4.
- [x] Existing joiner/remove/multi-epoch user-session cluster tests pass.
- [ ] Reviewer: please double-check the EndOfPublishV2 producer/consumer gating and the wire-author check (`handoff_signature.signer == authority`) is the right semantics.
- [ ] Reviewer: confirm the `PeerBlobFetcher` polling cadence (2s) and the gating (skip self, skip already-cached) are appropriate.

---

## Review-hardening update (multi-pass byzantine/restart review)

A multi-agent critical review (byzantine-validator, honest-mistake, restart, and doc-clarity passes) surfaced correctness gaps that are now fixed. All work is v4-gated; v3 paths are untouched at runtime.

### Fixes

1. **Write-through/read-through `BlobCache`** — the perpetual store and the in-memory cache backing the Anemo server were synced by convention at each call site; a forgotten mirror left a durably-stored blob (e.g. a DKG/reconfiguration output) unservable until restart. `BlobCache` owns both behind one `insert` (durable-then-cache) and one read-through `get` (cache, then perpetual fallback). The fallback is the structural fix: a perpetual-only blob is now servable without a restart.

2. **No BLS in the off-chain pipeline; explicit self/relayed announcement kinds** — the single `ValidatorMpcDataAnnouncement` consensus kind (with its implicit sender≠signer exemption and BLS payload signature) is split into `ValidatorMpcDataAnnouncement` (self-submission, no signature — consensus block-author authenticates, enforced by `sender == validator`) and `RelayedValidatorMpcDataAnnouncement` (carries the joiner's **Ed25519 consensus-key** signature, verified against its next-epoch consensus pubkey). `JoinerPubkeyProvider` now returns the Ed25519 pubkey, fetched from next-epoch validator info via the Sui client. `epoch` returns to the announcement body so the joiner signature binds it (no cross-epoch replay).

3. **Joiner announcement fan-out + P2P retry (was unwired)** — `submit_announcement_to_committee` existed but nothing called it, so a joiner never announced and could never enter the next epoch's working set. New `JoinerAnnouncementSender` signs (Ed25519) and fans out to current-committee peers with retry until `f+1` distinct peers accept (one honest relayer guaranteed) or a bounded budget is exhausted; wired into a node-startup watcher that fires when the node observes itself in `V_{e+1}` but not the current committee.

4. **Freeze delayed until joiners are attestable (closes the joiner-coverage gap)** — the freeze fired on the first quorum of ready signals, which validators emitted early (before `V_{e+1}` is published mid-epoch), so joiners were never in the frozen set, the next committee's class-groups map, or the handoff cert. Validators now withhold their ready signal until `V_{e+1}` is published and all its members are locally validated (or an epoch-clock deadline as a liveness backstop). This is a **producer emit-gate** change — the deterministic single-freeze mechanism is preserved; the deadline is wall-clock and only affects *when* each validator emits, while the freeze snapshot is still computed deterministically at the consensus-ordered quorum point.

5. **Producer self-heals via confirmation retries** — the producer marked itself done on submit *handoff*, not *confirmation* (`submit_to_consensus` returns Ok as soon as the background submit task is spawned, which can still fail to sequence at the epoch boundary or on crash). It now caches an idempotent announcement (same key on re-send → consensus dedup) and re-submits until its own entry appears in the table.

6. **Empty-input off-chain assembly + frozen-set-as-truth** — `assemble_committee_class_groups_off_chain` no longer returns `Complete` with empty maps; `try_assemble_class_groups` treats the frozen set as the post-freeze source of truth so a never-announcer can't stall assembly (the delayed freeze ensures the frozen set includes joiners).

7. **Smaller hardening** — sentinel `timestamp_ms == 0` rejected at sign + record; self-attestation gated on own-blob health; cert duplicate-signer / quorum-boundary tests; doc sweep fixing stale "chain fallback" / "NetworkKeyDKGReadySignal freezes" / plan-phase references.

### Tests

- Unit: 75 ika-core (`blob_cache`, `validator_metadata`, `epoch_tasks`) + 9 ika-types; clippy clean across ika-core/types/network/node.
- v4 cluster (exercise freeze-delay + joiner fan-out end-to-end): `off_chain_metadata` ✅ (271s), `joiner::test_joiner_added_at_epoch_2` ✅ (262s).
- v3 regression: `network_dkg` (incl. bwd-compat) 7/7 ✅ (1776s).
- CI green at HEAD `cd42e9c015` (Cargo Test Check, Format, all Move formatters pass).

---

## In-PR completion: cert-bootstrap consumer + F4-1 cycle-break (#4, #6)

Two items previously slated as follow-ups, now done in-PR.

### #4 — Joiner cert-bootstrap consumer wired in

The handoff-cert Anemo endpoint had no consumer. `JoinerBootstrapVerifier`
(`epoch_tasks/joiner_bootstrap_verifier.rs`) now fetches the prior-epoch
`CertifiedHandoffAttestation` from a peer (`HandoffCertSource` trait,
`P2pHandoffCertSource` impl) and verifies it against the prior committee
via `verify_joiner_bootstrap_cert(cert, expected_prior_epoch,
prior_committee, prior_consensus_pubkeys, expected_next_committee_pubkeys)`
— epoch-bound so a stale cert can't be replayed. Wired into
`monitor_reconfiguration`, gated on off-chain metadata + `epoch >= 1` +
self absent from the prior committee (i.e. an actual joiner). Non-halting:
a failed fetch/verify logs `error!` and retries on a bounded budget rather
than stalling startup. 4 unit tests cover fetch / retry / verify-loop.

### F4-1 cycle-break (the blocker the dedicated test caught)

The dedicated test surfaced a real deadlock: the joiner watcher and the
producer's freeze emit-gate both keyed off the **assembled** next-epoch
committee — but the assembly can't complete until the joiner announces,
and the joiner can't learn it's a joiner until the assembly publishes.
Fixed by publishing a lightweight **chain view** of the next-epoch
committee (members + stake, empty class-groups) over a new
`chain_next_epoch_committee_receiver` as soon as Sui selects it
(`sui_syncer::sync_next_committee`, before off-chain assembly). The joiner
watcher and the freeze emit-gate consume this chain view; the joiner now
demonstrably fans its mpc_data out (confirmed in logs — it never did
before). Pubkey-provider poll tightened 15s→5s and joiner fan-out retry
10s→3s so the relay path can complete inside the freeze window.

### #6 — Dedicated F4-1 cluster test (added, passing)

`test_joiner_lands_in_next_committee_class_groups` asserts a mid-epoch
joiner lands in the next committee's `class_groups_public_keys_and_proofs`.
It did its job — driving it to green surfaced **three** distinct, interacting
defects in off-chain joiner integration, all now fixed (commit
`5a241701d1`). It **passes** end-to-end (322s; epoch-1→2 freeze is
`frozen=5 excluded=0`):

1. **Ready-signal canonicalization stripped joiners (root cause).** Incoming
   `EpochMpcDataReadySignal`s were canonicalized against the **current**
   committee (`drop weight==0`). A next-epoch joiner has current-committee
   weight 0, so it was filtered out of the *recorded* signal even though
   every validator correctly attested it (emit-time `vcount=5`). The freeze
   partition — which decides *next*-epoch membership — then saw zero
   attestations for the joiner and excluded it (`frozen=4 excluded=1`). Fix:
   treat announcers as valid attestation targets in canonicalization; a
   joiner that announced has a signed announcement consensus-ordered before
   any ready signal attesting it, so this is padding-safe.

2. **The joiner's blob had no path to the committee.** The relay forwarded
   only the digest, and the peer blob fetcher pulls only from
   current-committee peers (excluding the joiner) — so no validator could
   obtain the joiner's blob to validate it. Fix: the joiner pushes its blob
   on the fan-out RPC; the relayer verifies (hash + structural decode) and
   caches it write-through, and the committee resolves it via the existing
   content-addressed P2P fetch (joiner→committee direction, no dial-back).

3. **Poll cadences too coarse for the freeze window.** The integration path
   must finish inside `[epoch/2, 3·epoch/4]`. The fixed multi-second cadences
   (10s chain-committee sync, 5s pubkey refresh, 3s fan-out, 2s blob
   fetch/producer) overrun that window in a short test epoch. New
   `epoch_scaled_poll_interval` scales each to ~1% of the epoch, clamped to
   the production default — a no-op at production epoch lengths.

A fail-fast timeout was added on the joiner's epoch-2 wait so an excluded
joiner fails the test with a clear message instead of hanging.

---

## Stale-share sign-failure fix: prepare-then-start handoff gate (+ epoch-advance robustness)

Heavy-load + slow-network testing surfaced two issues, now fixed (v4-gated where applicable; v3 runtime paths untouched).

### Root cause — stale network-key shares after reconfiguration

Per-party adoption-time tracing showed the off-chain reconfiguration output is adopted at **staggered times** across validators (60–130s lags under load). The new epoch's MPC service started immediately at the reconfigure seam while the handoff data was still arriving asynchronously, so a validator entering epoch N could begin signing while still holding epoch N-1's `(t,n)` decryption-key sharing — combined with peers who already adopted epoch N's sharing, the threshold sign round failed with `FailedToAdvanceMPC(InvalidParameters)` (~50–130s after each advance).

### Fixes

1. **prepare-then-start barrier** (`ika-node::monitor_reconfiguration`) — before starting the new epoch's MPC components, install the new epoch's network-key blob-source overlay first (else the barrier deadlocks: `network_keys_receiver` is fed from whichever overlay is installed, and the per-iteration install runs only in the *next* loop iteration), then **block until** (a) the cross-epoch handoff cert anchoring the new epoch is present **and re-verified** against the signing committee — a *second* verification at consume-time on top of the one done before it is persisted, fail-closed on a tampered local cert DB — and (b) every tracked network key has surfaced its reconfiguration output for the new epoch (`current_epoch >= next_epoch`, non-empty output). Blocks indefinitely (safety-first; a stalled validator that is visibly not signing beats one signing with the wrong shares). Only a validator in the new epoch prepares. Clear logs (INFO entry/exit, WARN every ~10s with the breakdown) + metrics `ika_handoff_prepare_{waiting,retries_total,duration_seconds}`.

2. **advance_epoch session-completion gate** (`sui_executor`) — under load `sessions_manager::advance_epoch` MoveAborts with `ENotAllCurrentEpochSessionsCompleted` (code 6) when a network-key system session starts after the quorum's `received_end_of_publish` snapshot; the notifier retried for an hour then `panic!`d the validator → quorum loss → cascade. Now re-checks `SessionsManager::all_current_epoch_sessions_completed()` (new, unit-tested) against just-synced state and holds the tick rather than submitting a doomed tx; the hour-long panic only guards genuinely fatal submission errors.

3. **Integration-test poll timeouts** — raised the SDK `#pollUntilCondition` default timeout (30s → 10m) and `retryUntil` default attempts (30 → 600), and removed the now-redundant per-call `{ timeout: 600000 }` overrides so every poll-site uses one default. MPC rounds legitimately take minutes under load; the short defaults were turning slowness into spurious `Timeout waiting for…` failures (never crypto failures).

4. Pre-existing test-compile fix (`new_validator_mpc_data_announcement` missing `blob` arg) that broke the ika-core test build.

### Tests

- Unit: `SessionsManager::all_current_epoch_sessions_completed` + `all_network_keys_ready_for_epoch` truth tables.
- v4 cluster: `off_chain_metadata` ✅, `protocol_version_transition` (v3→v4) ✅ — barrier engages, verifies, and clears at every transition; no `InvalidParameters`, no `WaitingForNetworkKey` wedge, no panic.
- TS integration (fresh networks): **`all-combinations-future-sign` 21/21** (previously produced 3 `InvalidParameters` sign failures), plus dwallet-creation 8/8, make-public-share-and-sign 7/7, global-presign 5/5, imported-key 6/6, dwallet-sign-during-dkg 10/10 — **zero sign failures across every run**, reaching epoch 8+ under the heavy suite that previously crashed at epoch 7. Every remaining intermittent test failure was a polling timeout under CPU starvation, not a crypto failure (addressed by the timeout-default change above).

---

### The 5x slowness was an env interaction, not hardware

The `ika-k8s-large` runners appeared ~5x slower than an M3 Max on class-groups crypto. Root cause: `RUST_BACKTRACE=1` (set workflow-wide) × class-groups constructing backtrace-carrying errors **eagerly on the success path** of every group operation (`ok_or(Error::from(..))`) — millions of globally-locked DWARF unwinds per network-key instantiation (~5.4x CPU, 23x sys-time, *negative* multi-thread scaling). "Apple Silicon is faster" was an artifact: dev Macs simply had `RUST_BACKTRACE` unset.

- Interim fix: `RUST_LIB_BACKTRACE: "0"` beside `RUST_BACKTRACE: 1` (panic backtraces kept, library captures off). Post-fix the EPYC runner is at **full per-thread parity** with the M3.
- Durable fix: cryptography-private **#575** (lazy error construction, 51 hot-path sites) — merged, pin bumped `a37297c → de3cddd` (exactly that one commit), and the `RUST_LIB_BACKTRACE` workaround **removed from all six workflows**. A/B-validated under the exact CI env on the instantiation path: 161.6s without the guard vs 162.3s with.

### Suites runnable on CI instead of locally

`integration-tests-ci.yaml` (45+ dwallet-MPC tests, ~35min at 4 threads), `test-cluster.yaml` (nextest process-per-test, 4-way; 8-way OOMs the 96Gi pod), `ts-integration-tests.yaml` (full SDK suite against one localnet), `simtest.yaml` — all dispatchable via `gh workflow run` with filters, documented in CLAUDE.md. Compilation now lives in its own (UI-collapsed) step — it was 57% of the cluster log volume — and failure replays stay inline deliberately: when a runner pod dies, the artifact upload never happens and the live log is the only surviving evidence.

### Test-infra hardening

- Cross-process cluster **boot lock** (fixed-port listener) serializing sui-swarm's probe-then-bind port window across parallel nextest processes — now also covering the **joiner spawn** (ports probed at config build, bound only after the candidate→stake→add tx sequence; that window raced and produced `Address already in use` joiner failures).
- TS poll budgets unified: `retryUntil` default raised to ~15min (an MPC operation astride the epoch-close lock legitimately waits out an epoch boundary), all 16 leftover 30-attempt per-call overrides removed, vitest per-case timeout 900s → 1200s and the local script default 120s → 1200s — a stuck wait surfaces as the helper's precise error, never an opaque vitest case kill.

---

## MPC service-loop responsiveness + internal-presign determinism

1. **Non-blocking network-key instantiation** — the service loop awaited the multi-minute class-groups instantiation inline, freezing all session processing per validator at every epoch boundary. Now spawn/poll split: instantiation runs on rayon, the loop polls completions **once per service iteration**.
2. **Key adoption moved out of the round drain** — adoption/instantiation/polling were gated on fresh consensus rounds, which deadlocks the key-arrives-after-request bootstrap (no key → no rounds → no adoption). Rule now encoded in the code: wall-clock-fed inputs (watch channels, perpetual store) run once per service ITERATION, never inside the per-consensus-round drain. This was the `missing_network_key` wedge.
3. **Internal-presign session-id determinism** — session identifiers baked in the instantiation-time consensus round (and HashMap iteration order), yielding 36 private session ids for 9 logical sessions (empty pool, "epoch 2 was blocked" joiner failures). The round is out of the preimage; iteration is ordered; ids are now byte-identical across the committee (verified live: every id instantiated exactly once per validator).

---

## Epoch-close wedge: two independent bugs, reproduced, fixed, spec'd, regression-tested

The TS-suite flake ("a session astride `lock_last_active_session_sequence_number` wedges the epoch; everything after cascades into timeouts") was cornered with a reproduction rig (5-min-epoch localnet + a global presign every 8s) and turned out to be **two separate bugs**, both confirmed against live chain state:

1. **Overshoot** — the global-presign serving path completed user sessions on-chain with **no lock check** (unlike MPC sessions, which gate computation on the synced target), and Move's end-of-publish predicate is a **strict equality** (`completed_sessions_count == locked target`) with a never-decreasing counter: one beyond-the-lock completion wedges the epoch permanently and unhealably. Reproduced: target frozen at 0, 97 sessions completed anyway. Admission rejections had the same hole (a quorum'd Rejected counts as completed). **Fix: gate consensus *submission* on the locally-synced lock target** (votes in `get_unsent_presign_requests`, rejections via a `pending_rejected_sessions` retry buffer). Gating must not happen at checkpoint build (checkpoint contents must be a pure function of consensus; the lock view is wall-clock fullnode state — it would fork checkpoints); quorum supplies the safety argument, since the chain target is monotone and frozen at lock, so any agreement implies an honest validator saw the target cover the session.
2. **Undershoot** — `handle_computation_results_and_submit_to_consensus` used `return` (not `continue`) in its missing/non-active-session guards, so **one stale result abandoned the whole batch**, silently dropping sibling sessions' round messages (pre-existing since #1588, 2025-10). Two validators hitting the race in the same boundary window starve the internal presign sessions below the message threshold → pool never refills → locked-set presigns unservable → `completed < target` forever. Reproduced live at an epoch boundary burst.

**Spec**: `specs/epoch-close-session-lock.md` — the frozen target, the strict-equality predicate's two failure directions, the gate-consensus-submission decision rule every user-session completion path must follow, and the batch-processing rule.

**Regression tests** (all green on CI):
- `global_presign_beyond_lock_target_held_until_target_covers_it` + `rejection_beyond_lock_target_held_until_target_covers_it` (integration) — held while beyond the target (with a stocked pool that would otherwise serve), served/submitted once covered.
- `computation_results_batch_survives_stale_entries` (integration) — six stale entries mixed into six live ones; every live round message must reach consensus (pre-fix survival probability < 0.2%).
- `test_global_presigns_complete_across_epoch_switches` (cluster) — streams global presigns across two epoch boundaries, asserts epochs keep advancing and the user-session counters fully drain.

**Validation**: locally, 10 consecutive lock windows under continuous traffic (303/303 completions, zero failures) where the unfixed binary wedged at window 0 (overshoot) and window 6 (batch-drop) respectively; on CI, the cluster suite passed 13/13 including the 5-epoch churn test that failed without the batch-drop fix, and the integration suite 48/48.

**Known issue surfaced (out of scope, tracked)**: ~64 minutes into max-rate traffic, one validator computed a byte-divergent output for one internal presign session and the output-quorum tally convicted it malicious (it even *self*-recognized — the tally is consensus-deterministic), silently downgrading the committee to zero-redundancy 3-of-4 until a later loss stalled MPC. Pre-existing, never observed in CI, untouched by this PR; diagnostic signature and reproduction recipe documented for a follow-up ticket.

---

## Observability + allocator

- **26 new metrics** across EpochMetrics / SuiConnectorMetrics / DWalletMPCMetrics / IkaNodeMetrics / blob store: freeze epoch, excluded validators, ready signals + stake, announcements, handoff cert (epoch/signatures/buffered/rejected), presign pool size/queue/served, instantiation in-flight/failures/sub-call durations, blob store size/evictions, fetch/bootstrap outcomes — all re-seeded at epoch-store open. Deduped warns (cert-digest mismatch once per digest, NOA starvation throttled), quorum anchors at info.
- **jemalloc compiled in** (Sui's pattern: `tikv-jemallocator` global allocator behind a default feature in ika-node + the ika CLI). The Dockerfile's `LD_PRELOAD` block never persisted past its RUN layer — production was silently on glibc malloc; the broken block is gone and the allocator is now in the binary.


ycscaly (commented):
The "ClassGroups" naming is stale across the off-chain pipeline added in this PR. The bundle each validator produces and propagates is no longer just class-groups — since PR #1707 it's a `ValidatorEncryptionKeysAndProofs` bundle containing class-groups + per-curve PVSS keys + proofs. `ValidatorMpcData` is already the convention elsewhere in this PR (`HandoffItemKey::ValidatorMpcData`, `validator_mpc_data_announcements` table) — only the assembly path wasn't renamed.

Inline comments mark the assembly-path sites. Out-of-diff sites worth tracking in a follow-up: the wire field `MPCDataV1.class_groups_public_key_and_proof` (`dwallet-mpc-types/src/dwallet_mpc.rs:431`) carries the whole BCS-encoded bundle, not class-groups; and `VersionedMPCData::class_groups_public_key_and_proof()` accessor mirrors it. BCS is positional, so renames there are source-only — no wire-shape impact.

Sites where "class_groups" *does* mean just-class-groups (e.g. `Committee.class_groups_public_keys_and_proofs`, sitting alongside `secp256k1_pvss_*`, `secp256r1_pvss_*`, `ristretto_pvss_*`) are correctly named and not flagged.

crates/ika-core/src/sui_connector/mod.rs
```diff
@@ -127,11 +157,34 @@ impl SuiConnectorService {
...
+        &self,
+        source: Box<dyn crate::validator_metadata::NetworkKeyBlobSource>,
+    ) {
+        self.network_key_blob_source.store(Some(Arc::new(source)));
+    }
+
+    /// Installs the off-chain class-groups assembler the
+    /// next-committee sync uses before falling back to the chain
+    /// `get_mpc_data_from_validators_pool` path.
+    pub fn install_class_groups_source(
```
ycscaly: Installs an `OffChainCommitteeClassGroupsSource` that resolves the full validator-mpc-data bundle. Rename to `install_mpc_data_source` to track the trait rename.

ycscaly:
## Heads-up: this PR makes a `mainnet-v1.1.8` → `dev` rolling upgrade structurally impossible (network-key handoff)

Surfaced while driving a real `mainnet-v1.1.8` → `dev` rolling binary swap through the out-of-process cross-binary harness (#1727). After clearing every other blocker (validator-key shape, graceful shutdown, the protocol-vote buffer-stake, and the internal-MPC-output assert in #1728), the upgrade **wedges at the first reshare the dev validators have to perform on a key they inherited from a v1.1.8-run epoch.**

### Trace (epoch 2 all-`dev`, off_chain disabled at v3, key in `AwaitingNetworkReconfiguration`)

1. The epoch 2→3 reconfiguration session is **parked**: the validator doesn't have the network key's public data (`requires_network_key_data && !key_public_data_exists` in `mpc_session.rs`).
2. `key_public_data_exists` is false because `network_encryption_keys` is empty — the key was never instantiated (no `Instantiating agreed network key` / `Updating network key` logs).
3. It was never instantiated because `agreed_network_key_data` is empty.
4. `adopt_cert_verified_keys` (`mpc_manager.rs`) **skips the reconfigured key**: a key carrying a `current_reconfiguration_public_output` is adopted only when the **prior epoch's handoff cert** pins both its DKG and reconfiguration digests (the branch that `continue`s when `reconfiguration_digests.get(key_id) != Some(...)`).
5. `get_certified_handoff_attestation(prior_epoch)` returns `None` — **the prior epoch was run by `mainnet-v1.1.8`, which has no handoff-cert mechanism**, so no cert was ever produced.

### Why it's structural

This PR removes the network-key consensus vote and replaces it with the handoff cert (`adopt_cert_verified_keys` docstring: *"replacing the now-removed consensus vote"*). The cross-epoch key handoff therefore requires **both** the outgoing and incoming epoch to speak the handoff-cert protocol. A `dev` epoch immediately following a `v1.1.8` epoch finds no cert and has **no fallback** to adopt the inherited key, so it can never reshare it — the network is wedged at the first post-upgrade epoch boundary.

A real `v1.1.8` → `dev` rollout needs a backward-compat path: adopt a key inherited from a pre-handoff-cert (pre-off-chain) epoch without requiring a prior-epoch cert (e.g. fall back to the old byte-identical-quorum/consensus-vote adoption when no cert exists for `prior_epoch`, gated on `prior_epoch` being pre-`off_chain_validator_metadata`).

Note: pure-`dev` works fine (every epoch produces a cert for the next). This only bites at the `v1.1.8` → `dev` boundary. Happy to share the harness logs/run if useful.

PR #1727, opened 2026-06-05
test(ika-upgrade-test): out-of-process cross-binary upgrade harness

ycscaly:
## Summary

New additive crate `crates/ika-upgrade-test` — an **out-of-process** test harness that spawns real, separately-compiled `ika-validator` binaries against an external `sui` localnet, swaps binaries on validators across epochs, and drives dWallet workloads. Unlike `ika-test-cluster` (in-process `IkaNode`, one binary), it can host genuinely different binaries in one committee. No changes to `ika-node` / `ika-swarm`.

Implements `docs/cross-binary-upgrade-testing*.md`; see `docs/cross-binary-upgrade-testing-results.md` for the full write-up.

## Tests (all opt-in via env flags; need real binaries + a workspace-tag `sui`)

| Test | Status | Proves |
|------|--------|--------|
| `tests/smoke.rs` (go/no-go) | ✅ ~396s | 4 out-of-process validators + notifier, network DKG, reach epoch 2 |
| `tests/cross_binary.rs` | ✅ ~722s | Boot 4 on a v3-only binary, swap all to dev, capability vote advances **v3 → v4** |
| `tests/workload.rs` | ✅ ~415s | Full user **DKG → Presign → Sign** completes on-chain |

The cross-binary run exercises, out of process: protocol-vote arithmetic, mid-epoch reconfiguration MPC across the swap, mixed-committee wire compat, and on-disk compat (restart on a new binary against the old RocksDB).

## Crate layout

- `sui.rs` — spawn external `sui start --with-faucet --force-regenesis` (waits for RPC **and** faucet).
- `cluster.rs` — chain bootstrap via `init_ika_on_sui` + `ValidatorConfigBuilder` + a notifier fullnode; `NodeConfig` → YAML → child; on-chain `wait_for_epoch` / protocol-version via `IkaClient`.
- `process.rs` — `ValidatorProcess`: spawn / stop / `swap_binary`, health via the admin RPC.
- `binary.rs` — `BinarySpec` (path / tag / sha / branch) + a sha-keyed `git worktree` build cache honoring each commit's pinned toolchain.
- `scenario.rs` — imperative DSL (start / wait_for_epoch / stop_and_swap / expect_protocol_version).
- `workload.rs` — drives a user dWallet lifecycle by orchestrating the canonical `ika` CLI; confirms completion on-chain.

## Key finding: `mainnet-v1.1.8` ↔ `dev` is not a naive rolling swap

Running the harness with the **real** `mainnet-v1.1.8` `ika-node` fails at boot for the **expected** reason, not a harness bug: v1.1.8 links `class_groups` from `dwallet-labs/inkrypto`, dev from `dwallet-labs/cryptography-private`, and v4 publishes the combined `ValidatorEncryptionKeysAndProofs` where v1.1.8 expects the bare `ClassGroupsEncryptionKeyAndProof` — the exact incompatibility flagged in `validator_initialization_config.rs` (`⚠️ MAINNET WIRE-FORMAT INCOMPATIBILITY ⚠️`). The v1.1.8 node loads its config, connects to Sui, reads the contracts, then fails decoding the on-chain validator record (`class groups public key … remaining input`). So the harness genuinely runs a different binary and fails on the documented wire-format divergence.

To demonstrate a **successful** heterogeneous upgrade, the green `cross_binary` run uses an OLD binary that is a `dev` build pinned to `MAX_PROTOCOL_VERSION=3` (same crypto, differs only in advertised protocol version) — disclosed in the test and results doc.

## Notes

- Builds use `--no-default-features` to drop `enforce-minimum-cpu` (panics on < 16-core hosts).
- The workload uses a dedicated funded user (faucet SUI + IKA transfer) to avoid contention with the notifier; `register-encryption-key` precedes `create`; v4 genesis for `internal_presign_sessions`; long epoch to clear the mid-epoch reconfiguration; sign confirmed via the coordinator's on-chain completed-session count.


ycscaly:
## Update: hardening + findings from a real `mainnet-v1.1.8` → `dev` run

Pushed `8672c96` — harness hardening surfaced by driving an actual `v1.1.8` → `dev` rolling swap (not the same-crypto MAX=3 stand-in):

- **Graceful SIGTERM swap.** `stop()` previously hard-SIGKILLed the node. That interrupts it mid-consensus-round and leaves dwallet-MPC replay state partial on disk, crashing the next binary on replay (`consensus round mismatch ...`). A swap is a planned restart — it now sends SIGTERM (handled by `node_runner`'s `wait_termination`), waits for clean exit, SIGKILL only as fallback.
- **Wait for the real epoch boundary.** The genesis network DKG runs *during* epoch 1, so the scenario waits for epoch 2 before swapping (the epoch counter advancing to N is the completion signal for N-1). Documented in `cluster.rs`.
- **`set_buffer_stake(0)` step.** With n=4 the default 50% buffer rounds up to requiring all four votes; a rolling swap can leave one validator's fresh capability uncommitted at the boundary tally. Dropping the buffer to a quorum is the realistic behavior on larger committees.

### What the real run established (in order)
1. validator class-groups key shape — needs bare-shape publication + shape-tolerant verify;
2. ^ the graceful-shutdown and epoch-wait fixes above;
3. protocol-vote buffer stake — handled by the new step;
4. internal-MPC-output dense round assert — **filed + fixed in #1728** (independent dev bug; affects any v3→v4 transition);
5. **network-key handoff cert** — **structural blocker, filed on #1721**: the off-chain-metadata PR replaces the network-key consensus vote with a per-epoch handoff cert that `v1.1.8` epochs don't produce, so `dev` validators can't adopt the network key inherited from a `v1.1.8` epoch. Not harness-fixable.

Net: the harness takes a real `v1.1.8` → `dev` swap through boot → genesis DKG → clean binary swap → epoch advance → protocol-vote-ready, and pinpoints the remaining blocker as a `dev`-side backward-compat gap (#1721), with #1728 fixing one dev bug along the way.

omersadika:
1. Remove PR-1721-action-plan.md
2. Don't use docs for internal dev docs - use dev-docs folder instead, follow guidance in claude md and README inside the folder
3. Create a new github action workflow - similar to the existing ones - to run ika upgrade test, try to make the input flexible and easy to run when needed

PR #1734, opened 2026-06-11
fix(sui-executor): notifier gas-handling and advance_epoch robustness under load

ycscaly:
Six commits, cherry-picked in their original order from `feat/ika-upgrade-test`, making the notifier's Sui-submission path robust under load. They chain on `sui_executor.rs`, so they ship as one PR; each commit message carries its full story. The distinct fixes:

1. **Track the gas coin from tx effects** (`593aef39`): the executor rebuilt every checkpoint/epoch-advance tx with gas fetched from the notifier's fullnode, which lags the validators by hundreds of object versions under checkpoint-heavy load — stale gas version, rejection, retry against the same lagging view, stalled epoch advance. Cache the gas `ObjectRef` from each tx's effects (the authoritative post-tx version) and build the next tx against it.
2. **Drop the cached gas ref on submission failure** (`34f70b99`): on a rejection the cache wasn't cleared, so `retry_with_max_elapsed_time!` spun on the identical stale ref for the full retry budget (an hour). Clear it on both failure paths.
3. **Stale-gas rejection robustness** (`7cae3fee`): recover when the notifier's gas coin is advanced by another holder of the key — re-fetch floored at the rejected version instead of spinning.
4. **Gate `request_advance_epoch` on session completion** (`caa2cde4`): the notifier submitted `request_advance_epoch` as soon as a quorum's checkpoint arrived; if just-synced state showed sessions still pending, the Move call aborted and the notifier died in a panic loop. Re-check the on-chain completion predicate before submitting. (Touches `system_inner_v1.rs` for the predicate read.)
5. **Apply stale-gas recovery to RPC-level submission rejections** (`01600b82`): the recovery only triggered on effects-level failures; RPC-level rejections (the common shape for stale gas) bypassed it. Factor the recovery into one path covering both.
6. **Document the `err.as_ref()` clippy trap** (`59b0db3e`): clippy's `unnecessary_to_owned` suggestion silently changes which error variant the stale-gas matcher sees; the comment prevents a "cleanup" from reintroducing the spin.

Validated on `feat/ika-upgrade-test`: the full dwallet-creation integration suite (8/8) passes with `sui_tx_err=0`, where the pre-fix binary stalled at epoch 5 with 100+ gas-version rejections; the cross-binary churn and literal v1.1.8 upgrade rehearsals are green on top of these.

### DIFF
--- a/crates/ika-core/src/sui_connector/sui_executor.rs
+++ b/crates/ika-core/src/sui_connector/sui_executor.rs
@@ -879,6 +879,14 @@ where
                 // the cached gas ref survives the failure and every retry
                 // rebuilds the identical stale tx until the one-hour panic,
                 // wedging checkpoint submission.
+                //
+                // Match the variant payload — do NOT pass `err.to_string()`
+                // and especially not clippy's `unnecessary_to_owned`
+                // "simplification" of it, `err.as_ref()`: `IkaError` derives
+                // strum's `AsRefStr`, so `as_ref()` returns only the variant
+                // name ("SuiClientTxFailureGeneric"), which never contains
+                // the rejection markers — it compiles fine and silently
+                // disables this recovery.
                 if let IkaError::SuiClientTxFailureGeneric(_, message) = &err {
                     state.handle_possible_stale_gas_rejection(message);
                 }
--- a/crates/ika-core/src/sui_connector/sui_executor.rs
+++ b/crates/ika-core/src/sui_connector/sui_executor.rs
@@ -41,7 +41,7 @@ use sui_json_rpc_types::SuiTransactionBlockEffectsAPI;
 use sui_json_rpc_types::{SuiExecutionStatus, SuiTransactionBlockResponse};
 use sui_macros::fail_point_async;
 use sui_types::MOVE_STDLIB_PACKAGE_ID;
-use sui_types::base_types::{ObjectID, TransactionDigest};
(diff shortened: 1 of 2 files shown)


omersadika:
Resolved the conflict with `dev` after #1721 merged. Relevance triage of the six fixes:

- **Items 1–4 are already on dev** via #1721 in their evolved form (effects-tracked gas, cache-drop on failure, floored stale-gas re-fetch with the 30s catch-up cap, and the `advance_epoch` session-completion gate — the `system_inner_v1.rs` predicate was byte-identical). The resolution takes dev's `sui_executor.rs` for all of these.
- **Items 5–6 are genuinely missing from dev and survive as this PR's effective diff**: RPC-level submission rejections (`execute_transaction_block_with_effects` returning `Err`) bypassed the stale-gas recovery that the effects-level `errors` branch had — the common rejection shape would spin the full one-hour retry budget on the identical stale tx. The recovery is factored into `NotifierSubmitState::handle_possible_stale_gas_rejection` covering both paths, with the `err.as_ref()` clippy-trap comment guarding it.
- **Dropped**: the `debug!`→`info!` bumps on per-checkpoint submission logs — dev's post-#1721 observability deliberately keeps those at debug to avoid per-checkpoint info spam.

Build + clippy clean on the merge. The effective diff is now ~60 lines in one file.

PR #1735, opened 2026-06-12
fix(dwallet-mpc): network-key adoption gates — no committee-unapproved parameter set may go live

omersadika:
## Summary

Fixes the **false-malicious conviction** failure mode surfaced during the epoch-close wedge investigation (see PR #1721's "known issue" note): a validator that installs a different network-key parameter set than its peers honestly computes byte-divergent MPC outputs, and the output-quorum byte-equality tally convicts it as malicious — observed live, including the node logging `node recognized itself as malicious`. One conviction silently converts a 4-validator committee into zero-redundancy 3-of-4; the network keeps running degraded with no error-level signal until any further loss freezes MPC.

Root context: at every epoch-N entry, early-starting validators transiently see stale/incomplete overlay data (the epoch-entry stale-mpc_data race). Two adoption holes let that transient state become an installed parameter set:

1. **Empty-reconfiguration-output adoption** — an overlay entry whose reconfiguration output is transiently empty slipped through the initial-DKG adoption branch even when the prior epoch's handoff cert pins a reconfiguration digest for the key, instantiating DKG-derived parameters the committee never agreed to run this epoch. Now skipped (warn once per cert digest, deduped) and retried until the cert-pinned bytes resolve locally — the overlay re-merges every sync tick and the prepare-then-start barrier installs the pinned blob by digest.
2. **Stale-epoch pre-spawn rejection** — adopted data whose `current_epoch` metadata mismatches the manager's epoch was only rejected ~10s *after* the parameter derivation burnt on the rayon pool, and the doomed in-flight instantiation blocked the same key's correct data behind it, widening the entry key gap during which sessions park. Now rejected before spawning.

Plus an observability fix that blinded two post-mortems: a session whose protocol-cryptographic-data generation fails was silently skipped every 20ms service tick (`.ok()?`). The skip semantics stay; it now logs once per session (deduped).

## Tests

- New `network_key_adoption.rs`: both gates with positive controls — the cert-pinned empty/mismatching/matching overlay progression (self-validating: if the cert weren't consumed, the mismatch assertion would fail via the blind-adopt path), and stale-vs-current epoch spawn discrimination.
- Regression battery green: `beyond_lock_target` 2/2, `computation_results_batch` 1/1, `missing_network_key` 1/1; clippy clean on touched code.
- Spec updated: `specs/handoff.md` certificate-consumption section documents both guards and why divergence ends in conviction.

## What this does NOT fix (tracked separately)

The residual epoch-entry race (validators still fetch the prior epoch's overlay at entry) and the CI-only "parked sessions never compute after the key installs" link — see the tracking issue.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/mod.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/mod.rs
@@ -8,6 +8,7 @@ mod message_before_event;
 mod missing_network_key;
 mod network_dkg;
 mod network_dkg_bwd_compat;
+mod network_key_adoption;
 mod network_owned_address_sign;
 mod noa_checkpoint;
 mod presign_consensus;
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/network_key_adoption.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/network_key_adoption.rs
@@ -1,0 +1,231 @@
+// Copyright (c) dWallet Labs, Ltd.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+//! Regression tests for the network-key adoption/instantiation
+//! determinism gates.
+//!
+//! Live-incident background (2026-06-12 localnet, run 3): at an epoch
+//! boundary one validator adopted overlay key data whose
+//! reconfiguration output didn't match what its peers instantiated,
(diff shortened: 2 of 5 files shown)


omersadika:
Tracking issue for the residual race + the CI-only parked-sessions link: #1736

PR #1739, opened 2026-06-12
fix(ika-node): prune per-epoch authority store directories

omersadika:
## Summary

Authority per-epoch stores (`<db-path>/live/store/epoch_<N>/`) were never pruned — validator disk usage grew without bound, one RocksDB directory per epoch, forever. Everything later epochs depend on lives in the `perpetual/` sibling by design (handoff certificates, epoch-keyed reconfiguration outputs, the blob mirror), so old epoch directories are dead weight once their epoch closes.

Reported via swykeai/ika#1 (credit: @mikroskeem); this is an independent implementation for the current `dev` architecture rather than a port — the design differences are deliberate and listed below.

## Design

- **Reuse the existing machinery**: `ConsensusStorePruner` is parameterized with a directory-name prefix and a metrics/log `kind`. Consensus layout = bare epoch number; authority layout = `epoch_<N>` with non-matching siblings (`perpetual/`, anything else) skipped silently. With kind `"consensus"` the metric names are byte-identical to the historical ones — dashboards keep working.
- **Default ON, retention 2** (`authority_db_retention_epochs`, config-overridable; very large value ≈ keep forever). The consensus store already prunes **by default with retention 0**, so a bounded default for the authority store is the conservative side of an *existing* node policy — leaving it opt-in would ship the bug to everyone who doesn't read release notes. Two prior epochs preserve a post-mortem window (2 days at mainnet's 24h epochs).
- **Safe ordering**: the epoch-transition notification fires only after `release_db_handles()` on the outgoing epoch store, so dropping is safe in every configuration including retention 0 (`safe_drop_db` + force-delete fallback as before).
- **Restart-dormancy fix for both pruners**: each pruner is now seeded with the node's boot epoch instead of 0. Previously the periodic tick pruned nothing until the first reconfiguration after every restart — up to a full epoch of accumulation — because the boundary started at epoch 0. (The original fork PR solved this with a disk scan; on `dev` the node knows its epoch at construction, so it's a parameter instead.)

## Differences from swykeai/ika#1 (for reviewers who compare)

1. Default-on retention instead of opt-in (rationale above).
2. Boot-epoch seeding via constructor parameter instead of scanning the directory for the max epoch.
3. No `node_config_builder` plumbing needed — serde defaults cover the swarm/test configs.
4. Adapted to dev's post-#1721 reconfiguration flow (prune notification placed after the handles release in `monitor_reconfiguration`).

## Tests

- Authority-layout pruning: prefix filtering, `perpetual/`/garbage/other siblings untouched, retention math.
- Existing consensus-layout cases unchanged and passing.
- Seeded-boundary tick path.
- 4/4 green in release mode; ika-node/ika-config build + clippy clean on touched code.

### DIFF
--- a/crates/ika-config/src/node.rs
+++ b/crates/ika-config/src/node.rs
@@ -294,7 +294,30 @@ pub struct NodeConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub run_with_range: Option<RunWithRange>,
+
+    /// Retention window for per-epoch authority store directories
+    /// (`<db-path>/live/store/epoch_<N>/`): the current epoch plus this
+    /// many prior epochs are kept; older directories are deleted at each
+    /// epoch transition and on a periodic tick. Defaults to
+    /// `DEFAULT_AUTHORITY_DB_RETENTION_EPOCHS` when unset — without
+    /// pruning these directories grow unbounded (everything a later epoch
+    /// needs lives in the `perpetual/` sibling, which is never touched).
+    /// Set a very large value to effectively retain everything.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub authority_db_retention_epochs: Option<u64>,
+
+    /// Period of the authority store pruner's periodic tick. Defaults to
+    /// one hour when unset.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub authority_db_pruner_period_secs: Option<u64>,
 }
+
+/// Keep the current epoch plus this many prior per-epoch authority store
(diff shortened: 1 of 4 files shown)


omersadika:
Correction to the PR description's difference-item 3: `node_config_builder` plumbing WAS needed after all — `NodeConfig` is directly instantiated there (two sites), which serde defaults don't cover; the first cluster dispatch caught it as a build failure (E0063). Both initializers now set the new fields to `None` (→ defaults), full-workspace check is clean, and the cluster suite is re-dispatched on the rebased branch (now atop #1734's merge).

PR #1741, opened 2026-06-12
fix(reconfiguration): carry forward stable mpc_data so a late member doesn't wedge the next epoch (#1742)

### LINKED ISSUE #1742: v4 liveness: a committee member that misses the mpc_data freeze permanently wedges the next epoch's reconfiguration
# v4 liveness: a committee member that misses the mpc_data freeze permanently wedges the next epoch's reconfiguration

## Summary

A current-committee validator that enters epoch N late (e.g. restart recovery near the N-1→N boundary — a routine production event) can miss the epoch-N mpc_data announcement freeze. The frozen set then lacks its bundle while the chain's next-epoch committee still lists it, and `ReconfigurationParty::generate_public_input` (`crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs:94`) requires full bundle coverage of the chain committee — so **every validator deterministically rejects the reconfiguration session**. The frozen set is immutable for the epoch, the re-pulled request re-rejects forever, `all_network_encryption_keys_reconfiguration_completed` never becomes true, and epoch N can never close. Permanent network wedge.

Found by the restart-mid-grace cluster test (PR #1741), run https://github.com/dwallet-labs/ika/actions/runs/27450785655 (artifact `cluster-tests-log-1`).

## Evidence chain (timestamps from the run)

- `00:31:44` validator X (`b7583c13`) stopped after the epoch-1 freeze (test-induced; down ~2.5 min total). Validator Y (`89047f09`) stopped `00:33:58` inside the EndOfPublish grace, restarted ~10s later.
- Epoch 1 closed correctly across the restarts — handoff cert formed (`signers=3 items=6`) and verified; this part of the protocol behaved.
- `00:35:23` the three fast validators submit their **epoch-2** mpc_data announcements (Y recovered in time). X is still bootstrapping epoch 2 — its last announcement submission is the **epoch-1** blob at `00:34:46`; its epoch-2 generation never completes in time.
- `00:35:54` epoch-2 freeze fires via the grace path: `freezing attestation-validated mpc_data input set current_epoch=2 frozen=3 excluded=0 excluded_set=[]` → `assembled committee mpc_data off-chain epoch=3 members=3`.
- `00:35:40+` all four validators (including X) reject the epoch-2 reconfiguration: `InvalidMPCPartyType("at reconfiguration_message_version == 3 every upcoming committee member must publish the post-PR-#1707 bundle shape, but only 3/4 class-groups, 3/4 secp256k1 PVSS, 3/4 secp256r1 PVSS, 3/4 ristretto PVSS keys decoded")` → `rejecting session.` (`dwallet_mpc_service.rs:1802`) → rejection output reaches quorum at `00:40:39`.
- The EndOfPublish gate (`sui_syncer.rs:1085`) reports `all_immediate_sessions_completed=false all_network_encryption_keys_reconfiguration_completed=false` from then to the end of the run (~8 min) with no recovery path: the frozen set cannot change within the epoch.

## Why the check fires

The coverage check at `reconfiguration.rs:94-107` was written as a migration guard (old-bundle-shape validators must not silently land with empty PVSS entries). But `expected = upcoming_committee.voting_rights.len()` is the **chain-side** membership (staking — 4), while the bundles come from the **off-chain frozen set** (3). Any committee member absent from the frozen set — missed window, not just wrong shape — now trips the same fail-loudly path, and unlike the migration case there is no operator action that can ever satisfy it within the epoch.

## Fix directions

**A (principled, matches existing semantics):** the reconfiguration's upcoming committee should be the chain next-committee **∩ frozen set** — the exact deterministic intersection `HandoffSignatureSender::send` already uses for handoff-attestation membership, and the same semantics as historical on-chain bad-mpc_data exclusion. The late validator gets no epoch-N+1 share (sits out MPC for one epoch), re-announces in N+1, self-heals for N+2. Absent-announcers should land in the same `excluded_set` that attestation-validation failures do, so all consumers (handoff membership, reconfiguration target, joiner verifier) read one consistent membership.

**B (defense-in-depth, optional):** treat current-committee members differently from joiners at the freeze: extend (not remove — a permanently dead validator must not stall every epoch) the wait for a missing *member* announcement beyond the joiner grace, e.g. a separate member-specific freeze grace.

**C (rejected):** just lengthening `mpc_data_freeze_grace_rounds` — narrows the window but any sufficiently late entry still kills the network.

## Scope

v4-only (`off_chain_validator_metadata_enabled`) — blocks the off-chain metadata rollout, does not affect live deployments. Related but distinct from #1736 (epoch-entry stale-mpc_data race: that's about *stale* data adoption at entry; this is about *absent* announcements at the freeze). PR #1741's test is the regression harness: it currently fails on exactly this wedge and should go green when the fix lands.

omersadika:
Fixes #1742. Bundles the regression test (which found the bug) with the fix.

## The bug

A committee member that enters an epoch late — a routine restart near the boundary — can miss that epoch's mpc_data announcement. The freeze then finalizes a frozen set that omits it while the chain still seats it, and every validator deterministically rejects the reconfiguration into the next epoch because a seated member's class-groups material is missing. The frozen set is immutable for the epoch, so the rejection repeats forever and the epoch never closes — a permanent wedge. The same latent wedge hits any excluded member, not only the missed-freeze case.

## Why the obvious fixes are wrong (investigated, see `dev-docs/plans/1742-mpc-data-carry-forward.md`)

- **Reducing the upcoming committee's `voting_rights` to the frozen set** reintroduces an already-fixed bug: narrowing committee membership makes honest handoff certs unverifiable by the joiners they certify (`handoff_cert.rs`).
- **Resharing to the frozen subset (M of N)** is unsafe: share decryption/signing uses the chain committee's access structure, not the key's own (which isn't even persisted), so a subset reshare decrypts/samples against the wrong structure.
- **Waiting for the chain to drop a dead member** doesn't work: committee selection is top-N by stake; there is no uptime eviction.

## The fix: carry forward stable mpc_data

A validator's mpc_data blob is a pure function of its root seed (`derive_mpc_data_blob`) — byte-identical every epoch. At the freeze, any committee member not freshly attested this epoch but present in the prior epoch's handoff certificate (its `ValidatorMpcData` items) is frozen at that prior digest; the bytes resolve from the perpetual `mpc_artifact_blobs` store. Only first-time joiners (no prior-cert digest) remain excludable — matching the announcements spec's intent.

Complete, not partial: because the carried digest re-enters this epoch's certificate, coverage chains across epochs, so even a permanently-down-but-staked member never wedges reconfiguration. Deterministic: the prior certificate is consensus-anchored, perpetual, and held by every validator before it processes this epoch's consensus (the prepare-then-start barrier). The reconfiguration guard stays as the safety net and should never fire in steady state.

## Changes

- `carry_forward_stable_mpc_data` (pure, unit-tested) + `prior_epoch_mpc_data_digests` reader on the epoch store, wired into `freeze_mpc_data_if_first`.
- Spec updated (`dev-docs/specs/validator-mpc-data-announcements.md`).
- **Regression test** `test_validator_restart_mid_end_of_publish_grace` (the restart-mid-grace cluster test) — strikes a validator mid-grace, restarts it, and asserts byte-identical close + handoff certs across all validators and one more full reconfiguration. It failed on exactly this wedge before the fix; CI run validating it goes green is linked in comments.
- Two read accessors (`end_of_publish_quorum_round`, `is_epoch_close_emitted`) for the test's strike-window polling.

### DIFF
--- a/.github/workflows/test-cluster.yaml
+++ b/.github/workflows/test-cluster.yaml
@@ -41,6 +41,11 @@ on:
         type: string
         required: false
         default: "4"
+      rust_log:
+        description: "RUST_LOG for the test run (default error; scope debug to specific modules — full-debug replays are multi-GB)"
+        type: string
+        required: false
+        default: "error"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -52,7 +57,7 @@ env:
   CARGO_NET_RETRY: 10
   RUSTUP_MAX_RETRIES: 10
   RUST_BACKTRACE: 1
-  RUST_LOG: error
+  RUST_LOG: ${{ inputs.rust_log || 'error' }}
   rust_stable: "1.94"
 
 jobs:
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -3303,6 +3303,21 @@ impl AuthorityPerEpochStore {
             .collect())
     }
 
+    /// The consensus leader round at which this validator observed the
+    /// `EndOfPublish` stake quorum — the persisted anchor of the
+    /// deferred-close grace countdown — or `None` if quorum hasn't been
+    /// reached this epoch (or the epoch closed inline under v3 rules).
+    pub fn end_of_publish_quorum_round(&self) -> IkaResult<Option<u64>> {
+        Ok(self.tables()?.end_of_publish_quorum_round.get(&0)?)
+    }
+
+    /// Whether this validator has already emitted the epoch-close
+    /// checkpoint message set. Persisted through the emitting commit's
+    /// batch so a restart cannot re-emit the close at a later commit.
+    pub fn is_epoch_close_emitted(&self) -> IkaResult<bool> {
+        Ok(self.tables()?.epoch_close_emitted.get(&0)?.is_some())
+    }
+
     pub async fn user_certs_closed_notify(&self) {
         self.user_certs_closed_notify.wait().await
     }
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -3221,6 +3221,49 @@ impl AuthorityPerEpochStore {
     /// announcer who withheld their blob from honest peers can't
     /// be smuggled into the working set, even if they signed a
     /// valid announcement digest.
+    /// The prior epoch's handoff-certificate `validator -> mpc_data
+    /// digest` map, used to carry forward stable mpc_data for committee
+    /// members that did not freshly announce this epoch (see
+    /// `carry_forward_stable_mpc_data`). Empty when there is no prior
+    /// certificate (genesis epoch, v3, or a read error) — carry-forward
+    /// then degrades to announce-only. The certificate is perpetual and
+    /// stake-quorum-signed, and the prepare-then-start barrier guarantees
+    /// every validator holds the prior-epoch certificate before
+    /// processing this epoch's consensus, so the returned map is
+    /// identical across honest validators — a precondition for the freeze
+    /// staying deterministic.
+    fn prior_epoch_mpc_data_digests(&self) -> HashMap<AuthorityName, [u8; 32]> {
+        let Some(prior_epoch) = self.epoch().checked_sub(1) else {
+            return HashMap::new();
+        };
+        let Some(perpetual) = self.perpetual_tables_for_handoff.load_full() else {
+            return HashMap::new();
+        };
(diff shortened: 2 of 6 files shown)


omersadika:
✅ Validated green on CI: https://github.com/dwallet-labs/ika/actions/runs/27468735994

The restart-mid-grace cluster test passes with the carry-forward fix. Protocol evidence from the run: `carried_forward=1` at epochs 2–3 (X attested=3 → frozen=4, coverage chaining while X was down), **zero** `InvalidMPCPartyType` reconfiguration rejections, and clean advance through epochs 2→5. Before the fix the same test wedged at epoch 2 forever.

The no-fork assertion is now the perpetual struck-epoch handoff cert byte-equality across all four validators (robust to the network advancing while polling), replacing the checkpoint walk-back that raced restart-induced gaps + checkpoint pruning.

PR #1750, opened 2026-06-14
test(cluster): widen mid-epoch-joiner freeze window to fix flaky exclusion

omersadika:
## Summary

Fixes the intermittent failure of `test_joiner_lands_in_next_committee_class_groups` (the assertion that flaked the Test Cluster suite): the mid-epoch joiner reached epoch 2 but its `class_groups_public_keys_and_proofs` entry was missing from the epoch-2 committee.

## Root cause — a test time-compression artifact, not a production bug

A mid-epoch joiner is captured only if its mpc_data is decode-validated by a **stake quorum** inside the freeze window `[epoch/2 (V_{e+1} published) → 3·epoch/4 (ready-signal deadline)]` = `epoch/4`. Inside that window the path is: fan out → relay → peer fetch → **class-groups decode-validate** → re-emit ready signals (`mpc_data_announcement_sender::decide_ready_to_finalize`; freeze in `authority_per_epoch_store.rs`). `epoch_scaled_poll_interval` compresses the *poll cadences* with the epoch, but the per-peer **decode-validate is a fixed multi-second crypto cost that does not compress**.

At a 120s test epoch that window is **30s**. Under the Test Cluster suite's 4-way parallelism — four full Sui+ika swarms contending for CPU — the quorum decode-validate intermittently overran 30s, so the `3·epoch/4` deadline forced the ready signals (and the freeze) **without** the joiner. That exclusion is **correct by design**: reconfiguration never blocks waiting on a joiner (the liveness rule hardened in #1741), and at production epoch lengths the window is *minutes*, so this never happens in production.

## Fix

Double the test epoch to **240s** (freeze window → 60s, registration window → a roomy 120s), staying well under the epoch length at which the network-key DKG starts to stall. This deterministically gives the propagation enough wall-clock — it is **not** a retry band-aid. Also adds a `dev-docs/learnings/pitfalls.md` entry: in time-compressed epoch tests, every epoch-*fraction* window must exceed the *absolute* (non-compressing) crypto cost inside it, with margin for CI contention.

No production code changes; the reconfiguration mechanism is untouched.

## Verification
- `cargo check -p ika-test-cluster --tests` → clean.
- Passes locally at 240s (394s, joiner captured).
- Flake resolution to be confirmed across **repeated** Test Cluster CI runs (a single pass can't prove a flake fix).

### DIFF
--- a/crates/ika-test-cluster/tests/joiner.rs
+++ b/crates/ika-test-cluster/tests/joiner.rs
@@ -90,16 +90,29 @@ async fn test_joiner_lands_in_next_committee_class_groups() {
     //      keygen (a fixed, multi-second cost) and land `add_validator`
     //      on-chain so it's selected into `V_{e+1}`. This is gated by
     //      crypto/tx time, NOT by poll cadence, so it needs absolute
-    //      wall-clock — a 60s epoch (30s window) is too tight.
+    //      wall-clock.
     //   2. Freeze `[epoch/2 → 3·epoch/4]`: fan out → relay → fetch →
-    //      decode-validate → re-emit, so the freeze captures its
-    //      mpc_data. `epoch_scaled_poll_interval` shrinks this path's
-    //      cadences to fit the window.
-    // 120s gives a 60s registration window and a 30s freeze window —
-    // both comfortable.
+    //      decode-validate → re-emit, so a stake quorum's ready signals
+    //      cover the joiner before the `3·epoch/4` deadline forces the
+    //      freeze without it. `epoch_scaled_poll_interval` shrinks this
+    //      path's cadences to fit the window, but the per-peer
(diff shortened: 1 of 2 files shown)


omersadika:
Validation complete: the widened freeze window holds under the 4-way contention that caused the flake.

- Local (240s epoch): pass (394s)
- Full Test Cluster suite on CI, 3/3 green:
  - https://github.com/dwallet-labs/ika/actions/runs/27493231311
  - https://github.com/dwallet-labs/ika/actions/runs/27494806233
  - https://github.com/dwallet-labs/ika/actions/runs/27496110856

The previously ~1-in-5 `test_joiner_lands_in_next_committee_class_groups` exclusion did not recur across these runs.

PR #1756, opened 2026-06-17
Strict bare-only on-chain + off-chain genesis PVSS (Sui keeps only the mainnet-v1.1.8 shape)

ycscaly:
## What

Brings the **strict, bare-only-on-chain** architecture: Sui stores **only** the bare mainnet-v1.1.8 `ClassGroupsEncryptionKeyAndProof` shape; the full validator bundle (3 PVSS HPKE keys + the Fast Schnorr VSS HPKE key) travels **off-chain** via validator P2P announcements; all on-chain decode sites are strict.

## Why

The hard constraint: **Sui must only ever hold the old mainnet-v1.1.8 bare shape — never the PVSS bundle.** `fast-schnorr` currently doesn't fully honor that:

- its CLI publish (`become-candidate` / `set-next-epoch-mpc-data`) still has the version-gated **bundle-on-chain** branch — by default it writes the PVSS bundle to Sui (constraint violation);
- it restored **shape-tolerant** on-chain decode — pointless if Sui is always bare;
- with a bare genesis publish and no off-chain genesis bootstrap, the **genesis network-key DKG wedges** (`InvalidMPCPartyType … 0/4 … PVSS keys decoded`) and the network never advances past epoch 1.

## How

- **Strict decode everywhere** (Sui is always bare); tolerant decoder removed.
- **Bare-only publish** at both genesis and the CLI (the bundle-on-chain branch is gone).
- **Off-chain genesis-PVSS fix** — `refresh_validator_keys_from_offchain`: a network DKG is a within-epoch system session, so once the off-chain-assembled committee (same announcements) is available with the same member set, the validators' PVSS keys are copied into the current per-party map. No-op at v3 / once complete.

## Verification

- **TS integration suite: green** on CI (`Failed: 0`).
- **Local:** fresh localnet reaches readiness — genesis DKG completes (`quorum=4`), the fix fires on all 4 validators (`refreshed=4`), **zero** `PVSS keys decoded` errors.
- Workspace compiles clean.

This branch merges `fast-schnorr` (keeping its history + review doc) but keeps our code on every conflict, so it's a clean base for `fast-schnorr` to adopt the bare-only-on-chain approach.


crates/ika-upgrade-test/tests/malicious_cross_binary.rs
ycscaly, line 1: How can this work - we need to actively change the code to compile the faulty binary, also its much better to be able to get the malicious parties and assert the value from the test rather than have to externally look at the logs.

crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
```diff
@@ -507,6 +507,18 @@ impl DWalletMPCService {
...
+        // before the announcements land, so refresh the per-party key map from
+        // the off-chain-assembled committee once it's available — otherwise the
+        // genesis network-key DKG wedges on a stale 0-PVSS snapshot. No-op at v3
+        // / once complete (see the method).
+        if let Err(e) = self
+            .dwallet_mpc_manager
+            .refresh_validator_keys_from_offchain()
+        {
+            error!(error = ?e, "failed to refresh validator MPC keys from off-chain announcements");
+        }
```
ycscaly: I dont like that we initialize wrong values from sui, then override them. It is much better to either initialize to None, or to move the field from the committee struct to be initialized at a later time when we have the infromation.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -730,6 +730,101 @@ impl DWalletMPCManager {
...
+    /// (`next_active_committee`, built from the same announcements) is available
+    /// AND has the same member set as the current committee, copy its
+    /// per-validator key map (by authority name → current party id) into the
+    /// current per-party map so the DKG sees the full PVSS set.
+    ///
+    /// No-op at v3 (Sui's bare shape is correct there), once the per-party PVSS
+    /// maps are already complete, while the assembled committee is unavailable
+    /// or still converging, or if its member set differs from the current one —
+    /// so it is safe to call every service-loop iteration.
+    pub(crate) fn refresh_validator_keys_from_offchain(&mut self) -> DwalletMPCResult<()> {
```
ycscaly: I really dont like this implementation. Firstly ti seems you conflate the current and next committee, and second the whole impl is ugly and weird and looks buggy

ycscaly (commented):
Everything besides the actual validator key refresh mechanism is ok

PR #1770, opened 2026-06-23
fix(ika-upgrade-test): bound spawned-node rayon pools so co-located validators don't starve the CI runner

omersadika:
## Symptom
The **Upgrade Test / cross_binary** run on `dev` died twice (runs [28023184063](https://github.com/dwallet-labs/ika/actions/runs/28023184063), [28026237741](https://github.com/dwallet-labs/ika/actions/runs/28026237741)) — identical signature each time: **"the self-hosted runner lost communication with the server … starves it for CPU/Memory"**, ~39 min in, no failed assertion, no artifacts. Reproduced deterministically, so not an infra flake.

## Root cause
The harness packs `num_validators` (4) validator processes **+ the notifier** onto one host, each running the real node binary built `--no-default-features` — which turns **off** `enforce-minimum-cpu`, so per [`ika-core/src/runtime.rs`](crates/ika-core/src/runtime.rs#L21) the node's global rayon pool defaults to **all cores** (no `num_threads`). That was harmless while crypto barely touched rayon — but **#1714 activated the `parallel` crypto feature in non-msim builds**, so those N unbounded pools now genuinely saturate the CPU N×, starving the async runtimes hard enough to kill the runner agent.

Corroboration that #1714's parallel crypto is the trigger: **v118_upgrade passed** on the same harness, because its OLD binary is the mainnet-1.1.8 tag (pre-`parallel`-crypto, doesn't saturate) — whereas cross_binary builds its OLD binary from the current branch (WITH parallel crypto), so *every* validator saturates.

This is the **out-of-process analogue of #1764** (which reserved cores for the *in-process* swarm); the out-of-process harness never got the equivalent bound. CI-harness/local only — production runs one validator per host and is unaffected.

## Fix
Give each spawned node a `RAYON_NUM_THREADS` budget — a fair slice of ~75% of the host's cores, reserving the rest for async/IO across the processes — set on every child `Command` in `ValidatorProcess::start` (so it also covers binary swaps and joiners). In the `--no-default-features` build, `runtime.rs` calls `build_global()` without an explicit thread count, so rayon honors `RAYON_NUM_THREADS`.

`cargo check -p ika-upgrade-test --all-targets`: clean.

## Validation
cross_binary dispatched on this branch ([28029297221](https://github.com/dwallet-labs/ika/actions/runs/28029297221)) — **in flight**. Expect it to clear the ~39-min starvation point and complete (crypto is slower under the bound but the host survives). I'll update once it lands.

### DIFF
--- a/crates/ika-upgrade-test/src/cluster.rs
+++ b/crates/ika-upgrade-test/src/cluster.rs
@@ -265,6 +265,10 @@ impl ClusterBuilder {
             .collect();
 
         // 4. Per-validator NodeConfig on a persistent data dir, written to YAML.
+        // Bound every co-located node's crypto rayon pool to a fair share of the
+        // host so the validators + notifier don't oversubscribe the CPU (see
+        // `rayon_threads_per_node`).
+        let rayon_threads = rayon_threads_per_node(self.num_validators + 1);
         let mut validators = Vec::with_capacity(self.num_validators);
         for (i, init) in validator_init_configs.iter().enumerate() {
             let data_dir = base.join(format!("validator-{i}"));
@@ -286,6 +290,7 @@ impl ClusterBuilder {
                 self.validator_binary.clone(),
                 &node_config,
                 data_dir.clone(),
+                rayon_threads,
             )
             .await?;
             validators.push(proc);
@@ -316,6 +321,7 @@ impl ClusterBuilder {
             self.notifier_binary.clone(),
             &notifier_config,
             notifier_dir,
+            rayon_threads,
         )
         .await?;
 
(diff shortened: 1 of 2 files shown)


omersadika:
✅ **Validation passed.** cross_binary on this branch ([28029297221](https://github.com/dwallet-labs/ika/actions/runs/28029297221)) completed **success in 38 min** — the runner survived (no "lost communication"), vs. two deterministic deaths at ~39 min on `dev`. The bound added no wall-clock cost (38m green vs. 39m death), confirming the failure was CPU starvation, not crypto throughput. Ready for review.

Issue #1736, opened 2026-06-12
Epoch-entry stale-mpc_data race: per-epoch silent dead-presign validator; two laggards wedge the epoch


omersadika:
## Summary

At **every** epoch-N entry, validators whose per-epoch MPC service starts ~1s early fetch the **prior** epoch's mpc_data overlay before the syncer publishes epoch-N's. Established by a five-angle forensic analysis of TS-suite run 27417439758's localnet log, cross-checked against three local reproduction runs.

Two fixes already landed/in-flight (see PR), but two parts remain open:

## 1. Parked sessions never compute after the key installs (CI-only, the actual wedge)

During the laggard's key gap, the entry presign top-up creates its internal-presign sessions in a parked state; after the correct key installs they are re-created active — and in the CI environment they then **never compute** (zero CPU, zero outputs, no errors). One dead-presign validator per epoch is routine and invisible (3-of-4 still quorums — the forensics found a *different* silently-dead validator in every healthy epoch). When TWO validators lose the race in one epoch: 2 outputs < quorum 3 → the presign pool never refills → completions stall → the epoch-close lock target pins at `completed+100` → undershoot → the epoch never closes → SDK-visible "Object does not exist" timeout cascades. This is almost certainly the historical ~1-in-3 heavy-TS-suite flake.

**Open question**: which link breaks — instantiate → spawn → complete — on the laggard. The sessions demonstrably recover in local reproduction rigs; only CI shows the dead state. Diagnosis needs one *failing* run with `ika_core::dwallet_mpc=debug` localnet logs (the workflow input `localnet_rust_log` exists for exactly this; a healthy-run debug baseline is in run 27424269629's artifact). Code-level suspects already located: the (now-logged) generation-failure skip in `perform_cryptographic_computation`, and orchestrator slot-leaks via completion-updates spawned onto a tokio handle (started-never-completed `currently_running` entries were observed locally).

## 2. The double-fetch itself

Even with the stale-epoch pre-spawn rejection landed, validators still fetch and adopt the prior epoch's overlay at entry before re-syncing the correct one. The clean fix is for the entry overlay publication to be epoch-gated at the source (syncer/prepare-barrier), so the adoption pass never sees epoch-(N-1) data inside epoch N.

## 3. Hardening note: the byte-equality tally converts honest divergence into conviction

[...]

PR #1777, opened 2026-06-29
feat(ocs): verify checkpoints against the committee before folding into the direct cache

omersadika:
## Why

A `sui-state-direct` validator serves OCS reads **cache-first**, and that cache is populated by `IkaCheckpointPusher` folding its **own fullnode's** checkpoints. Before this change, `build_entries` folded the fullnode's `CheckpointData` — summary, modified-objects tree, and inclusion proofs — into the cache with **no committee verification**.

So a direct node's cache-served reads — including the always-cached `System` / `DWalletCoordinator` singleton anchors on the MPC hot path — were rooted in its **own fullnode, not the Sui committee**. A compromised or buggy fullnode could seed the cache with forged state that cache-first reads served undetected; only mirrored/peer-only nodes (which verify every read) were protected.

This violated spec invariant 8 (*"cached state is committee-verified before it enters the cache"*) and the documented promise that `verify_summary` is the single chokepoint the direct-side folder shares with the reader — both were intent the code didn't enforce.

## What

`verify_before_fold`, run once per checkpoint at **fold time** (off the read hot path) before `absorb_entries`:

1. **Committee BLS** on the summary, via the shared `verify_summary` chokepoint (reader and folder can't drift apart).
2. **Artifacts-digest binding** (when objects are folded): the locally-built `tree_root` must reproduce the committee-signed `checkpoint_artifacts_digest` — a real root means every inclusion proof the pusher derived from that tree is over genuine state.

A verify failure is **refused, not cached**: the error propagates so the fold cursor stays put — a transient missing committee self-heals on the next poll, a bad signature halts the fold loudly (`security_critical`). **The direct trust root is now the committee, matching the peer-only path.**

Cost is one BLS verify per ika-touching checkpoint — *not* per read — so zero read-path latency added.

## Tests

Fault-injected per the test-testing rule (each proven to trip when its check is neutered, then restored):
- a foldable checkpoint signed by a **foreign committee** is refused end-to-end through `advance()` (head doesn't move, cursor stays, cache empty);
- a summary committing to a **mismatched artifacts digest** is rejected by `verify_before_fold`.

7/7 push_worker tests + 98/98 `sui_connector` tests green; clippy clean.

## Notes

- **Spec** (`ocs-verified-sui-reads.md`): cache-fast-path trust root flips from `own gRPC uplink` → `committee (verified at fold)`; invariant 8 cites the enforcing mechanism. Spec already declared this behavior; the code was the lagging side — now they agree.
- **Cargo.lock**: locks the cold-bootstrap archive deps (`object_store`/`sui-storage`/`sui-config`) the parent branch added to `ika-sui-client/Cargo.toml` but left out of the lock, so this branch builds `--locked`. Worth fixing on the parent PR too.
- Stacked on `feat/ocs-archive-cold-bootstrap` — review/merge that first.

### DIFF
--- a/crates/ika-core/src/sui_connector/ocs_metrics.rs
+++ b/crates/ika-core/src/sui_connector/ocs_metrics.rs
@@ -11,9 +11,10 @@
 use std::sync::Arc;
 
 use prometheus::{
-    HistogramVec, IntCounter, IntCounterVec, IntGauge, Registry,
-    register_histogram_vec_with_registry, register_int_counter_vec_with_registry,
-    register_int_counter_with_registry, register_int_gauge_with_registry,
+    Histogram, HistogramVec, IntCounter, IntCounterVec, IntGauge, Registry,
+    register_histogram_vec_with_registry, register_histogram_with_registry,
+    register_int_counter_vec_with_registry, register_int_counter_with_registry,
+    register_int_gauge_with_registry,
 };
 
 #[derive(Clone, Debug)]
@@ -39,6 +40,12 @@ pub struct OcsMetrics {
     pub pusher_pushed_total: IntCounter,
     pub pusher_skipped_irrelevant_total: IntCounter,
     pub pusher_fetch_failures_total: IntCounter,
+    /// Latency of the pre-fold committee verification the pusher runs before
+    /// folding a checkpoint into the local cache (committee BLS on the summary
+    /// + artifacts-digest binding). `_count` is the number of checkpoints
+    /// committee-verified before folding, `_sum` the total CPU time spent on
+    /// it — together they quantify the verify's cost on the fold path.
+    pub pusher_fold_verify_seconds: Histogram,
 
     // OcsVerifiedReader (consumer-side proof verification)
     pub proof_verify_total: IntCounterVec, // labels: ["kind"]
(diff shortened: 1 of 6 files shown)


omersadika:
## CI status + forensic verdict on the TS-integration failure

**Rust integration ✅ · Cluster ✅ · TS integration ❌ — the TS failure is pre-existing [#1736](https://github.com/dwallet-labs/ika/issues/1736), not this PR.**

The two heavy suites that actually exercise this change's fold path (Rust integration, Cluster) are green. The TS suite wedges in the heaviest file (`all-combinations-future-sign`) with `Object … does not exist` / state-never-reached timeouts — the documented #1736 epoch-boundary MPC-participation stall.

A 9-agent adversarial forensic sweep over the run logs (with a skeptic explicitly tasked to pin the failure on this change) exonerated it to high confidence:

- **The verify rejected nothing.** `verify_before_fold` never fired its reject path — 0 `refusing to fold`, 0 `failed committee verification`, 0 `security_critical`; the fold cursor advanced monotonically (569→21266). The new committee check never blocked a fold.
- **CPU-steal refuted.** The hypothesis that the added per-checkpoint BLS verify stole cycles and tipped the timing race fails on the data: at the wedge the runner was **idle, not contended** (effective_cpus ≈ 0.34, loadavg ≈ 0.75, ~15 cores free, cgroup quota never hit). Worst-case all-folds cost ≈ 7.5% of one core.
- **Branch-independent.** The immediate parent branch `feat/ocs-archive-cold-bootstrap` (run `28330463150`), which lacks this change, shows the **identical** wedge signature (`No last read consensus round`=20, `FailedToAdvanceMPC`=180 in per-boundary bursts, tops out at `run_epoch=5`).
- **Mechanism = #1736 residual variant:** epoch-6 reconfig prepares (`mpc_data … frozen=true epoch=6`), executor logs `Successfully locked last session in current epoch` but never calls `process_request_advance_epoch()`; sessions queue but never reach quorum; `run_epoch epoch=6` never appears. Silent liveness stall (0 ERROR, 0 panics).

Runs: original `28386601258`, re-run `28394110107`, parent baseline `28330463150` — all three wedge the same way, so re-running won't clear it; the blocker is org-wide #1736 (the genuine-laggard-participation variant that #1761's partial fix does not cure), independent of the OCS stack.

This PR also adds `ika_ocs_pusher_fold_verify_seconds` (count = checkpoints committee-verified before folding, sum = CPU the verify added) so the fold-verify cost is measured directly rather than inferred.

ycscaly (approved):

PR #1789, opened 2026-07-02
fix(dwallet-mpc): derive NetworkKeyId on demand to break the joiner adoption deadlock

omersadika:
Fixes the joiner network-key bootstrap deadlock that failed `v118_churn` in [run 28566288225](https://github.com/dwallet-labs/ika/actions/runs/28566288225) (release-review finding V6), plus the self-masking diagnostic that hid it.

## The deadlock

The handoff cert keys network keys by content-derived `NetworkKeyId`; the runtime keys them by `ObjectID`. The temporary translation map is populated only from the hardcoded `SEED` (testnet/mainnet keys) or by `register()` **at instantiation**. A validator consuming a cert for a key it never instantiated — a joiner on any fresh-DKG network, or any validator restarted mid-epoch on a non-seeded key — has neither:

- **adoption** needs the mapping to see the cert's digests (`cert_dkg_digest=None` otherwise → phantom "digest mismatch" → skip),
- the mapping registers at **instantiation**,
- instantiation requires **adoption**.

In `v118_churn` the joiner (validator-4) spun in exactly this cycle from its epoch-4 entry until the job timed out: never instantiated, canonical-version gauge stuck at 0. `cross_binary`'s joiner passed the same flow only by racing — its first adoption pass ran *before* its cert anchored, so it took the no-cert fallback and instantiated. Any joiner scenario on a non-seeded key can flake either way on that race.

A second hazard hid behind the same gap: an unmapped key was treated as "not pinned by the cert", so on the initial-DKG branch it could be **adopted without the cert gate** — parameters the committee never agreed on, the false-malicious-conviction divergence class.

## The fix

1. **Derive-on-demand** (`mpc_manager.rs` + `network_dkg.rs`): when the cert references keys and an overlay key's `ObjectID` has no mapping, the adoption pass spawns a background derivation of the `NetworkKeyId` from the locally-held blobs (same source instantiation uses: reconfiguration output when present, else DKG output — the id is identical either way, resharing preserves the collective encryption key) on the rayon pool, and **defers** adoption for that key. The derivation registers into the process-global map; the next pass proceeds through the normal cert gate. Spawned at most once per key per manager (per epoch); a successful registration persists for the process, so later epochs never re-derive.
2. **Memoization skip**: the adoption pass memoizes on the overlay `Arc` + cert presence — but a deferred unmapped key resolves via the global mapping, which the memo cannot see. A pass that deferred skips the memo update so it re-evaluates every tick until the registration lands.
3. **Honest diagnostics** (`ika-node/src/lib.rs`): the prepare-then-start breakdown computed "missing" with `object_id_for(key_id)?` inside a `filter_map`, silently dropping exactly the unmapped keys wedging the gate — the stuck joiner logged the contradictory `missing_locally=0 missing_key_ids=[]` for 890 retries. Unmapped cert keys are now reported as a separate `unmapped_cert_keys` field.

The derivation function is the one `print_deployed_network_key_ids` already used (promoted from `#[cfg(test)]`), so the derived ids match the seeded/instantiation-computed ones by construction.

## v3 safety

The new branch triggers only when a handoff cert with items exists — certs are produced only under protocol v4 (`off_chain_validator_metadata_enabled`). At v3 and at the cert-less v3→v4 boundary the pass is byte-identical to before.

## Testing

- New regression test `unmapped_cert_referenced_key_defers_and_spawns_derivation`: cert-referenced unmapped key → deferred (not adopted), derivation spawned exactly once across repeated passes, memoization skipped, and registration alone (overlay bytes unchanged) unwedges adoption through the cert gate — the v118_churn wedge in miniature, plus proof the pass re-evaluates.
- Existing adoption regression tests unchanged (their manual `register` calls legitimately simulate previously-instantiated state).
- End-to-end proof: `v118_churn` on this branch (the failing scenario from the run above).

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
+++ b/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
@@ -15,9 +15,13 @@ use commitment::CommitmentSizedNumber;
 use dwallet_classgroups_types::{
     ClassGroupsDecryptionKey, RistrettoPvssDecryptionKey, Secp256k1PvssDecryptionKey,
 };
+use dwallet_mpc_centralized_party::{
+    network_dkg_public_output_to_protocol_pp_inner,
+    reconfiguration_public_output_to_protocol_pp_inner,
+};
 use dwallet_mpc_types::dwallet_mpc::{
     DWalletCurve, NetworkDecryptionKeyPublicOutputType, NetworkEncryptionKeyPublicData,
-    SerializedWrappedMPCPublicOutput, VersionedDecryptionKeyReconfigurationOutput,
+    NetworkKeyId, SerializedWrappedMPCPublicOutput, VersionedDecryptionKeyReconfigurationOutput,
     VersionedNetworkDkgOutput,
 };
 use group::PartyID;
@@ -848,8 +852,7 @@ pub(crate) fn compute_all_network_owned_address_dkg_outputs(
 /// from the reconfiguration output). The result matches what
 /// instantiation computes because both seed `compute_noa_dkg` on the
 /// curve25519 encryption-key parameters.
-#[cfg(test)]
-pub fn derive_curve25519_network_owned_address_public_key(
+pub(crate) fn derive_curve25519_network_owned_address_public_key(
     curve25519_protocol_public_parameters: &[u8],
 ) -> DwalletMPCResult<Vec<u8>> {
     let protocol_public_parameters: twopc_mpc::curve25519::class_groups::ProtocolPublicParameters =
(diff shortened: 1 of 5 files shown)


omersadika:
End-to-end proof: dispatched the exact failing scenario (`v118_churn`) on this branch — [run 28573322323](https://github.com/dwallet-labs/ika/actions/runs/28573322323) — **green**, and the logs show the fix path itself carried it, not the timing race that let `cross_binary` pass:

- `validator-4/node.log:341` @ 08:02:32 — the joiner hit the exact pre-fix wedge condition: *"handoff cert references network keys but this key's ObjectID has no NetworkKeyId mapping"* (the cert-anchored branch was in force, the losing timing from [the failed run](https://github.com/dwallet-labs/ika/actions/runs/28566288225)).
- `validator-4/node.log:429` @ 08:02:34 — *"derived and registered the NetworkKeyId from locally-held key data"* — the background derivation completed in ~1.5s and unwedged adoption.
- Zero occurrences of the old wedge signature (*"skipping adoption, the key stays uninstantiated"*) in any node log.
- Step 11/13 `expect_network_dkg_output_version_at_least(3)` — the step that previously timed out after 20 minutes — passed in **45.3s**.

PR #1791, opened 2026-07-02
fix(ika-config): old-style config keeps legacy JSON-RPC for every role

omersadika:
Rollout-day fix from the 1.2.0 release review (finding G5 in the [review report](https://github.com/dwallet-labs/ika/pull/1787#issuecomment-4862501346)): an old-style config (`sui-rpc-url` only, no `sui-data-source`) now selects the legacy JSON-RPC transport for **every role**, not just validators.

## The problem

`select_sui_transport` routed an old-style **notifier/fullnode** to gRPC at the same URL. That means the 1.1.8 → 1.2.0 restart silently changes the notifier's Sui transport under an unchanged config — and a `sui-rpc-url` endpoint that serves JSON-RPC is not guaranteed to serve Sui gRPC (many managed providers don't expose it, or expose it on a different port). The notifier is the sole submitter of checkpoints and `advance_epoch`; a soft gRPC failure there is a network-wide epoch-advance stall on rollout day.

## The fix

Old-style config keeps the transport it selected on 1.1.8 — JSON-RPC — for all roles. Moving to gRPC is an explicit, operator-initiated config migration: add `sui-data-source`. No silent behavior change on upgrade.

Feasibility verified before the change: the notifier's submission path works over the JSON-RPC backend — `SuiBackend::JsonRpc` has a live `execute_transaction_block_with_effects` arm dispatching to the quorum-driver API (ika-sui-client `lib.rs`), which is exactly what 1.1.8's notifier ran. And the syncer already gates legacy event ingestion on `run_legacy_event_ingestion && mode.is_validator()` (`sui_syncer.rs`), so a legacy-path notifier skips event listening just like 1.1.8.

Changes:
- `select_sui_transport`: old-style → `LegacyJsonRpc` unconditionally; comment documents the invariant (a binary upgrade must never flip a node's transport under an unchanged config).
- `SuiTransportPlan` docs updated; the `old_style_routes_by_role` test becomes `old_style_keeps_legacy_json_rpc_for_every_role` (all roles asserted).
- The ika-node deprecation warning now speaks for any role.
- `dev-docs/specs/ocs-verified-sui-reads.md`: config-shape gate table row updated in the same PR, per the spec-maintenance rule.

## Coverage note / follow-up

All 15 `ika-config` transport unit tests pass. The test-harness notifiers are all new-style (`node_config_builder` always writes `sui-data-source`), so no CI flow exercises the legacy notifier path today — that's the remaining half of review finding G5: a rehearsal scenario booting the new binary from old-style YAMLs in Validator **and** Notifier modes across an epoch boundary, coming as a follow-up PR.

### DIFF
--- a/crates/ika-config/src/node.rs
+++ b/crates/ika-config/src/node.rs
@@ -346,16 +346,21 @@ pub struct SuiCheckpointArchiveConfig {
 /// flag can't halt running validators en masse at an upgrade boundary.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum SuiTransportPlan {
-    /// Old-style config (no `sui-data-source`) on a validator: the deprecated
-    /// JSON-RPC read path — the only one that serves `query_events`, which a
-    /// validator needs for MPC event ingestion. Sui is sunsetting JSON-RPC.
+    /// Old-style config (no `sui-data-source`), any role: the deprecated
+    /// JSON-RPC path — the 1.1.8 behavior the config shape used to select.
+    /// A binary upgrade must not change a node's transport under an
+    /// unchanged config (an endpoint that serves JSON-RPC need not serve
+    /// Sui gRPC, and for a notifier that mismatch surfaces as a
+    /// network-wide epoch-advance stall); moving to gRPC is an explicit
+    /// config migration (`sui-data-source`). Also the only path serving
+    /// `query_events`, which a validator needs for MPC event ingestion.
+    /// Sui is sunsetting JSON-RPC.
     LegacyJsonRpc,
     /// `sui-state-mirrored` with no `fallback-grpc-url`: the node has no direct
     /// Sui uplink, so every read crosses the verified OCS relay.
     PeerOnlyRelay,
-    /// A direct gRPC uplink: `sui-state-direct`, `sui-state-mirrored` with a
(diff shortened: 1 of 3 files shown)


omersadika:
TS suite triage ([failed run 28577745311](https://github.com/dwallet-labs/ika/actions/runs/28577745311)): **not caused by this branch — the change provably never executed.** Zero `DEPRECATED: old-style config` warns in the localnet log (that warn fires unconditionally on the legacy path this PR touches), and all nodes log the gRPC/OCS stack — the TS localnet runs new-style configs, which this PR leaves byte-identical. The failure is the pre-existing epoch-close wedge (issue #1736) — and usefully, it's the failing instrumented run that issue was waiting for; evidence posted [there](https://github.com/dwallet-labs/ika/issues/1736#issuecomment-4865057287). Re-dispatching the suite.

For the record, this branch's suite status: integration ✅ cluster ✅ TS re-running.

omersadika:
TS re-run [28585622657](https://github.com/dwallet-labs/ika/actions/runs/28585622657): **green** — same branch, zero changes. Confirms the earlier failure was the pre-existing #1736 wedge, not this PR. Final suite record for this branch: integration ✅ cluster ✅ TS ✅.

PR #1792, opened 2026-07-02
test(upgrade): legacy_config rollout rehearsal — new binary on old-style YAMLs for every role

omersadika:
The second half of release-review finding G5 ([report](https://github.com/dwallet-labs/ika/pull/1787#issuecomment-4862501346)): a CI scenario that rehearses the **actual rollout-day shape** — the new binary booting from old-style (1.1.8-shape) YAMLs. **Depends on #1791** (this branch is based on it); mergeable once that lands.

## Why

Operators upgrade the binary, not the YAML: on rollout day every mainnet node — validators *and* the notifier — runs `sui-rpc-url`-only configs on the legacy JSON-RPC transport. No harness covered that: every existing scenario's config builder writes `sui-data-source` (gRPC) for all roles, so a regression confined to the legacy path (validator MPC event ingestion over `query_events`, or notifier checkpoint/`advance_epoch` submission over the JSON-RPC quorum driver) would ship invisibly. This is also the scenario that would have caught the transport flip #1791 fixes, had it existed.

## The scenario — `legacy_config`

`Scenario::with_legacy_sui_config()` boots all four validators **and the notifier** from old-style configs (no `sui-data-source`, no trust anchor), then:

1. `wait_for_epoch(2)` — genesis network DKG completes; epochs advancing at all is proof the legacy-path notifier submits (nothing advances without it);
2. `run_workload("legacy-v3")` — full DKG → Presign → Sign at protocol v3: the exact post-swap, pre-upgrade mainnet state;
3. buffer drop → `wait_for_epoch(4)` → `expect_protocol_version_at_least(4)` — **the v3→v4 capability vote succeeds on legacy configs**, settling a real rollout-sequencing question: operators do not need to migrate to `sui-data-source` before the v4 activation vote;
4. `run_workload("legacy-v4")` — the lifecycle keeps working at v4 on the legacy transport.

## Mechanics

- `ValidatorConfigBuilder`/`FullnodeConfigBuilder` gain `with_legacy_sui_rpc_only()`: emit `sui_rpc_url: Some(url)` + `sui_data_source: None` (the exact 1.1.8 YAML shape). All other callers unchanged.
- `ClusterBuilder::with_legacy_sui_config()` threads it to every validator and the notifier, and skips the OCS genesis-committee anchor seeding (an old-style config with an anchor is one of the mixed shapes the node rejects at boot — correctly fail-loud).
- Workflow: `legacy_config` added to the scenario matrix (`all` now fans six jobs), env gate `RUN_LEGACY_CONFIG`, no OLD binary needed (current-build-only, like `smoke`/`workload`).

## Follow-up once #1790 also lands

One-liner hardening: bracket the `legacy-v3` workload with `expect_protocol_version_at_most(3)` (the ceiling step #1790 introduces) so timing drift can't silently degrade the v3 leg into a second v4 run. In practice the window holds — at n=4 the default 50% buffer stake requires all four capability votes at the tally, which is exactly why the flow drops the buffer only *after* the v3 workload.

### DIFF
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -29,6 +29,12 @@ name: Upgrade Test
 #                  so the v4 reshare of the 1.1.8-origin key includes a new
 #                  party (exercises the OCS joiner trust-anchor path from a real
 #                  1.1.8 origin). Same defaults as v118_upgrade.
+#   legacy_config — rollout-day config rehearsal: the current build on
+#                  old-style (1.1.8-shape) YAMLs — `sui-rpc-url` only, no
+#                  sui-data-source — for all validators AND the notifier
+#                  (the legacy JSON-RPC transport nothing else covers). Full
+#                  lifecycle at v3, the v3 -> v4 vote, lifecycle again at v4.
+#                  Current build only.
 #
 # Each scenario binds the fixed Sui localnet ports (9000/9123) and chdirs the
 # process during publish, so only ONE runs per job — `--test-threads=1`, one
@@ -116,7 +122,7 @@ jobs:
           TEST: ${{ inputs.test }}
         run: |
           set -euo pipefail
-          ALL="smoke workload cross_binary v118_upgrade v118_churn"
+          ALL="smoke workload cross_binary v118_upgrade v118_churn legacy_config"
           if [ "$TEST" = "all" ]; then
             SELECTED="$ALL"
           else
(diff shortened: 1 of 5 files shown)


omersadika:
Proof run: [28578561448](https://github.com/dwallet-labs/ika/actions/runs/28578561448) — **green in 1095s**, with the evidence that it exercised what it claims:

- **All five nodes** — validator-0..3 AND the notifier — logged `DEPRECATED: old-style config (no sui-data-source)`, i.e. every role booted the legacy JSON-RPC transport from an old-style YAML.
- Flow: `start_all` → epoch 2 (genesis DKG done ⇒ the legacy-path notifier is submitting) → `run_workload("legacy-v3")` → buffer drop → epoch 4 → `expect_protocol_version_at_least(4)` **passed** → `run_workload("legacy-v4")` → `test result: ok. 1 passed`.

That also settles the rollout-sequencing question empirically: **the v3→v4 capability vote works on legacy configs** — operators don't need to migrate their YAMLs to `sui-data-source` before the activation vote.

PR #1795, opened 2026-07-02
fix(joiner-bootstrap): tolerate a prior-committee member whose validator_info fails to decode

ycscaly:
## Why

Follow-up to #1762. `fetch_previous_committee` (the joiner-bootstrap path that chain-reads the prior committee to verify the handoff cert) fails the **entire** prior-committee construction when any single member's `validator_info` fails `verify()`:

- `ValidatorInfo::verify()` is a full well-formedness decode — it rejects on **any** malformed metadata field (network/p2p/consensus addresses, all `next_epoch_*` fields), not just the `consensus_pubkey` this path actually needs.
- The offending member may have **departed** the active set, so nobody is operationally motivated to fix its stale record.
- The failure is deterministic, so the bootstrap retry loop re-hits it forever: **one bad record blocks all joiner bootstrap**.

Before #1762 this same decode ran in `fetch_previous_committee_consensus_pubkeys`, but its caller was best-effort (warn + proceed with the current active set's keys) — a bad member cost only *its* signature toward the cert quorum. #1762 moved the decode inside the committee fetch, silently upgrading that degraded mode to a hard block.

## What

Per-member skip-and-warn — the same handling `trusted_peer_updater` already uses for this exact call. A member whose `validator_info` fails to decode is skipped: its handoff signatures drop as unverifiable (identical posture to a byzantine signer), and the cert still requires a quorum of the members that **do** verify, so nothing is weakened. If the skipped stake makes quorum unreachable, cert verification fails downstream with the honest error (insufficient stake) instead of "metadata didn't parse".

## Verification

- `cargo check --release -p ika-core` green; `cargo fmt --all` clean.
- Behavior change is liveness-only on the bootstrap path; no protocol impact.

### DIFF
--- a/crates/ika-core/src/sui_connector/pubkey_provider_updater.rs
+++ b/crates/ika-core/src/sui_connector/pubkey_provider_updater.rs
@@ -57,7 +57,9 @@ fn select_next_epoch_committee(system_inner: &SystemInnerV1) -> Vec<ObjectID> {
 /// epoch (so its `committee_store` has no entry for it). The membership
 /// is decoded with `read_bls_committee`; each member's consensus key is
 /// fetched and carried on the committee so its prior-epoch handoff
-/// signatures verify by name. The class-groups / PVSS maps are left empty:
+/// signatures verify by name (best-effort per member: one whose
+/// validator_info fails to decode is skipped and only its signatures are
+/// lost, never the whole committee). The class-groups / PVSS maps are left empty:
 /// handoff-cert verification (`verify_certified_handoff_attestation`) only
 /// needs membership, voting power, the quorum threshold, and the consensus
 /// keys.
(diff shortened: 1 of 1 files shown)


omersadika:
Reviewed — the fix is correct and the stake semantics are the strong version: membership/voting rights come from `read_bls_committee` untouched, so the quorum **denominator doesn't shrink**; the skipped member's signature just drops as unverifiable. `Committee::new` stores the partial `consensus_keys` map with no completeness invariant and lookups are per-signer `.get()`, so nothing downstream trips. 👍

One note: this covers one of the two wholesale-abort sites from the release-review finding (V5 in the [report](https://github.com/dwallet-labs/ika/pull/1787#issuecomment-4862501346)). The **periodic provider-install tick** still has the old shape — `pubkey_provider_updater.rs` `run_once` (≈ line 312 on this branch):

```rust
let verified = pool.validator_info.verify()
    .map_err(|code| anyhow::anyhow!("validator info verify failed: code {code}"))?;
```

One bad record fails the whole batch every tick — deterministically, so the retry loop never recovers and the joiner-announcement provider stays uninstalled for the epoch (the relay-suppression half of the finding). Same skip-and-warn shape would close it; happy to see it in this PR or a follow-up.

PR #600, opened 2025-02-06
543 presign

ItayLevyOfficial:

Issue #599, opened 2025-02-05
Use a dwallet-specific seed instead of being dependent on EDDSA


ItayLevyOfficial:
Right now we derive the user's encryption & decryption keys from its EDDSA private key. This logic can't work in the mainnet, for two reasons:

1. There are signing schemes in Sui that doesn't have a secret key (e.g. ZK login).

2. The wallet's secret key is not accessible from the TS SDK, leading to a complex UX.

To solve this issue, we introduce a new, dWallet dedicated seed, that will be used to generate both the signing & encryption key

ItayLevyOfficial:
already implemented

PR #1802, opened 2026-07-06
test(dwallet-mpc): add dwallet-mpc-unsafe-mock feature for fast integration tests

ycscaly:
## Summary

Adds a `dwallet-mpc-unsafe-mock` cargo feature that forwards to the cryptography-private `unsafe_mock` feature ([cryptography-private PR #599](https://github.com/dwallet-labs/cryptography-private/pull/599)): INSECURE deterministic 2PC-MPC protocol + validator-keygen mocks (constant signing key `x = 42`, constant nonce, neutral class-group/PVSS keys, no threshold crypto) so dwallet-mpc integration tests run without the multi-minute class-group protocol rounds. Signatures still verify on the real curves. **Never enable in production** — the feature is off by default and every hop of the chain is explicitly opt-in.

## Design: mock at the crypto level, behind high-level entry points

The validator keygen mock lives **inside the `class_groups` crate**, at high-level entry points that own the full keygen orchestration internally:

- `generate_class_groups_keypair(rng)` — the CRT class-groups keypair
- `generate_pvss_keypairs(secp256k1_rng, secp256r1_rng, ristretto_rng)` — the three per-curve PVSS keypairs

Their `#[cfg(feature = "unsafe_mock")]` twins return neutral encryption keys + `new_default` proofs, **memoized process-wide** (so only the first validator pays the one-time setup-parameter derivation; every later `from_seed`, e.g. a mid-epoch joiner, is instant).

`dwallet-classgroups-types::from_seed` therefore holds **no low-level crypto** — no setup-parameter derivation, no `derive_from_plaintext_parameters`, no per-curve const generics. It supplies the domain-separated RNGs and assembles the `class_groups`-native results into the committee types (which are those tuples verbatim). Enabling the ika feature simply turns on `class_groups/unsafe_mock`.

## Changes

- Pin the 7 cryptography-private crates to rev `6f52491b` ([cryptography-private PR #600](https://github.com/dwallet-labs/cryptography-private/pull/600), off latest `main` after #599 merged). The default build of that rev is behavior-identical to `main` — the mock is entirely feature-gated — so production builds are unaffected.
- `dwallet-mpc-unsafe-mock` forwarding chain: `dwallet-classgroups-types` / `dwallet-mpc-types` (→ `class_groups/unsafe_mock` + `twopc_mpc/unsafe_mock`) ← `ika-core` ← `ika-node` ← `ika-test-cluster`, plus `dwallet-mpc-centralized-party` forwarding directly.
- `dwallet-classgroups-types`: removed the former ~230-line ika-side `unsafe_mock` module and rewired `from_seed` to the high-level crypto entry points above; dropped the no-longer-needed optional `proof` dep.
- `ika-test-cluster` tests: always run under the mocks and shrink epochs accordingly (the crypto costs they were sized for are gone).

`cargo tree` confirms: with the feature on, `unsafe_mock` activates on `twopc_mpc`/`class_groups`; the default build activates none of it.

## Usage

```bash
cargo test -p ika-test-cluster --features dwallet-mpc-unsafe-mock --test <test> -- --test-threads=1
```

https://claude.ai/code/session_013VMVVVMou98FNxjB2gBgMN

### DIFF
--- a/crates/dwallet-classgroups-types/Cargo.toml
+++ b/crates/dwallet-classgroups-types/Cargo.toml
@@ -13,6 +13,15 @@ serde.workspace = true
 ika-types.workspace = true
 twopc_mpc.workspace = true
 
+[features]
+# No default features: the INSECURE `dwallet-mpc-unsafe-mock` below must be opt-in only.
+default = []
+# INSECURE test-only mock: replaces the expensive validator class-groups/PVSS
+# keygen (`from_seed`) with neutral keys + default proofs — consumed only by the
+# mocked protocol rounds. Never enable in production. See cryptography-private
+# `unsafe_mock`.
+dwallet-mpc-unsafe-mock = ["class_groups/unsafe_mock", "twopc_mpc/unsafe_mock"]
+
 # Production builds run crypto with rayon-backed parallelism. Under
 # `cargo simtest` (cfg msim), rayon worker threads have no msim node
 # context and panic at the first tracing/tokio call, so we drop the
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/utils.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/utils.rs
@@ -1793,6 +1793,81 @@ pub(crate) async fn advance_mpc_flow_until_completion_for_parties(
     }
 }
 
+/// The honest magic `u64` every mock party sends each round
+/// (`MOCK_HONEST_MESSAGE` in the cryptography crate's `2pc-mpc/src/mock.rs`). Mirrored
+/// here because it is `pub(crate)` in that crate and not exported; the mock `advance`
+/// flags any sender whose message value differs from this as malicious.
+#[cfg(feature = "dwallet-mpc-unsafe-mock")]
+pub(crate) const HONEST_MOCK_MESSAGE: u64 = 0xDEAD_BEEF;
+
+/// The non-magic `u64` a mock-corrupted MPC message carries in place of
+/// [`HONEST_MOCK_MESSAGE`], so the mock `advance` flags its sender malicious.
+#[cfg(feature = "dwallet-mpc-unsafe-mock")]
+pub(crate) const MALICIOUS_MOCK_MESSAGE: u64 = 0xDEAD_DEAD;
+
+/// Corrupt `party`'s most recent [`ConsensusTransactionKind::DWalletMPCMessage`] into a
+/// detectably-malicious one: rewrite the inner mock `u64` round message from
+/// [`HONEST_MOCK_MESSAGE`] to [`MALICIOUS_MOCK_MESSAGE`].
+///
+/// The mock parties are never wire-visible directly — every outgoing message is the
+/// bcs serialization of the guaranteed-output-delivery wrapper
+/// (`mpc::party::guaranteed_output_delivery::Message::MessageWithMetadata`), whose
+/// inner `message: M` (here `M = u64`) is the struct's **last** field. bcs encodes a
+/// `u64` as 8 fixed little-endian bytes with no length prefix and appends it with
+/// nothing after, so the message's trailing 8 bytes are exactly that inner `u64`
+/// regardless of how large the preceding metadata (round number, ignored-sender map,
+/// malicious list) is. Overwriting those 8 bytes therefore sets the inner message and
+/// nothing else, and the blob still deserializes.
+///
+/// We do not assume that layout — we prove it: an honest serialized message must end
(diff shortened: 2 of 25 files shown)


crates/dwallet-classgroups-types/Cargo.toml
```diff
@@ -13,6 +13,13 @@ serde.workspace = true
 ika-types.workspace = true
 twopc_mpc.workspace = true
 
+[features]
+# INSECURE test-only mock: replaces the expensive validator class-groups/PVSS
```
ycscaly, line 19: Make it explicit : use empty default features so it is not included.

ycscaly: Here and any other cargo.toml where it applies

crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
```diff
@@ -427,6 +427,51 @@ async fn test_internal_presign_stops_at_min_pool_size_when_not_idle() {
         "system should be non-idle after sending NetworkOwnedAddressSign requests"
     );
 
+    // Drain the sessions instantiated while the system was still idle: Phase 1 keeps
```
ycscaly: How is this test change related to this pr?

crates/ika-core/src/dwallet_mpc/integration_tests/utils.rs
```diff
@@ -1795,6 +1795,51 @@ pub(crate) async fn advance_mpc_flow_until_completion_for_parties(
...
+        .iter()
+        .rposition(|msg| matches!(msg.kind, ConsensusTransactionKind::DWalletMPCMessage(_)))
+        .expect("party should have a DWalletMPCMessage to corrupt");
+    let ConsensusTransactionKind::DWalletMPCMessage(ref mut message_content) =
+        submitted[index].kind
+    else {
+        unreachable!("index was verified to be a DWalletMPCMessage above");
+    };
+    let malicious_message =
+        bcs::to_bytes(&MALICIOUS_MOCK_MESSAGE).expect("serializing a u64 cannot fail");
```
ycscaly, line 1856: Isnt it wrapped by a struct enum in the mpc party? Esp. It should be wrapped by the generic asynchrounous mpc party message right? The mpc parties in this crate are never used directly, always through that intermediate. You can simply run once with a modified crypto private the actual tests with mocks where the party 1 sends malicious value DEADDEAD, and print its serialized value, and hardcode here, and explain the process done.

crates/dwallet-classgroups-types/Cargo.toml
ycscaly: Done — added an explicit `default = []` to every crate that defines `dwallet-mpc-unsafe-mock` (dwallet-classgroups-types, dwallet-mpc-types, dwallet-mpc-centralized-party, ika-core, ika-swarm, ika-test-cluster, sdk/ika-wasm), so the insecure mock is unambiguously opt-in. ika/ika-node already had a `default` line that excludes it.

crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
ycscaly: It wasn't — removed. That pool-quiescence loop was unrelated to this PR; `internal_presign.rs` is now back to its dev/main version (dropped from the PR entirely).

crates/ika-core/src/dwallet_mpc/integration_tests/utils.rs
ycscaly: You're right that the `u64` is wrapped — every outgoing message is `bcs(mpc::party::guaranteed_output_delivery::Message::MessageWithMetadata { mpc_round_number, inactive_or_ignored_senders_by_round, malicious_parties, message: M })`, with the inner mock `M = u64` as the struct's **last** field. bcs encodes a u64 as 8 fixed LE bytes with no length prefix and appends nothing after it, so the message's trailing 8 bytes are exactly that inner value regardless of the metadata sizes.

I didn't go with the capture-and-hardcode route: a hardcoded serialized blob would be protocol-specific (the wrapper's metadata differs per protocol/round) and would rot silently, whereas `corrupt_party_message` is a single generic helper used across all mock protocols. Instead I made it **self-validating** — before overwriting, it asserts the trailing 8 bytes equal `bcs(HONEST_MOCK_MESSAGE = 0xDEAD_BEEF)`. That empirically proves, every run, that we're hitting the inner message and not metadata; if the wrapper layout ever changes so the u64 is no longer the trailing field, the assert fires instead of silently corrupting the wrong bytes. Full rationale is in the doc comment on the fn. Happy to switch to a captured fixture if you'd still prefer it.

PR #1804, opened 2026-07-09
fix(ci): unblock main CI — Linux Docker SSH, Windows jemalloc, clippy needless_return

omersadika:
The `release/1.2.0-test1` dispatch ([run 29013707089](https://github.com/dwallet-labs/ika/actions/runs/29013707089)) failed on **Linux (x64+arm64)** and **Windows**, while macOS passed. Two independent, pre-existing bugs — neither caused by any code change; the test release just happened to exercise these paths.

## 1. Linux Docker builds — private-repo SSH auth
```
failed to authenticate when downloading repository: git@github.com:dwallet-labs/inkrypto
  * attempted ssh-agent authentication, but no usernames succeeded: `git`
```
`docker/builder/Dockerfile` writes the deploy key to `~/.ssh/id_ed25519` and sets an `insteadOf` SSH URL rewrite — the **git-CLI** auth path. But `cargo fetch` uses its default **libgit2** transport, which only tries **ssh-agent** (none in the container) and never reads the on-disk key. The workflow sets `CARGO_NET_GIT_FETCH_WITH_CLI=true` at *runner-env* level, but that doesn't cross into `docker build`.

**Fix:** set `ENV CARGO_NET_GIT_FETCH_WITH_CLI=true` inside the image, so cargo shells out to the git CLI (which honors the key + rewrite).

*Why macOS passed:* native runners inherit the runner env, so cargo already used the git CLI there — clean A/B confirmation.

## 2. Windows build — jemalloc compiled on MSVC
```
error: failed to run custom build command for `tikv-jemalloc-sys v0.5.4`
  configure: error: C compiler cannot create executables
```
The `#[global_allocator]` is already `cfg(not(target_env = "msvc"))`-gated in source, and `tikv-jemalloc-ctl` is target-gated in `Cargo.toml` — but `tikv-jemallocator` itself sat in plain `[dependencies]`, so its `-sys` crate still compiled on Windows (and jemalloc's autotools `configure` fails under MSVC).

**Fix:** move `tikv-jemallocator` into the existing `cfg(not(target_env = "msvc"))` section, mirroring `tikv-jemalloc-ctl`. jemalloc is unchanged on Linux/macOS.

## Verification
- `cargo tree --target x86_64-pc-windows-msvc -i tikv-jemalloc-sys`: **present before → absent after** (causal A/B).
- `cargo tree --target x86_64-unknown-linux-gnu -i tikv-jemalloc-sys`: still present (jemalloc kept).
- Host build still wires `tikv-jemallocator`; manifest parses; no `Cargo.lock` churn.

## Not included / follow-ups
- `crates/ika-node/Cargo.toml` has the **same latent** ungated `tikv-jemallocator` (line 35), but ika-node is **Linux-only** in the release so it doesn't fail today. Left out to keep this surgical — happy to gate it too for consistency if wanted.
- The proper validation is a re-dispatched release build (`network=testnet, version=1.2.0-test2`) after this merges.

### DIFF
--- a/crates/dwallet-mpc-centralized-party/src/lib.rs
+++ b/crates/dwallet-mpc-centralized-party/src/lib.rs
@@ -1214,11 +1214,9 @@ fn protocol_public_parameters_from_reconfiguration_output(
         bcs::from_bytes(&reconfiguration_dkg_public_output)?;
 
     match &reconfiguration_dkg_public_output {
-        VersionedDecryptionKeyReconfigurationOutput::V1(_) => {
-            return Err(anyhow::anyhow!(
-                "V1 reconfiguration outputs are no longer supported"
-            ));
-        }
+        VersionedDecryptionKeyReconfigurationOutput::V1(_) => Err(anyhow::anyhow!(
+            "V1 reconfiguration outputs are no longer supported"
+        )),
         VersionedDecryptionKeyReconfigurationOutput::V2(public_output_bytes) => {
             // bwd-compat reconfig output shape.
             let public_output: <twopc_mpc::decentralized_party_backward_compatible::reconfiguration::Party as mpc::Party>::PublicOutput =
--- a/crates/ika-node/Cargo.toml
+++ b/crates/ika-node/Cargo.toml
@@ -32,7 +32,6 @@ path = "src/bin/ika-notifier.rs"
 [dependencies]
 anemo.workspace = true
 dwallet-mpc-types.workspace = true
-tikv-jemallocator = { workspace = true, optional = true }
 anemo-tower.workspace = true
 arc-swap.workspace = true
 axum.workspace = true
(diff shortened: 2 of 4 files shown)


omersadika:
Added a third fix to this PR (scope now covers all currently-red `main` CI, not just the release builds):

**3. Clippy `-D warnings` failure on `main`** ([run 29017214620](https://github.com/dwallet-labs/ika/actions/runs/29017214620/job/86115084101))
```
error: unneeded `return` statement
  --> crates/dwallet-mpc-centralized-party/src/lib.rs:1218
```
`protocol_public_parameters_from_reconfiguration_output`'s V1 match arm used `return Err(...);` while the `match` is the function's tail expression and the V2/V3 arms yield a value. Changed it to a tail `Err(...)`.

Verified: `cargo clippy -p dwallet-mpc-centralized-party --all-targets --all-features -- -D warnings` → exit 0 (was: 1 error). `cargo fmt` clean.

omersadika:
Also folded in the **`ika-node` jemalloc gating** (the previously-noted follow-up): its `tikv-jemallocator` had the identical ungated pattern. Now gated to `cfg(not(target_env = "msvc"))` like `ika`. It's Linux-only in the release so it wasn't failing, but this keeps the two crates consistent and future-proofs a Windows/node build.

Verified same as `ika`: jemalloc-sys absent on windows-msvc, present on linux, still wired on host, manifest parses, no Cargo.lock churn.

Commits now on this branch:
- Docker git-CLI fetch (Linux SSH auth)
- `ika` jemalloc msvc-gate (Windows)
- clippy needless_return (main CI)
- `ika-node` jemalloc msvc-gate (consistency)

PR #1810, opened 2026-07-11
fix(ika-node): enforce minimum CPU only for validators on testnet/mainnet

omersadika:
## Why

`enforce-minimum-cpu` was a pure compile-time feature applied to **every** `ika-node` binary (validator, fullnode, notifier — they share the crate's feature set) on **every** network. With default features, any node on a <16-core host panicked at startup, and rayon reserved `TOKIO_ALLOCATED_CORES` (4) even where the MPC load doesn't warrant it.

## What

The feature stays as the outer compile-time gate (test/harness builds keep dropping it via `--no-default-features`, including for historical tags in the upgrade test). When compiled in, enforcement is now decided **at runtime** in `IkaRuntimes::new`, requiring **all** of:

1. feature compiled in,
2. the node runs as a **validator** (`NodeMode::Validator` — covers both the `ika-validator` binary and `ika-node` auto-detected as validator; fullnodes/notifiers don't do MPC computations),
3. the **IKA network** is testnet or mainnet — derived from the deployed ika system object's `ChainIdentifier` (`ika_system_object_id` vs the compiled-in ika mainnet/testnet identifiers, the same derivation `node_runner` uses for protocol-config selection). **Not** the Sui settlement chain: an ika devnet/staging network deployed on Sui testnet must not enforce. Anything unrecognized maps to `Chain::Unknown` → no enforcement.

The decision is made once and stored in a `OnceLock`, so the MPC orchestrator's later `calculate_num_of_computations_cores()` call agrees with the rayon pool sizing. Unset (tests that never build the runtimes) means no enforcement — matching the previous feature-off behavior.

`IkaRuntimes::new` gains a `NodeMode` parameter (its `NodeConfig` parameter was previously unused); the only caller (`node_runner.rs`) already has the mode resolved right above the call.

## Behavior matrix (feature compiled in)

| Node | IKA network | Before | After |
|---|---|---|---|
| Validator | ika mainnet / ika testnet | enforced | **enforced** |
| Validator | devnet / localnet / unknown | enforced (panic <16 cores) | not enforced |
| Fullnode / Notifier | any | enforced (panic <16 cores) | not enforced |

## Verification

- `cargo check -p ika-node` (feature on via defaults) — clean.
- Unit tests cover the mode × chain × feature decision matrix, plus a pin that an unrecognized ika system object maps to `Chain::Unknown` (`cargo test -p ika-core --lib runtime::` — 3/3 pass).

### DIFF
--- a/.github/workflows/system-tests-docker.yaml
+++ b/.github/workflows/system-tests-docker.yaml
@@ -4,9 +4,9 @@ on:
   workflow_dispatch:
     inputs:
       cargo_build_flags:
-        description: 'Build flags for cargo'
+        description: 'Extra build flags for cargo. Default features are kept: enforce-minimum-cpu is runtime-gated to validators on the ika testnet/mainnet networks, so it is inert in test deployments.'
         required: false
-        default: '--no-default-features'
+        default: ''
       imageTag:
         description: 'Tag for the Docker image'
         required: false
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -42,9 +42,12 @@ name: Upgrade Test
 #
 # The OLD binary for the upgrade scenarios is built here from `old_ref` in an
 # isolated `git worktree`, at that ref's OWN pinned toolchain (rustup honors
-# the worktree's rust-toolchain.toml). All `ika-node` binaries are built
-# `--no-default-features` to drop `enforce-minimum-cpu`, which panics on
-# cgroup-throttled / <16-core pods.
+# the worktree's rust-toolchain.toml). Only the OLD binary is built
+# `--no-default-features`: at historical refs `enforce-minimum-cpu` was
+# compile-time-only and panics on cgroup-throttled / <16-core pods. Current
+# binaries build with default features — enforcement is runtime-gated to
+# validators on the ika testnet/mainnet networks, so it is inert on the CI
+# localnet (its ika system object maps to Chain::Unknown).
 
 on:
   workflow_dispatch:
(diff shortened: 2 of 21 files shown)


omersadika:
Follow-up commit `2f9dc94089`: cleaned up the CI build workarounds that the runtime gating makes unnecessary.

**Cleaned (current-checkout builds only):**
- `upgrade-test.yaml` — NEW `ika-node` binaries now build with **default features**; mocks remain an additive `--features` flag. The old `--no-default-features --features jemalloc` dance is gone.
- `system-tests-docker.yaml` — default `cargo_build_flags` is now empty. Side benefit: the system-test image stops silently dropping **jemalloc** (the same historical bug the upgrade-test workflow had already fixed for itself).

**Deliberately kept:**
- The **OLD-binary** build (`old_ref` worktree) keeps `--no-default-features` — historical refs still have the compile-time-only enforcement that panics on cgroup-throttled / <16-core pods.
- `ika-upgrade-test`'s internal SHA builder (`binary.rs`) — builds historical tags, same reason.
- Harness `RAYON_NUM_THREADS` bounding comments updated: the bounding is still required (enforcement is inert on the localnet either way, so rayon would default to all cores per co-located node).

Safety check: no CI path sets `IKA_PROTOCOL_CONFIG_CHAIN_OVERRIDE`, so CI localnets always map to `Chain::Unknown` → the compiled-in feature never fires there.

omersadika:
Commit `26c5b885c3` completes the repo-wide consistency sweep — every `enforce-minimum-cpu` / `--no-default-features` reference now agrees with the runtime gate:

**Notifier dockerfile — answered:** nothing to clean. `docker/ika-node/Dockerfile` (builds validator/fullnode/notifier via `BIN=`) and `docker/builder/Dockerfile` always built with default features. The notifier was previously *wrongly enforcing* the 16-core assert in production; the mode gate in this PR is itself the fix.

**Cleaned in this commit:**
- Public docs `setup-localnet.mdx` + TS SDK README: dropped `--no-default-features` from the localnet commands (it never controlled enforcement on the `ika` CLI — its only default feature is jemalloc — and localnets are runtime-exempt now). Docs state the actual rule.
- Dev helper scripts `scripts/{run,rerun}_ika_*.sh`: same stale flag removed (side effect: local `ika start` now gets jemalloc, matching prod).
- `binary.rs` comment: was factually wrong (claimed enforce-minimum-cpu was the *only* default feature) — now explains the flag exists for HISTORICAL refs only.
- `swarm.rs` comment: in-process swarm never goes through `IkaRuntimes`; reworded to the runtime-gate reality.

**New dev-docs:** `dev-docs/conventions/enforce-minimum-cpu.md` (+ README index) — the 3-condition decision rule, the `OnceLock` mechanism and its two consumers, which builds must KEEP `--no-default-features` (old-ref builds in the upgrade test), and the `IKA_PROTOCOL_CONFIG_CHAIN_OVERRIDE` interaction.

**Deliberately untouched (verified accurate):** the OLD-ref build sites (`upgrade-test.yaml` OLD step, `ika-upgrade-test/src/binary.rs` flag itself, `v118_upgrade.rs` docs, `ci-suites.md` table) and the archived plan record.

Verified: `cargo check -p ika-upgrade-test -p ika-swarm` clean; final grep shows the only remaining `--no-default-features` are OLD-ref builds and the docs describing them.

omersadika:
## Heavy-suite validation — verdict

| Suite | Run | Result |
|---|---|---|
| **Upgrade Test** (`test=all`) — the suite that actually exercises this PR (real out-of-process binaries through the new `IkaRuntimes` gate, default features, <16-core pods, edited build steps, OLD-ref workaround) | [29161768027](https://github.com/dwallet-labs/ika/actions/runs/29161768027) | ✅ success |
| Integration Tests (incl. the new `runtime::` unit tests) | [29161765888](https://github.com/dwallet-labs/ika/actions/runs/29161765888) | ✅ success |
| TS Integration | [29161767300](https://github.com/dwallet-labs/ika/actions/runs/29161767300) | ✅ success |
| Test Cluster | [29161766603](https://github.com/dwallet-labs/ika/actions/runs/29161766603), re-run [29170787807](https://github.com/dwallet-labs/ika/actions/runs/29170787807) | ⚠️ 21/22 twice — **pre-existing flakes, not this branch** (triage below) |

### Cluster triage

Two runs, two *different* single-test failures, both with open-issue signatures, and each failing test passed in the other run:

1. Run 1: `test_global_presigns_complete_across_epoch_switches` — wedged ~2h on `NetworkDecryptionKeyNotReady` (`should_never_happen=true`) for `InternalPresign` at epoch-2 entry, then SIGSEGV at teardown. Signature of **open issue #1736** (epoch-entry stale-mpc_data race / dead-presign validator). The branch already contains the related merged fix #1807, so this is the documented residual race. Passed in run 2.
2. Run 2: `off_chain_metadata_v4_does_not_read_blobs_from_chain` — 3 unexpected chain blob reads across an epoch transition, matching the **residual double-fetch** noted on the same issue #1736. Passed in run 1.

**Why this branch cannot be the cause:** the cluster suite runs in-process and never calls `IkaRuntimes::new`; the only shared code (`calculate_num_of_computations_cores`) behaves identically because the test binary builds ika-core **without** `enforce-minimum-cpu` (verified via `cargo tree`) — `cfg!` was false before, the unset `OnceLock` is false after. The test-cluster workflow file is untouched. The rest of the diff is comments/docs/scripts.

Side finding while checking main's baseline: the **nightly suite orchestrator has been broken for ≥2 nights** (all 8 dispatch jobs die with `fatal: not a git repository` before dispatching), so main/dev have had no nightly heavy-suite signal — which is why these flakes weren't showing up on main. Flagged separately.

PR #1818, opened 2026-07-12
fix(dwallet-mpc): park internal presign sessions on missing network-key data instead of failing them terminally

omersadika:
## Problem

At epoch entry the VSS (Fast Schnorr) internal presign pools instantiate as soon as the network key installs — before the consensus-frozen off-chain validator key set is ingested — so their session-input construction fails with `NetworkDecryptionKeyNotReady`. `session_status_from_request` mapped every input error to terminal `SessionStatus::Failed`, killing the batch on every validator: `should_never_happen=true` error spam and terminally-dead sessions at every epoch entry, with the EdDSAVSS/SchnorrkelVSS/TaprootVSS pools starved until the stale-batch expiry (#1807) forgave the batch.

Proven live in CI run 29153674855: 222 `should_never_happen=true error=NetworkDecryptionKeyNotReady` errors at every epoch entry, 18 terminally failed sessions, EdDSA/Schnorrkel future-sign tests timing out while secp passes (vs green run 29162432790 on equivalent code). v4-gated (`internal_presign_sessions`), so no effect on the v3 mainnet rollout — but a v4-activation gate: at v4 it would starve EdDSA/Schnorrkel/Taproot NOA signing after every reconfiguration, and it makes the TS CI suite flaky today.

## Fix

Requests hitting the not-ready error class (`NetworkDecryptionKeyNotReady` / `WaitingForNetworkKey`, classified by the new `DwalletMPCError::is_network_key_data_not_ready()`) are **parked** in a new manager queue (`internal_presign_requests_pending_for_network_key_data`) and retried once per service iteration — per iteration, not per consensus round, because what they wait on (off-chain key ingest, async key install) completes on wall-clock time. On activation, an existing `WaitingForSessionRequest` entry is upgraded in place so peer messages that arrived while parked survive. Genuinely-fatal input errors remain terminal, and the reachable not-ready condition no longer logs `should_never_happen=true`.

**Why park rather than gate instantiation on readiness**: `next_internal_presign_sequence_number` is a single counter interleaved across all pools, and readiness is per-validator wall-clock — a validator that skipped a pool while another instantiated it would permanently desync every subsequent session identifier (sessions could never reach quorum). Parking consumes the sequence number at instantiation, so identifier derivation and the instantiated/completed batch counters stay committee-uniform. That also makes parking invisible to #1807's stale-batch expiry: no double-counting, and a pathological expiry-while-parked just adds a late-completing batch that the existing saturating-completion logic absorbs.

Also adds a parked-queue gauge (`ika_dwallet_mpc_internal_presign_requests_pending_for_network_key_data`), updates `dev-docs/specs/fast-schnorr-vss.md` (epoch-entry parking decision rule) and the mpc-stall playbook (the wedge signature is now a park line with no later activation line / a sustained non-zero gauge).

## Testing

- New integration test `test_internal_presign_vss_parks_until_off_chain_keys_ingested`: re-opens the epoch-entry window (empties the current-epoch key channel + resets the ingest state), asserts all three VSS pools park exactly one batch each with zero Failed sessions and no duplicate top-ups across extra rounds, then closes the window and asserts activation, completion, pool fill, and cross-validator counter uniformity.
- **Fault-injection validated** (dev-docs/playbooks/test-testing.md): with the parking guard disabled (pre-fix behavior), the test fails at the parking assert and the log reproduces the production signature exactly — 24 `should_never_happen=true error=NetworkDecryptionKeyNotReady` lines (6 sessions × 4 validators), zero park lines. With the fix: 24 park lines → 24 activate lines, zero errors, pass.
- All 5 `dwallet_mpc::integration_tests::internal_presign` tests green in release (1046s); clippy clean on ika-core + ika-types.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
@@ -197,6 +197,13 @@ pub struct DWalletMPCMetrics {
     /// pending-request gauges target.
     pub(crate) requests_pending_for_frozen_mpc_data: IntGauge,
 
+    /// Size of `DWalletMPCManager.internal_presign_requests_pending_for_network_key_data`
+    /// — internal presign sessions parked because the target network key's data
+    /// (typically the off-chain VSS validator key set at epoch entry) isn't
+    /// locally available yet. Transient non-zero values at epoch entry are
+    /// expected; sustained ones mean the key data never arrived.
+    pub(crate) internal_presign_requests_pending_for_network_key_data: IntGauge,
+
     /// One series per user session currently tracked in `DWalletMPCManager.sessions`,
     /// labeled by `session_seq` (as a string) and `state`. Value is 1 for the
     /// state the session is currently in, 0 for the other four states. Sessions that leave
@@ -489,6 +496,12 @@ impl DWalletMPCMetrics {
             requests_pending_for_frozen_mpc_data: register_int_gauge_with_registry!(
                 "ika_dwallet_mpc_requests_pending_for_frozen_mpc_data",
                 "Sessions parked waiting for the off-chain mpc_data freeze gate",
+                registry,
+            )
+            .unwrap(),
+            internal_presign_requests_pending_for_network_key_data: register_int_gauge_with_registry!(
+                "ika_dwallet_mpc_internal_presign_requests_pending_for_network_key_data",
+                "Internal presign sessions parked waiting for their network key's data to be locally available",
                 registry,
             )
             .unwrap(),
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
@@ -497,6 +497,13 @@ impl DWalletMPCService {
             error!(error = ?e, "failed to ingest off-chain validator MPC keys");
         }
 
+        // Retry internal presign requests parked on missing network key data —
+        // per ITERATION, not per consensus round, because what they wait on
+        // (the ingest above / an asynchronous key install) completes on
+        // wall-clock time, and no new consensus round is guaranteed to follow.
+        self.dwallet_mpc_manager
+            .retry_internal_presign_requests_pending_for_network_key_data();
+
         self.sync_last_session_to_complete_in_current_epoch().await;
 
         // Process any pending network-owned-address sign requests.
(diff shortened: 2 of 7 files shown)


omersadika (commented):
**Review verdict: would approve** (posted as a comment — GitHub forbids self-approval, and this review runs under the author account). This closes the bug root-caused during the 1.2.0 release review ([artifact](https://claude.ai/code/artifact/ac3c4e16-048b-46b9-946e-0c616b8133e5), CI run 29153674855), and the implementation survives the three traps I went looking for:

**Verified in the diff:**
1. **The sequence-number trap is handled correctly** — `next_internal_presign_sequence_number += 1` runs on every path including park, so identifier derivation and the instantiated/completed batch counters stay committee-uniform while a validator defers. The design note on *why* park-not-gate (a skipped pool would permanently desync the shared counter) is exactly right, and it's what makes parking invisible to #1807's expiry accounting.
2. **The retry runs per service iteration, not in the round drain** — with the comment citing precisely the rule this codebase learned from the `missing_network_key` wedge post-mortem (wall-clock-fed inputs must never wait on consensus rounds). Placed right after the off-chain key ingest, so a completed ingest activates parked sessions in the same tick.
3. **The error class is narrow**: only `WaitingForNetworkKey | NetworkDecryptionKeyNotReady` park; `InvalidParameters` and everything else stays terminal — it doesn't swallow the #1736-class failure signals, and `should_never_happen` is dropped only for the genuinely-reachable transient.
4. **Buffered peer messages survive activation** (in-place `WaitingForSessionRequest` upgrade + `set_request_metadata`), and the "quorum already resolved it without us" guard drops stale parked requests instead of double-running.
5. **Test quality meets the test-testing playbook standard**: fault-injection validated, reproducing the production log signature exactly (24 `should_never_happen` lines with the guard disabled → 24 park/activate lines with it).
6. No conflict with #1816; the new gauge emission lands inside the rate-limited refresh and inherits the 1/s cadence for free.

**Two non-blocking observations:**
- The per-validator **expiry-round desync** in #1807 (separate v4-gate finding from the review) remains open — this PR correctly neither fixes nor worsens it; don't let the "parking is invisible to expiry" note read as closing that item.
- Only presign *instantiation* parks; a NOA sign session hitting the not-ready class would still fail terminally. NOA is dark, so fine — just a known boundary of the fix.

**Merge timing**: recommend landing **after** the 1.2.0 tag — the bug is unreachable at v3 (its only live effect is CI flake), the final evidence batch is running on the current main lineage, and v4 activation needs a later release anyway.

PR #1819, opened 2026-07-12
fix(dwallet-mpc): keep internal presign sequence numbers committee-uniform when a network key's installation lags

omersadika:
> **Was stacked on #1818** (the not-ready parking fix — this change reuses its parking queue). #1818 merged as `3a88f2c99c`; this branch is rebased onto it and the diff is the single commit `c0e0ed6c81`.

## Problem

The internal presign top-up loop iterates every **adopted** network key, but a key's installation into `network_keys` completes asynchronously per validator, and the session-identifier preimage binds the installed key's content-derived `NetworkKeyId`. A validator in the adopted-but-not-installed window early-returned from `instantiate_internal_presign_session` **without consuming `next_internal_presign_sequence_number`** — while the caller still advanced the instantiated batch counter. Since that sequence counter is a single counter shared across all pools and keys, the validator's identifier derivation permanently desynchronizes from its peers': every subsequent internal presign identifier diverges and those sessions can never reach quorum, plus the instantiated/completed guard counters desync (instantiated advances with no session behind it).

Only reachable in multi-key epochs — with a single key the outer NOA-key gate returns before any counter is touched — which is why it hasn't bitten. All v4-gated (`internal_presign_sessions`), no deployed-network constraints.

## Fix

- **Sequence number is consumed unconditionally at instantiation**, before identity resolution. The top-up decision is committee-uniform, so every validator must account every session whether it can run it locally yet or not.
- **Not-installed keys defer instead of skipping**, in two shapes: the identity falls back to the pre-instantiation `ObjectID → NetworkKeyId` mapping (seeded deployed keys / handoff-cert background derivation) and the built request parks on the input's not-ready error (the #1818 queue); when even the mapping has no entry (a fresh key with no cert pin), an `AwaitingKeyIdentity` record parks carrying the **reserved** sequence number and the request is built at retry time — deriving the identical identifier peers derived at instantiation.
- **NOA-signing-key selection tie-breaks equal `dkg_at_epoch` by key id**: bare `min_by` resolved same-epoch ties by `HashMap` iteration order — per-process-random, so validators could disagree on which key is the NOA key and apply different pool configs (different batch sizes ⇒ the same divergence through another door).
- **Activation upgrades `WaitingForSessionRequest` entries in place** (shared by instantiation and retry), fixing the pre-existing `new_session` overwrite that dropped peer messages buffered before late local instantiation.

The alternative — skipping the whole (key, curve, algorithm) iteration including the counter increment until install — is ruled out in the code comments: install timing is per-validator wall-clock and the counter interleaves across pools, so any timing-dependent skip diverges it.

## Testing

- New two-key integration test `test_internal_presign_multi_key_install_lag_keeps_identifiers_uniform`: K0 bootstraps, K1 is a second same-epoch DKG; K1's installed data is removed on one validator (its adoption stays, and the manager's background re-install heals the window naturally — the production shape). Asserts the shared sequence counter reads identically on every validator **after every round**, parked batches drain, bound identifier→sequence maps stay subset-consistent, and no session ever fails terminally. A coda exercises the identity-less key: uniform `(sequence, curve, algorithm)` reservations across validators, then a registered mapping transitions them to fully-built parked requests with byte-identical identifiers.
- The run also organically hit the **natural** race: K1's first batches fired while K1 was adopted-but-not-installed on all four validators (12 identity-parks at the same round), with park→activate and park→already-resolved-drop paths both exercised.
- **Fault-injection validated** (dev-docs/playbooks/test-testing.md): restoring the skip-without-consuming behavior fails the test at the per-round counter assert with the exact production divergence (`left: 33, right: 31`).
- Full `dwallet_mpc::integration_tests::internal_presign` suite (all 6 tests, including #1818's) green in release; clippy clean.

Known multi-key uniformity holes **out of scope** and documented in `dev-docs/specs/fast-schnorr-vss.md` + tracked for follow-up: NOA-selection reads the installed set (divergent while installs lag), cross-pool sequence interleaving under adoption-timing skew, and guard counters keyed `(curve, algorithm)` shared across keys.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
@@ -7,10 +7,22 @@ use crate::dwallet_mpc::integration_tests::utils::{
     TEST_PRESIGN_POOL_MINIMUM_SIZE, apply_test_presign_pool_overrides, build_test_state,
     create_test_protocol_config_guard,
 };
+use crate::dwallet_mpc::mpc_manager::ParkedInternalPresignRequest;
 use crate::dwallet_mpc::mpc_session::SessionStatus;
 use crate::dwallet_mpc::{NetworkOwnedAddressSignRequest, ValidatorMpcKeysByPartyId};
-use dwallet_mpc_types::dwallet_mpc::{DWalletCurve, DWalletHashScheme, DWalletSignatureAlgorithm};
+use crate::network_key_id_mapping;
+use dwallet_mpc_types::dwallet_mpc::{
+    DWalletCurve, DWalletHashScheme, DWalletSignatureAlgorithm, NetworkKeyId,
+};
 use ika_protocol_config::ProtocolConfig;
+use ika_types::message::DWalletCheckpointMessageKind;
+use ika_types::messages_dwallet_mpc::{
+    DWalletNetworkEncryptionKeyData, DWalletNetworkEncryptionKeyState, SessionIdentifier,
+    SessionType,
+};
+use std::collections::{BTreeSet, HashMap};
+use std::sync::Arc;
+use sui_types::base_types::ObjectID;
 use tracing::info;
 
 /// The Fast Schnorr (VSS) signature algorithms — their internal presign pools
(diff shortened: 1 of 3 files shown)


omersadika (commented):
**Review verdict: would approve** (comment — GitHub forbids self-approval under the author account). Reviewed the standalone commit `2f8223feb3` (the stacked #1818 half reviewed separately there). This is the harder sibling of #1818 — same parking machinery, attacking the counter-uniformity invariant from the install-lag side — and it holds up:

**Verified in the diff:**
1. **Consume-first is enforced structurally, not by convention.** The sequence number is reserved as the *first* statement of `instantiate_internal_presign_session`, before identity resolution can bail — and the old tail-position `+= 1` (bypassable by early returns) is gone. The invariant ("the top-up decision is committee-uniform, so every validator must account every session") is stated exactly where it's load-bearing.
2. **`AwaitingKeyIdentity` is the right answer to the hard case.** When the identity isn't derivable yet, it parks the *reserved* sequence number and rebuilds the request at retry with the same `new_internal_presign(epoch, seq, curve, alg, key_id, identity)` call — byte-identical identifiers to what peers derived at instantiation. Boxing the large variant, silent re-park on the per-tick retry, and logging only real park→activate transitions (paired with the stall-playbook signature) are all considered touches.
3. **Closes the July-2 review's V8 tie-break finding verbatim** — `min_by` on `dkg_at_epoch` resolved same-epoch ties by `HashMap` iteration order; now `.then(a_id.cmp(b_id))`, with the divergence chain documented (different NOA key → different pool configs → divergent batch sizes → the same desync through another door).
4. **The shared `try_activate_internal_presign_request` fixes a pre-existing bug in passing**: the old `new_session` overwrite clobbered peer messages buffered before late local instantiation; both instantiation and retry now upgrade `WaitingForSessionRequest` in place.
5. **Test discipline**: per-round cross-validator counter asserts, the natural adopted-but-not-installed race hit organically (12 identity-parks in one round), fault-injection reproducing the exact production divergence (`33 vs 31`), and the out-of-scope multi-key holes written into `fast-schnorr-vss.md` rather than left as tribal knowledge.

**One exotic corner, non-blocking:** the identity-bytes fallback is asymmetric — a validator with the *installed* key whose `network_key_id()` derivation fails falls back to raw DKG-output bytes, while a lagging peer resolving via the cert-pinned mapping would use id bytes → divergent identifiers. Reaching it requires a cert to carry an id that local derivation can't reproduce, which is near-contradictory since cert ids come from the same derivation. Worth one defensive line someday.

**Bookkeeping for the v4-activation gate list** — after this PR, the session-identifier-uniformity family reduces to: the #1807 per-validator expiry-round desync (open major from the release review) plus the three holes this PR's spec update names (NOA selection over the still-installing set, cross-pool interleave under adoption skew, guard counters shared across keys).

**Merge timing**: same as #1818 — land **after** the 1.2.0 tag; v4-dark, keeps the running evidence batch valid for the tag commit.

PR #1820, opened 2026-07-12
fix(v4): pubkey-provider per-member robustness + prior-committee verification keying + SDK 0.5.0 changeset

omersadika:
First batch of the protocol-v4 gate-list fixes from the 1.2.0 release review. These paths ship dark in 1.2.0 but **activate automatically ~one epoch after the fleet finishes upgrading** (capabilities → `SetNextConfigVersion`, no separate vote), so they must be sound before a v4-advertising binary rolls out fleet-wide.

Scoped to the **isolated, low-risk** items; the interacting handoff-attestation cluster (V1/V2/V3/V4/V9b) and the presign-uniformity family land as separate PRs.

### V5 — one bad `validator_info` no longer suppresses joiner relay for the epoch
`pubkey_provider_updater::refresh()` aborted the whole joiner-provider install (`verify().map_err(...)?`) on any single unverifiable next-epoch member — retried every 5s forever, installing no provider, dropping every relayed joiner announcement. Now skips-and-logs per member (matching the reviewed `fetch_previous_committee` tolerance) with a loud incomplete-set warning. Design confirms the next-epoch path legitimately keys by current names, so the fix is skip-only, not re-keying.

### V9c — protocol-key rotation no longer silently drops a validator's cert stake
`fetch_previous_committee` keyed the consensus-key map by each member's **current** on-chain `protocol_pubkey`, while `voting_rights` uses the frozen `previous_committee` snapshot name. A validator that rotated its protocol key at the boundary signed its handoff attestation under the **old** name; the lookup by the new name returned `None`, its signature dropped as unverifiable, and its stake silently left the cert quorum — a benign rotation could fail the cert and fail-closed-halt a joiner. Both maps now key by the same prior-epoch snapshot name (resolved by validator id). The consensus-key *value* is fixed at registration, so reading the current one stays correct; only the map key changes. Out of scope (unfixable from chain state): consensus-key *value* rotation between signing and bootstrap.

### V7 (partial) — `get_committee` no longer panics on a decode error
`RocksDbStore::get_committee` unwrapped the `committee_map` read, crash-looping under `panic=abort` on any decode failure. Now propagates. The legacy-layout fallback decode for 1.1.8-written `Committee` records (mid-struct `consensus_keys` addition) is a **separate** change still in design — this PR only removes the panic.

### SDK 0.5.0 changeset
Adds the changeset (SchnorrkelSubstrate→Schnorrkel rename — key and runtime string; 30s→600s default poll timeout; V3-capable wasm **required before v4 activation**, since published `@ika.xyz/ika-wasm@0.2.1` throws on V3-tagged reconfiguration outputs) and fixes the stale `SchnorrkelSubstrate` row in the SDK README. npm 0.4.1 is wedged (= latest, breaking rename unpublished); this unblocks the bump.

## Validation
Compiles clean, clippy-clean, existing `pubkey_provider` tests pass. Per CLAUDE.md's reconfiguration/`sui_connector` verification rule, the cert-quorum and joiner-bootstrap behavior is validated by the **cluster suite** (dispatched on this branch) — the V5/V9c fault-injection scenarios (malformed member; boundary key-rotation) run through the `joiner.rs` harness. Fault-injection validation per the test-testing playbook is the pre-merge gate.

### DIFF
--- a/.changeset/ripe-schnorrkel-v4.md
+++ b/.changeset/ripe-schnorrkel-v4.md
@@ -1,0 +1,19 @@
+---
+'@ika.xyz/sdk': minor
+'@ika.xyz/ika-wasm': minor
+---
+
+Release 0.5.0 — protocol-v4 readiness and the `SchnorrkelSubstrate` rename.
+
+**Breaking**
+
+- `SignatureAlgorithm.SchnorrkelSubstrate` is renamed to `SignatureAlgorithm.Schnorrkel`. Both the enum key AND its runtime string value changed from `'SchnorrkelSubstrate'` to `'Schnorrkel'`. Update every reference; a value persisted or transmitted as `'SchnorrkelSubstrate'` is no longer recognized.
+
+**Required before protocol v4 activates**
+
+- The bundled WASM is rebuilt to parse V3-tagged network reconfiguration outputs. When the network advances to protocol v4, validators write V3-tagged reconfiguration outputs each epoch; the WASM shipped in `@ika.xyz/ika-wasm@0.2.x` only understands V1/V2 and throws a BCS "unknown variant" error from `getProtocolPublicParameters`. dApps must be on this release before v4 activates on their network, or user-side dWallet operations stop until they upgrade.
+
+**Behavior change**
+
+- The default poll timeout for the `*InParticularState` waiters increased from 30s to 600s. Flows that relied on the old 30s default failing fast will now wait longer before timing out; pass an explicit timeout to restore the previous behavior.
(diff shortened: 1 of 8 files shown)


omersadika:
CI green at head `4419ad9f` (the review-amended head):

- Test Cluster: ✅ [29282103749](https://github.com/dwallet-labs/ika/actions/runs/29282103749) (re-confirms the earlier green [29197873525](https://github.com/dwallet-labs/ika/actions/runs/29197873525) on the same head)
- Rust MPC Integration: ✅ [29282105649](https://github.com/dwallet-labs/ika/actions/runs/29282105649)

PR #1821, opened 2026-07-12
fix(v4): decode mainnet-v1.1.8 committee_map records via a LegacyCommittee fallback (V7)

omersadika:
Track A **prerequisite** of the protocol-v4 gate list (see `dev-docs/plans/v4-activation-gate-list.md`): the v3→v4 transition epoch can't proceed if a validator panics reading its own 1.1.8-written committee record — so this must land before the freeze/handoff work (V1/V3) and the presign machinery (#1818/#1819) that all run in that first v4 epoch.

## The bug

`Committee` gained a mid-struct `consensus_keys` field at v4 (#1762). It derives `Serialize`/`Deserialize` with no field skips, so its bcs layout is **positional** — a `committee_map` record written by mainnet-v1.1.8 has no `consensus_keys` bytes and does not decode under the v4 `Committee`. At the v3→v4 boundary a binary reads its **own** store written by its previous (1.1.8) version:
- `RocksDbStore::get_committee` unwrapped that error → `panic=abort` crash-loop (fixed separately in #1820);
- the joiner-bootstrap boundary read at `lib.rs` `.ok().flatten()`ed it into "absent", silently forfeiting the designed local trust anchor at exactly the transition epoch.

## The fix — one-time startup migration

`CommitteeStore::new` runs `migrate_legacy_records`: any `committee_map` record that fails the current decode but decodes as **`LegacyCommittee`** (an exact `Deserialize` mirror of the 1.1.8 layout) is rewritten in the current layout with **empty `consensus_keys`** (sound: a 1.1.8 prior epoch ran v3 and emitted no consensus-key-signed messages, so there's nothing to verify with it).

- The store converges to a single schema on first boot; the read path stays a plain decode (a decode error after migration is genuine corruption, propagated per #1820).
- Idempotent and crash-safe: already-migrated records pass the current decode and are skipped, so an interrupted migration just resumes next boot.
- **Fully deletable**: `LegacyCommittee` + the migration function are removed wholesale once no fleet upgrades directly from 1.1.8 data dirs — no permanent legacy-schema view in steady state.

(Earlier revision of this PR kept a permanent legacy fallback on the read path; reworked to the migration per review — smaller steady-state surface, no lasting backward-compat code.)

## Tests

- Unit (ika-types): a legacy record fails the primary `Committee` decode but recovers via `LegacyCommittee` with empty `consensus_keys` and preserved membership/stake; a current record does **not** decode under the legacy schema (bcs trailing-byte strictness) — the safety property behind "rewrite only records that fail the current decode".
- Store-level (ika-core): plants a 1.1.8-layout record in a real RocksDB, reopens the store, and reads it back migrated.

All green.

## Spec

`cross-binary-upgrade.md` invariant 5 extended: durable-local-record layout tolerance via the startup migration, plus the **reverse-direction rollback caveat** — a v4-written `committee_map` record is unreadable by 1.1.8; clear the column family before rolling back (it rebuilds from chain via `init_genesis_committee`/`insert_new_committee`).

Complements #1820's `storage.rs` unwrap→propagate; together they close V7.

https://claude.ai/code/session_01DPpwbWJcyVPEviCc941NBc

### DIFF
--- a/crates/ika-core/src/epoch/committee_store.rs
+++ b/crates/ika-core/src/epoch/committee_store.rs
@@ -8,7 +8,7 @@ use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use sui_types::base_types::ObjectID;
-use tracing::{info, warn};
+use tracing::info;
 use typed_store::rocks::{DBMap, DBOptions, MetricConf, ReadWriteOptions, default_db_options};
 use typed_store::rocksdb::Options;
 
@@ -19,13 +19,6 @@ use sui_macros::nondeterministic;
 
 pub struct CommitteeStore {
     tables: CommitteeStoreTables,
-    /// A second typed view over the SAME `committee_map` column family, whose
-    /// value type is the pre-`consensus_keys` (mainnet-v1.1.8) `Committee`
-    /// layout. Consulted only when a primary decode fails, so a record this
-    /// binary wrote (which decodes on the primary) is never read through the
-    /// legacy schema. Lets a v4 binary read a `committee_map` entry its own
-    /// prior version persisted, instead of panicking on the layout change.
-    legacy_committee_map: DBMap<EpochId, LegacyCommittee>,
     cache: RwLock<HashMap<EpochId, Arc<Committee>>>,
 }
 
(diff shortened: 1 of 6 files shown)


ycscaly (approved):

omersadika:
CI green at head `fbc92d1546` (the one-time-startup-migration redesign):

- **Upgrade Test `v118_upgrade`**: ✅ [29318023057](https://github.com/dwallet-labs/ika/actions/runs/29318023057) — the load-bearing run: boots the literal mainnet-v1.1.8 binary, swaps to this branch, and the scenario now **asserts the legacy-record migration end-to-end** as a hard test assertion (no more grepping logs for evidence).
- **Test Cluster**: ✅ [29318021536](https://github.com/dwallet-labs/ika/actions/runs/29318021536) — first attempt had one dead-at-spawn failure (`Address already in use` in the pinned sui-swarm's p2p startup, no ika code in the backtrace); the rerun on the same commit passed 22/22. That port-allocation race is now retried automatically by [#1833](https://github.com/dwallet-labs/ika/pull/1833).

PR #1822, opened 2026-07-12
fix(v4): make the VSS Shamir-cache derivation outcome explicit and traceable (V9a)

omersadika:
Track B, independent item from the protocol-v4 gate list (see `dev-docs/plans/v4-activation-gate-list.md`). No coupling to the handoff cluster or the presign-counter (C1) decision.

## The black hole

`derive_vss_shamir_cache_for_key` returned `Option`, and three distinct outcomes all collapsed to `None`:
- a **pre-V3 key** (VSS genuinely not applicable),
- a **bcs decode failure** of the V3 output,
- a **per-curve derivation failure**.

The call site inserted nothing and **logged nothing**. A VSS sign then read the cache via an accessor that returned `WaitingForNetworkKey` for a missing entry — indistinguishable from "derivation still running". With #1818's parking, the not-ready class parks and retries every service iteration, so a genuinely-failed or not-applicable key **parked forever, silently and untraceably** (the review's "untraceable `WaitingForNetworkKey` on every VSS sign").

## Fix — store the outcome, distinguish transient from terminal

- `derive_vss_shamir_cache_for_key` returns a tri-state `Derived / NotApplicable / Failed(reason)`; the two bcs decodes and three per-curve derivations map their error into `Failed` with a cause string instead of `.ok()?`.
- The cache field stores the **outcome** per key (`VssCacheEntry`), not only successes. The call site always records an entry and emits **exactly one** `error!` (with the cause) on `Failed`; it's overwritten on the next key update, so a transient failure re-derives.
- The accessor parks (`WaitingForNetworkKey`) only for an **absent** entry (derivation still running) or a **stale-epoch** `Derived` entry (the #1736 window — unchanged). A `NotApplicable`/`Failed` entry returns the new terminal `DwalletMPCError::VssShamirCacheUnavailable`, which is **not** in `is_network_key_data_not_ready()` — so a VSS sign **fails fast with a named error** instead of parking forever.

## Test

The load-bearing property: `VssShamirCacheUnavailable` is **not** classified as not-ready (so it doesn't park), while `WaitingForNetworkKey` is; plus a stable metric label. The six VSS-sign call sites are untouched (they inherit via `?`).

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
+++ b/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs
@@ -138,17 +138,29 @@ pub struct VssShamirCachePerKey {
 /// Storing the outcome (not only successes) lets the VSS sign path tell
 /// "still coming" (park and retry) from "will never come" (fail, traceably)
 /// instead of collapsing all three to a bare missing entry.
+///
+/// EVERY variant carries the epoch of the key data it was derived from, and
+/// the accessor treats an epoch mismatch exactly like a missing entry — the
+/// terminal variants included. A terminal entry derived from a PREVIOUS view
+/// of the key (e.g. the boundary window where the first ingest carried the
+/// pre-V3 key view and the fresh V3 re-derivation is still running on rayon)
+/// is "re-derivation in flight", not "will never come"; without the epoch
+/// tag a consumer trusting the terminal/transient distinction would drop out
+/// of VSS signing for the whole re-derivation window while its peers sign.
 #[derive(Clone, Debug)]
 pub enum VssCacheEntry {
-    /// The three-curve cache derived successfully. Boxed: it dwarfs the two
-    /// unit variants, so an unboxed enum would carry that size everywhere.
+    /// The three-curve cache derived successfully (carries its epoch as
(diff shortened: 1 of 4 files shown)


omersadika:
Review follow-ups pushed as `eb3bb07fab` (epoch-tagged terminal variants, staleness-first accessor, paging exemption, comment/spec fixes).

**Two corrections to this PR's original description:**

1. *"parked forever … now fails fast"* overstated the behavioral delta: sign-session instantiation never parked on `WaitingForNetworkKey` (every input-build error already produced `SessionStatus::Failed`), and the computation-layer read retries per tick for both error classes before and after this change. The internal-presign park path never reads the VSS cache. This PR is an **observability/semantics** improvement — a completed-but-unusable derivation is now a named, traceable outcome instead of being indistinguishable from "still deriving" — not a liveness fix.

2. Worth preserving deliberately: this PR also **removed a secret-material Debug-print**. The pre-existing `error!(error=?err, ...)` on the oneshot send-failure path Debug-formatted the `SendError` payload — `Option<VssShamirCachePerKey>` including the `secret_key_share_*` Shamir shares — into logs. The new `.is_err()` form drops the payload. Don't reintroduce the payload in that log line.

ycscaly (approved):

PR #1809, opened 2026-07-11
fix(sui-connector): stop permanently losing pruned checkpoints and bag entries (epoch-close wedge); add user-flow sim tests, re-enable all simtest reproducers

omersadika:
## What

Follow-up to #1808 (fault-simulation suite): that suite proved the machinery survives adversity; this PR proves the product works — and in doing so it cornered and fixed the open close-lock wedge, so **every previously-`#[ignore]`d simtest reproducer is re-enabled and the whole suite passes: 11/11, zero ignores** ([final gate](https://github.com/dwallet-labs/ika/actions/runs/29263090774)).

Contents, in dependency order: Rust-native drivers for every user-facing dwallet flow, two new happy-path sim tests that cross epoch boundaries, the simtest CI moved to the large runner, **two product fixes in the OCS/sui-connector pipeline**, and the re-enablement of all four reproducers (with one of them fixed as a test bug).

No existing tokio test is modified or removed.

## Product fixes (the close-lock wedge, root-caused)

The wedge — an epoch permanently pinned with `session_locked=true, all_epoch_sessions_finished=false` after a user session lands astride an epoch close — was two stacked state-loss bugs in the verified-state pipeline, cornered with deterministic debug runs of the new `sim_user_flows` test:

1. **The checkpoint pusher permanently lost the newest checkpoints of every pruner tick** (`push_worker.rs`). The fullnode's lowest-available watermark trails its executed head by roughly the pusher's old 2s poll, so every ~10s pruner tick vaporized the 2–3 newest checkpoints before the pusher fetched them — and a failed fetch **skipped the checkpoint forever**, leaving permanent gaps in the verified state cache. Fixed: 250ms poll, and failed fetches become *pending gaps* retried every tick and folded late when they materialize (safe: the cache is monotonic-by-version), never silent skips. Contract unit tests rewritten accordingly, including recovery of an end-of-epoch checkpoint on a previously failed seq. (An intermediate stop-and-retry shape was tried and rejected — it froze the whole cache behind one unfetchable checkpoint.)
[...]

PR #1823, opened 2026-07-12
fix(v4): make ready-signal canonicalization a pure function of the sequenced bytes (V2)

omersadika:
Track A consensus-determinism fix from the protocol-v4 gate list (see `dev-docs/plans/v4-activation-gate-list.md`). Precondition for V1's freeze determinism.

## The fork

`record_epoch_mpc_data_ready_signal` canonicalized the persisted `validated_peers` with a weight closure that consulted the local `validator_mpc_data_announcements` table: a next-epoch joiner (zero current-committee weight) was **kept only if present in that table**, otherwise **dropped**.

That table is populated for a relayed joiner announcement only after `verify_joiner_announcement` passes against the installed `JoinerPubkeyProvider` — a ~5s Sui-poll product installed on **wall-clock time** (buffered until then). So two honest validators processing the **same sequenced** ready signal persisted **different** canonical `validated_peers` depending on their provider-install timing: validator A (provider installed) keeps joiner J and can freeze it; validator B (provider lagging) drops J and excludes it. Divergent frozen set → divergent next-epoch committee mpc_data → forked reconfiguration participant set / handoff cert. This is exactly the taint `freeze_mpc_data_if_first` documents it avoids ("we deliberately do NOT read the local announcement table here") — laundered one step earlier, at receive-time canonicalization.

## The fix — pure function of the signal bytes

`canonicalize_ready_signal_peers` no longer consults any local table:
- **Keep** every deduped `(peer, hash)` pair — including zero-weight joiners — instead of dropping non-committee peers via the `announced`-table lookup. A joiner is frozen only when a stake-quorum of signers attest it in the tally (`compute_freeze_partition`); kept non-committee names are inert downstream (a garbage name never reaches a stake-quorum → lands in `excluded`, which is consulted only by committee-member key — verified).
- Measure quorum **coverage on current-committee weight only**, so a sparse signal still can't push the freeze trigger.
- A **deterministic length cap** (`K × current committee size`, K=4) bounds byzantine garbage padding; the deduped length is identical across honest validators, so the cap decision is itself consensus-deterministic.

Diagnostics `non_committee_dropped` → `non_committee_kept`; new `OverCap` outcome.

## Tests

Joiner retained in `Accept` while coverage stays committee-only; a signal of only zero-weight peers → `BelowQuorumCoverage`; over-cap dropped as a pure function of input length; the existing dedup/coverage/mixed-padding tests updated for the keep-not-drop semantics. All green.

## Spec

`validator-mpc-data-announcements.md`: receive-time canonicalization added to the ready-signal semantics and to invariant 1 (the "pure function of the consensus sequence" property now explicitly covers canonicalization, not just the freeze tally).

## Validation note

Per CLAUDE.md's reconfiguration/epoch-boundary rule, the determinism property (two validators with divergent provider timing persist byte-identical signals and compute the identical freeze partition) is validated by the **cluster suite** — the fault-injection scenario (force the provider absent on one node) is the pre-merge gate.

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -3282,23 +3282,52 @@ impl AuthorityPerEpochStore {
                 return Ok(());
             }
         }
-        // Surface byzantine-padding attempts. Placed AFTER the
-        // strict-superset gate so a byzantine signer re-submitting
-        // the same padded payload every consensus round doesn't
-        // log-flood: the gate drops the repeat above, so only the
-        // first padded payload (or a strictly-grown padded payload)
-        // makes it here. Honest emitters dedup + committee-filter
-        // before broadcast, so reaching this branch is a strong
-        // byzantine signal worth a `warn!` for operators.
-        if !diagnostics.non_committee_kept.is_empty() || diagnostics.duplicates_collapsed != 0 {
+        // Surface anomalies. Placed AFTER the strict-superset gate so a
+        // byzantine signer re-submitting the same payload every consensus
+        // round doesn't log-flood: the gate drops the repeat above, so only
+        // the first anomalous payload (or a strictly-grown one) makes it
+        // here. Two distinct signals, judged separately:
+        // - DUPLICATES are a genuine byzantine tell — honest emitters dedup
+        //   before broadcast — so they warn.
+        // - Non-committee names are NOT: honest emitters deliberately
+        //   include announced next-epoch JOINERS (zero current-epoch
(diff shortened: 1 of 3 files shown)


omersadika:
CI green at the main-merged head `04ddf6459e` (includes the merged #1809/#1820/#1821/#1822):

- Test Cluster: ✅ [29321828453](https://github.com/dwallet-labs/ika/actions/runs/29321828453) — 22/22 first-try (re-confirms the earlier green [29197874510](https://github.com/dwallet-labs/ika/actions/runs/29197874510) on the pre-merge head)

PR #1825, opened 2026-07-12
fix(v4): retry a failed NetworkKeyId derivation when its inputs change (networkkeyid Minor 1)

omersadika:
Small, standalone item from the protocol-v4 gate list (`dev-docs/plans/v4-activation-gate-list.md`). Follow-up to #1789's on-demand NetworkKeyId derivation.

## The bug

The cert-anchored adoption pass gated the expensive background NetworkKeyId derivation on a bare ObjectID set (`network_key_id_derivations_spawned: HashSet<ObjectID>`, insert-once). The spawned derivation **can fail** (its `Err` arm only logs and never `register`s), so `network_key_id_for` keeps returning `None` and the pass re-enters the unmapped-key block every overlay republish (~5s) — but the insert-once memo **suppresses re-spawn**, pinning the key to failure for the whole epoch.

The derivation is input-dependent (empty reconfiguration output → DKG path; non-empty → reconfiguration path), and the overlay legitimately republishes partial blobs during convergence. So a failure while the reconfiguration output is transiently empty/incomplete never self-heals.

## The fix

Memoize on the derivation-**input digest**: `HashMap<ObjectID, [u8;32]>` keyed by `Blake2b256(bcs(dkg_output, reconfiguration_output))`. Re-spawn iff the memoized digest differs.

- A deterministic failure on unchanged inputs is still spawned at most once (no rayon hammering).
- A new reconfiguration output changes the digest and re-derives — a transient-input failure self-heals on the next republish.
- A success registers in the global map and short-circuits before the gate, so the memo only governs the retry-on-failure case (exactly the bug).

## Test

The existing `unmapped_cert_referenced_key_defers_and_spawns_derivation` integration test is extended: an unchanged overlay does not respawn (len stays 1); a changed reconfiguration output updates the memo digest, proving a re-spawn happened. Green under real crypto.

> networkkeyid **Minor 2** (the barrier-side on-demand derivation, so an unmapped cert key isn't never-ready at the prepare-then-start barrier) folds into the barrier rewrite alongside V1/wedge/V4 per the cross-check, and is **not** in this PR.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/network_key_adoption.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/network_key_adoption.rs
@@ -430,17 +430,57 @@ async fn unmapped_cert_referenced_key_defers_and_spawns_derivation() {
         "an unmapped cert-referenced key must be deferred, not adopted"
     );
     assert!(
-        manager.network_key_id_derivations_spawned.contains(&key_id),
+        manager
+            .network_key_id_derivations_spawned
+            .contains_key(&key_id),
         "the background NetworkKeyId derivation must be spawned for the unmapped key"
     );
 
-    // Same overlay again: the deferral must have skipped the memoization
-    // (so the pass re-evaluates) and must not respawn the derivation.
+    // Same overlay again (identical derivation inputs): the memo is keyed on
+    // the input digest, so an unchanged overlay must not respawn.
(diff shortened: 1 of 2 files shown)


ycscaly (approved):
Note that its once per-change not per-input (we can do a -> b -> a and it runs 3 times not 2 times)

PR #1826, opened 2026-07-12
fix(v4): per-(key,curve,algorithm) internal-presign counters + consensus-agreed NOA key (multi-key-holes, expiry cross-key)

omersadika:
Track B of the protocol-v4 gate list (`dev-docs/plans/v4-activation-gate-list.md`), implementing the **per-key counter** decision (C1). Builds on #1818/#1819.

## The coupling bug

A single shared internal-presign sequence counter — plus shared `instantiated`/`completed` guard counters and a shared batch-instantiated round, all keyed only by `(curve, algorithm)` — coupled every pool's stream across keys. Because network-key **adoption** completes at a different consensus round on different validators (cert-gated, not round-uniform):

- a key adopted earlier on one validator shifted **every other pool's** sequence numbers → session identifiers (which bind the sequence number) diverged and never reached quorum (multi-key hole 2);
- one key's in-flight batch gated another key's pool of the same algorithm (hole 3);
- one key's stale-batch expiry perturbed another's batch-round (the cross-key dimension of #1807's expiry desync).

## PART A — per-`(key, curve, algorithm)` counters

Re-key four `mpc_manager` maps by `(ObjectID, DWalletCurve, DWalletSignatureAlgorithm)`: `next_internal_presign_sequence_number` (now a map, `entry().or_insert(1)`, still consumed **first and unconditionally** so #1819's parking stays aligned within a pool), `instantiated_`/`completed_internal_presign_sessions`, and `internal_presign_batch_instantiated_at_round`. Each pool's stream is now a pure function of that pool's own consensus-agreed inputs: a late-adopted key starts its pool fresh at 1 whenever it starts, identically on every validator, without perturbing any other pool.

## PART B (interim) — consensus-agreed NOA key

Select the NOA-signing key over the **adopted** set (`adopted_network_key_data`) instead of the wall-clock-**installed** set (`network_keys.network_encryption_keys`), so two validators with lagging installs can't pick different NOA keys and apply divergent pool configs (hole 1). The `dkg_at_epoch` + key-id tie-break is unchanged.

## Verification

Preserves #1818/#1819's consume-first/parking semantics (now per-pool). The **multi-key install-lag uniformity test passes** under real crypto (352s), and its cross-validator assertion is now the whole per-pool sequence **map** — a strictly stronger check that no pool's stream leaked into another's. clippy-clean, fmt.

## Out of scope (deferred to NOA-sign enablement)

The expiry-desync **root** fix — moving NOA-sign pool **consumption** onto the consensus round drain (it currently pops per wall-clock service iteration in `process_network_owned_address_sign_requests`). NOA sign is not live; the pool must not drain off the consensus timeline once it is, and that wiring lands with NOA-sign enablement. PART A already removes the cross-key dimension of the desync.

## Spec

`fast-schnorr-vss.md`: replaced the single-shared-counter invariant (from #1819) with the per-pool-counter invariant + the adopted-set NOA-key rule.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
@@ -10,17 +10,14 @@ use crate::dwallet_mpc::integration_tests::utils::{
 use crate::dwallet_mpc::mpc_manager::ParkedInternalPresignRequest;
 use crate::dwallet_mpc::mpc_session::SessionStatus;
 use crate::dwallet_mpc::{NetworkOwnedAddressSignRequest, ValidatorMpcKeysByPartyId};
-use crate::network_key_id_mapping;
-use dwallet_mpc_types::dwallet_mpc::{
-    DWalletCurve, DWalletHashScheme, DWalletSignatureAlgorithm, NetworkKeyId,
-};
+use dwallet_mpc_types::dwallet_mpc::{DWalletCurve, DWalletHashScheme, DWalletSignatureAlgorithm};
 use ika_protocol_config::ProtocolConfig;
 use ika_types::message::DWalletCheckpointMessageKind;
 use ika_types::messages_dwallet_mpc::{
     DWalletNetworkEncryptionKeyData, DWalletNetworkEncryptionKeyState, SessionIdentifier,
     SessionType,
 };
-use std::collections::{BTreeSet, HashMap};
+use std::collections::HashMap;
 use std::sync::Arc;
 use sui_types::base_types::ObjectID;
 use tracing::info;
@@ -72,11 +69,17 @@ const TEST_STALE_BATCH_EXPIRY_ROUNDS: u64 = 12;
 fn presign_batch_counters(
     test_state: &IntegrationTestState,
     service_index: usize,
-    network_key_id: ObjectID,
+    network_key_object_id: ObjectID,
     curve: DWalletCurve,
     algorithm: DWalletSignatureAlgorithm,
 ) -> (u64, u64) {
     let manager = test_state.dwallet_mpc_services[service_index].dwallet_mpc_manager();
+    // The counters are keyed by the content-derived NetworkKeyId; resolve it
+    // from the key's ObjectID the same way production does.
+    let Some(network_key_id) = manager.internal_presign_network_key_id(&network_key_object_id)
+    else {
+        return (0, 0);
+    };
     (
         manager
             .instantiated_internal_presign_sessions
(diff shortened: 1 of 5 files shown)


omersadika:
Amended per review feedback (*"avoid `network_key_id: ObjectID`, use `NetworkKeyId`"* + *"adopted keys always have a `NetworkKeyId`"*) — pushed as `c191610c41`.

**What changed**
- The four internal-presign per-pool counters (`next_internal_presign_sequence_number`, `instantiated_/completed_internal_presign_sessions`, `internal_presign_batch_instantiated_at_round`) are now keyed by `(NetworkKeyId, curve, signature_algorithm)` instead of `(ObjectID, …)`. The `NetworkKeyId` is the *same* content-derived identity bound into the session identifier, so the counter and the identifier can no longer key on divergent axes.
- The top-up loop resolves the key's `NetworkKeyId` **once** per adopted key before touching any counter. An adopted key always resolves (`adopt_cert_verified_keys` defers a key with no mapping), so a `None` is a should-never-happen the loop **skips** (logged), never a divergent fallback.
- Since the identity is resolved *before* instantiation, the request is always buildable there → the `ParkedInternalPresignRequest::AwaitingKeyIdentity` reserve-before-identity variant is unreachable and **removed**. `ParkedInternalPresignRequest` is now a one-field tuple struct wrapping the built request.
- `internal_presign_network_key_identity_bytes(&self) -> Option<Vec<u8>>` (with a raw-DKG-output fallback) → `internal_presign_network_key_id(&self) -> Option<NetworkKeyId>`.

**Spec + test**
- `dev-docs/specs/fast-schnorr-vss.md`: the per-pool-counter and deferred-instantiation rules rewritten for the `NetworkKeyId` axis and the skip-not-park behavior.
- `test_internal_presign_multi_key_install_lag_keeps_identifiers_uniform`: the Phase-4 coda now asserts an adopted-but-unresolvable key is **skipped uniformly** (no park, no counter movement, no failure) instead of parking `AwaitingKeyIdentity`.

**Validation**: test green in release (307s), `cargo clippy -p ika-core --tests` clean, `cargo fmt` applied.

Out of scope (unchanged): the NOA-sign session builder still derives its own identity bytes inline — that's the sign-session identifier basis, not a presign counter key, and must stay byte-stable.

omersadika:
**CI status for this branch — all suites green:**

- Rust MPC integration suite: ✅ [run 29197879713](https://github.com/dwallet-labs/ika/actions/runs/29197879713)
- TS integration suite: ✅ [run 29203182339](https://github.com/dwallet-labs/ika/actions/runs/29203182339) (fresh localnet)

**Flake triage for the record** — the first TS run ([29197880424](https://github.com/dwallet-labs/ika/actions/runs/29197880424)) failed with 6 timeouts and was investigated before re-running, since this PR's adoption/counter changes could plausibly produce a wedged-localnet symptom. The localnet logs ruled that out:

- Zero `should_never_happen` pages and zero unresolvable-NetworkKeyId skips — the invariant this PR enforces held for the entire 2-hour run; the new cert-less deferral fired exactly 4× (once per validator for the fresh DKG key) and resolved immediately.
- Zero sign failures, zero `InvalidParameters`, zero malicious convictions, 13 healthy epoch transitions.
- MPC output quorums were still landing at 17:37 — *after* the first test timeout at 17:09 — so the network was degrading in latency, not wedged.
- All 6 failures were timeouts confined to the **final** test file (~75 min in); the same flows passed earlier in the same run.

That is the documented cumulative-load pattern (one long localnet session slows progressively until the last heavy file times out, with `sign_failures = 0` — a runner-resource limit, not a code defect). The fresh-localnet re-run passing confirms the attribution.

PR #1835, opened 2026-07-14
fix(dwallet-mpc): level-triggered drain for requests parked on a network key — reshare across a restart no longer wedges the epoch (#1834)

### LINKED ISSUE #1834: Reshare in flight across a restart is stranded forever: chain-copy key adoption never drains requests_pending_for_network_key (wedges the epoch)
Found while validating #1821 with the `v118_upgrade` rehearsal; reproduced **twice back-to-back** — once on `main` + #1821, once on plain `origin/main` (`ce81f0…` lineage) — so it is a pre-existing main bug, not introduced by any open PR. It wedges epoch 2→3 permanently after the atomic 1.1.8→local swap.

## Symptom

After the swap, the end-of-publish gate sticks forever with exactly two false conditions:

```
end-of-publish gate STUCK ... epoch=2 stuck_secs=420
  all_immediate_sessions_completed=false
  all_network_encryption_keys_reconfiguration_completed=false
```

Both are two views of the same missing artifact: the 2→3 network-key reshare session's output, which is never produced. The reshare session appears on all 4 validators exactly once, `active=false`, and never advances a round. General session intake still works (the workload's user DKG reaches `active=true` and is correctly deferred past the epoch lock) — only the reshare is stuck.

## Root cause

The reshare request is the **only** session type that sets both `requires_network_key_data: true` **and** `requires_next_active_committee: true` (`sui_connector/sui_event_into_request.rs:311-325`).

Post-restart intake (`dwallet_mpc/mpc_session.rs:483-671`):
1. clears the next-active-committee gate when the committee arrives, then
2. parks in `requests_pending_for_network_key` (`:501-531`) because the key's public data isn't registered yet. Neither deferral logs at INFO — the strand is silent.

That parked queue is **edge-triggered**: it drains only for key ids in `newly_instantiated_network_key_ids` (`mpc_session.rs:425-441`), which is populated **solely** by the consensus-round key-instantiation path (`dwallet_mpc_service.rs:1231-1235`).

But on a fresh boot the key is materialized via the **cert-less chain-copy adoption path** (`mpc_manager.rs:~627-680`, log line "adopting reconfigured network key from the chain copy (off-chain metadata disabled; no handoff cert exists)"), which never emits that edge. The parked reshare request is never re-driven → session never becomes Active → no MPC output → both gate counters stay short → the epoch can never close.

Steady-state (no restart) reshares are unaffected: their key instantiates through the consensus path, so the edge fires. The bug bites specifically on **restart with a reshare in flight** — i.e. every real upgrade whose restart window overlaps the mid-epoch reconfiguration. In the rehearsal the collision is not even a narrow race: the harness's swap window and the notifier's mid-epoch reconfiguration land together on this epoch cadence, and both runs hit it identically.

## Evidence

- Failed run (main + #1821): `/mnt/nvme0n1p1/tmp/v118-run-1821.log`, cluster logs `/mnt/nvme0n1p1/tmp/ika-v118-upgrade/`
- Control run (plain origin/main): `/mnt/nvme0n1p1/tmp/v118-run-main.log`, cluster logs `/mnt/nvme0n1p1/tmp/ika-v118-upgrade-main/`
- Reshare requested on-chain by the executor mid-epoch step (notifier log "Running network encryption key mid-epoch reconfiguration") right as the swap begins; session added `active=false` on all validators; `Advanced session` count = 0 post-swap; the sole decode-looking errors are the benign #1821 migration probes.

## Fix direction

Either:
1. **Level-triggered drain** (preferred): retry `requests_pending_for_network_key` every service cycle, the way `requests_pending_for_frozen_mpc_data` already works (`mpc_session.rs:449-455`) — removes the edge/level trap for every future caller; or
2. Feed the boot/chain-copy adoption path's key ids into `newly_instantiated_network_key_ids` so the existing edge fires.

Also worth adding: an INFO log on both park branches — the current silence made this a multi-hour diagnosis.

## Relation to existing work

- Belongs on the protocol-v4 activation gate list (`dev-docs/plans/v4-activation-gate-list.md`): it blocks the exact 1.1.8→1.2.0 mainnet upgrade path, deterministically in the rehearsal.
- Adjacent to (but not covered by) the deferred barrier/close-gate cluster; the stuck-gate log's own pointer to #1736 appears to be this mechanism, or a sibling of it.
- The `v118_upgrade` rehearsal currently cannot pass until this is fixed — it is the pre-merge validation vehicle for several open v4 PRs (#1820–#1827).

https://claude.ai/code/session_01DPpwbWJcyVPEviCc941NBc

omersadika:
Fixes #1834 — **release blocker**: a reshare in flight across the upgrade restart wedged the epoch permanently on the exact mainnet 1.1.8→1.2.0 atomic-swap path (reproduced twice in the `v118_upgrade` rehearsal, on plain main and main+#1821; the batch's green rehearsal runs simply didn't overlap the swap window with the mid-epoch reshare).

## Mechanism

Post-restart intake parks the reshare in `requests_pending_for_network_key` (its key's public data isn't registered yet). The queue drained only for keys in the **newly-instantiated edge list**, emitted solely by the consensus-round instantiation path — but a fresh boot materializes the key via the **cert-less chain-copy adoption**, which never emits that edge. The parked request never re-drove → the session never activated → no reshare output → `all_immediate_sessions_completed` and `all_network_encryption_keys_reconfiguration_completed` stayed false forever → the epoch could never close.

## Fix (the issue's preferred direction)

- **Level-triggered drain**: every `handle_mpc_request_batch` cycle drains any parked key whose public data now exists — same discipline as `requests_pending_for_frozen_mpc_data`, and indifferent to *which* path installed the key. Consensus-path timing is unchanged (the old edge also took effect on the next service iteration).
- **Edge plumbing removed end-to-end** (it existed only to feed this drain): the id no longer threads through `run_service_loop_iteration` → `handle_new_requests` → `handle_mpc_request_batch`; `poll_pending_network_key_instantiations` no longer collects/returns ids; the test-harness per-party threading is deleted; 51 call sites mechanically updated.
- **Observability**: both park branches (network key / next active committee) now log at **INFO** once per parked request, and the drain logs what it releases — the silence made this a multi-hour diagnosis.

## Validation

- `network_key_received_after_start_event` (the existing integration test that reproduces the exact park-then-materialize flow) — ✅ green in release, real crypto, 334s.
- **Fault-injection proof** (per `dev-docs/playbooks/test-testing.md`): forcing the drain shut (`filter(|_| false)`) makes that test **FAIL at the party-advancement cap** — the pre-fix strand — then reverted. The test genuinely guards the #1834 mechanism.
- The true end-to-end vehicle is the `v118_upgrade` rehearsal at the wedging cadence — dispatching on this branch; @ycscaly, if your local cadence reproduces deterministically, a run of your repro against this branch would be the strongest confirmation.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
@@ -68,7 +68,6 @@ use prometheus::Registry;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use sui_types::base_types::ObjectID;
 use sui_types::messages_consensus::Round;
 #[cfg(any(test, feature = "test-utils"))]
 use tokio::sync::watch;
@@ -437,7 +436,6 @@ impl DWalletMPCService {
             "Spawning dWallet MPC Service"
         );
 
-        let mut newly_instantiated_network_key_ids = vec![];
         loop {
             match self.exit.has_changed() {
                 Ok(true) => {
@@ -470,18 +468,13 @@ impl DWalletMPCService {
                 break;
             }
 
-            newly_instantiated_network_key_ids = self
-                .run_service_loop_iteration(newly_instantiated_network_key_ids)
-                .await;
+            self.run_service_loop_iteration().await;
 
             tokio::time::sleep(Duration::from_millis(READ_INTERVAL_MS)).await;
         }
     }
 
-    pub(crate) async fn run_service_loop_iteration(
-        &mut self,
-        newly_instantiated_network_key_ids: Vec<ObjectID>,
-    ) -> Vec<ObjectID> {
+    pub(crate) async fn run_service_loop_iteration(&mut self) {
         debug!("Running DWalletMPCService loop");
 
         // Ingest the per-epoch off-chain validator MPC keys (3 PVSS + VSS HPKE)
@@ -510,13 +503,10 @@ impl DWalletMPCService {
         self.process_network_owned_address_sign_requests();
 
         // Receive **new** dWallet MPC events and save them in the local DB.
-        let rejected_sessions = self
-            .handle_new_requests(newly_instantiated_network_key_ids)
-            .await
-            .unwrap_or_else(|e| {
-                error!(error=?e, "failed to handle new events from DWallet MPC service");
-                vec![]
-            });
+        let rejected_sessions = self.handle_new_requests().await.unwrap_or_else(|e| {
+            error!(error=?e, "failed to handle new events from DWallet MPC service");
+            vec![]
+        });
 
         // Adopt locally-observed network-key outputs (cert-digest-gated)
         // and spawn instantiation for any not yet installed — once per
@@ -538,9 +528,12 @@ impl DWalletMPCService {
         self.process_consensus_rounds_from_storage().await;
         // Network-key instantiations complete asynchronously on the rayon
         // pool; poll them once per ITERATION (not per consensus round) so
-        // a completed key installs even when no new rounds arrived.
-        let newly_instantiated_network_key_ids = self
-            .dwallet_mpc_manager
+        // a completed key installs even when no new rounds arrived. Requests
+        // parked on a key drain via the level-triggered check in
+        // `handle_mpc_request_batch` on the next iteration — regardless of
+        // whether the key materialized here or through the chain-copy
+        // adoption path (issue #1834).
+        self.dwallet_mpc_manager
             .poll_pending_network_key_instantiations()
             .await;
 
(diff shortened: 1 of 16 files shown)


omersadika:
CI green on this branch:

- **Rust MPC Integration**: ✅ [29327887338](https://github.com/dwallet-labs/ika/actions/runs/29327887338) — includes `network_key_received_after_start_event`, the park-then-materialize flow this PR's level-triggered drain now guards (fault-injection-proven in the PR body).
- **Upgrade Test `v118_upgrade` + `v118_churn`**: ✅ [29327888791](https://github.com/dwallet-labs/ika/actions/runs/29327888791) — necessary but not sufficient for #1834: the wedge is cadence-dependent (CI's swap window doesn't reliably overlap the mid-epoch reshare). @ycscaly's deterministic local repro remains the strongest end-to-end confirmation vehicle for this fix.

PR #1837, opened 2026-07-14
fix(committee): marker-guarded one-time legacy migration — unbreaks msim/debug store reopens (#1836)

### LINKED ISSUE #1836: Committee legacy-record migration panics debug/msim builds on every store reopen (typed-store debug_fatal on schema-mismatched records)
Found by the post-merge simtest run on main ([29325949651](https://github.com/dwallet-labs/ika/actions/runs/29325949651)): `sim_two_laggards_one_epoch` panics during a node restart —

```
panicked at .../typed-store/src/rocks/safe_iter.rs:99:21
  4: ika_core::epoch::committee_store::CommitteeStore::migrate_legacy_records
  5: ika_core::epoch::committee_store::CommitteeStore::new
```

## Root cause

The #1821 migration (`migrate_legacy_records`, runs at **every** `CommitteeStore::new`) iterates the whole `committee_map` CF under the **legacy** schema and relies on "a current-layout record fails the legacy decode and is yielded as an `Err` item; skip it."

That contract only holds in release builds. The pinned typed-store calls `debug_fatal!` on any key/value that fails the view's schema — in `safe_iter` (safe_iter.rs:96/99) **and** in `get()` (mod.rs:1685). Under debug assertions (msim/simtest, plain `cargo test`, any debug validator) `debug_fatal!` **panics** before the `Err` item is ever yielded.

Consequences:
- **msim**: every node restart reopens the store; the CF now holds current-layout records; the legacy-view scan panics the node. `sim_two_laggards_one_epoch` (which restarts laggards) fails deterministically on main.
- The `tables.committee_map.get(&epoch).is_ok()` probe inside the loop has the mirror problem on genuine 1.1.8 records (current-schema decode of a legacy value → `debug_fatal`).
- Release builds are unaffected (`debug_fatal` logs instead of panicking) — which is why the `v118_upgrade` rehearsal passed while simtest fails.

## Fix direction

1. **Scan once, not on every open**: add a small schema-version marker (its own CF); when present, skip the scan entirely. First open of a fresh DB scans an empty CF (safe under msim), writes the marker; restarts never re-scan. A genuine 1.1.8 DB has no marker → scans once → all records decode as legacy (no mismatches, no `debug_fatal`) → rewrite + marker.
2. **Drop the current-decode probe** inside the loop: the two layouts are mutually exclusive on decode success (a current record leaves trailing bytes under the legacy schema; a legacy record exhausts input under the current one), so "legacy decode succeeded" already implies "needs rewriting".
3. Residual accepted corner: a crash mid-scan on a real 1.1.8 upgrade followed by a reopen in a DEBUG build would re-scan a mixed CF and panic; release re-scan handles it. Document.

omersadika:
Fixes #1836 — main's simtest is red: `sim_two_laggards_one_epoch` panics on node restart inside `CommitteeStore::new` (`typed-store safe_iter.rs:99`, run [29325949651](https://github.com/dwallet-labs/ika/actions/runs/29325949651)).

## Root cause

The #1821 migration scanned the whole `committee_map` CF under the **legacy** schema at **every** store open, relying on schema-mismatched records being yielded as `Err` items. That contract only holds in release: the pinned typed-store `debug_fatal!`s — a **panic** under debug assertions (msim, plain `cargo test`, debug validators) — on any record failing the reading view's schema, in both `safe_iter` **and** `get()`. Any store holding current-layout records (i.e. every store after its first reconfiguration) therefore panicked on reopen under msim. Release logs-and-continues — which is why `v118_upgrade` passed while simtest fails.

## Fix

1. **Scan once, marker-guarded**: a `committee_schema_version` singleton (own CF — consulting it never decodes a committee value) skips the scan when present-and-current. Fresh dir: scans an empty CF, writes the marker. Genuine 1.1.8 dir: no marker → scans once (all records decode as legacy — no mismatch, no `debug_fatal`) → rewrites → marker written **after** the scan (crash mid-migration re-scans). Marker read error → falls through to the scan.
2. **Current-decode probe removed**: the two layouts are mutually exclusive on decode success (current leaves trailing bytes under legacy; legacy exhausts input under current), so legacy-decode success already implies the rewrite — and the probe `debug_fatal`'d on every genuine 1.1.8 record in debug builds.
3. Documented residual: a marker-less **mixed** CF (crash mid-scan on a real 1.1.8 upgrade + reopen in a debug build) still panics; the release re-scan handles it.

## Validation

- Tests restructured for marker semantics, split by profile (the mixed-CF test is release-only *because the mixed scan panics by design in debug* — that's the bug class itself). Green in **both** debug and release.
- **Fault-injection proof**: disabling the marker gate makes the debug-profile tests fail with the *exact* production signatures — `safe_iter.rs:99` / `mod.rs:1685`, `Failed to deserialize value in cf committee_map: remaining input` — the same line as main's simtest failure. Reverted.
- Dispatching simtest (the reproducing suite) + cluster on this branch.

### DIFF
--- a/crates/ika-core/src/epoch/committee_store.rs
+++ b/crates/ika-core/src/epoch/committee_store.rs
@@ -8,7 +8,7 @@ use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use sui_types::base_types::ObjectID;
-use tracing::info;
+use tracing::{info, warn};
 use typed_store::rocks::{DBMap, DBOptions, MetricConf, ReadWriteOptions, default_db_options};
 use typed_store::rocksdb::Options;
 
@@ -27,7 +27,21 @@ pub struct CommitteeStoreTables {
     /// Map from each epoch ID to the committee information.
     #[default_options_override_fn = "committee_table_default_config"]
     committee_map: DBMap<EpochId, Committee>,
+    /// Singleton marker: the `committee_map` value-schema generation this
+    /// store was last opened under (see `COMMITTEE_SCHEMA_VERSION_CURRENT`).
+    /// Present-and-current ⇒ every record is current-layout and the legacy
+    /// migration scan is skipped at open. Absent ⇒ the CF may hold
+    /// mainnet-v1.1.8 records (or is brand new) and the one-time scan runs.
+    /// Its own column family, so consulting it never decodes a committee
+    /// value under a possibly-mismatched schema.
+    committee_schema_version: DBMap<(), u64>,
 }
+
+/// `committee_map` value-schema generations: the mainnet-v1.1.8 layout
+/// (pre-`consensus_keys`) is generation 1; the current layout is 2. Only
+/// the current value is ever written; the constant exists so the marker
+/// is self-describing when inspected.
+const COMMITTEE_SCHEMA_VERSION_CURRENT: u64 = 2;
 
 // These functions are used to initialize the DB tables
 fn committee_table_default_config() -> DBOptions {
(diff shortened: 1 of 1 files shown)


omersadika:
CI green on this branch — including the reproducing suite:

- **Simtest**: ✅ [29328776721](https://github.com/dwallet-labs/ika/actions/runs/29328776721) — `sim_two_laggards_one_epoch` **PASS (512s)**, the exact test that panicked on main ([29325949651](https://github.com/dwallet-labs/ika/actions/runs/29325949651), `safe_iter.rs:99` via `migrate_legacy_records` on node restart). All 11 sim tests green.
- **Test Cluster**: ✅ [29328778154](https://github.com/dwallet-labs/ika/actions/runs/29328778154)

Together with the fault-injection proof in the PR body (gate disabled → debug tests reproduce the exact production panic signature), this closes #1836: main's simtest is unbroken by this PR.

PR #1841, opened 2026-07-15
ci: gate releases on literal v1.1.8 mixed rollout

omersadika:
## Why

The upgrade suite did not reproduce a decentralized validator rollout. `cross_binary` built the candidate twice and only pinned one copy to protocol v3, while the literal-v1.1.8 scenarios replaced every validator before intentionally crossing their tested network-key reshare. Because `stop_and_swap([0, 1, 2, 3])` is sequential and on-chain epoch advancement needs only quorum, those green runs could not detect one upgraded validator producing different MPC output bytes, lagging locally, or recognizing itself as malicious while three old validators kept the network moving.

This matters because the testnet v1.1.8 -> v1.2.0 incident happened with both binaries selecting protocol config v3. A shared protocol number is not evidence that cryptography implementation, transcript construction, serialization, or output finalization is wire-compatible.

## What changed

- Adds `v118_mixed_rollout`: boot four validators from the literal `mainnet-v1.1.8` tag, complete genesis DKG, upgrade exactly validator 0 to the exact candidate SHA, then keep the 1-current/3-old committee mixed through two protocol-v3 network-key reconfigurations using real cryptography and production presign sizing.
- Adds deterministic DSL witnesses for reconfiguration-not-started, started, and completed state, plus fail-closed local epoch, health, protocol ceiling, malicious count, log absence, output convergence, and pending-session assertions.
- Adds protocol-generic MPC session observations for every identified protocol: per-authority canonical output digests, malicious-actor counts, rejected envelopes, and pending-session counts. The mixed-rollout gate filters these reusable metrics to `Network Encryption Key Reconfiguration`; future upgrade scenarios can apply the same machinery to DKG, presign, sign, or other protocols. Digests cover canonical BCS output bytes, exclude the sender/malicious envelope, preserve multiple digests per authority, and therefore fail on a 3-vs-1 split even if the majority result advances the chain.
- Makes metric reads fail closed on dead/unreachable validators and missing/malformed samples, with validator index and endpoint in every error. The intentionally-malicious scenario now uses the same scraper.
- Makes historical `MAX_PROTOCOL_VERSION` patching fail if the assignment is absent, duplicated, already equal (no substitution), or not exactly the requested value afterward. Old refs are resolved to immutable commits before worktree creation.
- Adds focused PR path filters, a reusable `workflow_call`, and a tag-release dependency that tests the exact candidate SHA. Manual dispatch and every existing scenario remain available; `test=all` now includes the mixed rollout.
- Bumps the workspace release version to `1.2.1` and refreshes both the root and independent WASM lockfiles.
- Updates the upgrade contract, CI playbook, metric inventory, and the misleading “atomic swap” documentation.

## Coverage flow

| Scenario | Binary topology at tested reshare | Protocol behavior | What it proves |
|---|---|---|---|
| `smoke` | current only | no upgrade | same-binary process plumbing |
| `workload` | current only | v3 -> v4 | DKG/Presign/Sign across activation |
| `cross_binary` | current source + current source max-v3 pinned | v3 -> v4 | current-source wire compatibility and churn |
| `v118_upgrade` | literal v1.1.8, then all validators sequentially current | v3 -> v4 | historical RocksDB/key continuity and pre-activation global presign |
| `v118_churn` | current only by churn reshare | v3 -> v4 | v1.1.8-origin key reshared to a new validator |
| **`v118_mixed_rollout`** | **1 exact candidate + 3 literal v1.1.8** | **held at v3 for two reshares** | **production rollout topology; canonical output convergence (upgraded validator + finalizing quorum, hard digest equality), zero false-malicious, no stranded member/session** |
| `legacy_config` | current only | v3 -> v4 | legacy JSON-RPC configuration |

## Mixed-rollout invariants

For both network-key reconfigurations, the test requires:

- all four expected validator processes running, admin-reachable, and locally in the expected epoch;
- on-chain and per-process protocol version at most 3;
- explicit on-chain start and completion of the epoch's reconfiguration;
- exactly zero malicious actors on the current observer and no self-malicious log marker on any validator;
- one identical canonical network-key output digest across every observed authority, where the observed set must include the upgraded validator and at least a finalizing quorum (reconfiguration finalizes at quorum and discards a still-computing straggler's output, so an honest old-binary straggler may legitimately never submit); any submitted divergent digest is a hard failure, with zero malicious actors and no rejected envelope in every report;
- no incomplete on-chain immediate/system session, no local pending reconfiguration session, and no in-flight network-key instantiation.

Epoch advancement and successful quorum workloads are intentionally insufficient assertions.

## Validation

Local:

- `cargo test --release -p ika-upgrade-test --lib` (11 passed)
- `cargo test --release -p ika-upgrade-test --tests -- --test-threads=1` (all test targets compiled/executed; external scenarios remain environment-gated)
- `cargo test --release -p ika-core dwallet_mpc::integration_tests::session::test_handle_mpc_request_with_invalid_protocol_data_returns_failed -- --test-threads=1` (real cryptography; 1 passed)
- `cargo clippy -p ika-upgrade-test --all-targets --all-features -- -D warnings`
- `cargo clippy -p ika-core --lib --all-features -- -D warnings`
- `cargo check --release -p ika-core --lib`
- workflow YAML parse and `git diff --check`
- after the release bump: `cargo check -p ika-upgrade-test --tests` plus locked root and independent-WASM metadata resolution at workspace version `1.2.1`

Large-runner evidence:

- Required literal/history batch (`v118_upgrade` and `v118_churn` passed; the mixed rollout reproduced the blocker): https://github.com/dwallet-labs/ika/actions/runs/29410941527
- Corrected `cross_binary` rerun (`1 passed; 0 ignored`): https://github.com/dwallet-labs/ika/actions/runs/29411679357
- Deliberate output-mutation run proving the convergence guard fails closed: https://github.com/dwallet-labs/ika/actions/runs/29426981676
- First unfaulted literal v1.1.8/current mixed rollout after the compatibility fix (`1 passed; 0 ignored`, two protocol-v3 network-key reshares and MPC workloads): https://github.com/dwallet-labs/ika/actions/runs/29481354745
- Exact `1.2.1` candidate SHA `4306469d06940ee632fe98b261ea64f601514748` release gate (queued): https://github.com/dwallet-labs/ika/actions/runs/29485803984

## Compatibility defect and fix

The first literal mixed-committee run reproduced the incident: the upgraded authority sent its first-round network-key reconfiguration message, all three v1.1.8 validators rejected its proof, the old-validator quorum advanced while marking the upgraded authority malicious, and the upgraded validator logged `recognized_self_as_malicious`.

The failure was after message deserialization and before output adoption. The incompatible transcript came from the backward-compatible reconfiguration proof using `SECRET_KEY_SHARE_LIMBS` where v1.1.8 used `SECRET_KEY_SHARE_WITNESS_LIMBS`. The fix is published in public `inkrypto` commit [`c3c20ac67baa862c38dfc0f43dcbd6b0e94ab92d`](https://github.com/dwallet-labs/inkrypto/commit/c3c20ac67baa862c38dfc0f43dcbd6b0e94ab92d), which this PR pins.

The passing mixed-rollout run above demonstrates the compatibility fix against the literal historical binary with real cryptography. The exact `1.2.1` candidate run is still queued, so merge readiness remains gated on that result.

Related: #1736, #1834, and #1835.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
@@ -154,6 +154,27 @@ pub struct DWalletMPCMetrics {
     /// assert exclusion programmatically instead of grepping logs.
     pub(crate) malicious_actors_count: IntGauge,
 
+    /// Number of network-key reconfiguration sessions that are still
+    /// non-terminal on this validator. This is deliberately unlabeled and
+    /// always exported so a release-gate assertion can distinguish a clean
+    /// zero from a missing metric.
+    pub(crate) network_key_reconfiguration_sessions_pending: IntGauge,
+
+    /// Per-authority network-key reconfiguration outputs observed through
+    /// consensus. The digest is over canonical BCS output bytes and excludes
+    /// the sender/malicious envelope. A current validator therefore exposes
+    /// the literal previous-release validators' submitted results without
+    /// requiring instrumentation in the historical binary.
+    pub(crate) network_key_reconfiguration_output_info: IntGaugeVec,
+
+    /// Malicious-actor count carried in each authority's network-key
+    /// reconfiguration output report.
+    pub(crate) network_key_reconfiguration_reported_malicious_actors: IntGaugeVec,
+
+    /// Whether each authority's network-key reconfiguration output report was
+    /// rejected (0 = canonical output, 1 = rejected output).
+    pub(crate) network_key_reconfiguration_output_rejected: IntGaugeVec,
+
     /// Histogram-by-bucket of how long each currently-Active session has been
     /// tracked for. Labels: `session_type` (`user` / `system` /
     /// `internal_presign` / `noa_sign`), `age_bucket` (`<30s`, `<5m`, `<30m`,
@@ -445,6 +466,34 @@ impl DWalletMPCMetrics {
                 registry
             )
             .unwrap(),
+            network_key_reconfiguration_sessions_pending: register_int_gauge_with_registry!(
+                "ika_dwallet_mpc_network_key_reconfiguration_sessions_pending",
+                "Network-key reconfiguration sessions still non-terminal on this validator",
+                registry,
+            )
+            .unwrap(),
+            network_key_reconfiguration_output_info: register_int_gauge_vec_with_registry!(
+                "ika_dwallet_mpc_network_key_reconfiguration_output_info",
+                "Per-authority canonical network-key reconfiguration output digest observed through consensus",
+                &["session_id", "authority", "output_digest"],
+                registry,
+            )
+            .unwrap(),
+            network_key_reconfiguration_reported_malicious_actors:
+                register_int_gauge_vec_with_registry!(
+                    "ika_dwallet_mpc_network_key_reconfiguration_reported_malicious_actors",
+                    "Malicious-actor count carried in each authority's network-key reconfiguration output report",
+                    &["session_id", "authority"],
+                    registry,
+                )
+                .unwrap(),
+            network_key_reconfiguration_output_rejected: register_int_gauge_vec_with_registry!(
+                "ika_dwallet_mpc_network_key_reconfiguration_output_rejected",
+                "Whether each authority's network-key reconfiguration output report was rejected",
+                &["session_id", "authority"],
+                registry,
+            )
+            .unwrap(),
             active_sessions_by_age: register_int_gauge_vec_with_registry!(
                 "ika_dwallet_mpc_active_sessions_by_age",
                 "Active session count by session type and age bucket",
@@ -607,6 +656,10 @@ impl DWalletMPCMetrics {
         self.user_session_local_output_rejected.reset();
         self.user_session_output_received_from.reset();
         self.user_session_distinct_output_digests.reset();
+        self.network_key_reconfiguration_output_info.reset();
+        self.network_key_reconfiguration_reported_malicious_actors
+            .reset();
+        self.network_key_reconfiguration_output_rejected.reset();
     }
 }
 
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
@@ -154,26 +154,24 @@ pub struct DWalletMPCMetrics {
     /// assert exclusion programmatically instead of grepping logs.
     pub(crate) malicious_actors_count: IntGauge,
 
-    /// Number of network-key reconfiguration sessions that are still
-    /// non-terminal on this validator. This is deliberately unlabeled and
-    /// always exported so a release-gate assertion can distinguish a clean
-    /// zero from a missing metric.
-    pub(crate) network_key_reconfiguration_sessions_pending: IntGauge,
+    /// Number of sessions still non-terminal on this validator, grouped by
+    /// protocol name. Completed protocols are exported with zero while their
+    /// session remains tracked, distinguishing a clean zero from a missing
+    /// observation.
+    pub(crate) protocol_sessions_pending: IntGaugeVec,
 
-    /// Per-authority network-key reconfiguration outputs observed through
-    /// consensus. The digest is over canonical BCS output bytes and excludes
-    /// the sender/malicious envelope. A current validator therefore exposes
-    /// the literal previous-release validators' submitted results without
-    /// requiring instrumentation in the historical binary.
-    pub(crate) network_key_reconfiguration_output_info: IntGaugeVec,
+    /// Per-authority session outputs observed through consensus. The digest is
+    /// over canonical BCS output bytes and excludes the sender/malicious
+    /// envelope. Labels identify the protocol and concrete session, allowing
+    /// targeted compatibility tests without protocol-specific instrumentation.
+    pub(crate) session_output_info: IntGaugeVec,
 
-    /// Malicious-actor count carried in each authority's network-key
-    /// reconfiguration output report.
-    pub(crate) network_key_reconfiguration_reported_malicious_actors: IntGaugeVec,
+    /// Malicious-actor count carried in each authority's session output report.
+    pub(crate) session_reported_malicious_actors: IntGaugeVec,
 
-    /// Whether each authority's network-key reconfiguration output report was
-    /// rejected (0 = canonical output, 1 = rejected output).
-    pub(crate) network_key_reconfiguration_output_rejected: IntGaugeVec,
+    /// Whether each authority's session output report was rejected
+    /// (0 = canonical output, 1 = rejected output).
+    pub(crate) session_output_rejected: IntGaugeVec,
 
     /// Histogram-by-bucket of how long each currently-Active session has been
     /// tracked for. Labels: `session_type` (`user` / `system` /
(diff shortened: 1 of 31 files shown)


crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
```diff
@@ -154,6 +154,27 @@ pub struct DWalletMPCMetrics {
...
+    /// Per-authority network-key reconfiguration outputs observed through
+    /// consensus. The digest is over canonical BCS output bytes and excludes
+    /// the sender/malicious envelope. A current validator therefore exposes
+    /// the literal previous-release validators' submitted results without
+    /// requiring instrumentation in the historical binary.
+    pub(crate) network_key_reconfiguration_output_info: IntGaugeVec,
+
+    /// Malicious-actor count carried in each authority's network-key
+    /// reconfiguration output report.
+    pub(crate) network_key_reconfiguration_reported_malicious_actors: IntGaugeVec,
```
ycscaly: Are we sure the metrics are the right way to test these values? And this is only for reconfiguration, but perhaps we wish to test all the protocols in the same manner? What if the malicious activity is in a sign session for example? Incompatabilities can be in different parts of the code.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
```diff
@@ -3483,7 +3484,39 @@ impl DWalletMPCManager {
             {
                 sessions_with_self_output_no_quorum += 1;
             }
+            if session.is_network_key_reconfiguration {
```
ycscaly: Again, this should be a generic code not tailored for reconfiguration only.

ycscaly (changes requested):

crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
omersadika: Addressed in b4c294d993. Per-authority output collection is now protocol-agnostic: every MPC session retains canonical digests, malicious counts, and rejected state, and the exported session metrics include protocol_name plus session_id. This supports the same convergence checks for Sign, Presign, DKG, and verification protocols without adding another protocol-specific node path. Detailed series retain the existing per-epoch manager lifecycle/reset. The v1.1.8 rollout scenario intentionally filters the shared observations to Network Encryption Key Reconfiguration because that is the release boundary specified by this test; broader protocol scenarios can reuse the generic evidence independently. I also added coverage proving unrelated Sign samples are separated from the network-key assertion and that a missing requested protocol fails closed.

crates/ika-core/src/dwallet_mpc/mpc_manager.rs
omersadika: Addressed in b4c294d993. The reconfiguration-only flag, observation map, and metric fields are removed. DWalletSession now stores a generic protocol_name and output_observations map, and the manager exports ika_dwallet_mpc_protocol_sessions_pending{protocol_name=...} plus generic session output/envelope metrics for every identified protocol. The network-key DSL assertion is now only a consumer that selects the reconfiguration protocol from that shared instrumentation.

crates/ika-upgrade-test/src/cluster.rs
```diff
@@ -493,6 +511,308 @@ fn parse_labelless_gauge(body: &str, metric: &str) -> Option<u64> {
...
+        // Reconfiguration finalizes at a Byzantine quorum and does not wait: a
+        // validator still computing when quorum forms marks its session
+        // complete (on seeing the quorum in consensus) and discards its own
+        // still-in-flight output, so an honest straggler legitimately never
+        // submits. Requiring the whole committee here would race that
+        // finalization. Require instead every validator under test (the
+        // upgraded observers) plus a finalizing quorum — enough to prove the
+        // upgraded node's output agrees with a quorum of peers.
+        if !required_authorities.is_subset(&observed_authorities)
+            || observed_authorities.len() < quorum
```
omersadika, line 791: [P1] Restore the 4-of-4 output-convergence invariant. This condition now accepts the upgraded validator plus any finalizing quorum (3-of-4), so the test can pass when the fourth honest validator is absent or produces different output bytes after the quorum has finalized. That is the same class of failure this release gate is meant to expose, and it contradicts both the stated requirement that every expected authority report an equivalent canonical result and the PR description’s 4-of-4 claim. If normal finalization discards a straggler’s later submission, capture each validator’s local canonical result before/disjoint from that discard (or add another deterministic harness witness); do not redefine missing output as acceptable quorum convergence.

ycscaly: The 4-of-4 relaxation isn't the harness being lenient — it's a protocol constraint. Reconfiguration finalizes at a Byzantine quorum and does not wait: a validator still computing when quorum forms marks the session complete on seeing the quorum in consensus and **discards its own in-flight output**. It never submits, so there is no "later submission" being dropped by the harness and no local canonical result to capture — the bytes are never produced. Capturing one would mean changing production to finish and record a computation the protocol has already abandoned, so the gate would be validating a code path real rollouts never run.

What the check still enforces, hard:

1. **The upgraded validator can never be the missing one.** `required_authorities` (every validator under test) must be a subset of the observed set — that subset check is exactly the failure class this gate exists for, and it fails closed.
2. **A divergent fourth output fails, it doesn't pass.** Any authority that *does* submit must match the single canonical digest (`digests.len() == 1` is a hard error, not a retry). The relaxation only tolerates an authority that never submits because finalization discarded its output — a straggler that shows up late with different bytes fails the run. The unit test with three matching + one divergent digest pins this.
3. Zero malicious reports and zero rejected envelopes for every observed authority, and out-of-committee authorities are a hard failure.

So the accepted weakening is narrow: an old-binary straggler whose output the protocol itself threw away. The cross-version claim — the upgraded node's output is byte-identical to a finalizing quorum of v1.1.8 peers — is still proven on every reshare. Forcing a literal 4-of-4 witness would require a harness barrier that delays finalization until all four submit, which changes the timing under test and makes the gate weaker as evidence about real rollout behavior.

You're right that the PR description contradicts the code — it still says "4-of-4". I've updated it to state the actual guarantee (all validators under test + finalizing quorum, with hard digest equality over every observed authority).

omersadika: The protocol explanation justifies why a literal 4-of-4 output requirement may be impossible, but it does not establish the stronger claim in point 2. The assertion returns immediately when the first scrape contains validator 0 plus a quorum (`if incomplete.is_empty() { return Ok(()) }`). A fourth output can already be submitted or propagating without being exposed in that scrape, then become visible only after the test has returned. The divergence unit test covers only four outputs already present before evaluation, so it does not exercise this timing case.

The current code proves upgraded-validator-plus-quorum equality at one observation; it does not prove that every output still capable of arriving agrees. Please add a deterministic witness that the observation set is closed before returning success (for example, a consensus/session-finality watermark, or a bounded stabilization/drain with repeated observations if no stronger primitive exists), or demonstrate from production state that an output not visible at this point can no longer be published. Until then, the P1 remains.

ycscaly: Accepted — the timing gap was real. Fixed in 3b01a38c21 with the bounded-stabilization option, after checking whether a stronger primitive exists:

- Production keeps recording late outputs after quorum: `complete_mpc_session` → `clear_data()` preserves `output_observations` and `protocol_name`, and `handle_output` still routes a late consensus output into the completed session (or recreates one), so it is exported on the next metrics refresh. The observation set is therefore open at first convergence, exactly as you said — but every output still capable of arriving *will* become visible.
- There is no production watermark that proves "no output message for this session remains uncommitted in consensus", so a closure witness has to be observational.

`expect_network_key_output_converged` now holds the first fully-converged canonical result through a 10s stabilization window (`NETWORK_KEY_OUTPUT_STABILIZATION`), re-scraping every second and re-running the full validation each pass. A late output that agrees grows the observed set without changing the canonical digests; a late divergent output trips the hard `digests.len() == 1` error on the next scrape; a changed canonical map (e.g. a session appearing mid-drain) fails closed; a fresh incomplete observation resets the window. Consensus commit latency in these clusters is well under a second, so the window is orders of magnitude past the in-flight bound it needs to cover.

ycscaly (approved):

PR #1842, opened 2026-07-15
feat(mpc): add anomaly-focused diagnostics

omersadika:
## Summary

- add a bounded, metadata-only per-session MPC diagnostic trace that is flushed only for abnormal sessions
- keep weighted-majority disagreeing voters separate from malicious authorities embedded in the winning MPC report, while preserving the existing final union and protocol behavior
- diagnose local computation/submission failures, rejected outputs, invalid/conflicting reports, quorum without local output, pending computations at completion, malicious attribution, and self-malicious service exit
- emit a versioned one-line structured snapshot on the dedicated `ika_mpc_diagnostics` target plus low-cardinality anomaly/trigger/drop metrics
- persist anomalies for any malicious authority and expose `local_authority_malicious` as an explicit boolean for self-attribution
- document local collection through journald or a file and optional Grafana Alloy forwarding to a validator-operator-owned Loki; no anomaly trace uses `ika-proxy`

## Behavior

This is observability-only for every representable MPC output: it does not change MPC voting, quorum thresholds, the winning candidate, malicious-authority union, consensus submission decisions, session finalization, or cached quorum output behavior.

The diagnostic digest helpers use fallible BCS serialization. If serializing an in-memory public output or report candidate fails, that report is classified as invalid and omitted instead of being inserted without diagnostics as before. The current output types are fully BCS-serializable, so this is an unreachable defensive edge for valid protocol values, but it is stated explicitly rather than hidden under the observability-only claim.

Normal sessions do not emit a new INFO/WARN/ERROR line. Their 64-event metadata trace is cleared on completion. Each anomaly kind is emitted once per session; malformed outputs without session state use a capped per-epoch deduplication set. Suppressed snapshots increment a dedicated drop metric, and diagnostic JSON serialization has a non-panicking fixed fallback.

## Privacy and local collection

The persisted schema is allow-listed. It excludes session preimages, raw protocol payloads, private inputs/outputs, secret shares, private MPC state, keys, tokens, and digests of private values. Arbitrary error strings are reduced to fixed error codes; only party IDs explicitly exposed by the MPC library are retained. Crypto-library call-stack backtraces are captured separately, bounded to 16 KiB, and never format the potentially private error value. Output digests cover public final output/report candidates.

Operators enable `RUST_LOG_JSON=1` and `RUST_BACKTRACE=1`, then retain the normal systemd journal (or configure `RUST_LOG_FILE`). Disabled or unsupported captures are stored as absent rather than as a misleading placeholder. The developer playbook includes a complete `journalctl` verification flow and an Alloy configuration that forwards only the dedicated anomaly target to same-host or operator-owned Loki. Same-host Loki requires no application-level encryption. Loki labels stay low-cardinality; session IDs remain structured metadata/body fields.

Mimir receives only these counters:

- `ika_dwallet_mpc_anomaly_snapshots_total{anomaly_kind,session_type,severity}`
- `ika_dwallet_mpc_anomaly_triggers_total{trigger,session_type}`
- `ika_dwallet_mpc_anomaly_snapshots_dropped_total{reason}`

## Verification

- `cargo fmt --all --check`
- `cargo check --release -p ika-core`
- `cargo clippy --release -p ika-core --all-targets -- -D warnings`
- `cargo test --release -p ika-core anomaly_diagnostics --no-fail-fast` (10 passed)
- `cargo test --release -p ika-core mpc_diagnostics --no-fail-fast` (6 passed)
- `cargo test --release -p ika-core test_some_malicious_validators_flows_succeed -- --test-threads=1 --nocapture` (passed, real crypto)
- `cargo test --release -p ika-upgrade-test --test malicious_cross_binary --no-run`

The only warnings were the repository's existing `serde_yaml` semver-metadata manifest warnings.

## Upgrade workflow review

Reviewed every scenario in `.github/workflows/upgrade-test.yaml`: `smoke`, `workload`, `cross_binary`, `v118_upgrade`, `v118_churn`, `v118_mixed_rollout`, and `legacy_config`. No flow deliberately injects one of these anomalies. Rolling restart and churn can legitimately reach quorum before a restarted validator reproduces and observes its local output, so asserting that those logs contain no anomaly snapshot would be invalid. No workflow change is included.


ycscaly (approved):
Add malicious case too, even if its not self

ycscaly:
The crypto crate has backtraces enabled for errors; make sure the anomaly reports include them

omersadika:
## Review (deep pass over the full diff + base repo)

**Verdict: sound — merge-worthy after a few minor fixes.** Verified the claims that matter given recent history:

- **No #1816-style hot-path regression.** All trace recording is event-driven (per message/output/computation), never per-tick — nothing touches `refresh_observability_metrics` or adds a scan over `self.sessions`. The 64-event trace is a true fixed ring buffer (`pop_front` + saturating drop counter, `with_capacity(64)`), single-digit KB per session, cleared via `clear_data() → clear_trace()` on completion. `running_computation_count_for_session` is bounded by available cores; `has_seen_computation` is O(1) lookups.
- **Cardinality is type-safe, not just convention-safe.** `MpcAnomalyContext.trigger_conditions: Vec<&'static str>` and the metrics API signatures make a dynamic string unrepresentable at the label site. Enumerated: 12 anomaly kinds × 2 severities (severity is functionally dependent on kind → ≤60 series/node), 21 triggers × 5 session types (≤105 series/node).
- **Privacy claims hold in code and tests**: `session_identifier.into_bytes()` is the keccak-derived digest, never the preimage (test asserts the preimage hex is absent from the snapshot); digests cover only consensus-public outputs; error strings collapse to static codes (`Serialization(_)`/`Bcs(_)` contents never formatted); status strings are enum variant names only.
- **Correctness isolation verified**: `final_malicious_authorities = malicious_voter_authorities ∪ reported_malicious_authorities` is provably the same set as the old chained collect; vote inputs (last-output-per-party over ascending rounds, BTreeMap order) unchanged; quorum first-write-wins preserved via `get_or_insert`.
- Healthy sessions emit nothing at INFO+ on the target; snapshot is compact single-line JSON with a non-panicking fixed fallback. Naming complies with `dev-docs/conventions/metrics.md`.

### Worth fixing before merge (none blocking)

1. **Silent blindness at the 1024 untracked-anomaly cap.** At `MAX_UNTRACKED_ANOMALIES` the early return emits *no log and no metric*. Since malformed outputs for nonexistent sessions are attacker-controllable (per the code's own comment), a flood blinds untracked-anomaly reporting for the rest of the epoch undetectably. Suggest still incrementing `anomaly_snapshots_total` (labels are static, cardinality-safe) or adding a dedicated drop counter.
2. **Severity mismatch on the untracked path**: it always logs `error!` even for warn-severity kinds (e.g. `invalid_output_received` carries `severity:"warn"` on an ERROR-level line), so log-level alerting and the `severity` field diverge. Use `warn!`/`error!` by kind severity like the tracked path.
3. **Playbook doc drift** (`dev-docs/playbooks/mpc-anomaly-diagnostics.md`): example strings operators will copy into alerts don't match code constants — `local_computation_returned_error` → actual `local_mpc_computation_returned_error`; `local_validator_submitted_rejected_output` → actual `local_validator_submitting_rejected_output`; example `error_code:"invalid_mpc_output"` is not an emitted code. Also the "stable top-level fields" list includes `session_id`, but the no-source-session service-exit line omits it and uses a singular `trigger` field with no `diagnostic_json` snapshot — and untracked lines serialize `MpcAnomalyContext` (a smaller shape than `MpcAnomalySnapshot`) under the same `schema_version: 1`. Parsers must branch on `tracked_session`; worth documenting explicitly (or bumping/splitting the schema tag).
4. **PR body nuance**: "no behavior changes" is not literally exact — an output whose `bcs::to_bytes` fails is now **not inserted** into the voting map (`AddOutputResult::Invalid` early-returns), whereas previously it was inserted with only digest tracking skipped. In-memory bcs serialization is effectively infallible and an invalid output adds no vote weight, so practical impact is nil — but worth one sentence in the description.

### Nits

- `LocalAuthorityMaliciousReason::Unknown` is unreachable (`final` is defined as the union of the two sets, so `(false,false)` + union membership is impossible) — dead arm, and the playbook documents "unknown" as a real state.
- `add_output` now performs a second full bcs serialization per received output (`report_digest` alongside the pre-existing `output_digest`); `build_outputs_to_finalize` clones every latest output per invocation. Event-driven and O(committee), so acceptable — but avoidable.
- Test coverage gaps: no test for the 1024-cap behavior; no emission tests for `VotingFailure`, `ProtocolMessageSubmissionFailed`, `ComputationUpdateAfterSessionCompletion`. (Dedup, bounds, redaction, healthy-session silence, metrics, split-vote separation, and self-malicious exit are all covered.)

### Infra side (dwallet-labs/infra)

Already adopted: [infra#94](https://github.com/dwallet-labs/infra/pull/94) makes the `ika_mpc_diagnostics` target explicit in the validator `RUST_LOG` (pod logs already reach Loki via the cluster pipeline; counters ride the existing metrics path), and [infra#92](https://github.com/dwallet-labs/infra/pull/92) adds snapshot/trigger panels, a Loki logs panel on the target, and an error-severity anomaly alert. The infra config uses label-based legends and no hardcoded trigger names, so item 3's doc drift can't bite it.

omersadika:
Addressed the review findings in `3d79a2b546` and `0840549766`:

- crypto `mpc::Error` backtraces are now included separately in anomaly JSON, capped at 16 KiB; the full error kind/value is never formatted, and the redaction test proves an arbitrary serialization-error secret is absent
- the 1,024-entry untracked cap now increments `ika_dwallet_mpc_anomaly_snapshots_dropped_total{reason="untracked_capacity"}`; a focused test verifies bounded state, drop accounting, and deduplication
- untracked warn-class anomalies now use `warn!`, while error-class anomalies use `error!`
- all anomaly lines declare `diagnostic_shape="snapshot"` or `"context"`; source-less self-malicious exit now carries structured context JSON, and the playbook documents the parser distinction and absent session ID
- corrected the example trigger/error values
- removed the unreachable `LocalAuthorityMaliciousReason::Unknown` branch
- updated the PR body to state the defensive BCS-serialization edge explicitly

Validation: 5 diagnostic unit tests passed, all 9 existing real-crypto anomaly tests passed, the new capacity test passed, release clippy passed with `-D warnings`, formatting/diff checks passed, and metric-name validation passed.

I left the accepted double-serialization performance nit unchanged: it is event-driven, bounded by committee output reports, and changing the digest representation now would add risk unrelated to the review defect.

omersadika:
## Fix verification (`3d79a2b546` + `0840549766`)

All review findings verified fixed in code:

1. **Cap observability** ✅ — `ika_dwallet_mpc_anomaly_snapshots_dropped_total{reason}` registered with a documented fixed-value label; incremented exactly on the capacity path (`len() >= MAX_UNTRACKED_ANOMALIES → inc + return`), while the dedup contains-check still returns plainly — correct semantics (dedup ≠ drop). Capacity test present.
2. **Severity mismatch** ✅ — untracked path now splits `warn!`/`error!` by `anomaly_kind.severity()`; levels and the `severity` field agree.
3. **Schema/parser distinction** ✅ — `diagnostic_shape = "snapshot" | "context"` on every line, source-less service-exit now carries structured context JSON, playbook documents the branch and the absent `session_id`.
4. **Doc drift** ✅ — trigger/error example values match the code constants (checked `local_mpc_computation_returned_error`, `local_validator_submitting_rejected_output`; stale names gone).
5. **`Unknown` branch** ✅ removed; **PR body** ✅ states the defensive BCS edge.

The new backtrace feature also checks out: rendering happens only on the anomaly path, `bounded_backtrace_string` truncates on a char boundary with no panic risk and ≤16 KiB output, the error kind/value is never formatted, and the redaction test forces a backtrace alongside a secret-bearing `Serialization` kind.

**One follow-up worth a playbook note:** the pinned inkrypto rev (`1d8b5a9`, `mpc/src/lib.rs:60`) captures via `std::backtrace::Backtrace::capture()`, which is **env-gated** — without `RUST_BACKTRACE=1` (or `RUST_LIB_BACKTRACE=1`) every snapshot's `error_backtrace` is `Some("disabled backtrace")` rather than a real trace. Suggest (a) documenting the env requirement in the playbook's operator setup, and optionally (b) mapping the disabled placeholder to `None` in `bounded_backtrace` so the field isn't misleading when the env is unset. Frame capture is cheap (symbol resolution is lazy) and `mpc::Error` construction is rare, so recommending the env var on validators is safe.

Infra side already updated: [infra#94](https://github.com/dwallet-labs/infra/pull/94) now sets `RUST_BACKTRACE=1` on ika validators, and [infra#92](https://github.com/dwallet-labs/infra/pull/92) plots the new drop counter on the anomaly panel.

omersadika:
Addressed the follow-up in `73e0121a26`.

- documented and added `RUST_BACKTRACE=1` to the validator systemd setup example
- `bounded_backtrace` now checks `BacktraceStatus`; disabled or unsupported captures are omitted instead of serializing `"disabled backtrace"`
- added a regression test for the disabled-capture case

Validation: 6 diagnostic unit tests passed, release clippy passed with `-D warnings`, formatting and diff checks passed. Infra already setting `RUST_BACKTRACE=1` means production validators will retain real bounded traces.

omersadika:
Verified `73e0121a26`: the backtrace follow-up is fully resolved. `bounded_backtrace` gates on `BacktraceStatus::Captured` (the right API — also covers `Unsupported`, not just `Disabled`), the disabled-placeholder test locks it in, and the playbook now documents the `RUST_BACKTRACE=1` requirement with the systemd drop-in example. Matches the infra side ([infra#94](https://github.com/dwallet-labs/infra/pull/94) sets the variable on managed validators). No remaining findings from my review — LGTM.

PR #1844, opened 2026-07-16
fix(upgrade-test): make the v1.1.8 mixed-rollout gate deterministic for honest stragglers

omersadika:
## Problem

The literal v1.1.8/current mixed-rollout release gate (#1841) produced contradictory results for the **same candidate SHA** `73e0121a26c30ff37cb1cf2a1b9cf51dc35df9d4`:

- Failed: https://github.com/dwallet-labs/ika/actions/runs/29508605742
- Passed: https://github.com/dwallet-labs/ika/actions/runs/29509220785

The diagnostics added in #1842 pinpointed the failing run precisely: the three literal v1.1.8 validators submitted the same canonical network-key reconfiguration output and a quorum finalized at consensus round 5902 (`local_authority_malicious=false`, zero malicious sets, zero rejections, `quorum_reached_without_local_output=true`, `local_computation_pending=true`). Roughly 195 ms later the upgraded validator's computation returned `finalize`, but the session was already non-active, so production discarded the result and never submitted it. The gate's `expect_network_key_output_converged` then waited 60 s for the upgraded authority to appear in the submitted-output set — which could never happen.

## Why this is a gate defect, not a compatibility failure

Production finalizes a reconfiguration at a Byzantine quorum and does not wait for stragglers: a validator whose computation finishes after it processed the quorum marks the session complete and discards its own result. **Any honest validator — including the upgraded one — can be that straggler.** Requiring the upgraded validator's submitted output at every boundary therefore races a ~200 ms scheduling window, which is exactly the same-SHA pass/fail behavior observed. The existing 10 s stabilization drain only covers submitted-but-still-propagating outputs; it cannot help an output production intentionally never submits.

At the same time, a missing candidate output **cannot itself prove compatibility**: quorum-only convergence is three v1.1.8 validators agreeing with each other, which says nothing about the upgraded binary's bytes. Passing on that alone would make the gate vacuous.

## Fix

**Node side — observability only** (no consensus, output-submission, session-lifecycle, or timing change; nothing is published, finalization is not delayed, the completed session is never re-activated):

- At quorum, `handle_consensus_round_outputs` stashes the winning network-key reconfiguration output's **raw-bytes digest** on the session before `complete_mpc_session` makes it non-active (reassembled from the in-order output chunks, so it equals the digest of the unsliced `public_output_value`; gated on the session's reconfiguration flag so the walk never runs for sign/presign outputs).
- At the discard site, a late `Finalize` for a completed reconfiguration session has its raw output bytes digested and compared with the stashed quorum digest. The verdict lands in the existing `computation_update_after_session_completion` anomaly (new trigger conditions), in the session trace, and in two scrapeable gauges: `ika_dwallet_mpc_session_late_output_info{output_digest,quorum_output_digest}` and `ika_dwallet_mpc_session_late_output_malicious_actors`. Scoped to reconfiguration sessions to keep the per-session cardinality bounded.

**Harness side — explicit per-validator evidence model.** Each boundary now distinguishes, for every validator under test:

1. candidate submitted inside the converged quorum and matched → **conclusive**;
2. candidate legitimately finished after quorum, discarded output's digest **equals** the quorum digest (zero late malicious) → **conclusive**;
3. candidate's bytes diverge (submitted or late, including cross-attempt disagreement) → **hard failure**;
4. candidate malicious/rejected (submitted envelopes or late malicious report) → **hard failure**;
5. clean quorum convergence with no comparable candidate output → **inconclusive**: not a failure (the ordering is legitimate), but never accepted as compatibility proof.

Every boundary still requires healthy validators, correct local epochs, protocol v3, quorum output convergence, zero malicious actors, zero rejected envelopes, no self-malicious logs, no stranded sessions; the existing 10 s stabilization drain for late-propagating **submitted** outputs is preserved unchanged. The scenario fails with "insufficient cross-version compatibility evidence" unless **every validator under test** is conclusively evidenced at **at least one** boundary — quorum-only progress can never pass the gate vacuously. Two whole-run log backstops close the metric-lifetime gaps: a divergent late output landing after a boundary's observation window still fails the run (its error line outlives the per-epoch metric reset), and an output consensus-submission failure on the candidate cannot hide behind the peers' quorum completing the session.

No historical binary is replaced, no mocks are introduced, and the rollout topology (1 current + 3 literal v1.1.8) is unchanged.

**Known residual (accepted, documented):** late-digest evidence proves the candidate *computes* identical bytes; it does not exercise the candidate's output-submission path when the candidate loses the race at every boundary (partially covered by the submission-failure log backstop), and a candidate whose computation hangs or errors *after* quorum at exactly one boundary is forgiven if the other boundary is conclusive (a deterministic cross-version computation failure hits both boundaries and fails the evidence gate; an error while the session is still Active submits a rejection and fails hard). Watch signal for operators: `ika_dwallet_mpc_anomaly_snapshots_total{anomaly_kind="local_computation_failed"}` on the upgraded validator.

## Regression tests

- `crates/ika-core/.../integration_tests/late_reconfiguration_output.rs` reproduces the exact ordering (three authorities finalize identical outputs → quorum closes the session → local `finalize` arrives afterwards) with envelopes built through the **real production chunker**, and asserts: nothing is submitted to consensus, the digest pair is recorded and exported, matching bytes are flagged as a match. A fault-injected variant returns **different** bytes and asserts the divergence trigger, trace event, and unequal exported digest pair; a third test proves sessions without the reconfiguration request record nothing.
- `mpc_diagnostics` unit tests pin the digest-domain invariant (in-order chunk reassembly ≡ unsliced raw bytes; rejected/foreign envelopes produce no digest).
- `cluster.rs` unit tests cover the full classification: straggler-without-evidence converges but is flagged; matching late evidence is conclusive; divergent late digest, late malicious report, cross-attempt digest disagreement, divergent submitted output, rejected envelope, and out-of-committee authority all fail hard; superseded `unknown`-quorum gauge children next to the compared pair are tolerated; racing scrapes stay retry-not-fail.
- `scenario.rs` unit tests pin the per-validator aggregate gate (every validator under test must be evidenced at some boundary, in either order; all-unevidenced fails; a validator only ever unevidenced fails even when another is conclusive).

## Validation

- Local: `cargo test --release -p ika-core --lib late_reconfiguration` → 3 passed; `mpc_diagnostics` → 8 passed; `cargo test --release -p ika-upgrade-test --lib` → 26 passed; `cargo fmt --all`; `cargo clippy --release -p ika-core -p ika-upgrade-test --all-targets` clean. Reviewed by the consensus-determinism checklist: no consensus message, submission, session-lifecycle, epoch-close, or malicious-accounting change; the quorum-stash input is the consensus majority vote and nothing derived from the new state feeds back into consensus.
- Literal v1.1.8/current mixed rollout on `ika-k8s-large` at the final SHA, three runs (same-SHA repeatability):
  - https://github.com/dwallet-labs/ika/actions/runs/29526059600
  - https://github.com/dwallet-labs/ika/actions/runs/29526061077
  - https://github.com/dwallet-labs/ika/actions/runs/29526063190
- Fault-injection run at the final SHA (`test_testing_fault=true`, deliberately faulty current validator — the gate must fail closed): https://github.com/dwallet-labs/ika/actions/runs/29526064935
- An earlier fault-injection run at the pre-hardening commit already failed closed as required, at the convergence assertion itself (incomplete quorum observations under the corrupted-message fault, flow 17/46): https://github.com/dwallet-labs/ika/actions/runs/29523299380

Results will be posted here when the runs conclude.

Related: #1841 (the gate), #1842 (the diagnostics that exposed the failing run's exact state).

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
@@ -173,6 +173,19 @@ pub struct DWalletMPCMetrics {
     /// (0 = canonical output, 1 = rejected output).
     pub(crate) session_output_rejected: IntGaugeVec,
 
+    /// A locally computed network-key reconfiguration output that returned
+    /// after the session completed via the peers' quorum, so production
+    /// discarded it without submission. The `output_digest` and
+    /// `quorum_output_digest` labels are both over RAW output bytes (before
+    /// any consensus envelope) — comparable with each other but NOT with
+    /// `session_output_info`'s envelope digests. Label equality proves the
+    /// discarded local bytes match the quorum-agreed output.
+    pub(crate) session_late_output_info: IntGaugeVec,
+
+    /// Malicious-actor count reported by the discarded late local computation
+    /// recorded in `session_late_output_info`.
+    pub(crate) session_late_output_malicious_actors: IntGaugeVec,
+
     /// Privacy-safe anomaly snapshots emitted to the validator's local log sink.
     /// Labels are fixed enums: `anomaly_kind`, `session_type` (or `unknown` when
     /// service termination has no source session), and `severity`.
@@ -505,6 +518,26 @@ impl DWalletMPCMetrics {
                 registry,
             )
             .unwrap(),
+            session_late_output_info: register_int_gauge_vec_with_registry!(
+                "ika_dwallet_mpc_session_late_output_info",
+                "Raw-bytes digest of a locally computed session output discarded after the quorum completed the session, next to the quorum output's raw-bytes digest",
+                &[
+                    "protocol_name",
+                    "session_id",
+                    "authority",
+                    "output_digest",
+                    "quorum_output_digest"
+                ],
+                registry,
+            )
+            .unwrap(),
+            session_late_output_malicious_actors: register_int_gauge_vec_with_registry!(
+                "ika_dwallet_mpc_session_late_output_malicious_actors",
+                "Malicious-actor count reported by the discarded late local computation",
+                &["protocol_name", "session_id", "authority"],
+                registry,
+            )
+            .unwrap(),
             anomaly_snapshots_total: register_int_counter_vec_with_registry!(
                 "ika_dwallet_mpc_anomaly_snapshots_total",
                 "Privacy-safe MPC anomaly snapshots emitted to the local log sink",
@@ -692,6 +725,8 @@ impl DWalletMPCMetrics {
         self.session_output_info.reset();
         self.session_reported_malicious_actors.reset();
         self.session_output_rejected.reset();
+        self.session_late_output_info.reset();
+        self.session_late_output_malicious_actors.reset();
     }
 }
 
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
@@ -18,6 +18,7 @@ use crate::dwallet_mpc::crytographic_computation::ComputationId;
 use crate::dwallet_mpc::dwallet_mpc_metrics::DWalletMPCMetrics;
 use crate::dwallet_mpc::mpc_diagnostics::{
     MpcAnomalyContext, MpcAnomalyKind, dwallet_mpc_error_diagnostic, output_digest,
+    raw_output_digest,
 };
 use crate::dwallet_mpc::mpc_manager::DWalletMPCManager;
 use crate::dwallet_mpc::mpc_session::{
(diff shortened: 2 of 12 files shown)


omersadika:
## Validation results — all concluded

**Four literal v1.1.8/current mixed-rollout runs at the final SHA (`ef284faab4`), all green** with `test result: ok. 1 passed; 0 failed; 0 ignored` and both reconfiguration boundaries logging `byte-level candidate evidence witnessed at this boundary`:

| Run | Outcome |
|---|---|
| https://github.com/dwallet-labs/ika/actions/runs/29526059600 | success |
| https://github.com/dwallet-labs/ika/actions/runs/29526061077 | success |
| https://github.com/dwallet-labs/ika/actions/runs/29526063190 | success |
| https://github.com/dwallet-labs/ika/actions/runs/29525976928 (auto PR-triggered, same SHA) | success |

In all four runs the upgraded validator happened to win the finalize race at all eight boundaries (submitted-output evidence; the late-evidence log line fired zero times). The straggler ordering is a scheduling race that cannot be forced in CI, which is exactly why the discarded-output path is pinned deterministically instead: the `late_reconfiguration_output.rs` integration tests reproduce the precise failing-run ordering (three peers finalize → quorum closes the session → local `finalize` arrives after) against real session/manager/service code and prove it records matching evidence without submitting, and the `cluster.rs` unit tests prove the harness accepts that evidence, stays inconclusive without it, and fails hard on divergence.

**Fault-injection runs (deliberately corrupted current validator; the gate must fail closed) — both failed as required, at the convergence assertion itself, not on an unrelated error:**

- Final SHA: https://github.com/dwallet-labs/ika/actions/runs/29526064935 — panicked at the scenario's convergence expect: `network-key output observations did not become complete within 60s: validator 0 ... is missing network-key output observations`
- Pre-hardening commit: https://github.com/dwallet-labs/ika/actions/runs/29523299380 — same assertion, at flow 17/46

The evidence model therefore does not weaken the gate: under the injected fault the reshare never produces clean converged observations and the boundary hard-fails on the incomplete-quorum path, exactly as before.

For contrast, the pre-fix contradiction at SHA `73e0121a26`: failed https://github.com/dwallet-labs/ika/actions/runs/29508605742 / passed https://github.com/dwallet-labs/ika/actions/runs/29509220785.

The `-val2`/`-val3`/`-fault` twin branches existed only to parallelize the runs across workflow concurrency groups and are deleted.

PR #1846, opened 2026-07-16
test(cluster): assert eventual joiner capture instead of first-freeze timing (#1772)

### LINKED ISSUE #1772: Flaky: test_joiner_lands_in_next_committee_class_groups (class-groups freeze under joiner-restart load)
## Summary
`ika-test-cluster::joiner::test_joiner_lands_in_next_committee_class_groups` is intermittently failing on the **Test Cluster** suite at a high rate (**2 of the last 3** runs on `feat/ocs-grpc-migration`). It is **non-deterministic / pre-existing**, not tied to any specific change — see evidence below.

## Failure
```
thread 'test_joiner_lands_in_next_committee_class_groups' panicked at crates/ika-test-cluster/tests/joiner.rs:154:5:
joiner k#… must appear in epoch-2 committee class_groups_public_keys_and_proofs (freeze must capture the mid-epoch joiner)
```
The mid-epoch joiner's `class_groups_public_keys_and_proofs` are not captured into the epoch-2 committee by the reconfiguration freeze.

## Evidence it is a flake (not a code regression)
- **Same commit, both outcomes.** On commit `de67e9eb92`: run `28236056630` **failed** this test; re-run `28242195424` (same SHA) **passed** all 22. Same code → opposite results = non-deterministic.
- Also observed failing again on a later commit (`28261559767`), while *other* runs of the same suite pass it. Different tests fail per run (port-collision, protocol-version, this one) — the flaky-suite signature.

## Likely root cause
In the failing runs the validator's **Sui fullnode becomes transiently unreachable** mid-test, e.g.:
```
ERROR ika_core::sui_connector::bag_event_pump: tcp connect error 127.0.0.1:9280 … Connection refused
```
A node that can't read/write Sui during the epoch-2 freeze window can't contribute its part of the class-groups capture → joiner omitted. This points at **localnet fullnode stability under the heavy joiner-restart load** these tests generate (and/or the known #1736 epoch-entry `mpc_data` race), rather than the assertion logic.

## Suggested next steps
- Make the test resilient to a transient fullnode blip during reconfiguration (retry/extend the freeze-capture wait), or stabilize the localnet fullnode under joiner-restart churn.
- Consider longer epochs for this test (the relay-topology / reconfiguration tests are sensitive to short epochs under CI load).
- Track alongside the #1736 epoch-entry `mpc_data` race investigation.

_Filed from CI triage on the #1744 (`feat/ocs-grpc-migration`) pre-merge review; the three fast-follow commits in that branch were ruled out as the cause (the flake predates them on the parent commit, and the class-groups freeze path does not touch the bag-event pump)._

omersadika:
## Problem

`test_joiner_lands_in_next_committee_class_groups` flakes (issue #1772): the mid-epoch joiner reaches epoch 2 but is missing from the epoch-2 committee's `class_groups_public_keys_and_proofs`.

## What the investigation actually found

**The test does not measure the off-chain mpc_data freeze — and never did.** Fault injection (disabling the joiner P2P fan-out entirely, so the freeze deterministically excluded the joiner: `frozen=4 … excluded_set=[]`, `assembled committee mpc_data off-chain epoch=2 members=4`) left the assertion **passing**: run [29527238305](https://github.com/dwallet-labs/ika/actions/runs/29527238305), debug-logged repeat [29528230442](https://github.com/dwallet-labs/ika/actions/runs/29528230442).

The map the test reads — `epoch_store.committee().class_groups_public_keys_and_proofs` — is built by `EpochStartSystem::get_ika_committee` from **chain state**: each member's on-chain `mpc_data` record, with a silent `filter_map` skip when the record is missing (`epoch_start_system.rs:189`). The joiner's class-groups key is on chain from candidate registration, so the off-chain pipeline is invisible here. The off-chain frozen set feeds a *different* committee object (the sui_syncer-assembled one consumed by the reconfiguration MPC).

So the #1772 flake was a **transient chain-read gap**: the joiner's on-chain record not visible in a node's epoch-2 `EpochStartSystem` read (per-run tx-ordering under load, aggravated by the observed fullnode outages). The issue's freeze-capture framing — and this test's own doc comment — described a mechanism the assertion never touched. Notably the chain-view map is still live state: `class_groups_keys_by_party_id` seeds the MPC manager's validator keys from it.

## Fix

Assert eventual consistency of the chain view with a bounded grace, instead of the lucky-timing boundary read:

1. **Continuing members must never be missing** from the epoch-2 map — their records predate the epoch, a gap is a real chain-read defect: fail immediately (the grace below cannot mask it).
2. **The joiner is present at epoch 2** (normal path), **or** its record was transiently invisible at the boundary and **must be present in the epoch-3 committee** — the record is durable, so the next boundary read has no excuse. Still absent at epoch 3 = registration/lifecycle bug → red. The grace engaging is logged at `warn` with a `joiner_in_voting_rights` disambiguator (record-read gap vs missed mid-epoch selection).

Also: corrected the stale fail-fast comment (an off-chain-excluded joiner keeps its chain seat and advances epochs), and extended the `dev-docs/learnings/pitfalls.md` entry with both lessons (no window width beats environmental stalls; verify *which object* an assertion reads before reasoning about its timing).

**Follow-up (not in this PR):** genuine off-chain freeze-capture coverage needs a handoff-cert-items assertion (the cert is the durable consensus-anchored record of the frozen set); and `get_ika_committee`'s silent member skip deserves a fail-loud look given spec invariant 3.

## Validation

- Happy path: full Test Cluster suite green 22/22, no flaky — run [29526037707](https://github.com/dwallet-labs/ika/actions/runs/29526037707).
- Grace-path plumbing: fault branch forces the grace path under injected off-chain exclusion — run [29528954534](https://github.com/dwallet-labs/ika/actions/runs/29528954534).
- The fault branch (`claude/1772-fault-injection`) is validation-only and will be deleted.

Closes #1772

### DIFF
--- a/crates/ika-test-cluster/tests/joiner.rs
+++ b/crates/ika-test-cluster/tests/joiner.rs
@@ -81,33 +81,38 @@ async fn test_joiner_added_at_epoch_2() {
 /// scales every cadence on this path to ~1% of the epoch (a no-op at
 /// production epoch lengths), so the path fits a bounded test epoch.
 ///
-/// Assertion semantics (issue #1772): capture by the FIRST freeze after
-/// the joiner registers — the one that fires during epoch 1 and fixes
-/// the epoch-2 committee — is timing-dependent BY DESIGN. The spec rule
-/// is "a joiner that misses the freeze window is excluded, not waited
-/// for" (`dev-docs/specs/validator-mpc-data-announcements.md`), and an
-/// environmental stall inside the window (CI contention, a transient
-/// fullnode outage) legitimately excludes the joiner — observed even at
-/// 240s test epochs (60s window). No window width makes a strict
-/// epoch-2 assertion sound, so this test asserts the protocol's actual
-/// guarantees instead:
+/// WHAT THIS TEST ACTUALLY MEASURES (established by fault injection
+/// while fixing the issue-#1772 flake): `epoch_store.committee()` is
+/// built by `EpochStartSystem::get_ika_committee` from CHAIN state —
+/// its `class_groups_public_keys_and_proofs` decodes each member's
+/// on-chain `mpc_data` record and silently `filter_map`-skips a member
+/// whose record is missing. It is NOT the off-chain assembled
+/// committee: disabling the joiner P2P fan-out entirely (so the
+/// mpc_data freeze deterministically excluded the joiner) still left
+/// the joiner present in this map, because its class-groups key went
+/// on-chain at candidate registration. So this assertion covers the
+/// CHAIN view — the map `class_groups_keys_by_party_id` seeds the MPC
+/// manager's validator keys from at epoch start — while capture by the
+/// off-chain freeze (which feeds the reconfiguration MPC and the
+/// handoff cert) is only indirectly exercised here and deserves its own
+/// handoff-cert-based assertion (follow-up).
+///
+/// Assertion semantics (issue #1772): whether the joiner's on-chain
+/// mpc_data record is visible in the epoch-2 `EpochStartSystem` read is
+/// a per-run chain-timing race under load (the #1772 flake reproduced
+/// even at 240s epochs, correlated with transient fullnode outages), so
+/// a strict epoch-2 assertion is unsound at any epoch length. Assert
+/// eventual consistency with a bounded grace instead:
 ///   1. Continuing members are NEVER missing from the committee's
-///      class-groups map (they are carry-forward-protected at the
-///      freeze) — a gap there fails immediately.
-///   2. The joiner is captured into the epoch-2 committee (the normal
-///      path), OR it was excluded there and MUST self-heal into the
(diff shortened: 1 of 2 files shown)


omersadika:
All validation legs green:
- Full Test Cluster suite on the first commit: 22/22, no flaky ([29526037707](https://github.com/dwallet-labs/ika/actions/runs/29526037707)).
- Filtered run on the final head `e3f09f08af`: PASS ([29529050326](https://github.com/dwallet-labs/ika/actions/runs/29529050326)).
- Grace-path plumbing under injected off-chain exclusion (fan-out disabled, fallback forced): PASS at 111.8s, traversing the epoch-3 wait + assertion ([29528954534](https://github.com/dwallet-labs/ika/actions/runs/29528954534)).

The validation-only branch `claude/1772-fault-injection` has been deleted.

Issue #1852, opened 2026-07-18
v1.2.1: mid-epoch restart wedges the MPC subsystem - sessions park on the network key and re-instantiation is never attempted


omersadika:
## Symptom

On testnet (epoch 355, fleet homogeneous on `1.2.1-d0c9abed7d7d`), **4 validators that restarted mid-epoch** — Encapsulate, Full Sail Validator, Mysten Labs, `bitszn.com | valopers` — have completed **zero MPC advances for 6–8.7h** while looking healthy at every other layer. A 5th (n1stake, uptime 21.9h, no recent restart) shows the same zero-advance symptom without the parked-session signature — possibly a distinct variant.

## Evidence (Mimir, all sampled 2026-07-18)

Identical signature on all 4 restarted validators:

| Signal | Value |
|---|---|
| `ika_dwallet_mpc_requests_pending_for_network_key` | **263 on each** (only these 4 network-wide; every other validator ~0) |
| `ika_dwallet_mpc_network_key_instantiations_in_flight` | 0 |
| `increase(ika_dwallet_mpc_network_key_instantiation_failures_total[6h])` | 0 — **not failing, never attempted** |
| `increase(ika_dwallet_mpc_data_blob_fetch_total[6h])` | no series — no fetch attempts at all |
| `ika_own_mpc_data_blob_unhealthy` / `ika_off_chain_assembly_wedged` / `ika_handoff_prepare_waiting` | all 0 |
| `ika_consensus_subscribed_to` | 43 (fully meshed) |
| consensus rounds | on par with healthy peers |
| `ika_current_epoch` | 355 (correct) |
| `ika_dwallet_mpc_network_key_loaded_epoch` | series **stopped exactly at process restart** (e.g. Mysten Labs: cut off ~8.7h ago, matching uptime ~8.67h) and never reappeared |
| `sum(increase(ika_dwallet_mpc_advance_completions[6h]))` | 0 across every protocol/curve/algorithm |

The epoch's network-key reconfiguration itself completed cleanly ~4h before sampling (pending-sessions gauge went 1→0 once, checkpoint aggregator at 1 distinct digest, chain reconfig-completed counter equal to key count).

## Hypothesis

[...]

PR #1854, opened 2026-07-18
fix(mpc): reclassify benign quorum races out of the anomaly taxonomy

### LINKED ISSUE #1853: #1842 anomaly diagnostics: quorum_anomaly and computation_update_after_session_completion flood on normal quorum races (~9 snapshots/sec network-wide)
## Symptom

On testnet (44 validators, all on `1.2.1-d0c9abed7d7d`), the #1842 anomaly counters show, over an 8h window:

- `ika_dwallet_mpc_anomaly_snapshots_total{anomaly_kind="quorum_anomaly", severity="warn"}`: **~265,000**
- `{anomaly_kind="computation_update_after_session_completion", severity="warn"}`: **~51,500**

That's **~9 snapshots/sec network-wide**, each also a JSON log line on the `ika_mpc_diagnostics` target (Loki ingest on our side, journald elsewhere).

## Why this looks like taxonomy noise, not real anomalies

The emissions are **uniform across healthy validators** (~20–23k each over 8h: Staking Facilities, Node.Monster, DTEAM, alphab.ai, Rhei, TrustedPoint, ...), not concentrated on the 5 currently-wedged validators (separate issue). During periods of high internal-presign churn, essentially **every validator emits `quorum_anomaly` for essentially every session**:

- `quorum_reached_before_local_output_observed` is the *expected* state for any validator not in the fastest ~2/3 of a session — quorum forming before a given validator's own output loops back is how threshold cryptography works, not an anomaly. The observed rate suggests it fires even more broadly than the natural 1/3.
- `computation_update_after_session_completion` similarly fires when a session completes at quorum while the local computation is still finishing — the same benign race from the other side.

The per-session dedup from #1842 works as designed (one emission per session per kind); the problem is that these two kinds classify normal races as anomalies, so "once per session" times high session churn equals a firehose. Anomaly counters/dashboards read as permanently alarming, which buries the kinds that carry real signal (`local_computation_failed`, `invalid_output_received`, `conflicting_output_digests`, ...).

## Suggestions (any of)

1. Gate `quorum_anomaly` on an actual defect signal — e.g. emit only when quorum was reached AND the local output was never submitted/observed within N rounds after, or when the validator also has zero participation in the session.
2. Reclassify these two kinds out of the anomaly taxonomy into plain (label-bounded) counters, keeping snapshots only for the defect-correlated cases.
3. Debounce per (validator, epoch) rather than per session for these two kinds.

## Context

Found while validating the anomaly pipeline end-to-end on infra dashboards; counter data above is from Mimir, 2026-07-18. Relates to the severity-taxonomy nuance raised in the #1842 review (severity is per-kind; both flooding kinds are `warn`, so the error-severity alerting stays quiet — correctly — while panels and Loki absorb the noise).

### LINKED ISSUE #1857: Add session-origin provenance and a sessions_reconstructed_total counter (follow-up to #1854)
## Summary

Follow-up to #1853 / PR #1854: sessions reconstructed from consensus output after a restart are indistinguishable — in metrics and in diagnostics — from sessions the validator participated in locally. Add explicit lifecycle provenance and a recovery counter.

## Motivation

`DWalletMPCManager::handle_output` correctly supports receiving an output before the corresponding session request (a `WaitingForSessionRequest` entry buffers reports until activation — mpc_manager.rs:1737, 2105-2125). After a restart this is the dominant path: the node reconstructs thousands of pre-restart sessions from consensus outputs it can never have local output for. Two consequences:

1. **Diagnostic noise:** every reconstructed session that reaches quorum trips `quorum_reached_before_local_output_observed`, producing tens of thousands of warn snapshots that mean "restarted recently", not "anomaly". (Post-#1854 these are gated/demoted, which fixes the volume — but also erases the distinction entirely.)
2. **Lost signal:** a *bounded* reconstruction burst after restart is healthy; an *unbounded* stream of no-local-output sessions is a validator that isn't participating at all. On testnet right now, the two wedge variants in #1852 / #1856 produce exactly that second signature ([NODERS] and bitszn: ~10k/8k such sessions in 2h with zero local advances, hours after restart). Today the only way to tell the difference is cross-referencing advance counters by hand.

## Proposal

- Tag each session with its origin at creation, e.g. `session_origin = local_request | reconstructed_from_consensus` (a `WaitingForSessionRequest`-born session that later receives its request and activates locally can upgrade to `local_request`).
- Export `ika_dwallet_mpc_sessions_reconstructed_total{session_type}` (a real `IntCounterVec` — see note below).
- Diagnostics: never count missing local output as anomalous for a still-reconstructed session; include `session_origin` in the snapshot summary fields so forensic triage can split the populations.
- Keep all correctness checks (digest agreement, rejection, malicious attribution) identical for both origins.

With this in place, the operational signals become clean:
- `sessions_reconstructed_total` spiking at restart and then flattening = healthy recovery.
- `sessions_reconstructed_total` still climbing N minutes after restart while `advance_completions` stays flat = the non-participation wedge (#1852 variants / #1856) — alertable directly, without depending on the anomaly flood #1854 removes.

## Note on existing metric types

`session_start_count` is an `IntGaugeVec` used as a monotonic counter (dwallet_mpc_metrics.rs:77). Fine in practice with `increase()`, but any new counters here should be genuine `IntCounterVec`s, and converting the existing one is worth doing opportunistically.

_Origin: this consolidates the observability-side recommendations from a local diagnostics investigation (quorum-tail/restart-reconstruction analysis) with live evidence from the #1852/#1856 wedge tracking. Filed from the monitoring side._

omersadika:
Fixes #1853.

## Problem

The #1842 anomaly pipeline classified two normal threshold-crypto races as anomalies, and per-session dedup times high internal-presign churn produced ~9 snapshots/sec network-wide on testnet (~265k `quorum_anomaly` + ~51.5k `computation_update_after_session_completion` warn snapshots over 8h, uniform across healthy validators). Every snapshot is also a JSON log line on `ika_mpc_diagnostics`, so panels and Loki read permanently alarming and the defect-correlated kinds are buried.

## Fix (issue suggestions 1 + 2 combined)

**New plain counter** `ika_dwallet_mpc_completion_races_total{race,session_type}` for the three benign races. The `race` label values reuse the former trigger strings verbatim (`quorum_reached_before_local_output_observed`, `local_computation_pending_at_session_completion`, `local_computation_update_received_after_session_became_non_active`) so dashboard queries migrate one-to-one.

**`quorum_anomaly` is defect-gated**: a snapshot is emitted only when a real defect trigger fired (rejected quorum, conflicting output reports, malicious voters / attribution / local-authority-malicious). The two race conditions still appear in `trigger_conditions` as context on those snapshots, and the dedicated snapshot fields (`local_output_observed`, `quorum_reached_without_local_output`, `running_computation_count`) are unchanged. Note this means `anomaly_triggers_total{trigger="quorum_reached_before_local_output_observed"}` now counts only defect-correlated races — raw race volume lives in the new counter.

**`computation_update_after_session_completion` is evidence-gated**: a snapshot is emitted only when the late network-key reconfiguration output capture ran (matched / diverged / unverified) — that is the #1844 mixed-rollout compatibility evidence and it is rare (bounded by reconfiguration sessions). The plain non-active computation update — the flood source — increments the race counter and logs at `debug` (demoted from per-occurrence `warn`).

The #1844 release-gate contract is untouched: `ika_dwallet_mpc_session_late_output_info` gauges, the `DIVERGES` error log line the harness greps for, and both `late_reconfiguration_output` regression tests are unchanged in behavior.

## Why not debounce per (validator, epoch) (suggestion 3)

Debouncing would still emit one misleading "anomaly" per epoch per validator per kind and would make the first snapshot of an epoch meaningless (it captures an arbitrary benign race, not a defect). Removing the benign kinds from the taxonomy while keeping their state as context on real snapshots preserves all diagnostic value at zero steady-state noise.

## Testing

- `cargo test --release -p ika-core anomaly_diagnostics` — 11 passed (reworked: benign race now asserts **no** snapshot + race counters; a new defect-variant test pins race-context triggers, running-computation state, and payload redaction on the emitted snapshot).
- `cargo test --release -p ika-core late_reconfiguration_output` — 3 passed (matched/diverged evidence snapshots unchanged; the non-reconfiguration late-finalize test now additionally pins no-snapshot + race-counter).
- `cargo test --release -p ika-core mpc_diagnostics::` — 8 passed.
- `cargo clippy --release -p ika-core --all-targets` clean; `./scripts/check-metric-names.sh` OK (223 literal).
- Docs updated in the same PR: `dev-docs/playbooks/mpc-anomaly-diagnostics.md` (triggers, alerting, example shapes, upgrade-test note) and the generated inventory in `dev-docs/conventions/metrics.md` (regenerated via the script, which also fixes pre-existing drift).

---

## Scope expansion: session-origin provenance (merged from #1858, closes #1857)

This branch now also carries the provenance follow-up: a `SessionOrigin` tag (`local_request` | `reconstructed_from_consensus`, upgraded in place when the request arrives), the `ika_dwallet_mpc_sessions_reconstructed_total{session_type}` counter, `session_origin` in diagnostic snapshots with race-context omitted for reconstructed sessions, and the opportunistic `session_start_count` IntGaugeVec→IntCounterVec conversion. This keeps the restart-wedge signature (#1852/#1856) directly alertable after the anomaly-flood demotion in this PR removes the accidental visibility.


omersadika:
## Review — solves #1853

Verified against the diff and the flood evidence from the issue:

**Both flood sources are correctly reclassified.**
- `quorum_anomaly`: the two race conditions are moved out of the emission decision — `emit_session_anomaly(QuorumAnomaly, ..)` is reachable only from the defect-gated branch (rejection / conflicting reports / malicious attribution), with the races attached to `trigger_conditions` only as context on those snapshots. Race occurrences increment `completion_races_total` unconditionally, so raw race volume stays measurable with the same label strings (1:1 dashboard migration, as advertised).
- `computation_update_after_session_completion`: the plain non-active update — the actual firehose — increments the race counter and demotes to `debug!`, which also kills the Loki/journald line volume (the `ika_mpc_diagnostics` target runs at info). The snapshot survives only for the late reconfiguration-output comparison, bounded by reconfiguration sessions. The #1844 gate contract (late-output gauges, the DIVERGES grep line, both regression tests) is preserved.

**Tests pin the semantics well**: benign race → 0 snapshots + both race counters at exactly 1; defect quorum → snapshot emitted carrying `malicious_authority_identified` *and* the race context, payload redaction re-asserted; non-active update → counter, no snapshot. The metrics.md inventory catch-up (dropped_total, presign-parked, late-output gauges) is welcome hygiene.

**One post-merge note worth adding to #1852**: the wedged-validator signature loses its (accidental, zero-specificity) presence in the anomaly flood — a #1852-wedged validator now emits no quorum snapshots at all, since nothing defect-gated fires. That's the right call (the flood was uniform across healthy nodes, so it never distinguished the wedge), but it makes #1852's dedicated signal ask slightly more important. Cheap interim heuristic the new counter enables: per-host `completion_races_total{race="quorum_reached_before_local_output_observed"}` rate ≈ session rate (ratio → 1.0) is exactly the wedge, whereas healthy validators sit at the natural straggler fraction. Worth a line in #1852.

Rate math sanity: at the observed ~9 snapshots/sec, this PR removes ~99.98% of snapshot volume (remaining: defect-correlated + rare late-output comparisons), and anomaly panels regain meaning. LGTM.

PR #1855, opened 2026-07-18
fix(dwallet-mpc): re-instantiate network keys from chain after a mid-epoch restart (#1852)

omersadika:
Addresses the restart-strand variant of #1852 (deliberately NOT auto-closing: live post-handoff monitoring split the issue into three variants — see https://github.com/dwallet-labs/ika/issues/1852#issuecomment-5012364385 — and this PR covers variant 1, the produced-this-epoch strand; variant 2 = missing canonical DKG anchor after restart, variant 3 = missed MPC-data freeze, tracked in #1856) — **testnet liveness**: 4 validators (~11% of stake) that restarted mid-epoch completed zero MPC advances for 6–8.7h, with 263 sessions parked on the network key and re-instantiation never attempted.

## Mechanism

Once an epoch's network-key reconfiguration completes, the off-chain overlay serves the just-produced reconfiguration output R(M) — the shares for the **next** committee. The adoption pass's `produced_this_epoch` guard correctly skips R(M) for a running validator (it already holds this epoch's output R(M-1) from epoch entry and is only pre-staging the boundary flip). A validator that restarts after that completion holds **nothing**: the overlay serves only R(M), the guard skips it forever, and the key is never instantiated. #1835's level-triggered drain releases parked requests once a key's public data exists — but nothing ever made the data exist. This exactly matches the issue's metric signature: `requests_pending_for_network_key` climbing, `instantiations_in_flight` 0, `instantiation_failures_total` flat (never attempted — the skip is upstream of any attempt), `network_key_loaded_epoch` cut off at process restart.

## Fix: stranded-key recovery

The manager is the only component that can see the strand, so it drives the recovery:

- When the `produced_this_epoch` guard skips a key the validator holds **nothing** for — not instantiated, no instantiation in flight, nothing adopted (the in-flight/adopted checks keep the healthy first-instantiation window out) — adoption flags the key in a node-lifetime `Arc<ArcSwap<HashSet<ObjectID>>>` shared with the sui-connector network-keys sync task.
- The syncer chain-reads **exactly the flagged keys**: full read instead of the synthesize-empty fast path (the Move contract writes reconfiguration outputs on-chain under `current_epoch + 1` regardless of protocol version, so `get_current_reconfiguration_public_output(M)` = R(M-1) with real bytes), and the new `overlay_network_key_data` helper does **not** overlay the off-chain reconfiguration blob for a flagged key (which would re-override R(M-1) with R(M)). A flagged key also bypasses the syncer's `(epoch, state)` fetch memo — the flag is raised after the strand-revealing entry was already recorded there.
- A confirmed instantiation un-flags the key, returning it to the off-chain read path. Recovery completes within one syncer tick (~5s) plus one instantiation (~10s).
- The **DKG blob still prefers the off-chain mirror** on the recovery read: the mirror holds the canonical (post-V2→V3) representation whose digest the prior epoch's handoff cert pins. Serving the chain's original V2 anchor would (a) fail adoption's cert DKG-digest gate — trading the silent wedge for a loud one — and (b) let the handoff hydration path file the stale anchor's digest under this epoch, diverging this validator's DKG handoff item from peers (`AttestationMismatch`).
- The previously-silent `produced_this_epoch` skip now logs at info with an `already_instantiated` field, and the strand-flagging logs explicitly — the observability the issue's monitoring note asked for.

The set is **empty in every healthy flow** (non-validators never run adoption at all), so the v4 no-steady-state-chain-read invariant holds exactly — asserted by the `off_chain_metadata_v4_does_not_read_blobs_from_chain` cluster test. An earlier draft of this fix keyed the syncer on an instantiated-keys set instead (the shape validated on the multichain feature branch as `80e028821e`); the cluster test caught it chain-reading during healthy first-instantiation windows, and the strand-set inversion is the result: the chain read now happens exactly when the wedge exists, and only then. Relative to that feature-branch fix, the NetworkKeyId-derivation-guard re-arm is also **not** ported — #1825's input-digest memo supersedes it (the recovery read changes the derivation inputs, which re-arms that memo naturally).

## No fork / epoch-close safety

Cleared by a dedicated consensus-determinism review pass:

- Both a recovered and a running validator install the identical canonical R(M-1) for the epoch (chain bytes are the same checkpoint-chunk bytes the local cache holds); the cert-digest equality gate rejects anything else.
- The stranded set is a local read-path signal that reaches no consensus artifact.
- EndOfPublish readiness (`snapshot_ready_for_signing`) gates the reconfiguration output on the epoch-keyed digest slice, not overlay bytes, so the R(M-1)-carrying overlay can neither withhold the vote nor sign a stale digest.

The review also surfaced one MEDIUM adjacent defect this PR fixes: the sync pass's chain-read `Err` arm aborted the **whole pass including the channel send**, after a sibling key that merged complete earlier in the same pass was already recorded as fetched — permanently withholding that sibling from the overlay until the next epoch. Pre-existing (the arm was unreachable for v4 cached keys), but the recovery read makes it hot on exactly the restart path. The arm now skips only the failed key.

Known residual (documented in the spec): a mid-epoch restart during the single V2→V3 canonical-migration epoch is not recovered (the prior cert pins V2 while the local mirror already flipped to V3) — fail-closed, warns loudly, self-heals at the next epoch boundary, identical to pre-fix behavior in that epoch.

## Spec/playbook

- `dev-docs/specs/handoff.md`: stranded-key recovery is now the third adoption guard; the "chain reads deprecated" sentence gains its one exception.
- `dev-docs/playbooks/mpc-stall-postmortem.md`: the #1852 metric signature added to the stall catalogue.

## Validation

- New manager-level regression test `this_epoch_reconfiguration_output_is_skipped_and_current_epoch_output_recovers`: an overlay serving R(M) is skipped by the produced-this-epoch guard even when the cert deliberately matches (isolating the guard), the skipped key is flagged as stranded, and an overlay serving R(M-1) — what the recovery read delivers — is adopted and spawns instantiation. **Fault-validated** per the test-testing playbook: disabling the guard makes the test fail at exactly the phase-1 assertion; reverted.
- New unit test `stranded_key_keeps_chain_reconfiguration_output_and_canonical_dkg` pins the overlay-merge contract (chain reconfiguration output kept, canonical DKG mirror preferred, chain copy untouched with no source).
- `network_key_adoption` suite (6 tests) and `missing_network_key` suite (incl. #1835's park-then-materialize vehicle) green in release with real crypto; `off_chain_metadata_v4_does_not_read_blobs_from_chain` green locally after the strand-set redesign.
- `cargo check --workspace --all-targets --release` and clippy clean.
- Heavy CI suites (integration / cluster / TS) dispatched on the final commit.

Follow-up (not in this PR): an explicit `sessions_parked_without_key_instantiation` anomaly kind, per the issue's monitoring note — the fix makes the state self-healing and observable via logs, but a first-class anomaly would let infra alert on any future variant.

---

## Scope expansion: variant 2 (merged from #1859)

This branch now also carries the **variant-2 fix** — the permanent never-instantiated wedge (`canonical_dkg_output_version=0`: Full Sail, n1stake, alphab.ai), root-caused in [#1852 (comment)](https://github.com/dwallet-labs/ika/issues/1852#issuecomment-5012466405):

- **Hydration guard:** the handoff sender's `hydrate_network_dkg_digests` pass is now fill-absence-only — it never overwrites an existing DKG digest, so a stale boot-race overlay can no longer clobber the durable perpetual canonical mirror.
- **Fetch-memo invalidation:** the network-keys sync task re-merges all keys when the off-chain blob source installs, closing the boot-race window minutes after startup instead of carrying a stale chain-anchor overlay all epoch.
- **Barrier DKG repair:** the prepare-then-start barrier's readiness predicate now digest-checks `NetworkDkgOutput` items against the handoff cert (previously unconditionally ready), and the existing peer-fetch installer repairs a missing/mismatching mirror — this is what recovers the three currently-poisoned validators at their first epoch boundary after deploy, with no operator action.

Rolling-upgrade safe: no protocol/Move/wire changes; the repair fetch uses the existing digest-verified blob protocol already served by v1.2.1 peers.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
@@ -30,6 +30,7 @@ use crate::dwallet_session_request::{DWalletSessionRequest, DWalletSessionReques
 use crate::epoch::submit_to_consensus::DWalletMPCSubmitToConsensus;
 use crate::noa_checkpoints::NOACheckpointHandler;
 use crate::request_protocol_data::ProtocolData;
+use arc_swap::ArcSwap;
 use commitment::CommitmentSizedNumber;
 use dwallet_classgroups_types::ValidatorMPCSecrets;
 use dwallet_mpc_types::dwallet_mpc::MPCDataTrait;
@@ -72,6 +73,7 @@ use prometheus::Registry;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+use sui_types::base_types::ObjectID;
 use sui_types::messages_consensus::Round;
 #[cfg(any(test, feature = "test-utils"))]
 use tokio::sync::watch;
@@ -191,6 +193,10 @@ impl DWalletMPCService {
         system_checkpoint_handler: Option<
             NOACheckpointHandler<noa_checkpoint::SuiSystemCheckpoint>,
         >,
+        // Shared set of instantiated network keys (created once at the node
+        // seam, also handed to the sui-connector syncer). The manager writes
+        // here on confirmed instantiation; the syncer reads it.
+        instantiated_network_keys: Arc<ArcSwap<HashSet<ObjectID>>>,
     ) -> Self {
         let network_dkg_third_round_delay = protocol_config.network_dkg_third_round_delay();
 
@@ -227,6 +233,7 @@ impl DWalletMPCService {
             epoch_store.clone(),
             network_owned_address_sign_output_sender,
             max_mpc_computation_cores,
+            instantiated_network_keys,
         );
 
         Self {
@@ -315,6 +322,7 @@ impl DWalletMPCService {
                 epoch_store,
                 network_owned_address_sign_output_sender,
                 None,
+                Arc::new(ArcSwap::from_pointee(HashSet::new())),
             ),
             exit: watch::channel(()).1,
             end_of_publish: false,
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
@@ -193,10 +193,11 @@ impl DWalletMPCService {
         system_checkpoint_handler: Option<
             NOACheckpointHandler<noa_checkpoint::SuiSystemCheckpoint>,
         >,
-        // Shared set of instantiated network keys (created once at the node
-        // seam, also handed to the sui-connector syncer). The manager writes
-        // here on confirmed instantiation; the syncer reads it.
-        instantiated_network_keys: Arc<ArcSwap<HashSet<ObjectID>>>,
+        // Shared set of restart-stranded network keys (created once at the
+        // node seam, also handed to the sui-connector syncer). The manager
+        // flags a stranded key at adoption and un-flags it on confirmed
+        // instantiation; the syncer chain-reads flagged keys.
+        stranded_network_keys: Arc<ArcSwap<HashSet<ObjectID>>>,
     ) -> Self {
         let network_dkg_third_round_delay = protocol_config.network_dkg_third_round_delay();
 
@@ -233,7 +234,7 @@ impl DWalletMPCService {
             epoch_store.clone(),
             network_owned_address_sign_output_sender,
             max_mpc_computation_cores,
-            instantiated_network_keys,
+            stranded_network_keys,
         );
 
         Self {
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/network_key_adoption.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/network_key_adoption.rs
@@ -709,3 +709,148 @@ async fn adopted_keys_always_resolve_a_network_key_id() {
     );
     assert_all_adopted_resolve(manager, "after the cert-less registration pass");
 }
+
+/// The mid-epoch-restart strand (#1852), manager side. Once this epoch's
+/// reconfiguration completes, its output digest is recorded under the CURRENT
+/// epoch and the off-chain overlay serves that output — which is encrypted to
+/// the NEXT committee. Adoption must skip it (a running validator already
+/// holds this epoch's parameters; adopting would fail decryption), and that
+/// skip is exactly what strands a validator that restarted mid-epoch with an
+/// empty instantiation map: nothing else re-delivers an instantiable output.
+/// The recovery contract this test pins: an overlay carrying the CURRENT
+/// epoch's output — what the syncer now sources from chain for a key the
+/// manager hasn't instantiated — must be adopted and spawn instantiation.
+#[tokio::test]
+async fn this_epoch_reconfiguration_output_is_skipped_and_current_epoch_output_recovers() {
+    let _ = tracing_subscriber::fmt().with_test_writer().try_init();
+    let (
+        mut dwallet_mpc_services,
+        _sui_data_senders,
+        _sent_consensus_messages_collectors,
+        epoch_stores,
+        _notify_services,
+        _network_owned_address_sign_request_senders,
+        _network_owned_address_sign_output_receivers,
+    ) = utils::create_dwallet_mpc_services(1);
+    let service = dwallet_mpc_services.first_mut().unwrap();
+    let epoch_id = service.epoch;
+    let prior_epoch = epoch_id - 1;
+
+    let key_id = ObjectID::random();
+    let network_key_id = NetworkKeyId([0x71; 32]);
+    crate::network_key_id_mapping::register(key_id, network_key_id);
+    let dkg_output = b"restart strand network dkg public output".to_vec();
+    // R(M-1): produced by the PRIOR epoch's reconfiguration FOR this epoch's
+    // committee — the output this epoch's parameters derive from, and what
+    // the chain serves as the current epoch's reconfiguration output.
+    let current_epoch_output = b"reconfiguration output for the current committee".to_vec();
+    // R(M): produced by THIS epoch's reconfiguration FOR the next committee.
+    let this_epoch_output = b"reconfiguration output for the next committee".to_vec();
+
+    // Record R(M)'s digest under the current epoch, exactly as output
+    // processing does when this epoch's reconfiguration completes.
+    let perpetual_dir = tempfile::tempdir().expect("tempdir");
+    let perpetual = Arc::new(
+        crate::authority::authority_perpetual_tables::AuthorityPerpetualTables::open(
+            perpetual_dir.path(),
+            None,
+        ),
+    );
+    perpetual
+        .insert_network_reconfiguration_output_digest_for_epoch(
+            epoch_id,
+            key_id,
+            mpc_data_blob_hash(&this_epoch_output),
+        )
+        .expect("digest insert");
+    epoch_stores
+        .first()
+        .unwrap()
+        .perpetual_tables
+        .lock()
+        .unwrap()
+        .replace(perpetual);
+
+    // Phase 1 — the strand shape. The cert deliberately pins R(M) so the
+    // cert-digest gate alone would ADOPT this overlay; only the
+    // produced-this-epoch guard can be the one skipping it. (In the real
+    // restart the prior epoch's cert pins R(M-1) and both gates skip R(M);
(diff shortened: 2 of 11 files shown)


omersadika:
**CI status on 2df15d01ae:** cluster suite ✅ ([29641109679](https://github.com/dwallet-labs/ika/actions/runs/29641109679) — includes `off_chain_metadata_v4_does_not_read_blobs_from_chain`, which caught the first design), Rust integration ✅ ([29641108862](https://github.com/dwallet-labs/ika/actions/runs/29641108862)), TS integration ❌ ([29641110659](https://github.com/dwallet-labs/ika/actions/runs/29641110659)) → triaged as **not branch-caused**, retry dispatched ([29644272838](https://github.com/dwallet-labs/ika/actions/runs/29644272838)).

TS-failure triage (per the MPC-stall playbook, from the run's `localnet-logs` artifact):

- The failure is the known cascading late-run shape: two `all-combinations-future-sign.test.ts` cases timed out ~90 minutes in with "Object does not exist" — the dwallet/sign objects were never created on chain.
- This PR's code paths were **provably inert** in the run: zero "network key stranded" flags, zero recovery chain reads (`get_network_encryption_key_with_full_data_by_epoch` never called), all 17 `produced_this_epoch` skips logged `already_instantiated=true` (the healthy pre-stage shape), 38/38 network-key instantiations succeeded, zero instantiation failures.
- The network stayed healthy throughout: epochs advanced on the regular 15-minute cadence for the whole run, no panics, no `should_never_happen`, no malicious convictions (the "malicious" log volume is the benign `quorum_anomaly` diagnostic flood tracked in #1853), and validators completed **2,223 sessions during the failure window** (11:30–12:20). The failing objects' ids never appear in the validator logs — the requests never reached the MPC plane.
- Shape matches the pre-existing late-run TS wedge tracked in #1736 (parked-sessions-never-compute / cumulative-load), which the playbook requires ruling out before attributing a TS failure to a branch.

omersadika:
**TS retry ([29644272838](https://github.com/dwallet-labs/ika/actions/runs/29644272838)) also failed — deeper triage, still pointing away from the branch; third run dispatched ([29647645821](https://github.com/dwallet-labs/ika/actions/runs/29647645821)) for a stronger sample.**

Per-file timeline vs the green main run ([29629809313](https://github.com/dwallet-labs/ika/actions/runs/29629809313), same base commit):

| file | green main | this branch (retry) |
|---|---|---|
| dwallet-creation | 3m34 | 3m31 |
| global-presign | 1m17 | 1m15 |
| transfer-dwallet | 7m58 | 7m53 |
| imported-key | 4m55 | 4m35 |
| make-public-share-and-sign | 6m05 | 6m35 |
| imported-key-make-public-share | 4m58 | 7m06 |
| dwallet-sign-during-dkg | 5m32 | 5m21 |
| all-combinations | 5m27 | 5m55 |
| **all-combinations-future-sign** | **18m37** | **45m27 (failed)** |

- No branch-wide slowdown: ~50 minutes of preceding tests run at green-main speed. The entire delta is the 1200s-per-case timeouts inside the final, heaviest file.
- **Different combinations fail in each run** (run 1: Secp256r1 imported-key SHA256 + Ristretto shared Merlin; run 2: three others incl. a DKG that never reached `Active`) while sibling cases in the same file pass — nondeterministic victims, not a curve/algorithm regression.
- This PR's code paths were again provably inert (zero strand flags, zero recovery chain reads, zero instantiation failures, all `produced_this_epoch` skips `already_instantiated=true`, no panics).
- The failing objects' ids again never appear in the validator logs; the network completed sessions normally throughout.
- The CPU sampler shows a heavily loaded, cgroup-throttled runner (load 14–20 at start), with the pod mostly idle-waiting (~0.4 CPUs) during the wedge — consistent with a load-sensitive race upstream of the MPC plane, the #1736 family shape. Main's daily 04:00 greens are single low-load samples.

Caveat stated plainly: two consecutive failures against a recently-green main can't be *fully* excluded as a subtle timing interaction, but there is no mechanism in the diff (its only always-on additions are a per-tick `HashSet` load and a few map lookups) and no trace of its involvement in either run's logs.

omersadika:
Reviewed against the #1852 evidence (I filed it from the monitoring side). **LGTM — the root cause matches the metric signature exactly, and the fix is sound.**

What I verified:

- **Strand detection is correctly scoped.** The flag is raised only when the produced-this-epoch guard skips a key the validator holds *nothing* for: `!already_instantiated && !pending_network_key_instantiations && !adopted_network_key_data`. The in-flight/adopted checks keep the healthy first-instantiation window out of the set, so a running validator pre-staging the boundary flip never gets flagged. This is exactly the #1852 signature (263 parked, 0 in-flight, 0 failures).
- **Un-flag only on confirmed instantiation.** The decrypt-fail / epoch-mismatch branches keep the flag, so a key that still can't instantiate keeps its chain-sourced overlay and retries rather than silently reverting to the wedged read path.
- **The recovery read is correct on both blobs.** Keeping the chain's canonical current-epoch reconfiguration output (R(M-1), real bytes — Move writes outputs under `current_epoch+1`) while still preferring the off-chain mirror for the DKG blob is right: overlaying the off-chain reconfiguration blob would reinstate R(M) and recreate the wedge, and the chain's pre-V3 DKG anchor would fail the cert's digest gate after the V2→V3 migration.
- **Cache bypass is necessary and present.** The flag is raised *after* the overlay entry was recorded in `last_fetched_network_keys`, so without the stranded-set check in the refetch filter the recovery read would never fire. Good catch documenting that ordering.
- **The `continue 'sync_network_keys` → per-key `continue` change is a real secondary fix**, not a refactor: previously one key's fetch error aborted the whole pass including the channel send, which could strand a sibling key that merged complete earlier in the same pass (recorded as fetched, never published). That arm becomes hot on exactly this recovery path, so fixing it here matters.
- **No steady-state chain reads added.** The set is empty in every healthy flow (and non-validators never run adoption), so the v4 invariant asserted by `off_chain_metadata_v4_does_not_read_blobs_from_chain` holds exactly.

Two notes, neither blocking:

1. **The V2→V3 migration-epoch residual is documented and fail-closed** — acceptable, but worth keeping in the ops playbook since a restart during that one epoch still wedges until the next cert.
2. **The n1stake variant from #1852 (zero MPC advances *without* the parked-sessions signature, no recent restart) is likely a different cause** and isn't covered here — suggest keeping #1852 open or splitting a follow-up for it rather than auto-closing on merge, unless it's confirmed recovered post-handoff.

Timing note: testnet's epoch-356 handoff already self-healed the five wedged validators (the guard is epoch-tagged, so the wedge is bounded to the epoch it occurs in), but this fix is what makes the **mainnet** rollout safe — ~61 validators restarting mid-epoch is the highest-density trigger window for this bug — and it stays load-bearing afterward for any mid-epoch crash/OOM/maintenance restart.

omersadika:
**Third TS run ✅ ([29647645821](https://github.com/dwallet-labs/ika/actions/runs/29647645821))** — full suite green with `all-combinations-future-sign` completing normally, supporting the triage that the two earlier failures were the pre-existing load-sensitive wedge (#1736 family), not this branch.

Final CI status on `2df15d01ae`: cluster ✅ · Rust integration ✅ · TS integration ✅ · PR checks (incl. `v118_mixed_rollout`, msim clippy) ✅. Ready for review.

omersadika:
**Correction to my timing note above**, now that I've live-verified against epoch-356 testnet data (details in https://github.com/dwallet-labs/ika/issues/1852#issuecomment-5012364385):

- The handoff **did** self-heal Encapsulate, Mysten Labs, and bitszn (~9.5k advances/h each) — the epoch-bounded recovery holds for the produced-this-epoch strand this PR fixes.
- It did **not** heal Full Sail Validator: 45h uptime (restarted in epoch 354), wedged through **two** handoffs, and `canonical_dkg_output_version=0` (fleet is at 3) — with no DKG anchor, adoption presumably never reaches the produced-this-epoch skip branch, so the stranded-set detection here wouldn't fire for it. Same for n1stake and alphab.ai (also dkg_version=0).
- Two additional validators ([NODERS], bitszn again) restarted mid-epoch 356 and are stuck on a **different** restart variant: key loaded fine, but `data_freeze_epoch=0` and zero advances — outside this PR's scope.

None of this changes the LGTM — the fix is correct for its variant and mainnet still wants it merged first. But I'd recommend **removing the `fixes #1852` auto-close** (or splitting the issue): the no-DKG-anchor and missed-freeze variants remain live on testnet right now.

PR #1858, opened 2026-07-18
feat(mpc): session-origin provenance and a sessions_reconstructed_total counter

### LINKED ISSUE #1857: Add session-origin provenance and a sessions_reconstructed_total counter (follow-up to #1854)

omersadika:
Fixes #1857. **Stacked on #1854** (base branch is `claude/issue-1853-0bb5fe`; retarget to `main` when #1854 merges — the diff shown here is the #1857 delta only).

## What

Sessions reconstructed from consensus artifacts after a restart were indistinguishable from sessions the validator participated in locally. After #1854 correctly silenced the benign quorum races, the unbounded no-local-output session stream — the #1852/#1856 mid-epoch-restart wedge signature ([NODERS]/bitszn: ~10k/8k such sessions in 2h with zero local advances) — went metric-dark except for advance counters.

- **`SessionOrigin` provenance tag** (`local_request` | `reconstructed_from_consensus`), derived from the creation status: `WaitingForSessionRequest` (stray message/output before the request) and `ComputationCompleted` (completion replayed from consensus) entries are reconstructions; every other creation site processes a local request. Activating a waiting entry upgrades it to `local_request` (the flip lives in `set_status`, so both the external-request and internal-presign activation paths get it for free).
- **`ika_dwallet_mpc_sessions_reconstructed_total{session_type}`** — a genuine `IntCounterVec`, incremented once at creation of a reconstructed session (a later upgrade doesn't retract it). Operational reading, now documented in the anomaly playbook: bounded spike at restart then flat = healthy recovery; still climbing at the network's session rate while `advance_completions` stays flat = the non-participation wedge, directly alertable.
- **Diagnostics rule**: missing local output on a *still-reconstructed* session is definitional, not a completion race — it no longer increments `completion_races_total{race="quorum_reached_before_local_output_observed"}` and is not attached as trigger context on defect snapshots. `session_origin` is carried inside the snapshot JSON and as a stable top-level field on tracked-session log lines, so triage can split the populations. The factual snapshot fields (`local_output_observed`, `quorum_reached_without_local_output`) are unchanged, and all correctness paths (voting, rejection, malicious attribution, output caching) are identical for both origins.
- **Opportunistic conversion** flagged in the issue: `session_start_count` was an inc-only `IntGaugeVec`; now a genuine `IntCounterVec`. Name kept (`_count`) for dashboard compatibility — noted as the one legacy exception in `dev-docs/conventions/metrics.md`; no in-repo consumer reads the exposition type.

## Interaction with #1854's tests

The #1854 benign-race and defect-context tests created their sessions *from outputs* — which under this PR makes them reconstructed. They now upgrade origin explicitly and keep pinning the local-request race semantics; two new tests pin the reconstructed side (quorum on a reconstructed session counts reconstruction, not a race, and emits nothing; a defect on one still snapshots, without race context and with `session_origin` split).

## Testing

- `cargo test --release -p ika-core anomaly_diagnostics` — 13 passed (2 new origin tests).
- `cargo test --release -p ika-core late_reconfiguration_output` — 3 passed (unchanged behavior; the #1844 evidence contract untouched).
- `cargo test --release -p ika-core message_before_event` — 1 passed, extended with real-path assertions: reconstruction counted at message-before-request creation, zero on prompt parties, origin upgraded to `local_request` on real activation with the creation count not retracted.
- `cargo clippy --release -p ika-core --all-targets` clean; `./scripts/check-metric-names.sh` OK (224 literal).
- Docs in the same PR: playbook section on session origin + the new counter's alerting reading; regenerated metrics inventory; convention note for the legacy counter name.

## Known labeling edge (deliberate)

The origin upgrade fires only on activation (`set_status(Active)`), per the issue's wording. Consequence: a `WaitingForSessionRequest` entry whose request arrives but **fails input construction** stays `reconstructed_from_consensus`, while the identical failure with no pre-existing entry is labeled `local_request` — so the label for "locally processed request that failed" depends on peer-message arrival order. Node-local labels only (metric/log JSON), never a gate; a Failed session definitionally lacks local output either way, and failures emit their own anomalies with full context. Flagged here so triage splitting populations by `session_origin` knows the edge exists.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
@@ -74,7 +74,10 @@ pub struct DWalletMPCMetrics {
     computation_duration_variance: GaugeVec,
 
     /// Tracks the number of MPC protocol sessions that have been started.
-    session_start_count: IntGaugeVec,
+    /// A genuine counter (monotonic, inc-only); the legacy `_count` name
+    /// predates the counter/`_total` suffix convention and is kept so
+    /// existing dashboards keep working.
+    session_start_count: IntCounterVec,
 
     /// Tracks the total number of completed MPC protocol sessions.
     ///
@@ -200,6 +203,16 @@ pub struct DWalletMPCMetrics {
     /// `anomaly_snapshots_total` meaningful. Labels: `race` (fixed condition
     /// strings), `session_type`.
     pub(crate) completion_races_total: IntCounterVec,
+
+    /// Sessions created from peer artifacts arriving through consensus (a
+    /// stray message/output before the request, or a completion replayed at
+    /// restart) rather than from a locally processed session request. A
+    /// bounded burst at restart is healthy recovery; a stream still climbing
+    /// long after restart while `advance_completions` stays flat is a
+    /// validator that is not participating at all (the mid-epoch-restart
+    /// wedge class). Incremented at creation; a later upgrade to
+    /// local-request origin does not retract it. Labels: `session_type`.
+    pub(crate) sessions_reconstructed_total: IntCounterVec,
 
     /// Individual reasons present in emitted anomaly snapshots. One snapshot can
     /// increment several triggers. Trigger values are compile-time static strings.
@@ -356,7 +369,7 @@ impl DWalletMPCMetrics {
         ];
 
         Arc::new(Self {
-            session_start_count: register_int_gauge_vec_with_registry!(
+            session_start_count: register_int_counter_vec_with_registry!(
                 "ika_dwallet_mpc_session_start_count",
                 "Number of MPC protocol sessions started",
                 &protocol_metric_labels,
@@ -566,6 +579,13 @@ impl DWalletMPCMetrics {
                 "ika_dwallet_mpc_completion_races_total",
                 "Expected races where a session completed via the peers' output quorum before the local validator caught up",
                 &["race", "session_type"],
+                registry,
+            )
+            .unwrap(),
+            sessions_reconstructed_total: register_int_counter_vec_with_registry!(
+                "ika_dwallet_mpc_sessions_reconstructed_total",
+                "Sessions created from peer artifacts arriving through consensus instead of a locally processed session request",
+                &["session_type"],
                 registry,
             )
             .unwrap(),
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/anomaly_diagnostics.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/anomaly_diagnostics.rs
@@ -3,7 +3,7 @@ use crate::dwallet_mpc::dwallet_mpc_service::DWalletMPCService;
 use crate::dwallet_mpc::integration_tests::utils;
 use crate::dwallet_mpc::mpc_diagnostics::{
     LocalAuthorityMaliciousReason, LocalComputationState, MpcAnomalyContext, MpcAnomalyKind,
-    output_digest, report_digest,
+    SessionOrigin, output_digest, report_digest,
 };
 use crate::dwallet_mpc::mpc_manager::MAX_UNTRACKED_ANOMALIES;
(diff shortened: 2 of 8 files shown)


omersadika:
Reviewed against the #1857 spec (I filed it). **LGTM — implements the issue exactly, plus the opportunistic metric-type fix.**

- `SessionOrigin` enum with in-place upgrade to `LocalRequest` when the request arrives and activates a `WaitingForSessionRequest` entry — tested (`activated_session.origin` assertion), so a reconstructed-then-activated session correctly leaves the reconstructed population.
- `ika_dwallet_mpc_sessions_reconstructed_total{session_type}` as a genuine `IntCounterVec`, incremented once at creation — the exact signal #1857 asked for. From the monitoring side this makes the wedge directly alertable (`reconstructed_total` still climbing N minutes post-restart while `advance_completions` stays flat), independent of the anomaly taxonomy #1854 quiets; I'll add that alert once this deploys.
- `session_origin` in the snapshot JSON and race-context omitted for reconstructed sessions — both pinned by tests (`reconstructed_session_quorum_counts_reconstruction_not_race`, `defect_snapshot_on_reconstructed_session_omits_race_context`), preserving all correctness/defect gating for both origins.
- Bonus: `session_start_count` converted `IntGaugeVec` → `IntCounterVec`. Confirmed no dashboard impact on our side: name and labels unchanged, and every Grafana/sweep query uses `increase()`/`rate()`, which are agnostic to the declared type. 

Merge sequencing: stacked on #1854 (base `claude/issue-1853-0bb5fe`) — merge #1854 first. CI still running at review time.

PR #1859, opened 2026-07-18
fix(handoff): stop the hydration pass clobbering the canonical DKG mirror; repair it at the barrier (#1852 variant 2)

omersadika:
Fixes variant 2 of #1852 — the **never-instantiated** wedge: three testnet validators (Full Sail Validator, n1stake, alphab.ai) with `canonical_dkg_output_version=0`, no `network_key_loaded_epoch` series, and zero instantiation attempts across **multiple epoch handoffs** (Full Sail: 45h, two handoffs). Root-cause analysis with confirmable predictions: https://github.com/dwallet-labs/ika/issues/1852#issuecomment-5012466405; monitoring verification (metric check confirmed, code chain independently verified): https://github.com/dwallet-labs/ika/issues/1852#issuecomment-5012484218.

Stacked on #1855 (variant 1); base auto-retargets to main when it merges.

## Mechanism

1. **Boot race** — after a cold start, the syncer's off-chain blob source is `None` until `monitor_reconfiguration`'s first iteration reaches the install, which sits behind a consensus submission (can block while consensus reconnects) and the joiner-bootstrap block. The syncer's first fetch then takes the full chain read and publishes an overlay whose DKG blob is the chain's **original pre-V3 anchor** (the chain never carries the reconstructed V3 canonical output).
2. **Stale overlay persists** — the entry is complete, so it's recorded in the fetch memo; if the restart landed after the epoch's reconfiguration completed, no chain-state change ever retriggers a re-merge, and installing the source didn't invalidate the memo. The V2-anchor overlay sits in the watch channel for the rest of the epoch.
3. **The clobber** — at EndOfPublish, the handoff sender's hydration pass cached the snapshot's DKG bytes unconditionally, overwriting the per-epoch digest table **and the durable perpetual canonical mirror** (`cache_protocol_output` writes both). The intact V3 mirror on disk is destroyed.
4. **Permanent wedge** — every later epoch serves the poisoned mirror (empty per-epoch table → perpetual fallback), adoption rejects the key at the handoff-cert DKG-digest gate *before* the `produced_this_epoch` branch (so #1855's stranded-key recovery is inert by design), and the barrier never repairs it: its readiness predicate returned `true` unconditionally for `NetworkDkgOutput` items, so the peer-fetch installer — which handles DKG items correctly — only ran while "not ready", i.e. never.

## Fix: three independent layers

**A — hydration is fill-absence only** (`hydrate_network_dkg_digests`, extracted for testability): the pass now skips any key that already has a digest in the merged view. Its purpose (a validator seeing EndOfPublish before its own MPC cached the output would build an attestation missing the item → `AttestationMismatch`) is fully preserved — filling absence is all it ever needed to do. The canonical overwrite paths (the instantiation mirror, including the one-shot V2→V3 flip, and the barrier's cert-anchored install) are deliberately NOT routed through this guard and still overwrite.

**B — the syncer clears its fetch memo when the blob-source identity changes** (install or per-epoch replacement, detected by `Arc` pointer identity): a source-less chain-read overlay from the boot window is re-merged within one tick of the install instead of persisting all epoch. The re-merge goes through the v4 fast path (no chain reads), so the no-chain-read invariant is unaffected.

**C — the barrier verifies and repairs DKG items** (`all_cert_network_key_outputs_held_locally`): epoch entry now blocks until every cert-pinned `NetworkDkgOutput` digest matches the local mirror, which makes the existing installer (already DKG-capable) actually run and re-cache the cert-pinned bytes from peers on mismatch or absence. The DKG digests are read from the **new** epoch store deliberately — its empty per-epoch table falls back to the perpetual mirror, the exact value the installer repairs; reading the outgoing store would consult the possibly-poisoned per-epoch table and deadlock the barrier. The every-10th-retry diagnostic warn now itemizes missing/unmapped keys across both item kinds.

**C is the layer that heals the three currently-wedged validators** — their poisoned perpetual mirror survives restarts, so A+B alone would prevent new poisonings but leave them stranded forever. With C deployed they recover at their next epoch boundary.

Migration-safety note (verified): the V2→V3 canonical flip happens at instantiation, *after* the barrier — so at the single migration boundary the cert pins V2 and the not-yet-flipped local mirror is V2 (match), and post-flip both sides are V3. The barrier's new DKG condition is satisfied on both sides of the migration for honest validators.

## Validation

- `hydration_fills_absent_digests_but_never_overwrites` — real epoch store + perpetual tables; asserts the existing canonical digest survives a conflicting snapshot while an absent key is still filled. **Fault-validated**: removing the guard fails the test at exactly the clobber assertion; reverted.
- `all_cert_network_key_outputs_held_locally_cases` — extended for DKG items: absent → not ready, poisoned (mismatching) mirror → not ready, canonical → ready.
- `network_key_adoption` suite (6), stranded-key overlay unit test, and the `off_chain_metadata_v4_does_not_read_blobs_from_chain` cluster test (layer B's blast radius) green in release.
- `cargo check --workspace --all-targets --release` and clippy clean. (Pre-existing, unrelated: `ika-node`'s `metrics::tests::test_metrics_endpoint_with_multiple_registries_add_remove` fails on the clean tree too.)
- Dedicated consensus-determinism review pass on the diff: no High findings; every barrier-liveness suspicion (missing perpetual handle, V2→V3 migration boundary, fresh-key closing epoch, all-poisoned committee, unmapped keys, layer-B ABA/cost) refuted with evidence. Its findings are addressed in this PR: the stale predicate/barrier docs that described the pre-fix reconfiguration-only gate are rewritten (they would have instructed a future maintainer to delete the fix), the fill-absence skip now WARNS when the snapshot contradicts an existing canonical digest (the #1852 signature, previously silent), the one-pass TOCTOU window is documented as accepted (healed by layer C), and a dead guard in the memo-clear was removed.

Spec: the barrier section and a new hydration-clobber entry under the adoption guards in `dev-docs/specs/handoff.md`.

### DIFF
--- a/crates/ika-core/src/epoch_tasks/handoff_signature_sender.rs
+++ b/crates/ika-core/src/epoch_tasks/handoff_signature_sender.rs
@@ -21,6 +21,7 @@ use crate::authority::authority_per_epoch_store::{
 use crate::consensus_adapter::SubmitToConsensus;
 use crate::validator_metadata::{HandoffItemsBuilder, next_committee_pubkey_set};
 use fastcrypto::ed25519::Ed25519KeyPair;
+use ika_network::mpc_artifacts::mpc_data_blob_hash;
 use ika_types::committee::Committee;
 use ika_types::dwallet_mpc_error::{DwalletMPCError, DwalletMPCResult};
 use ika_types::messages_consensus::ConsensusTransaction;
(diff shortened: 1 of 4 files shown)


omersadika:
Reviewed against the variant-2 root cause I independently verified on #1852 (code chain + live Mimir). **LGTM — all three layers implemented exactly as designed, and the one subtle invariant the repair depends on is preserved.**

- **Layer A (hydration guard):** fill-absence enforced in the *caller* (`hydrate_network_dkg_digests` checks `get_network_dkg_output_digests` first), while `cache_network_dkg_output` itself stays unconditional-write. That asymmetry is load-bearing — the barrier installer's cert-anchored write must still overwrite a poisoned mirror or layer C would be a no-op — and it's correctly preserved. The one-pass fill-absence race is honestly documented and bounded by C's next-boundary repair; accepting it over a lock is the right call. Test `hydration_fills_absent_digests_but_never_overwrites` pins the semantics.
- **Layer B (fetch-memo invalidation):** pointer-identity tracking of the blob-source Arc with a full `last_fetched_network_keys.clear()` on change is simple and correct; the re-merge cost is a handful of keys. This closes the boot-race window minutes after boot instead of end-of-epoch.
- **Layer C (barrier DKG repair):** the readiness predicate's unconditional `true` for `NetworkDkgOutput` (the "repair machinery exists but is never invoked" gap) is now a digest comparison against the cert, unified with the reconfiguration-item path so `missing_key_ids` drives the existing peer-fetch installer for both item kinds. `all_cert_network_key_outputs_held_locally_cases` covers the predicate.

Recovery claim verified in code: a validator whose perpetual mirror carries the pre-V3 anchor fails the digest compare at its first epoch boundary after deploy → barrier holds → installer fetches + caches the cert-pinned V3 bytes (unconditional write) → adoption's cert gate passes. That un-wedges Full Sail, n1stake, and alphab.ai without operator action — worth a line in the release notes.

Non-blocking suggestions: (1) an end-to-end integration test (poison the mirror, drive an epoch boundary, assert repair) would guard the A↔C asymmetry against future refactors — the two unit tests each cover one side but not the interlock; (2) post-deploy, the field confirmation is `ika_dwallet_mpc_network_encryption_key_canonical_dkg_output_version` flipping 0→3 on the three wedged validators at their first boundary — I'll watch it from the monitoring side.

Note for merge sequencing: stacked on #1855 (base `claude/issue-1852-ed2c8e`) — merge #1855 first. CI still running at review time.

omersadika:
**All three heavy suites green on the stack**: Rust integration ✅ ([29657978354](https://github.com/dwallet-labs/ika/actions/runs/29657978354)) · cluster ✅ ([29657979155](https://github.com/dwallet-labs/ika/actions/runs/29657979155)) · TS integration ✅ first-try ([29657979856](https://github.com/dwallet-labs/ika/actions/runs/29657979856)). Ready for review — merge #1855 first, this base auto-retargets to main.

PR #1860, opened 2026-07-18
feat(metrics): export validator host telemetry

omersadika:
## Summary

- export comprehensive validator host, process, storage, network, container, Kubernetes, and cloud telemetry through the existing ika-proxy push path
- add stable validator heartbeat and build, machine, deployment, binary-hash, cgroup, thermal, and collection-health information
- enrich validator heartbeat data with the authenticated validator identity, configured network, and proxy-observed source IP
- restrict environment collection to documented allow-listed values and avoid cloud credentials, tags, user data, and third-party geolocation
- document all validator metrics, metadata overrides, and cloud metadata behavior

## Verification

- cargo test --release -p ika-node validator_metrics
- cargo test --release -p ika-proxy consumer::tests
- cargo clippy -p ika-node -p ika-proxy --all-targets --all-features -- -D warnings
- cargo fmt --all -- --check
- ./scripts/check-metric-names.sh
- generated validator metric inventory matches documentation
- git diff --check


omersadika:
Reviewed from the monitoring side. **LGTM — I'd take this into v1.2.2.** The #1852/#1856 investigations spent real effort reconstructing exactly what this exports (restart times from `uptime`, cgroup limits, storage headroom questions we couldn't answer), and the slow-computation tail in the anomaly analysis (some validators at 15–34s per round vs 1s) is precisely where CPU/thermal/cgroup evidence was missing.

Verified in the diff:

- **The proxy `consumer.rs` change is a security hardening independent of telemetry**: `upsert_label` *replaces* node-supplied `network`/`host`/`metrics_source` labels instead of appending — the old `extend` allowed a node to ship a spoofed/duplicate `host` label alongside the proxy-derived one. Tests pin this (`spoofed-host`/`duplicate-host`). Worth having even if the telemetry were rejected.
- **Bounded by construction**: 15s refresh; network counters are aggregate (plain counters, no per-interface labels); storage is the DB volume only (plain gauges + one identity gauge with mount/fs/kind); info gauges carry small fixed label sets. Rough fleet cost ≈ 45 series/validator ≈ a few thousand series — trivial for Mimir.
- **IMDS handled carefully**: 300ms connect / 1s total timeouts, proxying disabled, no credentials/tags/user-data paths, `IKA_DISABLE_CLOUD_METADATA=1` kill switch, all documented. Env collection is allow-listed (`IKA_METRICS_*` + documented fallbacks), not an env dump.
- **Scoped correctly**: collection is gated on `node_mode.is_validator()` and runs on the metrics runtime; fullnode/notifier binaries only thread version strings.
- **Rolling-upgrade safe**: additive metric families over the existing push path; no protocol/Move/wire changes; v1.2.1 proxies would simply forward the new families (and the hardened proxy accepts old nodes unchanged).
- **`public_ip` only on the heartbeat**: attached to `ika_validator_up` alone, so it persists what the proxy already observes from the socket without stamping every series.

Two non-blocking suggestions:

1. **A module-level opt-out** (`IKA_DISABLE_HOST_TELEMETRY=1`, mirroring the cloud-metadata switch) would give operators who consider host details sensitive a documented alternative to firewalling — cheap goodwill for third-party validators.
2. The `host` label replacement in the proxy can change series identity for any node that previously shipped its own conflicting `host` label — cosmetic churn at worst, but worth one release-note line.

Monitoring side is covered: dwallet-labs/infra#95 (the v1.2.2 monitoring PR) is getting a Validator Host Telemetry row + a DB-volume disk-headroom sweep check keyed to these metrics.

PR #623, opened 2025-02-10
chore: Add system transactions parsing for all transactions

ItayLevyOfficial:

PR #1862, opened 2026-07-18
feat(mpc): protocol-v5-gated aggregated network-key public outputs (V4)

ycscaly:
## The problem

The network DKG and reconfiguration public outputs embed a threshold-encryption-to-sharing sub-output that persists **every dealer's full PVSS dealing** — per-receiver ciphertext plus EncDL proof per (curve, key-share part). That's O(n²) class-group material (~158 KB at 5×4 parties, ~100 MB projected at n=100), even though the proofs are only ever consumed *in-protocol*, during public verification and the malicious-dealer vote. Since all dealers encrypt receiver *i*'s share under *i*'s single PVSS key and Shamir sharings add coefficient-wise, each receiver's ciphertexts can be homomorphically summed into **one ciphertext per (curve, key-share part)** — O(n) storage, and share derivation drops from n class-group decryptions to one.

## The constraint that shaped everything

Testnet runs protocol v4 and already persists old-format outputs under the `V3` tags of ika's versioned enums. Two hard facts follow: MPC outputs must reach **byte-identical quorum**, so a mixed-binary committee where new binaries emit a new format splits the output vote at every epoch boundary; and old binaries can't decode an unknown BCS enum variant at all. So the format change cannot be binary-driven — it must flip at a **protocol version boundary**, which in turn means a new binary must still be able to produce the old format bit-for-bit at v4. (CLAUDE.md's "v4 not active on any deployed network" gotcha was stale; fixed here.)

## The design

**Crypto side (cryptography-private#623) — don't aggregate in-protocol.** Round 4 and the persisted `PublicOutput` are untouched from main, so wire identity with deployed binaries is trivial rather than something to prove. The aggregated shape is a purely additive *derived* type: `AggregatedPublicOutput` (for dkg, reconfiguration, and the sharing sub-output), reached via `PublicOutput::upgrade()` — the per-receiver summation done on demand. It carries the same core prefix, the same share-derivation and polynomial-commitment API (now one decryption), and an aggregated `new_from_reconfiguration_output` for anchor reconstruction. Upgrade is deterministic given identical input bytes — which quorum already guarantees — and derivation semantics are preserved exactly (a receiver with no aggregated entry derives the masked secret, matching the old empty-dealings behavior, which also keeps the mock's neutral output safe). Eager Round-4 aggregation is deferred until old-format production is dropped network-wide.

**ika side (this PR) — new tag, protocol-gated flip.**
- `V4` variants on `VersionedNetworkDkgOutput` / `VersionedDecryptionKeyReconfigurationOutput` carry aggregated bytes; `V3` arms are untouched and stay readable forever (testnet's persisted history).
- Protocol **v5** (`MAX_PROTOCOL_VERSION = 5`; NOA checkpoints move to planned v6) carries the `aggregated_network_key_public_outputs` flag. Producers write the protocol's output as `V3` below the gate and `upgrade()` → `V4` above it (flag threaded manager → `ComputationRequest` → `compute_mpc`), so the whole committee flips in the same epoch.
- The V1/V2-anchor reconstruction keys its output version off the **reconfiguration output's version** (V3 prior → V3 anchor, V4 prior → V4 anchor) — input-driven, so it stays byte-identical across validators without consulting config.
- Consumers accept both tags permanently: core-only sites decode the shared `PublicOutputCore` prefix of either shape; the VSS Shamir cache always derives on the aggregated form (upgrading a V3 output once per epoch per key), so signing gets the one-decryption win even before v5. SDK/WASM protocol-public-parameter extractors gain V4 arms (TS passes bytes to WASM opaquely).

## What each rollout stage looks like

At v4, a new binary behaves byte-identically to the old one. When v5 activates (only after all validators run the new release *and* an SDK release with the new WASM has shipped, since deployed SDKs decode V3 only), the first reconfiguration consumes the V3 prior through the old decode arm and emits V4. The storage and derivation wins materialize from that epoch on.

## Cross-tested against the deployed testnet binary

The `v121_rollout` upgrade-harness scenario (in this PR, wired into the upgrade-test workflow) boots four literal `release/testnet-v1.2.1` validators, swaps one to this branch, and witnesses two mixed-committee v4 reshares with **byte-identical per-authority output digests** (zero malicious reports, full user lifecycle after each) — then swaps the rest, rides the capability vote to v5, and confirms the first aggregated V4 reconfiguration output installs on every validator (via the new `..._latest_reconfiguration_output_version` gauge) while continuing to serve users. Green on CI, alongside the v118 mainnet gate.

## Tests

- New `test_pre_aggregation_to_aggregated_reconfiguration_migration`: the v4→v5 boundary — reconfiguration from a (V3 anchor, V3 prior output) pair must instantiate, produce a **V4-tagged** output, and install. Fault-injection-verified non-vacuous: disabling the V3-prior decode arm fails the test with the injected error visible in every validator's phase-2 session.
- `test_v2_to_v3_reconfiguration_migration` / `test_v1_anchor_main_reconfiguration_and_anchor_migration` pin the flag **off** — they model deployed testnet (v4), which must keep producing V3 until v5 activates.
- Default-config tests (MAX = v5) exercise V4 production/installation/VSS-cache consumption end to end, including the NOA-VSS sign flows.
- Crypto-side equivalence (`derive(old) == derive(upgrade(old))`) is pinned in cryptography-private#623's `dkgs_reconfigures_signs`.
- All four targeted real-crypto integration tests green locally; Integration Tests, Test Cluster, TS Integration Tests, and the v1.1.8 mixed-rollout Upgrade Test (with MAX=5 capability signaling) green on CI.

## Merge blockers

- [ ] cryptography-private#623 merged + published to inkrypto; swap the temporary cryptography-private pin (commit `e2782fcd93`) to the published inkrypto rev and drop the setup-ssh rewrite.
- [ ] SDK release with the rebuilt WASM sequenced before v5 activation on testnet.

https://claude.ai/code/session_015sMBcixNTaHLxR1vyeNvQW


omersadika:
> **Update 2026-07-19:** two positions below were revised after operator-side discussion — see the [decisions comment](https://github.com/dwallet-labs/ika/pull/1862#issuecomment-5014952807): existing DKG anchors **will migrate to V4** at the v5 flip on both networks (V3-forever is not the endstate), and mainnet jumps **protocol 3→5 directly** (skipping v1.2.2 entirely). The struck text below is kept for history.

Reviewed from the monitoring/rollout side (the crypto in cryptography-private#623 is outside what I can verify — this review covers the gating, migration, and operational surface). **Design LGTM, one determinism question, and a release-placement recommendation.**

What holds up well:

- **The gating architecture is right.** Additive `V4` enum variants; the persist path branches solely on `aggregated_network_key_public_outputs()` (on at protocol 5); below the gate V3 is persisted as-is, so wire identity with deployed v1.2.1 binaries is structural rather than something to prove. Readers accept V2/V3/V4 with V3 upgraded *locally* on read — consumption gets the O(n) win immediately at v4 without touching quorum bytes. ~~Existing DKG outputs staying V3 forever sidesteps a repeat of the V2→V3 canonical-migration hazard (#1855's residual)~~ *(revised — see update above: DKG anchors migrate to V4 at the flip, running the proven canonical-migration pattern with #1855's barrier repair as the safety net)* — the canonical DKG mirror, cert digest gates, and barrier repair are untouched by this change as written, which I verified against the #1855 machinery that just merged.
- **The rollout scenario is exactly the right test**: `v121_rollout_reaches_aggregated_outputs_at_v5` pins protocol ≤4 through every mixed phase and reaches v5/aggregated only after full upgrade, and `test_pre_aggregation_to_aggregated_reconfiguration_migration` exercises the V3-anchor → aggregated path with real bytes. The stale CLAUDE.md gotcha fix is appreciated.

**The one thing I'd want pinned before the v5 flip ever activates: `upgrade()` determinism as a fixed-vector test.** At the flip epoch every validator independently derives and persists the V4 output, so byte-identical quorum rests entirely on the homomorphic aggregation being canonically ordered (receivers, curve-parts, summation order) across binaries and refactors. The migration test proves one binary round-trips; a snapshot-style test (fixed V3 input → pinned V4 output bytes) in ika or #623 would make iteration-order nondeterminism un-introducible. If #623 already has this, ignore me — just say so on the PR for the record. *(Expanded in the decisions comment: the fixed-vector ask now also covers the V2-start compound path for mainnet's 3→5 jump.)*

Monitoring note: the new `ika_dwallet_mpc_network_encryption_key_latest_reconfiguration_output_version` gauge is exactly what we need to watch the flip fleet-wide — I'll add a version-distribution panel next to the canonical-DKG one on the infra side once this merges.

**Release placement recommendation: not in v1.2.2.** The wedge-fix release (#1854/#1855/#1860 + bump) is fully validated as a set — PR CI plus all four heavy suites executed per branch — and 5–6 testnet validators are wedged *now*, with the mainnet upgrade window next. This PR's feature is dormant until a protocol-5 committee quorum exists, so it gains nothing by riding the urgent release, while adding a MAX_PROTOCOL_VERSION bump and ~2k lines to a tree that would need re-validation (this PR's own upgrade-test runs are still in flight as I write). Suggest: cut v1.2.2 as staged, land this next, and let it be the headline of the v5-max release with its own rollout comms — including the explicit note that once 2f+1 stake runs the new binary, v5 activation excludes any remaining old binaries at the next boundary. *(Superseded in part: v1.2.2 is now testnet/devnet-only — mainnet stays on v1.1.8 and takes the v5-max release directly; see the decisions comment.)*

omersadika:
## Release-path decisions + two concrete asks

After reviewing the gating design (LGTM above stands), we've settled two decisions on the operator side that shape this PR's remaining scope:

**1. Existing DKG anchors migrate to V4 on both networks — no V3-forever operational state.** The upgrade-on-read design is sound, but we don't want live state pinned on a legacy encoding indefinitely. Since the canonical anchor's real pin is the off-chain mirror + handoff-cert digest chain (not chain immutability), this is a rerun of the proven V2→V3 canonical-migration pattern, gated at the same protocol-5 flip: derive `V4 = upgrade(V3)` deterministically, re-cache the mirror, build that epoch's handoff items with the V4 digest. The V3 enum arm stays for historical/chain decode either way — this migrates state, not code. Side benefit: `upgrade()` becomes a one-time migration tool instead of a permanent consensus-adjacent read path, and #1855's barrier repair gives this migration a boundary-repair safety net the V2→V3 one didn't have.

**2. Mainnet jumps protocol 3→5 directly.** `choose_highest_protocol_version_and_move_contracts_upgrades_v1` already advances to the highest quorum-supported version in one flip, so mainnet (protocol 3, V2 outputs, v1.1.8 fleet) skips v1.2.2 entirely and goes v1.1.8 → the v5-max binary → single 3→5 flip. Consequence: mainnet's flip epoch activates the full v4 feature set *and* v5 aggregation together, so the canonical migration must handle a **V2 starting mirror → V4 end state in one epoch** (chained V2→V3 reconstruction + V3→V4 upgrade). As written, the PR assumes a V3 start — this is the main scope addition.

**The two asks:**

1. **Fixed-vector determinism tests for the migration path**, covering both starting states: pinned V3 input → exact V4 bytes (testnet's path), and pinned V2 input → exact V4 bytes (mainnet's compound path). At the flip every validator derives and persists these bytes independently, so byte-identical quorum rests entirely on canonical ordering in `upgrade()` — if cryptography-private#623 already pins this, a pointer on this PR is enough for the V3 case, but the V2-start compound case is new.
2. **A mainnet-shaped rollout scenario** alongside `v121_rollout_reaches_aggregated_outputs_at_v5`: protocol-3 cluster with V2 persisted outputs → rolling binary upgrade → single 3→5 flip → byte-identical quorum through the mixed window, V4 canonical anchor + V4 reconfiguration output after, all validators healthy across the boundary. (The v118 mixed-rollout gate already proves the new binary behaves at protocol 3, which is the hard half.)

Sequencing on our side: testnet rehearses the smaller event first (4→5, V3→V4) and soaks at least one full epoch boundary before mainnet's jump; operator comms will carry the no-restarts-during-the-flip-epoch rule for both networks.

omersadika:
Reviewed `de0e145088` + `7c3ee71dc1` against the decisions comment — **the input-driven design is accepted, and it's cleaner than what we asked for.**

Making the anchor version a pure function of the (protocol-gated) reconfiguration output version gives us the V4 migration on both networks with no separate flag, no explicit migration machinery, and structural uniformity for mainnet's compound 3→5 jump. Scoping the unconditional V4 persist to *fresh* DKGs checks out on both premises: no deployed network ever persisted a V3-tagged fresh DKG (deployed keys carry V1 chain anchors; testnet's V3 anchors are reconstructions), and the V3 decode arms staying matches the "migrate state, not code" intent. The dropped mixed-phase gauge expectation is also correct — old binaries don't export the gauge, so only unanimous new-binary values appear during mixed windows (verified this keeps our infra-side version-split alerts false-positive-free).

Three things to close out:

1. **"No DKG session runs during a rollout window" is an operational assumption, not an invariant.** Nothing in code prevents a new network-key creation mid-rollout, and if one ran, old binaries would persist the fresh output V3-tagged while new ones persist V4 — a split output vote. Blast radius looks small (a failed key creation, not consensus damage), but please confirm the failure mode is *loud and retryable* (rejected/timed-out session) rather than a silently parked key, and we'll add "no network-key creation during binary rollout windows" to the operator runbook either way.
2. **Fixed-vector determinism tests still open** (from the decisions comment): pinned V3 input → exact V4 bytes, and pinned V2 input → exact V4 bytes. The boundary test exercises real crypto but doesn't pin bytes — the property that actually protects the flip is cross-binary/cross-refactor byte identity of the derivation, and a snapshot vector is what makes that un-regressable.
3. **Mainnet-shaped rollout scenario still open**: protocol-3 cluster with V2 persisted outputs → rolling binary upgrade → single 3→5 flip → V4 anchor + V4 reconfiguration output, byte-identical quorum throughout. The new boundary test covers testnet's v4→v5 shape; mainnet's compound start is the higher-stakes path and deserves its own gate.

With those three, the PR fully implements the agreed release path.

ycscaly:
@omersadika Thanks — both decisions are now implemented and validated, and one ask turned out to be already delivered. Point by point:

**Decision 1 (anchors migrate to V4 — no V3-forever) — implemented in `7fb7dbcde6`.** The anchor reconstruction gained the (V3 anchor, aggregated V4 reconfiguration output) arm: it rebuilds the canonical anchor with the **same construction as the proven V1/V2 migrations** (the anchor's reconfiguration-invariant class-group DKG output + the cert-pinned reconfiguration output) rather than a literal `upgrade(V3 anchor)` — deterministic either way, but this keeps one construction for every migration arm and the anchor's sharing data tracking the latest reshare, exactly like the V2→V3 migration did. It stays input-driven (fires when the reconfiguration output's format is ahead of the anchor's), so validators derive byte-identical bytes without consulting protocol config, on the same timing as the proven pattern. The mirror/handoff machinery needed no changes (bytes+digest, idempotent — verified against the #1855 wiring). Coverage: the boundary integration test now injects a testnet-shaped (V3 anchor, V3 reconfiguration output) pair and asserts the V3→V4 migration on the run's real bytes (green, 898s real crypto), and `v121_rollout` now asserts the canonical-anchor gauge reaches **4** after the flip (dispatched).

**Ask 1 (fixed-vector determinism) — implemented in `7fb7dbcde6`**, in ika (`integration_tests/fixed_vectors.rs`), covering **both starting states**: committed fixtures captured from one real 4-party run pin, byte-for-byte — (a) `upgrade()` of the pre-aggregation reconfiguration output (testnet's 4→5 flip; the fixture pair also makes the size win concrete: 229KB → 36KB), (b) the **V2-start compound path** (mainnet's 3→5 direct jump: V2 core's class-group output + aggregated reconfiguration output → exact V4 anchor), and (c) the V3-start anchor migration (which must land on the same pinned V4 anchor bytes, and does). All three run in ~3s; the `#[ignore]`d generator's doc explicitly warns that regenerating fixtures to fix a red pin defeats the tests. cryptography-private#623 does not pin vectors itself — these ika-level pins cover the actual persisted bytes, which is the quorum-relevant surface.

**Ask 2 (mainnet-shaped rollout scenario) — already delivered**, since your review predated the runs: `v118_upgrade` and `v118_churn` were strengthened into exactly this shape and are green ([run 29689667976](https://github.com/dwallet-labs/ika/actions/runs/29689667976)) — literal `mainnet-v1.1.8` fleet at protocol 3 with V2 outputs → rolling swap → **single 3→5 flip asserted** (`≥5`; v4 never activates) → V2→V4 canonical anchor + aggregated reconfiguration output asserted on every validator via the gauges (churn's mirrored joiner included) → full user lifecycles after. The mixed-window byte-identity half is `v118_mixed_rollout` (green on every push, per-authority digest convergence at v3). One note on your framing: the V2-start path needs no chained V2→V3→V4 — the reconstruction rebuilds V4 directly from the class-group output + the aggregated reconfiguration output, which is what those runs exercised.

Monitoring: yes — `ika_dwallet_mpc_network_encryption_key_latest_reconfiguration_output_version` is set at key instantiation next to the canonical-DKG gauge (0 = none installed, 3 = pre-aggregation, 4 = aggregated), and the canonical-DKG gauge now reports the reconstructed anchor's actual version (was hardcoded 3), so your version-distribution panel gets both axes of the flip.

Release placement is yours to call — the branch is self-contained and the feature stays dormant below a protocol-5 quorum either way.

https://claude.ai/code/session_015sMBcixNTaHLxR1vyeNvQW

omersadika:
Verified all three against the branch — **the checklist from the decisions comment is complete**:

1. ✅ Fixed vectors: `integration_tests/fixed_vectors.rs` + committed `.bcs` fixtures, all three pins present (reconfiguration `upgrade()`, the V2-start compound anchor, V3-start landing on identical anchor bytes), generator `#[ignore]`d with the anti-regeneration warning. Agreed that ika-level pins on the persisted bytes are the right surface — the quorum votes on these bytes, not on #623 internals.
2. ✅ Mainnet-shaped scenario: `v118_upgrade`/`v118_churn` assert the single 3→5 flip (v4 never activating) and per-validator V2→V4 anchors + aggregated outputs via the gauges; run 29689667976 green. Point taken on the framing correction — direct V4 reconstruction from the class-group output + aggregated reconfiguration output, no chained V2→V3→V4; cleaner and it's what the green runs exercised.
3. ✅ The anchor migration reusing the V1/V2 migration construction (rather than literal `upgrade(anchor)`) keeps one code path for every migration arm — accepted.

The canonical-DKG gauge now reporting the *actual* anchor version (was hardcoded 3) is exactly what our flip-epoch watchboard needed — the version-distribution panels + split alerts on the infra side (dwallet-labs/infra#95) now get true readings on both axes: mainnet's mixed window will show unanimous 2s, the flip moves both gauges to 4 fleet-wide, and any host left behind pages within a scrape interval.

Release placement per the operator decision stands: this is the headline of the v5-max release after testnet's v1.2.2 soaks — not in v1.2.2. From the monitoring/rollout side this PR is done pending final CI.

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: No, you must hardcode this to the nonaggregated struct, because in the future we will change the output type here

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
```diff
@@ -1644,12 +1645,26 @@ impl ProtocolCryptographicData {
...
+                        // Reconfig writes V3 (pre-aggregation shape) below the
+                        // aggregated-outputs protocol gate and V4 (aggregated
+                        // shape, via upgrade()) above it; the bwd-compat path
+                        // in `advance_network_reconfiguration_bwd_compat`
                         // writes V2.
-                        let public_output_value = bcs::to_bytes(
-                            &VersionedDecryptionKeyReconfigurationOutput::V3(public_output_value),
-                        )?;
+                        let public_output_value = if aggregated_network_key_public_outputs {
+                            let public_output: <twopc_mpc::decentralized_party::reconfiguration::Party as mpc::Party>::PublicOutput =
```
ycscaly: Again here and elsewhere, be precise about which type you are using instead of using the ambigious-soon-to-change related types out of the interface. This logic is only for the nonaggregated public output. Fix everywhere

crates/dwallet-mpc-centralized-party/src/lib.rs
ycscaly: Fixed in `bc1096f9e1` — hardcoded to `reconfiguration::NonAggregatedPublicOutput`.

crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations.rs
ycscaly: Fixed everywhere in `bc1096f9e1`: all five sites that decode the non-aggregated shape via the Party's associated output type now name `NonAggregatedPublicOutput` concretely — the centralized-party V3 arm, the instantiation V3 arm, the decrypt-shares V3 arm, the VSS-cache V3 decode, and the reconfiguration advance's decode-and-upgrade wrap (that one with a comment marking it for revisit when the Party migrates to the aggregated output, so the flip fails loudly there instead of misinterpreting). The `decentralized_party_backward_compatible` sites keep the associated type deliberately — that module is frozen v1.1.8-compat, so its output type is stable by definition. Verified no other main-party `as mpc::Party>::PublicOutput` decode remains.

PR #1863, opened 2026-07-19
fix(dwallet-mpc): preserve buffered messages when a NOA sign instantiates late

omersadika:
## What

Preserve buffered consensus messages when a network-owned-address (NOA) sign session instantiates late, and harden the late-instantiation path against a poisoned placeholder type and an already-resolved session.

`instantiate_network_owned_address_sign_session` ended in a bare `new_session(...)` that unconditionally overwrote any existing session entry. When a validator instantiates a NOA sign **late** — its network key or presign becomes available only after peers have already begun the signing round — the message-receipt path has already created a `WaitingForSessionRequest` placeholder and buffered peers' round-1 messages in it. The overwrite replaced that placeholder with an empty message buffer, dropping the buffered messages. Consensus rounds are delivered once and never redelivered, so the late validator was permanently stuck at round 1 with only its own message — below the signing threshold. The sign never assembled, and any NOA checkpoint gated on it (e.g. an epoch-close checkpoint) wedged the epoch.

The other two activation paths (`handle_mpc_request`, `try_activate_internal_presign_request`) already upgrade the placeholder in place with explicit "must survive" comments; the NOA sign path was written without that care.

## Fix

- **Upgrade the placeholder in place** (`set_status` / `set_request_metadata` / `set_protocol_name` / `set_request_diagnostic_metadata`), creating a new session only when no entry exists — composing the resolved-session short-circuit from `try_activate_internal_presign_request` with the computation-type normalization from `handle_mpc_request`.
- **Normalize a poisoned `Native` computation type** on upgrade. A pre-instantiation placeholder can be typed `Native` from a sender-controlled output report, and the NOA session identifier is predictable from public inputs — so without this guard a single byzantine consensus message would route instantiation to `try_new_native` (which errors for NOA sign) forever, on every honest validator, a permanent unattributed wedge. Reset a non-`MPC` placeholder type to a fresh `MPC` buffer before activating.
- **Short-circuit an already-resolved session** (not `WaitingForSessionRequest`), placed after the presign pop so pool consumption order stays network-uniform.

## Why this matters on `main`

The NOA settlement system is not live, so this code path is not exercised in production today — but the defect is in shared session machinery `main` already ships, and it is a correctness bug (the un-hardened upgrade would be a one-byzantine-message permanent epoch wedge). Fixing it now removes a latent landmine before NOA activation on any `main`-derived branch.

## Validation

- New regression test `test_late_instantiating_validator_preserves_buffered_sign_messages` — four validators, two instantiate only after the other two's round-1 messages are delivered; the sign completes only if the buffered messages survive. Fault-checked: reverting to the overwrite reproduces the exact production signature (`NetworkOwnedAddressSignOutput not received after 300 consensus rounds`).
- New regression test `test_instantiation_normalizes_byzantine_native_placeholder` — delivers a byzantine `Native` output report for the predictable session id before instantiation, asserts the placeholder is `Native` (attack precondition met), then asserts instantiation activates it as `MPC`. Fault-checked: removing the type-reset fails the assertion.
- Consensus-determinism reviewed: both branches yield a session built from consensus-uniform inputs; the local branch taken affects only whether this validator can participate, never a consensus-visible output.

Found and validated via the multichain-migration rehearsal, whose staggered rolling restarts are the natural trigger for the late-instantiation window.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/network_owned_address_sign.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/network_owned_address_sign.rs
@@ -13,7 +13,7 @@
 use crate::authority::authority_per_epoch_store::AuthorityPerEpochStoreTrait;
 use crate::dwallet_mpc::NetworkOwnedAddressSignRequest;
 use crate::dwallet_mpc::crytographic_computation::mpc_computations::network_owned_address_sign_dkg_emulation::network_owned_address_sign_dkg_session_identifier;
-use crate::dwallet_mpc::mpc_session::SessionStatus;
+use crate::dwallet_mpc::mpc_session::{SessionComputationType, SessionStatus};
 use crate::dwallet_mpc::integration_tests::network_dkg::{
     create_network_key_test, create_reconfigured_network_key_test,
 };
@@ -24,7 +24,10 @@ use crate::dwallet_mpc::integration_tests::utils::{
 use dwallet_mpc_types::dwallet_mpc::{
     DWalletCurve, DWalletHashScheme, DWalletSignatureAlgorithm,
 };
-use ika_types::messages_dwallet_mpc::{SessionIdentifier, SessionType};
+use ika_types::message::{DWalletCheckpointMessageKind, MakeDWalletUserSecretKeySharesPublicOutput};
+use ika_types::messages_dwallet_mpc::{
+    DWalletMPCOutput, DWalletMPCOutputReport, SessionIdentifier, SessionType,
+};
 use std::collections::HashSet;
 use itertools::Itertools;
 use tracing::info;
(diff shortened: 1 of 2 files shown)


omersadika:
Reviewed against the session-lifecycle machinery that merged this weekend (#1854/#1855/#1858). **LGTM — and I'd pull this into v1.2.2 ahead of the version bump.**

Verified:

- **The in-place upgrade preserves exactly what the bug dropped**: the `WaitingForSessionRequest` placeholder's `messages_by_consensus_round` buffer survives activation, matching the "must survive" discipline of `handle_mpc_request` / `try_activate_internal_presign_request`. The fallback `new_session` only runs when no entry exists.
- **The resolved-session short-circuit is correct and subtle in the right way** — not flipping a peer-completed session back to `Active`, while keeping the presign pop so pop-order stays network-uniform with peers that ran the session.
- **The Native-normalization guard is quietly a security fix**: the placeholder's computation type derives from a *sender-controlled* flag on a session id predictable from public inputs, so one byzantine output report could route every honest validator's NOA sign to `try_new_native` and wedge the epoch unattributably. Normalizing on upgrade discards the poison and, correctly, only resets the buffer in the poisoned case — a legitimate MPC placeholder keeps its messages.
- **#1858's `session_origin` interplay is handled by construction** — the `ReconstructedFromConsensus → LocalRequest` flip lives inside `set_status` (mpc_session.rs:656), so this path inherits it; late-activated NOA signs count and diagnose correctly.
- Both failure modes are pinned by tests (`test_late_instantiating_validator_preserves_buffered_sign_messages`, `test_instantiation_normalizes_byzantine_native_placeholder`). Non-blocking: a third case asserting the resolved-session short-circuit wouldn't hurt, though it mirrors the already-tested internal-presign path.

**Why it belongs in v1.2.2**: the #1855 recovery path *increases* late instantiations by design — a validator recovering from the network-key wedge instantiates its parked NOA sessions late, which is precisely the window this bug fires in. Shipping the recovery without this fix makes the message-loss race more likely on the exact validators v1.2.2 is rescuing. It's small, green, and self-contained; heavy suites are being dispatched on this branch now per the release bar, merge on green and then #1861.

PR #1865, opened 2026-07-19
fix(dwallet-mpc): assign NOA-sign presigns in consensus order, not local instantiation order

omersadika:
## What

Assign NOA (network-owned-address) sign presigns in **consensus order**, not local instantiation order.

NOA settlement signs consume presigns from a network-uniform internal pool via `pop_presign`, in **local instantiation order** (the order the service loop pulls requests off its channel). That order diverges across validators — staggered restarts, burst timing — so two validators pair **different** presigns with the **same** sign. The threshold decryption shares then can't combine: every validator aborts with `FailedToAdvanceMPC` / `InvalidParameters` at the sign's decryption round, and the sign never certifies. The pool itself is network-uniform (keyed by the consensus-assigned presign sequence); only the *consumption order* diverged.

## How

Announce each sign demand through consensus and assign each demand a presign in **consensus-delivery order**:

- A new `NOAPresignDemand` consensus message carries a network-uniform demand id (checkpoint TxRef + retry round, or a gRPC-attestation session id).
- On consensus delivery, every validator drains the demands in consensus order (an in-order `Vec::retain`, mirroring the existing global-presign drain) and **atomically** assigns each one a presign into a per-epoch `demand_id -> (session id, blending index, raw presign, network key id)` table. The pop and the record commit in a **single rocksdb batch** (`assign_noa_presign`, mirroring `assign_presign`), so a crash between them cannot let a re-drain (after a consensus-round replay) pop a different presign than peers assigned.
- Assignment is idempotent (an already-assigned demand returns without popping) and keeps a demand when the pool is momentarily empty, preserving order across rounds.
- The sign session instantiates by **reading** its presign and the key it was popped from out of that table — it never pops the pool locally, and it signs under the same key the presign came from (never a locally re-resolved key that could have shifted since announce).

**Determinism:** the assignment step reads only the consensus-delivered demand queue and the consensus-generated pool — no wall-clock, channel-arrival, or local-instantiation-order input. Consensus delivery order is a total order identical on every validator, so all validators pair the same presign with the same demand.

## Latent repair — inert until NOA turns on

NOA is **not live on any deployed network**. The entire NOA sign path — announce, submit, drain, assign, instantiate — is gated behind the `noa_checkpoints` protocol feature flag, which is **off in main's version schedule**. Validators that predate the `NOAPresignDemand` consensus kind never send or process it, so there is **no wire-compatibility hazard**. This change is a latent repair that becomes active only when `noa_checkpoints` enables NOA — the same posture as the message-loss fix in #1863.

## Tests

Because the fix is gated off, its tests would not exercise the new path on this base. The integration tests that drive a NOA sign enable `noa_checkpoints` through **scoped** test-config guards (`create_test_protocol_config_guard_with_noa_checkpoints`, and the flag set in `idle_status_voting`'s own guard) — turned on **only for the modules that exercise NOA sign**, not globally, since the broader NOA cluster on main is partially wired.

- **New:** `test_presign_assignment_is_consensus_ordered_not_local` sends two messages in divergent local order across four validators and asserts both signs complete **and** each demand's presign is identical across validators that received them in opposite order. Fault-validated: reverting the instantiate to the local `pop_presign` reproduces the exact wedge (`InvalidParameters` at the decryption round, no output in 300 rounds).
- The full `network_owned_address_sign` module (14 tests) passes single-threaded with the flag scoped on.
- The message-loss tests from #1863 were re-choreographed for the multi-round demand flow and re-fault-checked for non-vacuity: with the placeholder-upgrade reverted they still fail with the original signature, so the property they guard did not evaporate in the re-choreography.
- Blast-radius smoke: the remaining integration-test modules (70 tests) pass with the flag off — the gated path changes nothing for them.

## Behavior delta: idle voting during the demand window

Under the demand flow, a validator holding a pending-but-uninstantiated NOA sign request may vote **idle** for the few consensus rounds the demand takes (previously the prompt local pop made it non-idle immediately). The sole consumer of the idle majority vote is the opportunistic internal-presign pool top-up (`network_is_idle && pool < max` in `mpc_manager`), so the worst case is a bounded extra top-up during the window — no epoch-close, checkpoint, or session-lifecycle path reads the idle vote. Considered and accepted rather than compensated for.

## Note for maintainers (reconciliation)

Enabling `noa_checkpoints` makes NOA sign instantiation depend on the consensus-ordered flow, which takes several rounds (vs the old prompt local pop). Two modules drive a NOA sign as part of their setup and are re-choreographed to pump the demand flow to completion: `network_owned_address_sign` (including the #1863 message-loss tests) and `idle_status_voting` (its 2-2 split test now pumps until the split settles; the final assertions are unchanged). `internal_presign` only needed the new `demand_id` field on the request it constructs — no choreography change. The scoped-enable of `noa_checkpoints` is applied per-module, not globally. One store change rides along: `assign_noa_presign` marks the popped presign used, since the consumption point moved from instantiation to the assignment drain. Everything else is test-only.


omersadika:
Reviewed against the NOA family (#1863) and this weekend's consensus/wire-safety bar. **LGTM — this is the most rigorously validated PR of the set.**

What holds up:

- **The root cause is exactly right and the fix uses the proven pattern.** The pool was already network-uniform; only consumption order diverged (local instantiation order = channel-arrival order = nondeterministic across restarts/bursts). Announcing demands through consensus and draining in delivery order mirrors the existing global-presign drain — one total order, same presign↔sign pairing everywhere. Signing under the key the presign was popped from (never a re-resolved key) closes the adjacent divergence too.
- **The atomic pop+record batch is load-bearing and correctly identified as such**: `pop_presign` self-commits, so a separate record write would leave a crash window where a replayed drain pops a *different* presign than peers recorded. Single-batch `assign_noa_presign` + idempotent re-entry kills both the crash and the replay case.
- **No wire hazard, no protocol-version claim.** Everything — including the new `NOAPresignDemand` kind — sits behind the existing `noa_checkpoints` flag, off in every deployed schedule; the new kind inherits its NOA siblings' validation posture in both the validator and the handler routing. No collision with #1862's protocol-5 surface. Same latent-repair posture as #1863.
- **The test rigor is the standard the rest of the codebase should copy**: fault-validated in both directions (reverting the fix reproduces the exact `InvalidParameters`-at-decryption wedge; the #1863 message-loss tests were re-choreographed *and* re-fault-checked so the guarded property didn't evaporate), a real divergent-order 4-validator assertion, and a 70-test flag-off blast-radius smoke.
- The idle-vote delta analysis (bounded to an extra pool top-up; sole consumer identified) is the right way to document an accepted behavior change.

Two sequencing notes, neither about the code:

1. **Merge after the v1.2.2 tag is cut.** The release tree through #1863 is what the heavy suites validated; this PR is inert-by-flag but there's no reason the tag shouldn't point at the exact validated commit. Land this immediately after tagging.
2. Heavy suites are being dispatched on this branch now per the release-train bar; merge on green.

PR #1867, opened 2026-07-20
fix(reconfiguration): don't let never-announcing members force the 3/4-epoch freeze deadline every epoch

### LINKED ISSUE #1866: Freeze gate: carry forward latest validated MPC data; first-half window; don't let never-alive members force the 3/4-epoch deadline every epoch
## Summary

The MPC-data freeze gate re-requires a fresh announcement from **every** next-committee member **every** epoch, even though the announced blob is a deterministic, epoch-free function of the validator's root seed — and its liveness backstop (freeze at 3/4 epoch, silent members dropped) has quietly become the **every-epoch path** on testnet: two permanently-offline committee members make the happy path unreachable, so the network pays the full 6-hour courtesy delay each epoch, and a *continuing* member that misses the window loses a whole epoch of participation despite the network provably holding its validated data.

Three changes are proposed: carry forward the latest validated MPC data, move the announcement window to the first half of the epoch and freeze at mid-epoch + grace, and stop letting never-alive members extend any wait.

## Verified facts (v1.2.1 source + live testnet, epochs 355–357)

1. **The blob is epoch-invariant and deterministic.** `derive_mpc_data_blob(seed: &RootSeed)` (validator_metadata.rs:294) derives the full `ValidatorEncryptionKeysAndProofs` bundle from the root seed alone — no epoch input; determinism is unit-tested. A member's epoch-N blob is byte-identical to its epoch-N+1 blob absent key rotation (which is chain-visible).
2. **The gate waits for everyone, bounded by 3/4 epoch.** `decide_ready_to_finalize` (mpc_data_announcement_sender.rs) withholds ready signals until every V(e+1) member is locally validated, with the `ReadyViaDeadlineMissing` backstop at `epoch_start + 3/4 × duration`. Sound liveness — a silent validator delays but can never block (verified live: epoch 357's freeze fired within minutes of the 18.00h deadline).
3. **The backstop is now the every-epoch path.** Swyke and hereWeGo have been committee members for 3+ epochs while never announcing (zero-valued in the committee-uptime view throughout; committee 46–47 members). Epoch 357's freeze fired at exactly 18.06h elapsed; epoch 356's is bracketed to the same deadline (not fired at 14h elapsed, fully fired by 18.3h — and its happy path was impossible with those two in V(357)). **Prediction: every future epoch freezes at exactly the 18h mark until those two come online or leave the committee.** Reconfiguration + pricing + lock are thereby compressed into the last quarter of every epoch, permanently.
4. **A continuing member lost an epoch it didn't need to lose.** Synergy Nodes announced in epoch 356 and was hash-validated by the whole network (its pair is in 356's consensus-agreed frozen set). It was offline during 357's window, got dropped at the deadline, and is now excluded from the epoch-358 class-groups map / handoff cert — a full epoch of non-participation — even though its 357 blob would have been byte-identical to the data every validator already holds.

## Proposed fixes

**A. Carry forward the latest validated MPC data.** When building the ready signal, for each V(e+1) member with no fresh validated announcement, include the `(member, blob_hash)` pair from the **prior epoch's frozen set** — consensus-agreed state every validator holds identically, so the freeze tally stays a deterministic pure function of consensus data. Blob bytes remain fetchable by hash from the content-addressed store. Guard rotation by comparing the member's chain-registered key commitment; a rotated member needs a fresh announcement (correct). This alone fixes the Synergy class: transient outages stop costing epochs.

**B. First-half window; freeze at mid-epoch + grace.** Accept announcements from epoch start (candidates know their registration; the current gate makes joiners wait for V(e+1) publication unnecessarily), and fire the freeze shortly after mid-epoch committee publication (e.g., +1h grace) using latest-known data per member. Genuinely-new members get a ~12h first-appearance window — *more* generous than today's 6 — while reconfiguration gains the entire second half of the epoch as runway instead of the last quarter.

**C. Never-alive members extend nothing.** No fresh announcement AND no prior frozen entry → dropped at the grace boundary, with an operator warning naming them. With A+B+C: a dark member delays nothing, a transient outage costs nothing, and only a first-ever member missing a 12-hour window is dropped.

**D. Operational (independent of code):** Swyke and hereWeGo should be chased or removed via stake governance — until then the 18h freeze delay is arithmetic, not incident. Monitoring side (ours): an alert naming committee members missing announcements by mid-epoch is being added, since this leading indicator has now preceded two epoch-lifecycle investigations.

## Determinism note

A+B keep the freeze snapshot a pure function of consensus-ordered signals: the carried pairs come from the prior epoch's frozen set (identical on every validator), and the deadline only affects *when* each validator emits, exactly as today (per the existing comment in `send_epoch_ready_signal`).

_Filed from the monitoring side; timeline evidence gathered live during epochs 356–357 (freeze-gate wait, deadline firing, and the dropped-member consequences)._

omersadika:
Fixes #1866.

## What was happening

The ready-signal emit gate (`decide_ready_to_finalize`) required a fresh, locally-validated announcement from **every** next-epoch committee member, bounded only by the 3/4-epoch backstop. With two permanently-silent members seated on testnet, the happy path was unreachable, so every validator held its ready signal until the 18h mark — compressing freeze + reconfiguration + pricing + lock into the last quarter of **every** epoch (issue #1866 fact 3, verified live at epoch 357's 18.06h freeze).

## Reconciliation with the issue's proposals

- **Proposal A (carry forward validated MPC data)** is already live: `carry_forward_stable_mpc_data` (merged for #1742, present in v1.2.1) re-freezes any committee member with a prior-epoch handoff-cert digest, deterministically from consensus-anchored state. This means the issue's **fact 4 is most likely a misread**: Synergy Nodes was in epoch 356's frozen set → in 356's cert → carried into 357's frozen set. What the monitoring side saw was the deadline **warning** naming Synergy ("may be excluded"), which fired even for members carry-forward keeps. If the epoch-358 class-groups map truly lacks Synergy, that's a separate carry-forward bug worth its own investigation with node logs — please verify against the freeze log line (`carried_forward = N`).
- What was **not** fixed anywhere is the emit gate: it didn't know about carry-forward coverage, and its only deadline was the 3/4 mark. That's this PR.

## Changes (both emit-timing only — freeze determinism untouched)

1. **Carry-forward-aware gate**: a member with a digest in the prior epoch's handoff cert counts as covered — the freeze will keep it regardless, so waiting for its fresh announcement buys nothing. Only uncovered members (first-time joiner still propagating, or never-announced-in-any-epoch) hold the gate, and only they are named in the deadline warning (fixing the false exclusion alarm above).
2. **Publication-relative deadline** (proposal B's freeze-timing half + proposal C): once a validator first observes `V_{e+1}` published, the wait for uncovered members tightens from the fixed 3/4 backstop to `min(backstop, first-observed-publication + grace)`, `grace = clamp(epoch/24, 30s, 1h)`. The `min` guarantees the gate can only fire **earlier** than before; short test epochs keep the exact 3/4 behavior (the clamped grace overshoots them), so the 10–20s-epoch joiner cluster tests' freeze windows are unchanged.

Net effect on a 24h epoch with mid-epoch committee publication: freeze at ~13h instead of ~18h; a dark-but-previously-frozen member delays nothing; a never-alive member costs publication + ≤1h instead of 6h; a live first-time joiner still has a generous window (propagation takes minutes, plus quorum formation and the 50-round freeze grace on top of the deadline).

The gate's cert read degrades to "no coverage" on error (strictly the old, longer wait) — fail-loud stays a freeze-commit-only requirement, since the emit gate never influences signal content. Per-validator skew in observing publication (including restart resetting the anchor) only shifts when that validator emits.

**Deferred** (explicitly out of scope): proposal B's announce-from-epoch-start half (candidates announcing before `V_{e+1}` publication). It requires relayers to verify joiner signatures against the candidate registry instead of the published next committee — a trust-model change to the relay path that the above doesn't need to deliver the issue's outcomes. Proposal D (chasing/unstaking the two silent members, monitoring alert) is operational.

## Verification

- New unit tests: cert-covered member doesn't hold the gate and isn't reported missing; uncovered member blocks until the deadline; deadline formula across 10s/5min/24h/7-day epochs (backstop cap, grace floor and ceiling, unpublished-committee case).
- `cargo test --release -p ika-core mpc_data_announcement_sender` (6/6) and `validator_metadata` (79/79) green; clippy clean.
- Spec updated in the same PR: `dev-docs/specs/validator-mpc-data-announcements.md` gains the emit-gate rule.

Heavy suites (cluster + integration) should run on CI before merge; the joiner tests are the ones exercising this gate end-to-end.

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -3448,7 +3448,17 @@ impl AuthorityPerEpochStore {
     /// freeze-path realization of handoff.md invariant 4 ("fail open with
     /// retry on read errors"), and it mirrors the ready-signals read in
     /// `freeze_mpc_data_if_first`, which already `?`-propagates.
-    fn prior_epoch_mpc_data_digests(&self) -> IkaResult<HashMap<AuthorityName, [u8; 32]>> {
+    ///
+    /// Second consumer: the ready-signal emit gate
+    /// (`MpcDataAnnouncementSender::ready_to_finalize`) reads the same map
+    /// to know which members carry-forward covers, so they don't hold the
+    /// gate open. THAT read is emit-timing only, so the caller degrades an
+    /// `Err` to an empty map (wait longer) instead of propagating —
+    /// fail-loud is a freeze-commit requirement, not a property of this
+    /// function.
+    pub(crate) fn prior_epoch_mpc_data_digests(
+        &self,
+    ) -> IkaResult<HashMap<AuthorityName, [u8; 32]>> {
         let Some(prior_epoch) = self.epoch().checked_sub(1) else {
             return Ok(HashMap::new());
         };
--- a/crates/ika-core/src/epoch_tasks/mpc_data_announcement_sender.rs
+++ b/crates/ika-core/src/epoch_tasks/mpc_data_announcement_sender.rs
@@ -18,10 +18,12 @@
 //! 3. Once the announcement is confirmed AND local blob coverage
 //!    meets stake quorum AND `ready_to_finalize` holds (the
 //!    next-epoch committee is published and all its members are
-//!    locally validated, or the 3/4-epoch deadline elapsed), submits
-//!    an `EpochMpcDataReadySignal` (the first quorum of which freezes
-//!    the input set). Re-emits with an incremented `sequence_number`
-//!    as `validated_peers` grows, until `is_mpc_data_frozen()`.
+//!    covered — locally validated or carry-forward-covered by the
+//!    prior epoch's handoff cert — or the announcement deadline
+//!    elapsed), submits an `EpochMpcDataReadySignal` (the first
+//!    quorum of which freezes the input set). Re-emits with an
+//!    incremented `sequence_number` as `validated_peers` grows,
+//!    until `is_mpc_data_frozen()`.
 //!
 //! Without this task running, no validator would broadcast its
 //! mpc_data — leaving `frozen_validator_mpc_data_input_set` empty
@@ -54,20 +56,23 @@ use tracing::{debug, info, warn};
 /// Outcome of the ready-signal emit gate ([`decide_ready_to_finalize`]).
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum ReadyToFinalize {
-    /// Don't emit yet — keep waiting (V_{e+1} unpublished, or not all
-    /// of its members validated, and the deadline hasn't passed).
+    /// Don't emit yet — keep waiting (V_{e+1} unpublished, or some of
+    /// its members neither validated nor carry-forward-covered, and
+    /// the deadline hasn't passed).
     NotYet,
-    /// Emit: the next-epoch committee is published and every member's
-    /// blob is locally validated, so a freeze triggered by these
-    /// signals captures all of them.
+    /// Emit: the next-epoch committee is published and every member is
+    /// covered — its blob locally validated, or its digest present in
+    /// the prior epoch's handoff cert (the freeze carries such members
+    /// forward, so a freeze triggered by these signals loses no one).
     Ready,
     /// Emit because the epoch-clock deadline elapsed, but some
-    /// next-epoch members were NOT locally validated. They will be
-    /// excluded from this validator's `validated_peers` and risk
-    /// being dropped from the frozen set / next committee's
-    /// class-groups map — i.e. blob propagation is too slow for the
-    /// epoch length. The missing members are surfaced for an operator
-    /// warning + metric.
+    /// next-epoch members were neither locally validated nor covered
+    /// by the prior epoch's handoff cert. They will be absent from
+    /// this validator's `validated_peers`, the freeze cannot carry
+    /// them forward, and they risk being dropped from the frozen set /
+    /// next committee's class-groups map — i.e. a first-time joiner
+    /// whose blob never propagated (or a never-alive member). The
+    /// missing members are surfaced for an operator warning.
     ReadyViaDeadlineMissing(Vec<AuthorityName>),
 }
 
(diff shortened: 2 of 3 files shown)


omersadika:
Reviewed with the verification the PR asked for. **LGTM — and the reconciliation section is correct: issue #1866's fact 4 was my misread, confirmed against primary evidence.**

**The requested verification, done:** our validator's freeze log at the epoch-357 freeze (04:54:07Z — the 3/4 deadline to the minute) reads `frozen=44, attested=43, carried_forward=1` — Synergy Nodes was carried into the 358 frozen set by `carry_forward_stable_mpc_data` exactly as this PR's analysis predicted. I've also confirmed the carry-forward function is present in deployed v1.2.1 with the "frozen even once stays covered" chaining semantics. I'll correct #1866 accordingly; proposal A was already live, and the emit gate's unawareness of it — this PR — was the actual gap.

On the changes:

- **The covered-member filter is the right minimal fix**: `missing` excluding prior-cert members simultaneously un-holds the gate and fixes the false deadline warning that named carried members (which is what misled my issue's fact 4 in the first place — appreciated).
- **The publication-relative deadline is safe by construction**: `min(backstop, first_seen + grace)` can only fire earlier; the grace clamp keeps short test epochs on exact legacy behavior; the CAS'd first-seen anchor is idempotent and restart-shift only affects local emit timing.
- **Conservative error fallback** (cert read failure → empty coverage → the *longer* old wait) is the correct degradation direction for an emit-timing-only path.
- **Rolling-upgrade safe without a flag, and I checked the mixed-fleet case explicitly**: patched validators emit at ~publication+grace, v1.2.1 validators at the 3/4 backstop; the freeze still fires at the first consensus-ordered quorum of signals, and the snapshot remains a pure function of those signals — so a mixed fleet just freezes somewhere between ~14h and 18h depending on stake mix, deterministically. No wire changes, no protocol claim.
- Tests cover both pure functions including the case that matters (deadline warning names only the never-alive member).

Net effect to expect on testnet once deployed: freeze at ~14h instead of 18h every epoch (publication ~13h + 1h grace, with Swyke/hereWeGo as the never-covered holdouts), and the deadline warning finally naming only them. The monitoring side's announcements-incomplete alert stays correct as a leading indicator; I'm fixing its description to distinguish covered from uncovered members.

Deferring proposal B's announce-from-epoch-start half is reasonable — it's an optimization on top of a no-longer-painful baseline. Heavy suites dispatched on this branch per the release-train bar; merge on green (after the v1.2.2 tag, same sequencing as #1865).

omersadika:
Follow-up refinement filed as #1868: denominate the emit-gate deadline in ika consensus time (running max of `commit_timestamp_ms`, epoch-first-commit backstop anchor, quorum-crossing of the sequenced publication observation as the grace anchor). Fate-shared clock, zero new liveness dependency, eliminates the per-validator first-seen skew this PR's anchor still carries — but explicitly non-blocking for this PR; land as-is.

PR #1887, opened 2026-07-23
feat(sui-connector): default to the public Sui checkpoint archive on mainnet/testnet

ycscaly:
## What

When no `sui-checkpoint-archive` is configured, the verified OCS connector now defaults to the network's public Sui checkpoint store, resolved from the config's `sui-chain-identifier`:

- `mainnet` → `https://checkpoints.mainnet.sui.io`
- `testnet` → `https://checkpoints.testnet.sui.io`

These public stores retain the complete **end-of-epoch** checkpoint history (root `epochs.json` + `{seq}.binpb.zst` blobs) needed for committee-chain proofs, even though they don't promise retention of every ordinary checkpoint. With the default in place, the committee ratchet's archive-backed paths — cold bootstrap from genesis and the pruned-epoch-boundary fallback — work out of the box against pruning fullnodes, instead of requiring every operator to hand-configure the archive.

## How

A pure resolver in `ika-config` (`resolve_sui_checkpoint_archive(chain_identifier, configured)`), applied at the single point where the verified connector stack builds its archive (`build_sui_connector_stack`). The resolved config feeds the existing `OcsVerifyingClient::with_archive(...)` wiring, which already covers both cold bootstrap and the runtime pruned-boundary fallback.

Precedence and scope:

- An explicit `sui-checkpoint-archive` config wins **verbatim** (URL and options untouched) — the default is never merged into or substituted over it.
- No default is guessed for `Devnet`/`Custom` (localnet) chains; swarm/localnet configs use `Custom`, so tests see no behavior change.
- Legacy JSON-RPC configs are untouched: they never reach `build_sui_connector_stack` (it errors with `MissingDataSource` before archive resolution), and no unrelated config is mutated.
- Startup logs once which default archive URL was selected.

## Security model (unchanged)

- Genesis remains the trust root; the archive is an **untrusted availability source** only.
- `epochs.json` stays a location hint; every fetched end-of-epoch summary is BLS-verified against `committee[E]`, and the store's `next.epoch == head + 1` contiguity assertion still gates every install.
- `allow_unverified_committee_fallback` is untouched (default `false`).
- A malformed, forged, reordered, incomplete, or unreachable archive can stall bootstrap (the archive paths are best-effort and fall through with warnings) but can never install an unverified committee.

## Testing

- New unit tests in `ika-config`: explicit config wins on every chain, mainnet/testnet default to the public store, devnet/custom resolve to no archive.
- `cargo test --release -p ika-config` — 18 passed.
- `cargo test --release -p ika-core sui_connector::setup` / `sui_connector::ocs_verifier` — all passed (existing archive backfill / pruned-boundary coverage).
- `cargo test --release -p ika-sui-client archive` — passed.
- `cargo clippy --release -p ika-config -p ika-core --all-targets` — clean.

Spec updated in the same PR: `dev-docs/specs/ocs-verified-sui-reads.md` (archive resolution + staleness-bound sections).

https://claude.ai/code/session_01NnRjdvthoSVjPG5UtGtj57

### DIFF
--- a/crates/ika-config/src/node.rs
+++ b/crates/ika-config/src/node.rs
@@ -350,6 +350,41 @@ pub struct SuiCheckpointArchiveConfig {
     pub options: Vec<(String, String)>,
 }
 
+/// The public Sui checkpoint stores for the well-known networks. They retain
+/// the complete **end-of-epoch** checkpoint history (a root `epochs.json` plus
+/// `{seq}.binpb.zst` blobs) needed for committee-chain proofs, even though they
+/// don't promise retention of every ordinary checkpoint.
+pub const SUI_MAINNET_CHECKPOINT_ARCHIVE_URL: &str = "https://checkpoints.mainnet.sui.io";
+pub const SUI_TESTNET_CHECKPOINT_ARCHIVE_URL: &str = "https://checkpoints.testnet.sui.io";
+
+/// Resolve the end-of-epoch checkpoint archive the verified OCS connector
+/// stack should use. An explicit `sui-checkpoint-archive` config always wins,
+/// verbatim (URL and options untouched). With none configured, the known
+/// public chains fall back to their public Sui checkpoint store; there is no
+/// default to guess for `Devnet`/`Custom` (localnet), so those resolve to
+/// `None`.
+///
+/// Trust is unaffected either way: the archive is an untrusted availability
+/// source, and every checkpoint fetched from it is BLS-verified against the
+/// genesis-rooted committee chain (see [`SuiCheckpointArchiveConfig`]).
+pub fn resolve_sui_checkpoint_archive(
+    chain_identifier: SuiChainIdentifier,
+    configured: Option<&SuiCheckpointArchiveConfig>,
+) -> Option<SuiCheckpointArchiveConfig> {
+    if let Some(explicit) = configured {
+        return Some(explicit.clone());
+    }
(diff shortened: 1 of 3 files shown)


ycscaly:
Consensus-review pass (trust model, epoch-boundary liveness, legacy-path isolation): no findings requiring code changes. Two bounded-latency notes worth having in release notes for egress-filtered operators:

- On mainnet/testnet, a node whose egress to `checkpoints.{mainnet,testnet}.sui.io` is blackholed now pays a bounded delay before reaching the same outcome as before: up to ~11 s (`epochs.json` enumerate) plus up to ~60 s (checkpoint-fetch retry budget, from Sui's object-store util) per pruned epoch boundary, and once per process on the first ratchet (the one-shot cold-bootstrap attempt). Every failure warns and falls through — boot cannot fail and the ratchet cannot wedge because of the default archive, and the ordering strictly narrows use of the unverified fallback.
- The mainnet default is dormant in practice today: the OCS stack requires Sui protocol v122+ (`probe_artifacts_digest`), which mainnet hasn't reached — so the behavior change is effectively testnet-only until then.

Verified intact: both archive paths go through the same `install_next_from_verified_summary` chokepoint (head guard, BLS against `committee[head]`, `next.epoch == head + 1` contiguity, keyed by the committee's own epoch); a partial/forged store stalls but can never install a committee; legacy JSON-RPC configs never reach the resolver.

PR #1893, opened 2026-07-24
fix(notifier): stop loading validator root seed

omersadika:
## Summary

- validate node configuration and discard legacy root-seed descriptors before non-validator runtime initialization
- gate validator components, joiner announcements, and epoch promotion on the selected validator process role
- warn when a committee protocol key is running in a non-validator process
- keep root-seed-key-pair mandatory for ika-validator while preserving legacy notifier YAML parsing
- remove the fake generated notifier key placeholder

## Security rationale

Validator root-seed material is highly sensitive and is not needed by ika-notifier. Notifier workloads should never mount validator identity material merely to satisfy shared configuration validation. After this change, notifier startup does not open, parse, derive from, or otherwise use the configured validator root-seed file.

## Operator migration

Operators may remove root-seed-key-pair from notifier configuration and remove the root-seed.key file or secret mount from notifier workloads. Existing notifier configurations containing the field remain compatible; the field is ignored without reading the referenced file. Fullnodes also discard this validator-only field. Validator configurations missing it are rejected during role validation.

## Validation

- cargo fmt --all -- --check
- cargo test --release -p ika-config notifier
- cargo test --release -p ika-config fullnode_preparation_discards_root_seed_without_reading_it
- cargo test --release -p ika-config validator_config_without_root_seed_is_rejected
- cargo check --release -p ika-node --bin ika-notifier --bin ika-validator
- cargo clippy -p ika-config -p ika-node --all-targets --all-features -- -D warnings


omersadika (commented):
## Review

**Verdict: sound fix, minor suggestions.** Verified the central security claim against the code, checked all callers/modes (including swarm/test-cluster paths), and reviewed the epoch-boundary changes for consensus/liveness invariants. CI is fully green.

### Correctness — verified

- **"Discard without reading" holds structurally.** `RootSeedWithPath` is lazy: the seed lives in a `#[serde(skip)] OnceCell` and the file is only read on `root_seed()` access. Both entry paths (`node_runner.rs` and `start_with_mode`) sanitize before any consumer runs; the double invocation is idempotent and harmless. All four workspace consumers of the seed are validator-component paths post-PR.
- **The real pre-existing read path is closed.** Previously `monitor_joiner_announcements` was spawned unconditionally, so a notifier with a legacy config (validator config + notifier key) would read the seed file and run the expensive class-groups `derive_mpc_data_blob` derivation at startup. Now gated on process role, and the seed is discarded besides — defense in depth.
- **Mode gates don't wedge legitimate joiners.** A joiner must run Validator mode (it needs `consensus_config`, which is exactly what `detect_from_config` keys on), so the joiner monitor and promotion gates pass for it. The ungated `state.is_validator` sites at `lib.rs:2897/2907` are inside the was-already-validator branch, unreachable for notifiers/fullnodes since both producers of `validator_components` are now mode-gated. A Fullnode-mode node could never promote successfully pre-PR anyway (`construct_validator_components` errors on missing consensus config and kills the reconfiguration loop) — the gate strictly improves that path.
- **Swarm/tests unaffected.** Swarm validators set both `consensus_config` and the seed (`node_config_builder.rs:189`), so auto-detection yields Validator and the new requirement passes; the fullnode builder sets neither.
- **Template removal is safe.** `NodeConfig` has no `deny_unknown_fields` (only `SuiDataSource` does), so `class-groups-key-pair-and-proof` in `fullnode.template.yaml` was silently-ignored dead YAML for a field that no longer exists.
- **No mixed-version quorum risk.** Notifiers have no consensus adapter; the only seed path they could reach was the P2P joiner fanout, whose blob is a pure function of the seed and byte-identical to the real validator's own announcement — dropping the duplicate changes nothing consensus-visible.

### Findings

1. **Silent demotion (low, worth fixing).** The new gates turn a loud boot failure into a silently-degraded node when the process role and committee state disagree — e.g., an operator's systemd unit pointing at `ika-fullnode` after a migration boots cleanly, syncs, and looks healthy while its stake is absent from every MPC quorum and checkpoint signature set. Pre-PR the same misconfiguration crashed at `construct_validator_components`. Suggest a `warn!` at both gate sites (`lib.rs:1196` and the promotion branch) when `!mode.is_validator() && state.is_validator(..)`: "this node's protocol key is in the committee but it is not running in validator mode".
2. **Inconsistent discard scope (minor).** `validate_and_prepare_config` discards the seed only for Notifier mode. By the PR's own rationale ("key material the selected role must never use"), Fullnode mode should discard it too. No read path remains after the gating, so this is consistency rather than a hole.
3. **Now-dead guard (nit).** The `let Some(root_seed_kp) ... else { return }` at `lib.rs:1353` is dead for the only mode that still spawns the monitor (Validator mode now guarantees the seed at boot). Fine as defense in depth, but its comment ("we can't be a joiner — nothing to do") describes a state that can no longer occur.
4. **Ops-visible tightening (flag in release notes).** An `ika-validator` without a root seed now fails at startup instead of panicking later on committee entry. Strictly better, but it's a deliberate behavior change the migration doc could mention alongside the notifier guidance.
5. **Test maintenance tax (nit).** `config_for_mode` hand-builds the full `NodeConfig` literal, so every future field addition breaks these tests. Also, `prepared_notifier_config_omits_root_seed_when_serialized` is essentially implied by the discard test plus `skip_serializing_if` — harmless redundancy.

### Conventions

Style follows the codebase (kebab-case config, doc comments explaining constraints, tests in-module). No spec conflict: `dev-docs/specs/validator-mpc-data-announcements.md` doesn't pin process-level spawn conditions, so no spec update is required. Docs page and migration guidance are clear; `sidebar_position: 3` in the new mdx is redundant since `meta.json` orders pages explicitly — trivial.

Issue #1892, opened 2026-07-24
Epoch close has no resilience to a stalled checkpoint writer: the notifier is a single point of failure that blocks EndOfPublish network-wide, indefinitely and near-silently


omersadika:
## Class, not instance

Three different checkpoint-writer outages in one week — gas exhaustion (07-19), sui-state-direct verification bootstrap failure (#1882, 07-23 and 07-24), and the sync-side stall described below — each produced or would produce the same network-level outcome: **the epoch cannot close, ever, until a human intervenes**. The instances have their own fixes (gas top-up, config revert); this issue is about the class.

## Observed live (twice on testnet, near-miss on mainnet)

- 2026-07-23: testnet epoch 360 overdue ~30m+, `user_sessions_lag` asserted on all 40 validators. The final user-session completions were certified in dwallet checkpoints that never landed on Sui. Resolved by restarting the notifier.
- 2026-07-24: identical recurrence at epoch 361 (overdue 1,700s+ when caught).
- Same day: the **mainnet** notifier showed the same stall signature (known−synced gap 12,276, growing) ~21h before mainnet's first v1.2.5 epoch close — caught by an infra alert added after the second testnet incident, fixed before the boundary.

## Mechanism (code-verified, v1.2.5)

1. **Single writer by design**: dwallet/system checkpoint submission runs only where `sui_notifier: Some(SuiNotifier)` — one funded Sui address on the notifier node (`sui_executor.rs:60` "Submission state for the notifier's single Sui address"). Validators embed the identical submission machinery (`sui_connector` write metrics exist on every validator, permanently zero) but are role-gated to `None`.
2. **The close gate is chain-side**: `all_epoch_sessions_finished` compares the coordinator's on-chain `completed_sessions_count` against the locked close target (`sui_syncer.rs` ~1320). Completions only reach the chain through the sole writer's checkpoints. Writer stalled ⇒ every validator asserts `user_sessions_lag` ⇒ no EndOfPublish ⇒ no close. There is no timeout, no fallback, no escalation.
3. **The failure is near-silent in-protocol**: the only defense is an error log after 1h of *failed submissions* (`sui_executor.rs:786`). In both incidents it never fired — the stall was on the checkpoint-**sync** side (synced=0 while known=21k), so there was nothing to submit and therefore no submission failures. From the binary's own telemetry, a fully-stalled writer is indistinguishable from an idle one.
[...]

PR #1894, opened 2026-07-24
fix(#1892): make a stalled checkpoint writer loud and correctly attributed (writer-lag blocked-reason, sync-stall self-report)

omersadika:
Addresses #1892 under the reviewed design constraint: **the network keeps exactly one transaction-publishing node (the notifier); validators never submit.** The fallback-writer mechanism from earlier revisions was removed on review (see comments); what ships is the observability pair that makes a writer stall minutes-to-detect and correctly attributed — both July incidents were near-silent in-protocol and lost their first debugging phase to `user_sessions_lag` ambiguity.

## B. The close gate names the writer (validators, read-only)

`sync_dwallet_end_of_publish` reads the local certified dwallet-checkpoint head and exports:
- `ika_sui_connector_chain_dwallet_checkpoint_writer_lag` — highest locally-certified checkpoint − the coordinator's on-chain cursor;
- a new `checkpoint_writer_lag` label on the EndOfPublish blocked-reason gauge — set when sessions are unfinished on chain **while certified checkpoints sit unlanded**, lighting up alongside `user_sessions_lag` and naming the writer directly. Also carried in the gate-stuck WARN.

Observability only: validators submit nothing; the gate's firing conditions are unchanged.

## C. The writer self-reports a sync stall (the nothing-to-submit blind spot)

State sync (both checkpoint streams) detects "same-chain peers advertise checkpoints ahead of us and our synced watermark is not advancing" (120s grace), exports `ika_dwallet_checkpoint_sync_stall_seconds` / `ika_system_checkpoint_sync_stall_seconds` (0 = healthy), and errors at 1/min. This covers the exact incident shape (`synced=0`, `known=21k`): nothing reaches the submission path, so the pre-existing 1h submission-failure log can never fire. The `highest_known_*` gauges now refresh every 5s tick from peer heights instead of only when a sync job runs — which is exactly what stops happening during this stall.

## Docs (same PR, per dev-docs policy)

- `dev-docs/specs/checkpoint-writer-observability.md` — the deliberate single-writer design, the two stall shapes, and their signals; indexed in `dev-docs/README.md`.
- `dev-docs/playbooks/mpc-stall-postmortem.md` — new diagnosis class: `all_epoch_sessions_finished=false` with `checkpoint_writer_lag > 0` ⇒ fix the notifier, not the MPC pipeline.

## Out of scope, by direction

Writer redundancy. The liveness gap remains operational (restart the notifier). The forward-looking option preserving a single writer *identity* is SIP-58 address-balance gas (`ValidDuring`): byte-identical, deterministically-constructed submissions from replicated notifier processes collapse to one tx digest — active-active redundancy without a second address or any takeover logic — and it would delete the gas-object caching machinery outright. Left for a follow-up once address-balance gas is enabled on the target networks.

## Validation

- `SyncStallTracker` unit tests (grace, rate-limit, recovery, never-synced bootstrap shape): 6/6.
- Full cluster suite ran green (22/22 pre-existing tests) on an earlier revision whose diff was a strict superset of this one; a fresh run on this exact head is linked in comments.
- `cargo clippy --all-targets` clean on touched crates; `cargo fmt --all`.


omersadika:
Cluster suite (full, includes the new `epoch_close_survives_notifier_death_via_fallback_writer`) dispatched on this branch: https://github.com/dwallet-labs/ika/actions/runs/30109710608

omersadika:
**Cluster suite result (run 30109710608):** 22/23 passed — the only failure was the new `notifier_failover` test itself, i.e. no regressions across the existing cluster suite.

The test's failure was an epoch-numbering bug in the test, not in the failover: with 30s epochs and a multi-minute boot the cluster is far past epoch 2 by phase 3, so `wait_for_epoch(2)` returned instantly, no takeover was ever needed, and the counter assertion (correctly) flagged that nothing had been exercised. Fixed in ff5bf615: phase 3 now waits for `epoch_at_kill + 2` — two closes guarantee one full epoch cycle (mid-epoch → reconfig → pricing → lock → advance + all checkpoints) happened entirely after the primary's death.

Re-dispatched:
- fixed test (must pass): https://github.com/dwallet-labs/ika/actions/runs/30111285527
- test-testing fault injection, fallback removed (must FAIL, proving the test catches the pre-#1892 outage shape): https://github.com/dwallet-labs/ika/actions/runs/30111334510

omersadika:
**Validation complete — the regression test is proven in both directions:**

- ✅ **Must-pass** (fallback configured, notifier killed): run 30113080474 — `epoch_close_survives_notifier_death_via_fallback_writer` **PASSED** in 581s, first try. Two full epoch closes completed after the notifier's death, and the in-test assertion confirmed the takeover counter went positive (only the fallback-key validator can increment it).
- ❌ **Must-fail** (test-testing fault injection: fallback removed, everything else identical): run 30111334510 — **FAILED on both tries with exactly the expected evidence**: `epoch did not close after the notifier died — the fallback writer never took over: Elapsed(())` after the 900s ceiling. That is the pre-#1892 outage shape, reproduced and caught. Fault branch deleted.
- Full cluster suite on the branch (run 30109710608): 22/22 pre-existing tests passed — no regressions from the writer changes.

Branch history was rewritten for commit authorship only (`--reset-author`, trees identical to the SHAs CI ran on): `22e65e5c → 1a69b7c1`.

omersadika:
Reworked per review direction (single notifier; validators never publish): the fallback-writer mechanism, its config surface, swarm plumbing, and the failover test are removed in 129f038d — the branch now carries only the observability changes (writer-lag attribution on validators + the notifier's sync-stall self-report). Net diff vs main is now +249/−170ish lines across sui_syncer/state_sync/metrics/docs.

The earlier CI evidence still stands as an engineering data point (the failover design passed its must-pass/must-fail pair), but it's no longer part of this PR's scope. Fresh full cluster suite on the reworked head: https://github.com/dwallet-labs/ika/actions/runs/30116288637

omersadika:
Full cluster suite is green on the reworked single-notifier head (129f038d): https://github.com/dwallet-labs/ika/actions/runs/30116288637 — all 22 tests passed. PR is ready for review.

omersadika:
Correction found while root-causing the cold-start bug in #1896: the stall tracker was watching the SYNCED watermark, which is fed only by the consensus output path — on a notifier/fullnode it never advances even when pull sync is perfectly healthy, so the detector would have reported a permanent false stall on exactly the node it protects. Switched to the VERIFIED watermark (bumped by every pull-path insert). Spec updated. This also retroactively explains the incident telemetry's `synced=0`: that gauge is structurally 0 on notifiers.

omersadika:
Revalidation suite green on the final head (bb337855, includes the verified-watermark tracker correction): https://github.com/dwallet-labs/ika/actions/runs/30130746855 — this PR is merge-ready.

PR #1895, opened 2026-07-24
chore!: remove v3→v4 migration scaffolding and drop protocol v3/v4 support (MIN=5)

### LINKED ISSUE #1751: Remove v3→v4 migration scaffolding once all network keys are reconfigured under v4
## Context
v4 (`off_chain_validator_metadata_enabled`) moved network-key DKG/reconfiguration blobs off chain into the handoff plane. During the v3→v4 rolling upgrade, temporary scaffolding bridges keys whose DKG or last reconfiguration ran under v3 — their authoritative blobs exist only on chain. Once every network key has reconfigured under v4, this scaffolding is dead code and must be removed. Currently tracked only by `TODO(v3->v4 migration)` comments; this issue is the external anchor.

## Trigger condition
Every network key is in the off-chain handoff plane: the syncer no longer takes the chain-read fallback for any key (`dkg_in_handoff` true for all; the `key_blobs_already_cached`-false branch is never hit).

## Cleanup set (coupled — remove together)
Greppable: `grep -rn "TODO(v3->v4 migration)" crates`
1. `crates/ika-core/src/sui_connector/sui_syncer.rs:819` + `:852` — delete the `key_blobs_already_cached` blob-read branch; collapse `chain_fetched` to the unconditional `off_chain_on` synthesize-empty fast path (anchor removal).
2. `crates/ika-core/src/dwallet_mpc/mpc_manager.rs:966` — tighten cert-less adoption of a reconfigured key to reject rather than blindly adopt from chain.
3. `crates/ika-core/src/dwallet_mpc/mpc_manager.rs:973` — remove the transiently-empty-overlay downgrade guard.
4. `crates/ika-core/src/dwallet_mpc/mpc_manager.rs:2081` — remove the `off_chain_validator_metadata_enabled()` gate on DKG mirroring (always mirror).

## NOT in scope (permanent guards, untagged)
- `crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs:941` — internal-presign history-replay guard.
- `crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/reconfiguration.rs:79` — fail-loud partial-map check in the v3-callable reconfig path.

## Suggested hardening
Add a log/metric when the syncer stops taking the chain-read fallback for the last pre-v4 key, so "migration complete → clean up" is a signal, not a memory item.

## Related
`dev-docs/plans/handoff-sui-decoupling.md`.

omersadika:
Closes #1751. With both deployed networks **verified on-chain at protocol v5** (mainnet: epoch 360, v5, no pending upgrade; testnet: epoch 363, v5, no pending upgrade — via publicnode RPC), `MIN_PROTOCOL_VERSION` moves to **5** and everything the binary kept only to run v3/v4 is removed — then the upgrade-test suite is rebuilt around the deployed v1.2.5 release.

## Commits 1–2 — issue #1751: the v3→v4 migration scaffolding

The four coupled `TODO(v3->v4 migration)` sites: the `dkg_in_handoff` chain-read fallback in the syncer (stranded-key recovery from #1852 preserved), cert-less adoption of a reconfigured key tightened to **reject** instead of blindly adopting from chain, the transiently-empty-overlay downgrade guard, and the v4 gate on DKG mirroring. A consensus-review pass on this commit produced the commit-2 keep-guard, later removed again when v3 stopped existing entirely.

## Commits 3–4 — protocol v3/v4 support removal (`MIN_PROTOCOL_VERSION = 5`)

- **ika-protocol-config**: MIN 3→5. The per-version history arms (`for cur in 2..=version.0`) are **kept** as readable history; the three v5 snapshots are untouched and their passing pins that the config is byte-identical (pre-v5 snapshots deleted).
- **ika-core network-key flows**: the backward-compatible (class-groups-only) DKG/reconfiguration parties, their `PublicInput`/`ProtocolCryptographicData` variants and dispatch arms, and pre-aggregation (V3-tagged) output **production** removed — reconfiguration outputs are always aggregated (V4). Off-chain-metadata gates inside the network-key flows collapse to unconditional.
- **Persisted-state compat retained**: `Versioned*` enum variants keep their BCS indices; V1/V2/V3 **decode** paths and the anchor-reconstruction machinery stay (deployed networks persist pre-v5 state); old validator-key-shape reading untouched. The retained decode surface is pinned by `fixed_vectors.rs` against committed fixtures.
- **Tests**: `network_dkg_bwd_compat` (subject = the deleted v2-crypto path) and `protocol_version_transition` (no boundary below MAX until v6) deleted — resurrection SHAs recorded in #1909; v4-pinned cluster tests repinned to `ProtocolVersion::MAX`; the fixed-vector generator deleted (fixtures are frozen historical state, pinned tests stay).

## Commits 5–7 — the v1.2.5 upgrade-test suite (successor to the retired rehearsals)

The v3/v4-era scenarios (`cross_binary`, `malicious_cross_binary`, `v118_upgrade`, `v118_churn`, `v118_mixed_rollout`, `v121_rollout`) cannot boot against this binary and rehearsed completed rollouts. Their still-relevant coverage is **ported, not dropped**:

- **`v125_rollout`** — the deployed-release gate and the **PR/workflow default scenario**: literal `release/mainnet-v1.2.5` committee at genesis v5, one validator upgraded to current, two mixed reshares with per-authority byte-equality + zero-malicious gates, then full swap + convergence. A pure binary swap — no protocol transition.
- **`v125_churn`** — full sequential swap (on-disk state continuity), a peer-only-mirrored OCS joiner folded into the reshared v1.2.5-origin key (4→5), MPC timing snapshot, and a shrink reshare (5→4); direct/mirrored read-topology split.
- **`malicious_v125`** — test-testing counterpart: honest v1.2.5 committee + one faulty current validator (the `test-testing` fault hook is **re-homed** into the main reconfiguration path — it silently died with the bwd-compat party); detection asserted via the malicious-actors gauge. CI-dispatchable (the workflow builds the faulty binary).
- **`upgrade-test.yaml` retains and rewires the OLD-binary build machinery** (worktree build of `old_ref`, defaults `release/mainnet-v1.2.5`/`ika-validator`, real-crypto guards, `test_testing_fault` input for expected-fail gate validation).

## Verification

- **CI green on this branch**: cluster suite (×3 runs), Rust integration suite (real crypto), standard CI/clippy/fmt, `workload` end-to-end, **`v125_rollout` full pass** (literal v1.2.5 ↔ current convergence), **`malicious_v125` full pass** (detection fires — the gates are not vacuous). `v125_churn` + TS integration + rebased-head re-runs in flight at time of writing; merge on current-head green per review.
- Local: protocol-config snapshot tests (value-preserving fold proven), targeted real-crypto ika-core tests 10/10, clippy clean.

## Follow-ups — #1909
Anchor-reconstruction removal (gauge-gated), always-true flag sweep, v6 boundary-test resurrection (SHAs recorded), BLS-key cleanup (#1751 thread).

https://claude.ai/code/session_01RmR92cFA9bz3MPfb9z6urf


omersadika (commented):
## Review: remove v3→v4 scaffolding, drop protocol v3/v4 (`MIN=5`)

**Verdict: looks good to merge, with two things to confirm first.** The
load-bearing premise checks out, the riskiest claim is pinned by a test rather
than asserted, and upgrade coverage is retargeted rather than dropped.

### The premise — verified independently

Everything here rests on "mainnet and testnet both run protocol v5". I queried
the chain rather than take it on faith:

```
MAINNET  epoch = 360   protocol_version = 5   next_protocol_version = None
```

Mainnet — the network that was furthest behind at v3 — **is on v5 with no
pending upgrade**. That is the strict constraint and it holds.

⚠️ **Testnet unverified**: the public testnet RPC didn't respond for me.
Testnet was on v4 and upgrades ahead of mainnet, so it is almost certainly ≥5
— but "almost certainly" isn't the bar for a change that makes the binary
unable to join a sub-v5 network. Worth a 10-second confirmation before merge.

### What's done well

**The value-preserving fold is proven, not asserted.** Folding the v3/v4 arms
into the base config is the change most likely to silently alter a config
value. Deleting the v1/v3/v4 snapshots while leaving all three v5 snapshots
**untouched** is exactly right — the unchanged v5 snapshot passing *is* the
proof. Confirmed no `version_5` snapshot appears in the diff.

**Coverage was retargeted, not deleted.** (My first pass misread this because
I stripped rename targets out of the file list — flagging in case anyone else
skims the same way.)

| Retired | Successor |
|---|---|
| `v121_rollout` | → `v125_rollout` (retargeted to `release/mainnet-v1.2.5`) |
| `malicious_cross_binary` | → `malicious_v125` |
| `cross_binary`, `v118_*` | → `v125_churn` (new) |

`v125_churn`'s doc comment explicitly picks up what `cross_binary` uniquely
covered — the 5→4 reshare-to-fewer-parties direction and the mixed
direct/mirrored OCS topology. That's careful, not deletion by convenience.

**The security posture improves.** A reconfigured key with no prior handoff
cert is now **rejected** (warn-once, retried) instead of blindly adopted from
chain, closing the one documented bypass of the cert-digest gate.

**Persisted-state compat is correctly retained** — `Versioned*` variants keep
their BCS indices and the decode paths stay. The distinction between "stop
*producing* V3" and "keep *decoding* V3" is the right one, and the follow-up is
properly gated on the fleet-wide version gauge rather than done speculatively.

### Findings

**1. The PR description materially misdescribes the change.** It says
*"upgrade-test.yaml drops the OLD-binary build machinery and its inputs; the
PR/release default scenario is now `workload`"*. The workflow actually defaults
to **`v125_rollout`** and **retains** the OLD-binary machinery — the
description appears to describe an earlier state of the branch. It understates
the work in the direction that makes a reviewer think mixed-binary coverage was
abandoned. Worth updating, since this is what reviewers and future
archaeologists read.

**2. Dangling reference to a deleted file.**
`crates/ika-upgrade-test/src/bin/upgrade-test.rs:73` still says *"Mirror the
proven-good config from `tests/cross_binary.rs`"* — that file no longer exists.
One-line fix. (The other `v118_*` / `cross_binary` mentions I checked are
deliberate historical prose — "the successor to the retired X" — which are fine
and worth keeping.)

**3. Raising `MIN` turns stale local state into a boot panic.**
`get_for_version` asserts `version >= MIN`, and the single production caller
reads the **current** epoch's version from `epoch_start_configuration`, so
there is no historical-config reconstruction and the blast radius is bounded.
But a validator restored from a pre-v5 backup, or one offline since before
activation, will panic at boot with:

> `Network protocol version is 4, but the minimum supported version by the binary is 5. Please upgrade the binary.`

That message is exactly backwards for this case — the operator needs to catch
up chain state or re-sync, not upgrade the binary. Worth a release-note line,
or an assert message that names both directions.

**4. The green `v125_rollout` is on a stale SHA.** It passed on `812544de9b`;
PR head is `609f83e1aa`, whose Upgrade Test is still in flight. Don't merge on
the old green.

### Minor

- `network_dkg_bwd_compat.rs` (−1,313) is the largest single deletion. Its
  subject is the deleted v2-crypto path so removal is right — worth a
  confirmation that nothing in it was incidentally covering the **retained**
  decode paths.
- Deleting `protocol_version_transition` leaves no upgrade-boundary cluster
  coverage until v6. The reasoning (no boundary exists below MAX) is sound and
  the PR says to resurrect from history — please record the commit SHA in the
  follow-up issue so that's a lookup rather than an archaeology exercise.

### Risk summary

The irreversible part — a binary that cannot join a sub-v5 network — is
justified by verified on-chain state. The reversible parts (deleted tests,
folded config) are either retargeted or pinned by a passing snapshot. The main
residual risk is operational (finding 3), not protocol.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

omersadika:
Re: review — all findings addressed:

1. **PR description** — rewritten; it described a pre-`v125` state of the branch (the workflow retains the OLD-binary machinery and defaults to `v125_rollout`, as you noted).
2. **Dangling `cross_binary.rs` reference** in `upgrade-test.rs` — fixed in `e9d9c4456d`.
3. **Backwards panic message** — the below-MIN assert now names both directions (stale local state → catch up/re-sync vs. genuinely older network → different binary), same commit.
4. **Stale-SHA green** — agreed; the current-head `v125_rollout` run is queued/in-flight and merge waits for it.

⚠️ **Testnet — verified**: `protocol_version = 5`, `next_protocol_version = None` at epoch 363, queried via `sui-testnet-rpc.publicnode.com` (the default fullnode endpoint is unresponsive, matching your experience). Object path: `ikaSystemObject 0x2172…dec6` → dynamic field → `SystemInner`.

Minor items: the retained decode surface (V1/V2/V3 anchor reconstruction + reconfiguration-output upgrade) is pinned by `fixed_vectors.rs` against committed fixtures; the one decode arm `network_dkg_bwd_compat` covered incidentally (instantiation from a **V2-tagged current** reconfiguration output) is unreachable on any v5 network — every live key's current output is V4 — and fails loudly if ever hit. Resurrection SHAs for `protocol_version_transition` (both flavors) are recorded in follow-up #1909.

omersadika:
**`v125_churn`'s first run caught a real #1751 regression** — worth the scenario's existence on its own.

The mirrored joiner never instantiated the v1.2.5-origin network key (canonical-version gauge stuck at 0): a joiner holds nothing for a key DKG'd before it existed, its producer cache can never fill, and the cert-pinned blob install (barrier) covers only continuing validators — while adoption skipped the empty overlay entry *before* the #1852 stranded-key detection could fire. Pre-#1751, the removed chain-read fallback covered this shape implicitly (`dkg_in_handoff == false` is a joiner's initial state, not just a pre-v4 key's).

Fix (`882e945cde`): a second trigger on the **existing** stranded-key recovery — an empty overlay entry for a key DKG'd in a *prior* epoch, on a validator with nothing instantiated/in-flight/adopted for it, flags the key for the syncer's chain-sourced read; the cert digest gates still verify the returned blobs, confirmed instantiation un-flags, keys DKG'd this epoch are excluded (fresh-key bootstrap window), and restarting continuing validators never hit it (the barrier fills their cache first). Regression-pinned in `network_key_adoption::empty_overlay_for_prior_epoch_key_flags_stranded`; `handoff.md`'s stranded-recovery contract now names both triggers. `v125_churn` re-dispatched on the fix.

`malicious_v125` passed on its first run, and testnet is verified at v5 (see previous comment) — so both of the review's pre-merge confirmations are done; merge now additionally waits on the churn re-run.

Issue #1882, opened 2026-07-22
sui_state_direct cannot bootstrap committee verification: MissingCommittee retry-forever on validator boot (needs historical committee/checkpoint serving or a persisted verification anchor)


omersadika:
## Severity: high — any validator booting on `sui_state_direct` never brings up its Sui connector; its entire MPC stack stays dead

Live occurrence 2026-07-22: dWallet Labs 1, n1stake, and Full Sail Validator (all v1.2.3 `194a9db7fe`, all switched to `sui_state_direct`) restarted mid-epoch-360 and show `canonical_dkg_output_version=0`, no `network_key_loaded_epoch` series, no advances — while ika-side consensus keeps working (`ika_current_epoch=360`, `announcements_received=42`).

## Log signature (from dWallet Labs 1, repeating every 30s since boot — 602 attempts over ~5h)

```
WARN ika_core::sui_connector::sui_executor (sui_executor.rs:215)
  "verified_system_inner still failing — likely a pruned-and-uncached anchor
   (Sui-fullnode retention gap); raise fullnode retention or re-anchor"
  error=MissingCommittee(1168) attempts=602 backoff_secs=30

WARN ika_node (lib.rs:926)
  "Sui committee ratchet failed; will retry"
  error=Transport(Network("code: 'Some requested entity was not found', message: \"Epoch 0 not found\""))
```

## Mechanism

`verified_system_inner` needs the Sui committee for the current Sui epoch (1168) to verify system-object reads. The committee ratchet walks committees forward from an anchor; on this source the walk fails twice over:
- the ratchet's cold-start read asks for epoch 0 (genesis committee) — the state-direct source answers "Epoch 0 not found";
- the incremental path needs prior-epoch end-of-epoch checkpoint data to step to 1168 — state_direct serves current state + a change stream, not historical epochs.

So a **fresh process** on state_direct can never verify anything: no verified system object → the syncer's `system_object_receiver` stays `None` → key delivery, canonical-mirror hydration, and epoch-start MPC data never happen. A long-running process that verified before the switch keeps working from its in-memory ratchet — the failure is boot-only, which is why this looks exactly like a wedge and only appears after a restart. Note the OCS pusher itself is fine (`ika_ocs_chain_latest_epoch=1168`, current, on all three) — the gap is verification bootstrap, not the data feed.

[...]

PR #1896, opened 2026-07-24
fix(state-sync): notifier cold-start — pull-mode sync starts from the on-chain processed cursor

omersadika:
Fixes the actual root cause behind the notifier-side stalls in #1892 (the failure that forced the gRPC → JSON-RPC revert; #1884 fixed the validator committee-ratchet boot, but not this).

## Root cause

Pull-mode checkpoint sync demanded contiguous history from sequence 1. On a long-lived network **no peer can serve deep history**: validators never pull (they only hold what consensus handed them since their own last redeploy), and nothing backfills. A notifier deployed on a fresh database therefore looped forever on `no peers were able to help sync checkpoint 1` — `synced` pinned at 0 while `known` grew (the exact incident telemetry: `known` was only ever updated inside the dying sync job), the writer had nothing to submit, and every epoch close was silently blocked. The gRPC correlation was the fresh deploy that came with the migration, not the transport; the JSON-RPC revert "worked" because it restored the old database.

## Fix

The chain already defines what a writer still needs: everything at or below the coordinator's / system object's `last_processed_checkpoint_sequence_number` has landed on Sui and is submit-useless.

- The node's Sui connector feeds those two cursors into p2p state sync (`OnChainCheckpointCursors`: `watch` channels of `Option<u64>`, `None` until the first successful chain read, monotonically forward).
- Pull-mode sync (both checkpoint streams) starts from `max(local_next, cursor + 1)`, logging the skip.
- **Cold-start deferral**: an empty store does not start its first sync job until the cursor is known — otherwise the first job races the connector's first read straight back into the from-1 wedge. At genesis the first read is `Some(0)` within seconds (`Some(0)` ≠ `None` is why the channel carries `Option`), so brand-new networks are unaffected — every existing cluster test exercises that path from boot.
- Validators never pull; their behavior is untouched. Builder users without a cursor feed keep today's behavior.

## Regression test

`notifier_cold_start.rs`: boots a 4-validator cluster, waits for a healthy close, **wipes the notifier's entire database**, restarts it, and requires (a) two further epoch closes — only the history-less notifier can drive them — and (b) checkpoint 1 staying absent from its store. Localnet peers *do* hold full history, so (b) fails if sync silently falls back to the old from-1 path: the assertion pins the floor engaging, not just liveness.

## Docs

`dev-docs/specs/checkpoint-sync-floor.md` (+ README index): the failure, the floor rules, and the new gap-tolerant store invariants.

Related: #1892 (observability for this stall class is in #1894), #1882/#1884 (the separate committee-ratchet boot fix).


omersadika:
Cold-start regression test green: https://github.com/dwallet-labs/ika/actions/runs/30127086906 — post-restart snapshot confirmed an empty store, the floor started sync at cursor+1, checkpoint 1 stayed absent, and two epoch closes completed on the gap-started store. Full suite on the final head dispatched below.

omersadika:
Full cluster suite green on the final head (6344e4e4): https://github.com/dwallet-labs/ika/actions/runs/30130793467 — all 23 tests including the new cold-start regression. This PR is merge-ready. Note for merge order: #1894 and this touch different regions of state_sync/mod.rs; whichever merges second may need a trivial rebase.

PR #1900, opened 2026-07-25
feat(sui-connector): opt-in SIP-58 address-balance gas for the notifier

omersadika:
Follow-up to #1892/#1894/#1896, implementing the direction discussed on #1894: move the notifier's gas off owned coin objects onto SIP-58 **address-balance gas** — the building block for future same-identity writer redundancy (deterministically-constructed submissions from replicated notifier processes collapse to one tx digest), and an immediate deletion of the stale-gas failure class.

## What it does

New `notifier-gas-from-address-balance` config flag, **default OFF** (flag off = byte-identical writer behavior). When enabled, every writer submission (checkpoints + all epoch-switch calls) is built with:

- **empty gas payment** — the SIP-58 protocol trigger for paying gas from the sender's SUI address balance (`is_gas_paid_from_address_balance`);
- **`ValidDuring { min_epoch: current, max_epoch: current + 1 }`** — a submission built just before a Sui epoch boundary stays valid into the next epoch instead of expiring mid-flight; the one-epoch extension is the maximum `is_replay_protected` allows, and multi-epoch expiration (protocol 105) predates every network's balance-gas enablement, so the window is legal wherever the transaction is. Carries the chain's **full** genesis-rooted `ChainIdentifier` (resolved once at boot; compiled-in for mainnet/testnet, genesis-checkpoint digest for localnets; boot fails loudly if unresolvable) and a random `u32` nonce.

## Boot sequence with the flag on

1. **Chain-identifier resolution** — loud boot failure if unresolvable.
2. **Funds read** — new `get_sui_funds` API returns the address-balance / coin-object breakdown (gRPC `GetBalance` carries both fields; JSON-RPC derives the coin part). Unreadable (accumulators disabled on the target chain) ⇒ loud boot failure.
3. **Automatic migration sweep** — if the address holds ≥ 1 SUI in coin objects (the state of every existing notifier when the flag is first flipped), one transaction deposits them into the address balance: all owned gas coins as the sweep's own gas payment (protocol-merged into the first), `SplitCoins(GasCoin, total − sweep budget)`, then `coin::send_funds<SUI>(split, sender)` — split-then-deposit because the gas coin cannot be passed by-value to a Move call. The budget remainder stays as one small coin below the re-sweep threshold (no boot-loop churn). Best-effort: a failed sweep warns and falls through. **Net effect: flipping the flag on the live notifier is a zero-manual-step migration.**
4. **Balance preflight** — refuses to boot below one gas budget of address balance, with the exact remediation in the error; counts a just-swept amount directly rather than re-reading through a possibly-lagging fullnode view. A writer that cannot pay must not take the writer role (the alternative is an hour of failed submissions, a panic, and a blocked epoch close — the #1892 outage shape).
5. **Live funds gauge** — `ika_sui_connector_gas_coin_balance` (previously registered but never written by anything) is seeded at boot and refreshed every 60s from the address balance, so drain is alertable before submissions fail. Mid-flight exhaustion after a healthy boot still ends in the historical hour-retry-then-panic path, same as coin-mode gas exhaustion — the gauge is the alert window that was missing.

## What it deletes (when enabled)

No transaction references a gas object, so the entire `NotifierSubmitState` apparatus is bypassed: post-effects gas-ref caching, the `min_gas_version` re-fetch floor, and stale-gas rejection recovery — machinery that exists solely because the notifier's fullnode view lags the validators, and that has wedged epoch advance before. Node-view staleness can no longer invalidate a submission. The checkpoint fee reimbursement, which merged into `Argument::GasCoin`, is transferred to the writer address as an owned coin instead.

## Preconditions (documented on the flag + spec)

- Address-balance gas enabled on the target Sui network: **testnet since protocol 108, mainnet since 124** (verified in the pinned `mainnet-v1.73.2` protocol config); localnets at max version have it.
- Funding is handled by the boot sweep for a migrating notifier; a brand-new address needs an explicit balance deposit (plain coin transfers don't fund the balance). Each submission reserves the full 10-SUI budget for its validity window; the writer submits serially.

## Rollout intent

Testnet canary of the flag on the live notifier before any mainnet use; the gas-coin path remains the default until then. With #1894's stall metrics and #1896's sync floor in place, a misbehaving canary is loud within ~2 minutes and recoverable by flipping the flag back.

## Validation

- Unit tests pin both transaction contracts: the balance-gas submission (empty payment ⇒ balance-gas classification, `ValidDuring [current, current+1]`, `is_replay_protected()`, budget/sender/nonce) and the sweep (gas-coin payment, `SplitCoins(GasCoin, …)`, `coin::send_funds<SUI>` deposit to sender).
- **Flag-on end-to-end, green on CI**: `epochs_close_with_notifier_on_address_balance_gas` boots the cluster with the flag enabled — the notifier funded only with coin objects (the production migration state) — and passes: chain-id resolution, funds read, the automatic sweep (real transaction), the balance preflight, then two epoch closes where every checkpoint and epoch-switch submission is an empty-gas-payment `ValidDuring` transaction (no gas-coin fallback exists in this mode), with the funds gauge asserted positive. Run: https://github.com/dwallet-labs/ika/actions/runs/30161728015 — its first two attempts each caught a real client bug (short-id chain identifier via `Display`; `GetBalance` queried with the coin object type reading silently as zero), both surfaced as loud boot refusals by the preflight design.
- Flag-off default path: full cluster suite **green** — https://github.com/dwallet-labs/ika/actions/runs/30158782641 (re-run on the final head linked in comments).

Spec: `dev-docs/specs/notifier-balance-gas.md` (both gas modes, boot behavior, decision rules, invariants).

### DIFF
--- a/crates/ika-config/src/node.rs
+++ b/crates/ika-config/src/node.rs
@@ -618,6 +618,27 @@ pub struct SuiConnectorConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub notifier_client_key_pair: Option<KeyPairWithPath>,
 
+    /// Pay the notifier's gas from its SUI **Address Balance** (SIP-58)
+    /// instead of owned gas-coin objects. Submissions then carry no gas
+    /// `ObjectRef` at all (`ValidDuring` expiration, empty gas payment), which
+    /// removes the whole class of stale-gas-version wedges the coin path
+    /// fights: the transaction is valid regardless of how far this node's
+    /// fullnode view lags.
+    ///
+    /// Requirements before enabling:
+    /// - The target Sui network must have address-balance gas payments
+    ///   enabled (Sui protocol >= 108 on testnet, >= 124 on mainnet; always
+    ///   on for localnets at max version).
+    /// - The notifier address must hold its SUI in the ADDRESS BALANCE, not
+    ///   (only) in coin objects — deposit via the Sui CLI/SDK balance
+    ///   transfer. Each submission reserves the full gas budget from the
+    ///   balance for the transaction's validity window.
+    ///
+    /// Default `false`: the gas-coin path remains the production default
+    /// until this mode has been canaried on testnet.
+    #[serde(default)]
+    pub notifier_gas_from_address_balance: bool,
+
     /// Override the last processed EventID for sui module `ika_system`.
     /// When set, SuiSyncer will start from this cursor (exclusively) instead of the one in storage.
     /// If the cursor is not found in storage or override, the query will start from genesis.
@@ -1293,6 +1314,7 @@ mod tests {
             ika_dwallet_coordinator_object_id: ObjectID::ZERO,
             verified_cache_retention_checkpoints: None,
             notifier_client_key_pair: None,
+            notifier_gas_from_address_balance: false,
             sui_ika_system_module_last_processed_event_id_override: None,
         }
     }
--- a/crates/ika-core/src/sui_connector/mod.rs
+++ b/crates/ika-core/src/sui_connector/mod.rs
@@ -22,19 +22,24 @@ use ika_types::messages_dwallet_mpc::{
 use ika_types::sui::{
     DWalletCoordinator, DWalletCoordinatorInner, System, SystemInner, SystemInnerTrait,
 };
+use move_core_types::ident_str;
 use shared_crypto::intent::{Intent, IntentMessage};
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::Duration;
+use sui_json_rpc_types::{SuiExecutionStatus, SuiTransactionBlockEffectsAPI};
+use sui_types::SUI_FRAMEWORK_PACKAGE_ID;
 use sui_types::base_types::{ObjectID, ObjectRef, SuiAddress};
 use sui_types::crypto::{Signature, SuiKeyPair};
 use sui_types::digests::{
     ChainIdentifier as SuiNetworkChainIdentifier, get_mainnet_chain_identifier,
     get_testnet_chain_identifier,
 };
+use sui_types::gas_coin::GAS;
+use sui_types::programmable_transaction_builder::ProgrammableTransactionBuilder;
 use sui_types::transaction::{
-    GasData, ProgrammableTransaction, Transaction, TransactionData, TransactionDataV1,
-    TransactionExpiration, TransactionKind,
+    Argument, Command, GasData, ProgrammableTransaction, Transaction, TransactionData,
+    TransactionDataV1, TransactionExpiration, TransactionKind,
 };
 use tokio::sync::watch;
 use tokio::sync::watch::{Receiver, Sender};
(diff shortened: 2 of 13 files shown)


omersadika:
Hardening added in a899c05f, answering "what happens with an empty address balance?" — previously: Sui rejects every submission, the writer retries for an hour, panics, restarts, repeats, with the epoch close blocked throughout (the gas-exhaustion outage shape). Now:

- **Boot preflight**: with the flag on, the notifier reads its ADDRESS BALANCE and refuses to start below one gas budget, with the exact remediation in the error (deposit into the balance — coin-object transfers don't fund it — or unset the flag). An unreadable balance (accumulators not enabled on the target network) also fails boot loudly.
- **Live funds gauge**: `ika_sui_connector_gas_coin_balance` — which previously had NO writer at all — is now seeded at boot and refreshed every 60s from the address balance in this mode, so exhaustion is alertable before submissions start failing.

Runtime exhaustion mid-flight (balance drains after a healthy boot) still ends in the hour-retry-then-panic path — same policy as coin-mode gas exhaustion — but the gauge now gives the alert window that was missing.

Full cluster suite re-dispatched (previous run was a runner startup_failure, nothing ran): https://github.com/dwallet-labs/ika/actions/runs/30158438833

omersadika:
Added the coin-object migration sweep (439f2887): with the flag on, boot now runs — chain-id resolution → funds read → **automatic sweep** of any >= 1 SUI held in coin objects into the address balance (one tx: all gas coins as the sweep's own payment, `SplitCoins(GasCoin, total − sweep budget)`, `coin::send_funds<SUI>(split, sender)` — the gas coin can't be passed by-value to a Move call, hence split-then-deposit) → balance preflight (counting a just-swept amount directly, since a fullnode re-read can lag the finalized sweep) → 60s funds-gauge refresher.

Net effect: flipping the flag on an existing notifier is now a zero-manual-step migration — its coin funds move themselves. A failed sweep is best-effort (warn + fall through to the preflight verdict). Unit test pins the sweep tx shape. Suite re-dispatched on this head: https://github.com/dwallet-labs/ika/actions/runs/30158782641

omersadika:
Full cluster suite green on the final head (439f2887): https://github.com/dwallet-labs/ika/actions/runs/30158782641 — all tests pass on the flag-off default path (byte-identical writer behavior). PR is ready for review; flag-on validation is the documented testnet canary.

omersadika:
Review feedback applied in 30f632a4: the `ValidDuring` window is now `[current, current + 1]` instead of single-epoch. The single-epoch shape was defensive against the legacy `min == max` validity rule, but that caution was moot — multi-epoch expiration is enabled unconditionally from Sui protocol 105, strictly below every network's address-balance-gas enablement (testnet 108 / mainnet 124), so wherever balance gas is legal the wider window is too. Benefit: a submission built just before a Sui epoch boundary stays valid instead of expiring mid-flight and burning a rebuild-retry. It's the maximum `is_replay_protected` allows. Documented trade-off: an abandoned signed-but-never-executed tx can hold its balance reservation up to two epochs instead of one (irrelevant to a serial writer with normal float). Unit test updated; flag-off path untouched, so the existing green suite (30158782641) still covers the default behavior.

omersadika:
**Flag-on E2E is green**: https://github.com/dwallet-labs/ika/actions/runs/30161728015 — `epochs_close_with_notifier_on_address_balance_gas` passed in 88s: boot resolved the chain identifier, swept the notifier's coin-object funding into its address balance, passed the preflight, and drove two epoch closes where every checkpoint and epoch-switch submission was an empty-gas-payment `ValidDuring` transaction (no gas-coin fallback exists in this mode), with the funds gauge asserted positive.

The two attempts before it each caught a real bug that unit tests could not see and that would otherwise have burned testnet canary cycles:
1. caa2f599 — the ika transport stringifies the chain identifier via `Display` = the 4-byte short id; the typed identifier is now read through the writer uplink.
2. bcb9168c — `GetBalance` was queried with the coin OBJECT type (`Coin<SUI>`) instead of the coin type (`0x2::sui::SUI`), silently reading zero and skipping the sweep.

Both were surfaced as loud boot refusals with exact messages — the preflight design working as intended.

Final full suite (now including both new tests) dispatched on this head: https://github.com/dwallet-labs/ika/actions/runs/30162167607

omersadika:
Final full cluster suite green on the merge head (bcb9168c): https://github.com/dwallet-labs/ika/actions/runs/30162167607 — all 24 tests, including both new ones (flag-on balance-gas E2E and the flag-off default-path coverage). PR is merge-ready.

PR #1908, opened 2026-07-25
refactor(config): compiled-in ika identity on public chains; unsafe override struct for the rest

omersadika:
## Summary

The seven ika package/object IDs (`ika-package-id`, `ika-common-package-id`, `ika-dwallet-2pc-mpc-package-id[-v2]`, `ika-system-package-id`, `ika-system-object-id`, `ika-dwallet-coordinator-object-id`) leave the flat config surface:

### Mainnet / Testnet: hardcoded, not overridable
The identity is compiled into the binary (`compiled_in_ika_identity`, keyed off `sui-chain-identifier`). **Previously a config-supplied value silently won over the compiled-in identity even on public chains** — an operator-level identity swap could redirect a mainnet node to arbitrary packages/objects. Now a public-chain config carrying an override is rejected at startup, and old flat YAML keys parse but are ignored (serde-skipped fields).

### Localnet / Devnet / Custom: the new `ika-unsafe-identity-override` struct
Chains with no compiled-in identity (IDs freshly generated each genesis) must supply them via the new nested struct — required there, and the sole source. `-v2` is optional inside it ("no upgrade published" stays a first-class `None`). The `unsafe_` prefix follows the repo convention for opt-outs that skip a safety property.

```yaml
sui-connector-config:
  sui-chain-identifier: custom
  ika-unsafe-identity-override:
    ika-package-id: '0x...'
    ika-common-package-id: '0x...'
    ika-dwallet-2pc-mpc-package-id: '0x...'
    ika-system-package-id: '0x...'
    ika-system-object-id: '0x...'
    ika-dwallet-coordinator-object-id: '0x...'
```

### Mechanics
- The flat fields remain as **runtime-resolved outputs** (`serde(skip)`), so the ~20 in-tree readers are untouched; the config-file surface is exactly the new struct.
- Resolution is wired into `NodeMode::validate_and_prepare_config` so it runs **before `IkaRuntimes::new`** — whose mock-crypto refusal keys off `ika_system_object_id`; this ordering makes that gate derive the chain from the compiled-in identity rather than an operator-writable field. `IkaNode::start` still resolves too (idempotent) for in-process swarm nodes.
- Deletes the dead per-chain dwallet-v2 hardcode in `ika-node/lib.rs` (duplicated the compiled-in literals; unreachable since resolution already filled v2 on public chains).
- Migrates: swarm/fullnode config builders, the TS system-test fullnode template + both genesis scripts (yq paths now write the override block), and the ika-operator skill docs (which still referenced the long-removed `sui-trusted-anchor`).

### Operator impact
- **Mainnet/testnet configs need no change**: flat ID keys are ignored; the compiled-in identity takes over (same values).
- Anyone running a **custom/localnet** chain must move their IDs into `ika-unsafe-identity-override` — a config without it fails at startup with an actionable error.

## Test plan

- `cargo check --release --all-targets` across all eight touched crates — clean.
- `cargo test --release -p ika-config` — 26 passed, including new tests: override rejected on public chains (no partial resolution), required + sole source on private chains, ZERO ids inside the override rejected, flat YAML keys ignored end-to-end, kebab-case YAML round-trip with absent `-v2` staying absent.
- Suggest the cluster + TS suites on CI before merge — they boot localnets entirely through the migrated override path.


omersadika:
## Review: design verified sound — but DO NOT MERGE: `ika_system_package_id` is wrong for BOTH networks (fleet-breaking, and this PR removes the config escape hatch that would mitigate it)

### Verified good (against the branch, not just the description)

- **All three production binaries run the resolution**: `ika-validator`/`ika-fullnode`/`ika-notifier` are bin targets of `ika-node`, each calls `run_node(Some(NodeMode::…))` → `node_runner.rs:134` `validate_and_prepare_config(&mut config)` **before** `IkaRuntimes::new` — so the notifier (which loads the same `sui-connector-config` surface) is covered, and the mock-crypto gate ordering claim holds.
- The compiled-in `-v2` constants byte-match the dead lib.rs hardcode being deleted (testnet `0x6573a6c1…`, mainnet `0x23b5bd96…`) ✓.
- 12 of 14 compiled-in constants byte-match the configs running in production **right now** (pulled live from the testnet + mainnet notifier pods minutes ago).

### Blocking: the two that don't match

| network | field | live production config (now) | PR compiles |
|---|---|---|---|
| testnet | `ika_system_package_id` | `0xae71e386fd4cff3a080001c4b74a9e485cd6a209fa98fb272ab922be68869148` | `0xde05f49e5f1ee13ed06c1e243c0a8e8fe858e1d8689476fdb7009af8ddc3c38b` |
| mainnet | `ika_system_package_id` | `0xb874c9b51b63e05425b74a22891c35b8da447900e577667b52e85a16d4d85486` | `0xd69f947d7ee6f224dd0dd31ec3ec30c0dd0f713a1de55d564e8e98910c4f9553` |

Same single field wrong on both networks ⇒ a systematically stale source (likely genesis-era values — see below). If this ships, every upgraded node resolves the system package to a non-current package: sui-connector event filters and system entry calls target the wrong package, fleet-wide, on both networks. And the sharpest part: **because this PR (correctly) rejects config overrides on public chains, operators would have no config-side escape** — the only remedy for a wrong compiled-in constant is another binary release. The hardening removes the workaround for exactly the class of mistake it itself contains. The "Mainnet/testnet configs need no change (same values)" claim in the description is false for this field.

### Why it happened, and the design question it exposes

`ika_system_package_id` **rotates**: Move-contract upgrades publish a new system package (both networks did this at the protocol 4→5 flip — which is why live differs from the genesis-era IDs). A compiled-in constant for a rotating id goes stale at every contracts upgrade unless every such release also bumps the constant. Options, strongest first:

1. **Compile in only the immutable identity** (object ids + original package ids) and resolve the rotating package ids (system, current 2pc-mpc) **from the system object at startup** — chain-authenticated, can't go stale, strictly better than both config and constants for these fields.
2. Keep full compile-in but add a **startup cross-check**: resolved `ika_system_package_id` vs what the system object reports; loud fail on mismatch. (This check alone would have caught both bad constants.)
3. Minimum: fix the two constants to the live values above and add a release-pipeline step that diffs compiled-in identity against live chain state for testnet/mainnet, so drift can't ship silently.

Option 1 (or 2) also answers the maintenance question this PR otherwise creates for every future Move upgrade.

PR #1910, opened 2026-07-25
fix(config): compiled-in ika_system_package_id must be the ORIGINAL package, not the v2 upgrade (fleet-breaking if released)

omersadika:
## Severity: release blocker for the next tag — main currently compiles fleet-breaking identity constants for BOTH networks

Follow-up to the [review on #1908](https://github.com/dwallet-labs/ika/pull/1908#issuecomment-5079917244), which flagged these two constants pre-merge; the PR merged unchanged. The provenance is now fully pinned down and the fix is chain-verified.

## The bug

`compiled_in_ika_identity` sourced `ika_system_package_id` from the **`ika_system_package_id_v2`** line of `deployed_contracts/{mainnet,testnet}/address.yaml` instead of the base key one line above it:

| network | correct (original / base key) | shipped in #1908 (`_v2` upgrade) |
|---|---|---|
| testnet | `0xae71e386fd4cff3a080001c4b74a9e485cd6a209fa98fb272ab922be68869148` | `0xde05f49e5f1ee13ed06c1e243c0a8e8fe858e1d8689476fdb7009af8ddc3c38b` |
| mainnet | `0xb874c9b51b63e05425b74a22891c35b8da447900e577667b52e85a16d4d85486` | `0xd69f947d7ee6f224dd0dd31ec3ec30c0dd0f713a1de55d564e8e98910c4f9553` |

## Why the v2 id is wrong by ROLE, not just by value

Every consumer of the config/compiled field matches **Sui type tags** — event filters, the system object's own type check, and the OCS push-worker relevance set (`push_worker.rs:96`). Sui type tags carry the **defining** package forever: the system objects themselves are typed `{original}::system::System` on both networks today, even though the v2 code executes. A node resolving this field to the v2 id filters on types that tag nothing → the sui connector goes deaf to system events fleet-wide. Because #1908 (correctly) rejects config overrides on public chains, there would be **no config escape hatch** — only another binary release.

The current *executable* package is already resolved dynamically — `sui_executor.rs:700` reads `system.package_id` from the live system object — which is precisely why production works with the original ids in config and why **no compiled constant ever needs bumping at a contracts upgrade** (this also retires the rotation-maintenance concern from the #1908 review).

## Chain receipts (sui_getObject, 2026-07-25)

- Testnet system object `0x2172c648…`: type `0xae71e386…::system::System`, content `package_id: 0xde05f49e…`, `version: 2`
- Mainnet system object `0x215de95d…`: type `0xb874c9b5…::system::System`, content `package_id: 0xd69f947d…`, `version: 2`
- Both networks' production node configs (validators + notifiers, pulled live from the pods) carry the original ids — the values the working fleet runs on.
- All 12 other compiled-in constants were re-verified against live configs: byte-identical, untouched here.

## Changes

- The two `ika_system_package_id` constants → the original (base) ids.
- Struct doc rewritten to state the role rule: **all package ids here are defining/original publications; the current executable resolves at runtime; never bump on upgrades; source from the BASE keys of address.yaml.**
- New pin test `compiled_in_system_package_id_is_the_original_not_the_v2_upgrade`: asserts both networks equal the original AND not-equal the v2 (the two ids sit on adjacent lines in address.yaml — pinning both directions blocks the exact copy mistake from recurring).

## Deliberately unchanged (correct for their role)

- `ika_sui_config.yaml` — documented as flattening to latest; latest is its semantic.
- `sdk/typescript/network-configs.ts` `ikaSystemPackage` — the SDK builds transactions; entry calls target the current package.
- `network_dkg.rs` test-only `DEPLOYED_ENVS` table — latest-flavored live-read fixtures (uses the 2pc-mpc v2 in its base slot too).

## Test plan

- `cargo test --release -p ika-config`: 28 passed, including the new pin test.
- Recommended follow-up (from the #1908 review, not in this PR): a startup assert that the fetched system object's type address equals the resolved `ika_system_package_id` — that check would have caught #1908's constants before any human did.

### DIFF
--- a/crates/ika-config/src/node.rs
+++ b/crates/ika-config/src/node.rs
@@ -222,12 +222,20 @@ impl fmt::Display for SuiChainIdentifier {
 }
 
 /// ika's on-chain identity (Move package + object IDs) for a public chain.
-/// Deployment constants: the object IDs are immutable, and the package IDs are
-/// the exact values the node config expects — the dwallet base id is the
-/// **original v1** with the upgrade in `_v2` (event filtering accepts both),
-/// and the system id is the **current/latest**. Sourced from
-/// `deployed_contracts/{mainnet,testnet}/address.yaml` (NOT `ika_sui_config.yaml`,
-/// which flattens packages to their latest under the base name).
+/// Deployment constants: the object IDs are immutable, and every package ID is
+/// the **original (defining) publication** — the dwallet base id is the
+/// original v1 with the upgrade in `_v2` (event filtering accepts both), and
+/// `ika_system_package_id` is likewise the ORIGINAL system package, never a
+/// later upgrade. These ids feed type/event matching and the OCS relevance
+/// set, and Sui type tags carry the *defining* package forever — the system
+/// object itself is typed `{original}::system::System` even after upgrades.
+/// The *current executable* system package is resolved at runtime from the
+/// system object's `package_id` field (`sui_executor` reads
+/// `system.package_id`), so NO constant here needs bumping when contracts
+/// upgrade. Sourced from the BASE keys of
+/// `deployed_contracts/{mainnet,testnet}/address.yaml` (`ika_system_package_id`,
+/// not `ika_system_package_id_v2`; and NOT `ika_sui_config.yaml`, which
+/// flattens packages to their latest under the base name).
 #[derive(Clone, Copy, Debug)]
 pub struct IkaOnChainIdentity {
     pub ika_package_id: ObjectID,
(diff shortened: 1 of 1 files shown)


omersadika:
## Re-verification on chain (fresh queries, 2026-07-25 ~21:20 UTC) — including the decisive probe the original review lacked

The first verification established the object metadata; this pass adds **live event type tags**, which is the direct empirical test of the claim that this field's consumers need the defining id.

**1. System objects re-read (both networks):** testnet type `0xae71e386…::system::System` / content `package_id: 0xde05f49e…`; mainnet type `0xb874c9b5…::system::System` / content `package_id: 0xd69f947d…`. Unchanged: type address = original, recorded current = v2.

**2. Live emitted events** (`suix_queryTransactionBlocks` over the system objects' recent transactions, `showEvents`):

- **Testnet, 46 events**: `0xae71e386…::system_inner::EndOfPublishEvent`, `::system_inner::SystemEpochInfoEvent`, `::system_inner::SystemCheckpointInfoEvent`, 43× `::validator_set::ValidatorEpochInfoEventV1` — **zero events tag `0xde05f49e…`**, despite the v2 package being the executing code.
- **Mainnet, 119 system events across 25 txs**: 112× `0xb874c9b5…::validator_set::ValidatorEpochInfoEventV1`, 2× each of `EndOfPublishEvent`/`SystemEpochInfoEvent`/`SystemCheckpointInfoEvent`, 1× `SetNextProtocolVersionEvent` — **100% tag the original; zero tag `0xd69f947d…`**. (Common-package events likewise tag its original `0x9e1e9f8e…`.)

These are exactly the epoch-machinery events. Note also `ika-sui-client/src/lib.rs` `query_events_by_module` **asserts** `event.type_.address == package_id` on every page — with the v2 id, a node either sees nothing or trips that assert.

Conclusion unchanged and now triple-grounded per network: the constants in this PR (the originals) are the only values consistent with on-chain reality.

PR #1911, opened 2026-07-25
feat: AuthorityName → consensus key at protocol v6 (Rust-only, upgrade-gated)

omersadika:
## Summary

A validator's `AuthorityName` — its canonical committee identity — becomes its **Ed25519 consensus key** instead of its BLS12-381 protocol key, activated at **protocol version 6**. Rust-only: **no Move changes**. The BLS key stays on chain and is carried on `Committee`, so BLS aggregate-certificate (checkpoint) verification is unaffected.

The wire encoding does not change: the 32-byte consensus key is zero-padded into the existing 48-byte container, and the deserializer accepts both the padded and unpadded encodings.

## Why this was hard, and what the upgrade test found

A previous attempt at this flip was implemented end-to-end and **reverted** because it wedged the rolling upgrade (`dev-docs/plans/authority-name-consensus-key.md`). The root cause is an asymmetry that still exists: the next-epoch committee is assembled **mid-epoch under the current epoch's protocol version**, but rebuilt by the next epoch **under its own**. At the single activation boundary those disagree about how to name the same members.

Nothing in the repo tested a protocol-version transition — the v3/v4-era rehearsals were retired with v3/v4 support, and their successor `v125_rollout` is explicitly *"a pure binary swap, no protocol transition."* So this PR adds that gate, and it immediately paid for itself. Bugs it caught, none of which were visible to unit tests or review:

1. **Halt — self-identity frozen at startup.** `AuthorityState.name` is minted once and was carried into every later epoch store, so at the flip a sitting validator's own name was absent from its committee: *"no longer a validator after reconfiguration."* Now re-derived per epoch; `is_validator` compares the epoch store's name.
2. **Halt — the handoff cert's committee hash.** `next_committee_pubkey_set_hash` is computed by the outgoing epoch and re-verified by the incoming one; across the flip the hashes differ and the node fail-closed halts. The verifier now also accepts the same membership expressed in the other basis.
3. **Halt — the chain-read prior committee** (a true joiner, or any node missing that epoch's `committee_store` record). It was built under the *current* version's basis, making every cert signer weight-0 — which `verify_certified_handoff_attestation` rejects **outright**, not through quorum. The prior epoch's basis isn't recoverable from chain, so `fetch_previous_committee` returns one candidate per basis from a single read and the verifier takes whichever verifies.
4. **Silent — mpc_data carry-forward.** Prior-cert names looked up against current-epoch names: at the boundary every seated member that hadn't freshly announced would fall into the excluded set, removing exactly the safety net that stops reconfiguration being rejected for a still-seated member's missing key material. Visible only as `carried_forward = 0`.
5. **Silent — prior-cert key ingest and blob repair.** A node that restarted mid-epoch would stay MPC-dead for the epoch; the blob repair reported "nothing missing" **with no log at all**.

(4) and (5) go through `Committee::name_translation`, which maps a member's name under either basis to the one that committee uses, derived from the committee's own key maps — no chain read. Away from the boundary every entry maps a name to itself.

Also fixed: a per-tick `get_validators_info_by_ids` I had introduced in the sync loop whose `continue`-on-error could starve next-committee assembly during restarts — the same shape as a wedge this repo already paid for once. Consensus keys are registration-fixed, so they're now cached by validator id and steady-state ticks do no chain read.

## A sweep bounded the blast radius

Every `AuthorityName` artifact that crosses an epoch boundary flows through **one** carrier: the perpetual `certified_handoff_attestations` table. Checked and cleared: the seven name-keyed `AuthorityEpochTables` are physically per-epoch (one basis each); `committee_map` records are each stored and read under their own epoch's basis; checkpoints carry no `AuthorityName` in any signed payload (`signers_map` is a positional bitmap) and MPC party IDs are positional.

## Validation

All heavy suites green on this branch:

| Suite | Result |
|---|---|
| **`v125_v6_upgrade`** (new gate) | ✅ 32/32 flows, `got=6`, committee size 4, boundary reshare converged, post-flip workload served |
| **`v125_rollout`** (deployed-release gate) | ✅ mixed literal-v1.2.5/current committee, two aggregated reshares at v5 — the regression check for networks that never flip |
| **Cluster** | ✅ 23/23 tests (genesis at v6, so consensus-key names end to end) |
| **Rust integration** | ✅ |
| **TypeScript integration** | ✅ |

**The v6 gate is fault-validated.** Removing the identity-basis tolerance — one token — fails it at the flip epoch with `next_committee_pubkey_set_hash mismatch` and a fail-closed halt in **all four** validators' logs, while the v5 phases and the binary swap stay clean. A green run is therefore evidence, not faith. (The halt arrives via `prepare_handoff_anchor`, so grep the hash-mismatch string, not a call site.)

Unit coverage: deserializer round-trips/leniency/rejections (BCS + JSON), `from_consensus_key` ↔ unpadded-wire agreement, `EpochStartSystem::V2` basis carriage, V1 BCS records surviving the V2 enum addition, and cert tolerance in all three arms (strict without an alternate, accepts the genuine alternate, rejects an unrelated one). Workspace clippy clean.

### Test-side fixes worth knowing about

Four assertions encoded the *old* identity invariant and failed once genesis actually ran at v6 — each reporting it as a production defect. Reviewers re-reading these should know the failures were in the assertions, not the code they accused:

- the upgrade harness built its expected committee from BLS keypairs only (`cluster.rs`);
- it then derived the Byzantine quorum from that set's length, which doubled when I made it span both bases;
- `JoinerHandle::authority_name` returned the BLS name unconditionally — reported as *"its on-chain mpc_data record never became readable — a registration/lifecycle bug"*;
- `cluster.validator_names` likewise — reported as originals *"missing from the epoch-1 handoff cert"*.

The swarm's node registry deliberately still keys on the protocol key: it is a local addressing scheme, never compared against a committee.

## Operator-facing fix found by the injection

The fail-closed halt asserted *"the local handoff-cert DB is tampered or corrupted"* — false for a basis mismatch, and it would have sent every operator to debug storage. Both halt sites now name the alternative cause for a hash mismatch.

## Docs

`dev-docs/specs/committee-consensus-keys.md` now specifies the v6 identity rule and its boundary caveat; the plan doc records the reinstatement; `cross-binary-upgrade.md` and `ci-suites.md` document the new gate and its fault validation.

## Review guidance

The dual-basis tolerances are the parts worth the most scrutiny: each trades strictness for boundary survivability. My argument that they're safe is that both name sets describe the *same* membership — an encoding ambiguity, not a widened acceptance, and a cert still needs a stake quorum of valid signatures — but that judgement deserves a second pair of eyes rather than my own plus CI.

https://claude.ai/code/session_01WMKeqyqHbEexXehGQgXX67


omersadika:
## Review: the boundary engineering is excellent — but DO NOT MERGE yet: the design's stated foundation ("consensus keys are registration-fixed") is false on chain, and it breaks the PR in three distinct ways

I reviewed the dual-basis tolerances you asked scrutiny for (verdict below: sound), and then went after the chain-side assumptions this Rust-only change rests on. That's where it fails.

## The blocker: consensus-key rotation is a live chain feature

`system::set_next_epoch_consensus_pubkey_bytes` (system.move:477) is a **public, ValidatorCap-gated entry point**, wired through `validator_set` → `validator_info.move:224`, and **effectuated at epoch advance** — `rotate_next_epoch_info` at validator_info.move:295 does `consensus_pubkey_bytes = next_epoch_consensus_pubkey_bytes.extract()`. One operator transaction, applied at the next boundary. Under v5 this is harmless (names are BLS; per-epoch committee reads pick up the new consensus key). Under v6, where the consensus key IS the identity, a single rotation breaks the PR three ways:

**(a) Identity discontinuity.** A rotating validator changes its `AuthorityName` at the boundary. `name_translation` is derived from the NEW committee's key maps, so the validator's OLD consensus key aliases nothing — its prior-cert `ValidatorMpcData` entry misses translation and it lands in the excluded set. That is your bug #4's exact silent shape, except permanent machinery, not a one-boundary caveat: carry-forward, prior-cert key ingest, and blob repair all break for that member, every time anyone rotates.

**(b) The never-invalidated cache serves a stale identity fleet-wide.** `committee_names_for_basis` documents the cache as safe because values are "registration-fixed." After a rotation, every validator's syncer keeps naming the rotated member under its old key — the assembled next-epoch committee diverges from chain reality on all nodes at once, and the member's signatures no longer verify against the name the committee carries. This is a fleet-wide wedge candidate triggered by one routine operator action.

**(c) Rotation-time identity collision — the sharpest one.** `is_duplicate` (validator_info.move:463) enforces consensus-key uniqueness thoroughly — **at registration**. The rotation setter calls only per-validator `validate()`; I found no cross-validator duplicate re-check on the rotation path. So an active validator can rotate its next-epoch consensus key to ANOTHER active validator's consensus key. At effectuation under v6, two committee members hold the **same `AuthorityName`**: `voting_rights` carries duplicate names, `index_map`/`name_translation` silently collapse (HashMap last-write-wins), and stake/attribution semantics are undefined. That's an identity-collision attack for the price of one transaction from any registered validator.

## What I'd require before this merges

1. **Reconcile rotation with v6 identity — explicitly, not by assumption.** Options in rough preference order: (i) a companion contracts change that aborts `set_next_epoch_consensus_pubkey_bytes` once the protocol version is ≥6 (rotation was BLS-era flexibility; identity keys don't rotate — Sui's protocol key doesn't either), shipped in an upgrade that activates before or with v6; (ii) Rust-side handling that treats a rotated key as leave+join with a once-per-epoch cache revalidation — much more machinery, and identity continuity is still lost.
2. **Rust-side fail-closed on name collision regardless of (1)**: committee construction (`new`, `new_with_protocol_keys`) must reject duplicate `AuthorityName`s loudly (`report_invariant_violation!` is sitting right there from #1925) instead of silently collapsing maps. This is Rust-only, cheap, and turns scenario (c) from undefined behavior into a refused committee.
3. **A regression in the v6 gate for the rotation shape**: rotate one validator's consensus key at the flip epoch in `v125_v6_upgrade` and assert whatever behavior (1) chooses. Given your one-token fault-validation standard, this is the missing injection.

## The parts you asked about: the dual-basis tolerances are sound

- **Wire encoding**: one canonical in-memory form (48-byte padded, enforced at the single constructor and by the deserializer), serializer always emits canonical — so signed/hashed payloads built locally are byte-stable across honest nodes. Cross-basis ambiguity of the 48-byte container needs a valid BLS G1 point with 16 zero tail bytes AND a known dlog — ~2^128 grinding past the registration proof-of-possession. Worth one sentence in `committee-consensus-keys.md` so the argument is on record.
- **Cert hash tolerance**: exact equality against exactly two candidate hashes, both derived from the caller's own trusted chain read of the SAME membership — no acceptance widening, and signature+stake quorum requirements are untouched. The invariant worth documenting at the parameter: the alternate set must always be locally derived, never remote-supplied.
- **Try-both prior committee**: both candidates carry identical membership/stake and identical consensus-key verification material; only labels differ. Sound.
- **`load_inner`'s silent `filter_map`** (minor): a name that fails BLS decode is either consensus-basis (16-zero tail — expected) or corrupt data (previously a loud panic, now silently absent until `public_key()`). Distinguish by the zero tail and fire `report_invariant_violation!` on the garbage case.

## Also worth saying

The engineering standard here is the best of the month: the new `v125_v6_upgrade` gate found five real boundary bugs review couldn't, the one-token fault validation makes green runs evidence, the blast-radius sweep (single perpetual carrier, positional signer maps) is exactly the analysis this class of change needs, and the cache design encodes a wedge this repo already paid for. That's why the rotation gap stands out: it's the one chain-side assumption the (excellent) test infrastructure never injects — and it happens to be operator-reachable in production today. MAX=6 also means merge starts the activation clock on the next release; the rotation story must land before any MAX=6 binary reaches quorum on either network.

omersadika:
## Refinement of the rotation finding + the Rust-only design for supporting it (no Move changes — decided)

Discussion follow-up: per-epoch key freshness is real — rotation effectuates at the boundary and every epoch's chain reads carry the latest keys, which is exactly why rotation is harmless under v5, where the consensus key is *data attached to* a stable BLS identity. Under v6 the key IS the identity, so "latest each epoch" means the identity itself changes at the boundary — and that bites in two places per-epoch freshness cannot reach:

1. **The straddle recurs forever.** Epoch E assembles E+1's committee mid-epoch, *before* effectuation, so it reads the rotating member's current (old) key and computes `next_committee_pubkey_set_hash` under the old name; E+1's entrants re-derive under effectuated (new) keys. That's this PR's own root-cause asymmetry — solved once for the v5→v6 basis flip — reappearing at **every** rotation epoch, and the alternate-basis tolerance can't cover it (the mismatch is consensus-vs-consensus; widening to "accept any hash where one member's key differs" would be genuine acceptance-weakening).
2. **Cross-epoch artifacts are keyed by the old identity.** Prior-cert `ValidatorMpcData` items name the rotated member under the old key; `name_translation` built from E+1's committee aliases only the new consensus key and the BLS key → the member silently falls into the excluded set. Bug #4's shape, reintroduced by any rotation.

## The Rust-only design (constraint: no Move changes)

1. **Effectuation-aware next-committee assembly**: when re-keying the *next*-epoch committee, use `next_epoch_consensus_pubkey_bytes` when set, falling back to the current key. E's view of E+1 then matches post-effectuation reality by construction — the hash agrees, no tolerance needed for rotation.
2. **Per-epoch key cache**: the current never-invalidated cache is only correct under the registration-fixed assumption. Replace with a cache that distinguishes current vs pending values and resets (or revalidates) at epoch advance — steady-state ticks still do zero chain reads within an epoch, preserving the anti-wedge property the cache exists for.
3. **`name_translation` spanning prior + current committees**: both are in the `committee_store`; including the prior epoch's consensus-key aliases makes old-name artifacts (carry-forward, prior-cert ingest, blob repair) resolve to the continuing member instead of silently excluding it.
4. **Fail-closed on `AuthorityName` collision — now load-bearing, not defense-in-depth.** Since the missing rotation-time duplicate check cannot be added on-chain, Rust must refuse to construct or adopt any committee carrying duplicate names (`Committee::new` / `new_with_protocol_keys` / the assembly path), with `report_invariant_violation!` so it pages (#1925). Honest trade-off to state in the spec: without the Move check, a hostile rotation-onto-someone-else's-key becomes a **liveness** event (the fleet halts loudly on an unconstructible committee) rather than an **identity corruption** (two members silently collapsing into one name). A visible halt with both colliding validators identifiable on chain is the best available outcome under the no-Move constraint — and strictly better than HashMap last-write-wins.
5. **Gate injection**: add a rotation flow to `v125_v6_upgrade` — rotate one validator's consensus key at a post-flip epoch, assert the boundary completes with the rotated identity continuous (items 1–3), plus the collision variant asserting the loud refusal (item 4). Same one-token fault-validation standard as the existing gate.

Items 1–3 make rotation *work* under v6; item 4 contains the one abuse the chain currently permits; item 5 makes all of it evidence rather than intention.

omersadika:
## Rotation review: verified, one fix landed, the rest scoped — with one point where I don't think the "fresh reads" argument closes the gap

Thanks — this was the right thing to go after. I verified all three chain-side claims against the contracts rather than taking them on trust, and they hold:

- `system::set_next_epoch_consensus_pubkey_bytes` (`system.move:477`) is public, `ValidatorOperationCap`-gated.
- `rotate_next_epoch_info` (`validator_info.move:292`) effectuates it, and it is reached from `validator.advance_epoch` (`validator.move:447`) via `validator_set::advance_epoch` (`:687`/`:995`) — i.e. **at the epoch boundary**, not at mid-epoch reconfiguration.
- The rotation setter (`validator_info.move:224-230`) calls only `self.validate()`, which is length/format checks. `is_duplicate` is reachable only from `validator_set.move:207/274/299` — the registration and pending-validator paths. **There is no cross-validator uniqueness check on rotation.**

So "consensus keys are registration-fixed" is false, and that sentence was mine. Correcting the record on each of (a)/(b)/(c).

### Fixed now: (b), the never-invalidated cache — this was the fleet-wide one

`consensus_key_cache` was declared outside the syncer loop, and the syncer is spawned once per node lifetime, so a key cached before a rotation was never re-read — every node would keep naming the rotated member under its old key indefinitely. Under v5 that is a stale value; under v6 it is a stale *identity*, held fleet-wide. Agreed this was the sharpest of the three in practice.

It is now **dropped at every epoch change**. That keeps the property the cache exists for — steady-state ticks do zero chain reads, so RPC flake cannot starve next-committee assembly during restarts, the wedge this repo already paid for once — because a rotation can only effectuate at a boundary, so *within* an epoch the value genuinely cannot change. I also corrected two comments in the prior-committee reconstruction that repeated the same false premise; their logic is unchanged and still assumes no rotation since the prior epoch, but that is now stated rather than asserted as fact.

### Deferred by decision: (a), prior-epoch artifacts keyed by the old identity

Accepted as a known gap, on this reasoning: each of the three consumption sites degrades **per member**, and quorum absorbs it.

- Carry-forward only bites a member that rotated *and* did not freshly announce that epoch — two conditions — and the outcome is that one member lands in the consensus-agreed excluded set.
- Prior-cert key ingest affects only the rotated node itself, and only if it also restarted mid-epoch; it recovers the next epoch.
- Blob repair loses one member's repair path while the announcement path still covers it.

Rotation is not something a whole fleet does at once, and handoff certs carry more signatures than quorum, so a single member's temporary loss is tolerable. Tracked for a follow-up rather than this PR.

### Where I don't think the gap is closed: the cert-hash straddle

We discussed resolving this with per-epoch freshness — everyone reads the latest `consensus_pubkey_bytes` at epoch start, so why care about `next_epoch_consensus_pubkey_bytes`. **For committee construction that argument is right**: every node re-reads at epoch start, so epoch E+1's committee, peer maps and self-identity are consistent fleet-wide under the new key with no pending-field handling at all.

It does not extend to `next_committee_pubkey_set_hash`, because that artifact is produced on the other side of effectuation:

1. Mid-epoch E — `next_epoch_committee` appears on chain.
2. Still in E — the syncer re-keys it from the **current** consensus keys (rotation not yet effectuated → old key). Fresh and correct *for its moment*.
3. Still in E — the handoff attestation is signed over that set's hash.
4. Boundary — `advance_epoch` → `rotate_next_epoch_info` → new key.
5. E+1 — fresh read yields the **new** key; re-verifying E's cert gives a different digest.

Both sides read the latest available at read time; the value changed *between* the reads. Re-reading more often cannot fix it, because the producer must run before the boundary (that is what a handoff attestation is) and the consumer after. And unlike (a), this one is **not** per-member and quorum does not absorb it: the hash covers the committee *set*, so one rotated member changes the digest and every node rejects the cert — the same fail-closed halt the fault injection reproduced on all four validators.

Options, none of which are in this PR:

1. Read `next_epoch_consensus_pubkey_bytes` when assembling the next-epoch committee (available already as `VerifiedValidatorInfo.next_epoch_consensus_pubkey`). Closes the common case; leaves a residual window if the rotation is submitted *after* assembly/freeze.
2. Make the hash identity-agnostic — hash validator IDs instead of names. Rotation and the basis flip both stop mattering; costs plumbing (`Committee` does not carry IDs today) and its own transition, which the existing dual-basis tolerance would cover.

My preference is (2) as the durable answer, with (1) as the cheap interim if v6 activation is near.

### Still open, agreed valuable: fail-closed on duplicate `AuthorityName`

Not yet implemented. I agree it should land regardless of how the straddle is resolved — it is Rust-only, cheap, and converts the collision from silent HashMap last-write-wins into a refused committee. Say the word and I will add it to this PR rather than a follow-up.

### On your closing point about the activation clock

Agreed, and worth stating plainly: `MAX = 6` means merging starts the clock, but it does **not** activate anything — v6 requires a capability-vote quorum. The rotation story and the duplicate check must land before any v6 vote on a live network; they do not have to land before this merges. If that distinction is uncomfortable, the alternative is holding `MAX` at 5 in this PR and raising it in the activation PR, which I am happy to do.

### On the smaller items

- The wire-encoding and tolerance arguments you validated: I will add the 2^128-grinding argument and the "alternate set must always be locally derived, never remote-supplied" invariant to `committee-consensus-keys.md` and to the parameter doc, so both are on record rather than in a review thread.
- `load_inner`'s silent `filter_map`: agreed — distinguish the 16-zero-tail (expected consensus-basis) case from genuine garbage and fire `report_invariant_violation!` on the latter, restoring the loudness the old panic had without the panic.

omersadika:
## Correction: finding (c) — the rotation-time identity collision — is WRONG. The chain does enforce uniqueness on rotation.

I need to retract part of my previous reply. I confirmed finding (c) as real, and it isn't. The error was mine: I verified the claim only at the layer the review named (`validator_info.move`) instead of tracing the whole call chain, then reported the absence as fact.

**The rotation path does check for duplicates.** The chain is `system.move:477` → `system_inner` → **`validator_set.move:559`**, and that layer does:

```move
validator.set_next_epoch_consensus_pubkey_bytes(consensus_pubkey_bytes, cap);
self.assert_no_pending_or_active_duplicates(validator_id);   // validator_set.move:568
```

`assert_no_pending_or_active_duplicates` (`:291`) runs `is_duplicate_with_pending_validator` (`:766`) across every other validator in the pending-active set, aborting with `EDuplicateValidator`.

And `is_duplicate` (`validator_info.move:463`) covers precisely the case in question, in all three cross-products:

- `next_epoch_consensus_pubkey_bytes` vs the other's `next_epoch_consensus_pubkey_bytes`
- **my** `next_epoch_consensus_pubkey_bytes` vs the other's **current** `consensus_pubkey_bytes`
- the other's `next_epoch_consensus_pubkey_bytes` vs **my current** `consensus_pubkey_bytes`

plus the consensus↔network key cross-checks. So "an active validator can rotate its next-epoch consensus key to ANOTHER active validator's consensus key" **aborts on chain**. It is not reachable, and it is not "an identity-collision attack for the price of one transaction."

The observation that the `validator_info`-layer setter calls only `self.validate()` is correct — enforcement simply lives one layer up, in `validator_set`, which neither of us checked before I confirmed it.

### What this changes

- **(c) is void.** No rotation-reachable identity collision.
- **The Rust-side duplicate-name check drops from load-bearing to defence-in-depth**, and we are **not** adding it in this PR. The argument for it in the follow-up comment was explicitly "since the missing rotation-time duplicate check cannot be added on-chain, Rust must…" — that premise is false, so the conclusion goes with it. The liveness-vs-corruption trade-off described there does not arise either. (For the record, if two members ever *did* share a name, `index_map` at `committee.rs:311` collapses last-write-wins while `total_votes` counts both — worth knowing, not worth a guard against an unreachable state.)
- **(a) and (b) are unaffected.** They concern a *legitimate* rotation by a single validator, not a collision:
  - **(b), the never-invalidated cache — fixed.** Now dropped at every epoch change; steady-state ticks still perform zero chain reads, since a rotation can only effectuate at a boundary.
  - **(a), prior-epoch artifacts keyed by the old identity — deferred by decision.** Per-member degradation that quorum absorbs; tracked for follow-up.
  - **The cert-hash straddle stands as the open item**, unchanged by this correction, and remains the thing to resolve before any v6 activation vote. It is not a collision problem — it is the producer/consumer ordering: the hash is signed during epoch E from pre-effectuation keys and re-verified in E+1 against post-effectuation keys.

Apologies for the noise — a false "operator-reachable attack" is exactly the kind of claim that gets quoted downstream, so I would rather retract it loudly than leave it in the thread.

omersadika:
## Re-verified the retraction — (c) was my error, confirmed void. Answers on the four open decisions.

**On (c):** I checked `validator_set.move` directly this time: the rotation setter is immediately followed by `assert_no_pending_or_active_duplicates(validator_id)` (set-then-assert, aborting on `EDuplicateValidator`), and `is_duplicate` covers all three next-vs-next / next-vs-current cross-products plus the consensus↔network crosses. The collision is unreachable on chain. The error was mine originally — I traced `system.move` and the `validator_info` setter, saw only `validate()`, and asserted the absence of a check I had simply not found; the enforcement lives in the middle layer I skipped. "Operator-reachable identity-collision attack" is retracted on my side too — thank you for tracing it fully before it got quoted anywhere else.

**The straddle — endorse option (2), hash validator IDs.** Your producer/consumer analysis is exactly right and settles why option (1) can only shrink the window (a rotation submitted after assembly still straddles; the producer must run before the boundary by definition). Validator object IDs are the chain's actual stable identity anchor — immutable, already the join key of every chain read this PR touches — so binding the cert to the ID set removes BOTH the basis flip and rotation from the hash's sensitivity, permanently, and the existing dual-basis tolerance covers its own transition epoch. The cert's security obligation (bind the attestation to a specific next membership) is fully preserved: the consumer knows its committee's IDs from the same trusted read it gets everything else from. (1) only as interim if an activation vote gets scheduled before (2) lands.

**(a) deferral — accepted**, the per-member/quorum-absorbs reasoning is sound and all three degradations self-heal within an epoch. One ask: state the gap in `committee-consensus-keys.md` alongside the boundary caveat, so a rotated-member exclusion during some future epoch reads as a documented degradation, not a discovered incident — the freeze-laggard alert will fire for it and the on-call should find the explanation in the spec, not in this thread.

**Duplicate-name guard — your call stands; one paragraph for the record and then I'll drop it.** The premise correction is real: it's defense-in-depth now, not load-bearing. The case for one cheap assert at committee construction anyway: the invariant's enforcement lives in a different codebase than its consumers, and a future contracts upgrade could weaken it without any Rust reviewer noticing — exactly the cross-codebase drift class from #1929. `index_map` collapsing last-write-wins while `total_votes` counts both is quiet corruption if the unreachable ever becomes reachable. A one-line assert + `report_invariant_violation!` prices that at nearly zero. Non-blocking either way.

**MAX — take your own offer: hold MAX=5 in this PR.** With the straddle explicitly open and (a) deferred, the clean sequencing is: this PR merges inert (all v6 code ships, nothing advertises 6), and the activation PR — raising MAX alongside the straddle resolution and the spec updates — becomes the single deliberate, reviewable event that starts the clock. That also makes the activation PR the natural home for the rotation flow in the `v125_v6_upgrade` gate. Merge-readiness from my side then reduces to: MAX held at 5, the (a) spec note, and the two doc items you already accepted.

omersadika:
## All four decisions implemented — pushed in `40b9abba52`

1. **MAX held at 5.** The whole consensus-key path ships inert: `MAX_PROTOCOL_VERSION` back to 5, the version-6 arm commented out in the replay loop, v6 snapshots dropped. `consensus_key_authority_names()` is false everywhere and the identity flip is unreachable at runtime. Raising MAX is now a single deliberate act in the activation PR.

   `max_version_test` caught the half I'd have missed: with MAX=5 a *live* version-6 arm makes `get_for_version(MAX + 1)` succeed where that test requires a panic. Commenting the arm out — the repo's existing idiom for a planned version — is what makes "defined but unadvertised" actually true rather than nominally true.

   One consequence, documented rather than papered over: `v125_v6_upgrade` asserts the network reaches v6, so **it cannot pass until activation**. It's opt-in (`RUN_V125_V6_UPGRADE`) and never in a default CI selection — PRs run `v125_rollout` — so nothing goes red, but a manual dispatch before activation will fail at the version assertion. That's stated in the scenario's doc comment, the spec, and the ci-suites playbook, and its green run plus fault validation on a MAX=6 build stay on record.

2. **(a) spec note — done, written for the on-call reader** as you asked. `committee-consensus-keys.md` now carries the not-advertised status, both open gaps, and explicitly: a freeze-laggard or excluded-validator alert naming a validator that rotated its consensus key at the preceding boundary is *this accepted degradation*, self-healing the following epoch — not an incident. The three affected sites and their exact preconditions are listed so the reasoning is checkable rather than assertion.

3. **Straddle — recorded as agreed**, with the ID-hash resolution named in the spec as the thing the activation PR must land, plus why read-freshness cannot substitute (producer structurally pre-boundary, consumer post-boundary).

4. **Duplicate-name guard — skipped**, as decided. Your drift argument is on the record in this thread; if #1929's cross-codebase concern turns into a general policy, this is a natural early candidate.

Also landed: the 2^128-grinding argument for why the 48-byte container isn't ambiguous across bases (in the spec), and the invariant on the dual-basis parameter that the alternate set must always be locally derived and never remote-supplied — since that's precisely the property that keeps the tolerance from becoming an acceptance widening.

Merged `origin/main` in first; the only conflict was in `ika-protocol-config`, where main deleted two dead accessors (#1915, #1918) in the same block this branch extended — took main's deletions, kept the live one.

Full suites are running on the merged tree; I'll post the results. Cluster is already green.

omersadika:
## Two decisions reversed by the author — recording them so the thread isn't stale

Both of these overturn things agreed earlier in this discussion, so flagging them explicitly rather than letting them land quietly in a diff.

**1. `MAX` goes back to 6 — this PR advertises v6.** The intent is for this change to actually switch the network to v6, not to ship the path inert. So `MAX_PROTOCOL_VERSION = 6`, the version-6 arm is live, snapshots restored (`2dd861b9d5`). Validators advertise 6 and the capability vote carries a network to v6 once a quorum upgrades: **the identity flip arrives with the rollout, not as a separate decision.** The earlier agreement to hold MAX at 5 and make activation its own reviewable event no longer applies.

**2. The validator-ID hash is declined.** We are not binding `next_committee_pubkey_set_hash` to validator IDs. The rotation straddle is accepted as an **operational hazard** instead of a code fix.

### What that means concretely, for the record

On a v6 network, if any validator rotates its consensus key, the handoff cert's digest is computed during epoch E from pre-effectuation keys and re-verified in E+1 against post-effectuation keys. The digest covers the whole committee set, so this is **not** a per-member degradation that quorum absorbs — it is a fleet-wide fail-closed halt, the same one the fault injection reproduced across all four validators. One `set_next_epoch_consensus_pubkey_bytes` transaction is enough to trigger it at the next boundary.

So: **consensus-key rotation must be treated as an operational hazard on a v6 network, not a routine action.** That is now stated in `committee-consensus-keys.md` next to the boundary caveat.

What I did instead, since a documented hazard is a weak control for something one transaction can trigger (`2613d9d5ab`, diagnostics only): the fail-closed halt messages now name a recent rotation as the **first** thing to suspect on a hash mismatch, ahead of the v5→v6 flip and well ahead of DB corruption, and point at `set_next_epoch_consensus_pubkey_bytes` by name. An operator hitting this reads "check for a recent rotation" rather than being sent to debug storage. It does not reduce the probability of the halt, only the time to diagnose it.

### Validation

`v125_v6_upgrade` passed on the merged tree (`60de2a64b`) — full v5→v6 upgrade with the identity flip, on top of `main`, not just on the earlier standalone commit. Cluster, Rust integration and TypeScript are green there too. Suites are re-running on the MAX=6 HEAD.

Everything else agreed in this thread stands: the per-epoch cache fix, the (a) deferral with its on-call spec note, the 2^128-grinding argument, the locally-derived-alternate invariant, and skipping the duplicate-name guard.

PR #1914, opened 2026-07-25
fix(ocs): assert the chain's anchor objects are typed by the compiled-in package

### LINKED ISSUE #1913: Startup assert: fetched system object's type address must equal the resolved ika_system_package_id (would have caught #1908's constants)
## Follow-up from #1908/#1910

#1908 shipped the system package's **v2 upgrade id** in the compiled-in `ika_system_package_id` — a value that matches zero live event type tags (verified on both networks: 165/165 sampled system events tag the original defining package) — and #1908's design deliberately removes the config escape hatch on public chains, so a wrong constant can only be fixed by another release. #1910 corrected the constants and pinned them in a unit test, but a unit test can only pin what a human once verified: it cannot catch the *next* drift between compiled identity and chain reality.

## Ask

At startup, after the first successful system-object fetch (`verified_system_inner`), assert:

```
system_object.type_.address == resolved ika_system_package_id
```

- On mismatch: fail startup loudly with both values and the remediation ("compiled-in identity does not match the chain's system object type; this binary was built with a wrong or wrong-role package id").
- This is chain-authenticated (the object type comes from verified state), costs one comparison on data the node already fetches, and turns the #1908 failure mode from "fleet silently deaf to system events" into "node refuses to start with an actionable error."
- Optional second assert, same shape, for the coordinator object vs `ika_dwallet_2pc_mpc_package_id` (its type tags the original dwallet package).

## Test

Unit-testable without a chain: feed the verifier a system object typed `A::system::System` while the resolved identity says `B` → startup error naming both. The existing `mid_epoch_restart_keys`-style integration harness covers the happy path implicitly (nodes boot).

omersadika:
Closes #1913.

## Why

#1908 shipped the system package's **v2 upgrade id** as the compiled-in
`ika_system_package_id` — a value matching zero live event type tags — and
because that PR removed the config escape hatch on public chains, the only
remedy was another release. The symptom was a fleet silently deaf to system
events. #1910 corrected the constants and pinned them in a unit test, but a
unit test can only pin what a human once verified; it cannot catch the *next*
drift between compiled identity and chain reality.

## What

Every OCS-verified read of the two singleton anchors now asserts the chain's
object is typed by the package this binary compiled in for that role:

| anchor | expected package |
|---|---|
| `System` | `ika_system_package_id` |
| `DWalletCoordinator` | `ika_dwallet_2pc_mpc_package_id` |

The type comes from OCS-verified state, so the check is chain-authenticated,
and it costs one comparison on data the node already fetches.

Three details worth calling out:

- **Terminal, not retried.** `must_get_system_inner` /
  `must_get_dwallet_coordinator_inner` retry *forever*, so returning a plain
  error would spin with warnings — recreating the silent-failure mode this
  check exists to replace. `ReaderError::is_terminal()` marks the mismatch and
  those loops panic with both values plus remediation.
- **Checked on every read, not only at startup.** The first read happens at
  startup (so the issue's ask is covered), and keeping it on the steady-state
  path costs nothing while also catching a mid-life drift.
- **The expectation comes from `compiled_in_ika_identity`, not the config
  field.** A Sui type tag carries the **defining** package forever, so the
  system object stays `{original}::system::System` across upgrades — this is a
  stable equality check, not one to bump per contract upgrade. That function
  returns the original package and is `None` on Devnet/Custom, where ids are
  generated per genesis and a config-supplied `ika_system_package_id` may
  legitimately name a later upgrade (which would false-positive).

## Verified against the live networks first

So the assert cannot false-positive today:

| network | system object type | compiled-in | |
|---|---|---|---|
| mainnet | `0xb874c9b5…::system::System` | `0xb874c9b5…` | ✅ |
| testnet | `0xae71e386…::system::System` | `0xae71e386…` | ✅ |

## Tests

5 unit tests: the decision rule, the error naming **both** values, the
localnet skip, terminal-ness (and that no other error is terminal), and the
metric label.

One deliberate limitation: the comparison is split into
`check_identity_addresses` so it is testable without constructing a typed
`Object`. The safe Sui test constructors (`treasury_cap_for_testing`) wrap the
caller's `StructTag` in an outer type, making the top-level address `0x2`, and
the direct constructors are `unsafe` — denied workspace-wide. So the tests
cover the decision rule and error contents but not the `type_().address()`
extraction, which is the existing `push_worker.rs:554` pattern. Happy to add
cluster-level coverage instead if you'd rather have it end-to-end.

- `cargo test -p ika-core --lib identity_tests` — 5/5
- `cargo clippy -p ika-core --all-targets` clean; `cargo fmt --all` (verified
  no unrelated churn)

## Docs

`dev-docs/specs/ocs-verified-sui-reads.md` gains invariant 10, recording the
rule, why the type tag makes it upgrade-stable, and the #1908 backstory.

## Open question

Worth also emitting an `ika_ocs_identity_mismatch` counter, or is the
fail-closed panic sufficient? A node that refuses to start is visible by
absence, but a counter would name the cause on a dashboard.

### DIFF
--- a/crates/ika-core/src/sui_connector/ocs_metrics.rs
+++ b/crates/ika-core/src/sui_connector/ocs_metrics.rs
@@ -72,6 +72,14 @@ pub struct OcsMetrics {
     // BagEventPump — omission detection via verified parent state.
     pub bag_omission_suspected_total: IntCounterVec, // labels: ["bag"]
 
+    /// A singleton anchor's on-chain type names a package other than the one
+    /// this binary compiled in for that role. Always fatal (the node refuses
+    /// to run — the constant is baked into the build), so any non-zero value
+    /// names the cause of a node that will not start. Scrape it from the
+    /// crash-loop window or the last scrape before exit; a node down for this
+    /// reason otherwise presents only as absent series.
+    pub identity_mismatch_total: IntCounterVec, // labels: ["anchor"]
+
     /// End-to-end verify latency on the consumer side (transport
     /// round-trip + proof verify). Captures what consumers actually
     /// experience.
@@ -186,6 +194,13 @@ impl OcsMetrics {
                 "ika_ocs_bag_omission_suspected_total",
                 "Bag walk returned fewer children than the verified parent's `Bag.size` claimed (suspected relay omission; could also be a benign race when entries are removed mid-walk — only a hard signal if it persists)",
                 &["bag"],
+                registry,
+            )
+            .unwrap(),
+            identity_mismatch_total: register_int_counter_vec_with_registry!(
+                "ika_ocs_identity_mismatch_total",
+                "A singleton anchor object's on-chain type names a package other than the one this binary compiled in for that role (System -> ika_system_package_id, DWalletCoordinator -> ika_dwallet_2pc_mpc_package_id). Always fatal: the node refuses to run, since a compiled-in constant cannot be fixed by retrying",
+                &["anchor"],
                 registry,
             )
             .unwrap(),
--- a/crates/ika-core/src/sui_connector/setup.rs
+++ b/crates/ika-core/src/sui_connector/setup.rs
@@ -36,7 +36,8 @@ use std::sync::Arc;
 
 use anemo::PeerId;
 use ika_config::node::{
-    SuiConnectorConfig, SuiDataSource, SuiStateMirrorPeer, resolve_sui_checkpoint_archive,
+    SuiConnectorConfig, SuiDataSource, SuiStateMirrorPeer, compiled_in_ika_identity,
+    resolve_sui_checkpoint_archive,
 };
 use ika_network::proof_provider::{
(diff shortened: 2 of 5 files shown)


omersadika:
Reviewed — approve, all green. This implements #1913 better than the issue asked: the assert rides **every** OCS-verified read rather than only the first fetch (continuous, chain-authenticated, one address compare of already-fetched data), covers both singleton anchors, and sources the expected id from `compiled_in_ika_identity` with the correct original-defining-package rationale — including the localnet `None` case where a config-supplied id may legitimately name a later upgrade. The terminal-error class (`is_terminal` → panic instead of retry) replaces exactly the silent 30s-retry-forever failure mode #1882 demonstrated, and `ika_ocs_identity_mismatch_total` with the crash-loop-scrape note makes it visible from outside. The unit-testable decision-rule split (given the pinned Sui's constructor constraints) is a nice touch. Merge-ready.

Issue #1881, opened 2026-07-22
v1.2.3 permanent wedge: cert-path RetryLater deadlock when prior-cert mpc_data blobs are missing locally (10 live validators; no peer refetch; re-hits at every epoch flip)


omersadika:
## Severity: urgent — 10 committee validators are MPC-dead with no self-healing path

Follow-up to #1879/#1880, from live testnet verification 2026-07-22 ~16:30 UTC (epoch 360, ~5.5h in).

## Live evidence

36 committee validators restarted onto v1.2.3 (`194a9db7fe`) around/after the 359→360 boundary. **23 recovered and advance normally (~190 completions/30m) — the #1880 fix works for them.** These 10 have **zero `ika_dwallet_mpc_advance_completions` since boot** (no series at all, 3–5h of uptime):

DTEAM, Perfect-nodes, Ruby Nodes, StakeUp, StakingCabin, Stardust Staking, Studio Mirai, Synergy Nodes, TrustedPoint, bitszn.com | valopers

Everything else on them is healthy — this isolates the failure to exactly one stage:
- `ika_current_epoch` = 360, `announcements_received` = 42 (consensus processing fine)
- `ika_dwallet_mpc_network_key_loaded_epoch` = 360 (network-key adoption fine)
- `canonical_dkg_output_version` = 3 (mirrors fine)
- validator-MPC-keys bundle: never ingested → every session defers

(Separate issue being filed for dWallet Labs 1 / n1stake / Full Sail — that trio is a `sui_state_direct` connector-bootstrap failure, `MissingCommittee` retry-loop, unrelated to this path.)

## Code trace of the deadlock (all in `mpc_manager.rs` @ 194a9db7fe)

1. `try_ingest_current_epoch_keys_from_prior_handoff_cert` assembles cert-359's `ValidatorMpcData` ∩ committee against perpetual `mpc_artifact_blobs`.
2. On `OffChainMpcDataAssembly::Incomplete` it warns (10s-throttled) — **operators can confirm by grepping: "prior-cert mpc_data blobs missing or undecodable in the perpetual store"** (the `missing` field names the exact digests) — and returns `RetryLater`.
3. `ingest_offchain_mpc_keys`: `RetryLater => {}` deliberately does NOT fall back to the freeze channel (correct superset-divergence rationale). So tonight's post-freeze channel delivery will be skipped too.
[...]

Issue #1868, opened 2026-07-20
Freeze-gate deadline: denominate in ika consensus time (fate-shared clock), not machine or Sui-chain time


omersadika:
Follow-up to #1867 (and the tail of #1866): move the freeze-gate deadline from machine wall-clock to **ika's own consensus clock**. Emit-timing-only, no wire change — a correctness-by-construction refinement, not a bug fix.

## Current state (post-#1867)

The ready-signal emit gate compares `now_ms()` (`SystemTime::now()`, validator_metadata.rs:671) against a deadline built from chain-anchored terms: `min(epoch_start_ts + 3/4·duration, first_seen_publication + grace)`. Safety never depended on the clock — the frozen set is a pure function of consensus-ordered signals, and the coverage gates reject premature emissions — but emit timing is a function of 46 NTP opinions, and #1867's publication anchor is each validator's *local observation* time (per-validator skew by construction).

## Why ika-consensus time, specifically (not Sui-chain time)

**Fate-sharing.** The gate's inputs (announcements, ready signals) and output (the freeze at the consensus-ordered quorum point) all ride ika consensus. A deadline denominated in ika commit timestamps can neither outrun nor outlive the machinery it times: if consensus stalls, the deadline pauses — and no clock choice could have produced a freeze anyway, since the signals travel through that same consensus. Zero new liveness dependency, by the same argument that justified the original wall-clock backstop.

**Decoupling.** Sui-chain time ("now" = latest processed Sui checkpoint timestamp) was considered and rejected: it would couple gate timing to Sui liveness, checkpoint cadence, and sui-connector health — an unnecessary external dependency, and against the standing decoupling direction (#1758).

## Design — every term internal and fleet-identical

- **now** — running max of processed `commit_timestamp_ms()` (already read at consensus_handler.rs:199 in the very path that processes ready signals). The max makes it monotone under replay/restart.
- **Backstop anchor** — the commit timestamp of the epoch's **first ika commit**, recorded at epoch-store open; backstop = that + 3/4·duration. Duration stays a config parameter — a number, not a clock, so no domain coupling.
- **Grace anchor** — the commit timestamp at which the **consensus-sequenced publication observation** (`SuiChainObservationUpdate` is already a consensus kind) crossed quorum. Identical on every validator by construction — this eliminates, rather than bounds, the one per-validator term #1867 left (local first-seen).

## Properties gained

[...]

PR #1916, opened 2026-07-25
feat(observability): aggregate operational races and stalls

omersadika:
## Summary

Healthy mainnet validators were producing operational warning noise despite advancing normally:

- 8,353 `tried to add a message to a non-active MPC session` warnings over two hours
- approximately 741 of those warnings in the latest ten-minute sample across five validators
- 553 `Skip scheduling new commit fetches: consensus handler is lagging` warnings over two hours
- periodic off-chain `validator-mpc_data` assembly warnings during ordinary convergence

This change replaces per-event warning noise with bounded counters, fixed-size aggregation, state gauges, transition/recovery logs, and last-success/progress timestamps. It does not change MPC validation, consensus, reconfiguration, checkpoint, OCS proof verification, or message acceptance behavior.

Automatic SuiStateMirror peer discovery, peer selection, and pinned-peer behavior are deliberately outside this PR.

## Why terminal messages can be expected

Consensus can deliver an MPC message after the local session has already completed through output quorum. The message remains rejected by the terminal session, but normal completion makes that arrival an expected asynchronous race rather than an immediate correctness warning.

`DWalletSession::add_message` now returns a structured result. The manager counts ignored terminal messages and handles logging:

- every ignored message is counted
- completed-session arrivals are debug events
- completed-session aggregate warnings require at least 100 events in a 60-second window
- failed-session arrivals remain distinct and trigger an immediate aggregate warning
- warning reminders are limited to one per 60 seconds
- a warned burst emits one recovery transition after a quiet 60-second interval
- an isolated event does not emit recovery noise

The session remains terminal and does not store the message.

## Immediate diagnostics preserved

The existing privacy-safe anomaly diagnostic system remains intact. Rejected outputs, conflicting outputs/split votes, output-quorum rejection, malicious-self recognition, self-output without quorum, and late divergent reconfiguration outputs continue to produce immediate high-severity diagnostics. The permanent `EverythingExcluded` assembly wedge remains an immediate once-per-epoch error. OCS proof-verification and high-water failures retain their existing counters and failure paths.

Expected completion races are not added to anomaly snapshots.

### MPC

- `ika_dwallet_mpc_messages_after_terminal_session_total{terminal_status,session_type}`
  - cumulative count of consensus-delivered messages ignored because the local session was already terminal
  - `terminal_status`: `completed|failed`
  - `session_type`: `user|system|internal_presign|noa_sign|unknown`

### Off-chain validator-mpc_data assembly

- `ika_off_chain_assembly_incomplete`: 1 after the latest incomplete attempt, 0 after success
- `ika_off_chain_assembly_consecutive_incomplete_ticks`: current consecutive incomplete attempts, reset on success
- `ika_off_chain_assembly_incomplete_duration_seconds`: whole seconds since the current incomplete period began, reset on success
- `ika_off_chain_assembly_missing{reason}`: current missing count for one fixed reason; all other reason series are zero
  - reasons: `announcement|blob_missing_or_invalid|source_unavailable|no_input|everything_excluded`
- `ika_off_chain_assembly_last_success_timestamp_seconds`: process-local Unix timestamp of the latest successful off-chain assembly; zero until the first success after restart

Existing `ika_off_chain_assembly_incomplete_ticks_total` and `ika_off_chain_assembly_wedged` semantics remain unchanged.

Assembly retries log at debug during normal convergence. Sustained incompleteness warns after five minutes, with reminders no more than every five minutes. Incompleteness after the epoch's three-quarter backstop or after freeze escalates immediately. Successful assembly resets current-state gauges and emits one recovery transition if the condition had warned.

### OCS

- `ika_ocs_last_successful_relay_timestamp_seconds`: process-local Unix timestamp of the latest untrusted relay read that completed proof verification successfully; zero until the first success after restart

The existing `ika_ocs_proof_verify_failures_total` and `ika_ocs_high_water_violations_total` metrics already covered the requested failures and were not duplicated.

### Consensus

- `ika_consensus_last_committed_timestamp_seconds`: consensus Unix timestamp of the latest commit processed at the Ika commit boundary; zero until this process handles a commit

The value changes only while processing commits, never during metric collection.

## Alert interpretation

- A low completed-session terminal-message rate is expected. Alert on a sustained rate change, especially when paired with output-quorum, rejection, or conflict diagnostics.
- Any failed-session terminal-message rate deserves investigation because the status is separately labeled and warned immediately.
- Alert when `ika_off_chain_assembly_incomplete == 1` and duration keeps growing; page on post-deadline/frozen incompleteness or `ika_off_chain_assembly_wedged == 1`.
- Treat a stale `ika_off_chain_assembly_last_success_timestamp_seconds` only in conjunction with current incomplete state and epoch progress; zero means no success in this process lifetime.
- Treat a stale `ika_ocs_last_successful_relay_timestamp_seconds` only when relay traffic is expected. Pair it with relay/read volume and proof/high-water failures.
- Evaluate consensus progress by changes in `ika_consensus_last_committed_timestamp_seconds` and the existing last-committed leader round, not by scrape time.

The remaining `Skip scheduling new commit fetches: consensus handler is lagging` warning is emitted by pinned upstream Sui Mysticeti (`consensus/core/src/commit_syncer.rs`). This PR does not patch external dependency code merely to silence it.

## Cardinality and privacy

All new labels are fixed enums. There are no session IDs, object IDs, digests, authorities, senders, arbitrary protocol names, error strings, or user-controlled values in metric labels. Aggregation state is fixed size; no unbounded map was added.

Logs contain aggregate counts, durations, fixed reason/status/type codes, and transition state only. They do not contain MPC messages, shares, private keys, decrypted values, payloads, or secret protocol state. Existing privacy-safe anomaly snapshots are preserved.

## Protocol behavior

No protocol behavior changes:

- terminal sessions still reject new messages
- no message required for correctness is dropped earlier than before
- validation and quorum decisions are unchanged
- consensus and reconfiguration decisions are unchanged
- assembly deadlines affect logging severity only
- OCS proof and high-water validation are unchanged

## Tests

Passed:

- `cargo fmt --all --check`
- `cargo check --release -p ika-core --all-features`
- `cargo clippy --release -p ika-core --all-targets --all-features -- -D warnings`
- terminal-message aggregation and bounded-label unit tests
- terminal completed/failed counter, immutable-state, and anomaly-exclusion integration test
- all 14 `dwallet_mpc::integration_tests::anomaly_diagnostics` real-crypto tests
- all 3 `dwallet_mpc::integration_tests::late_reconfiguration_output` real-crypto tests
- assembly state, warning interval, recovery, post-deadline escalation, missing-reason, and committee-assembly tests
- operational metric co-registration and safe-initialization test
- OCS verified-relay timestamp test
- consensus/freeze metric tests, including commit timestamp progress and scrape immutability

Not run locally:

- full `cargo test --release -p ika-core`
- full workspace test suite
- full MPC integration, cluster/simtest, Sui connector, reconfiguration, and consensus suites

Those broad CPU-bound suites are left to CI; the narrow release-mode suites covering every modified behavior passed locally.


omersadika:
## Review: approve — behavior preservation, boundedness, and incident-signal retention all verified against the diff. Two small non-blocking asks.

### Verified

- **Behavior preservation is real, and tested.** The terminal branches of `add_message` return without storing (identical to the old early-return); the manager treats `Stored` as a no-op; the integration test pins terminal statuses unchanged AND asserts zero anomaly snapshots for the race. All immediate diagnostics (rejection, conflicting outputs, quorum failure, malicious-self, `EverythingExcluded`'s once-per-epoch error + wedged gauge) are untouched in the diff.
- **Fixed-size state, no identifiers.** Both state machines (`TerminalMessageLogState`, `AssemblyHealthState`) are scalar-only; the assembly log even *removes* the `?missing` authority list from the previous debug line, replacing it with the bounded reason enum — a privacy improvement on top of the noise fix.
- **Metric hygiene.** All label combinations pre-registered at construction (no absent-series ambiguity — as someone who spends his days distinguishing "0" from "no series" in Mimir: thank you), a test pinning label-set boundedness, and the metrics.md inventory updated.
- **Incident signal is retained, not just noise removed.** The escalation tiers are the right ones: ordinary convergence at debug; 5-minute sustained stall warns (5-min throttle); incomplete assembly **after the 3/4 backstop or when the freeze is already locally visible warns immediately** (`post_deadline || frozen_at_assembly`) — that immediate path is exactly the #1866/#1881-shaped situation where a stall is never benign. The spec note that the wall-clock backstop comparison is severity-only and doesn't feed the consensus-clock freeze decision heads off the #1868-class objection correctly.
- **The Mysticeti line handled at the right boundary**: not patched or rate-limited upstream; instead an Ika-owned progress gauge (`consensus_last_committed_timestamp_seconds`) with the staleness expr documented in the playbook. The "scraping must not change it" test guard is a nice touch.
- Cross-checked our observability infra: nothing greps the removed/demoted log lines (`tried to add a message to a non-active MPC session`, the per-tick assembly wrappers) — no downstream breakage.

### Two small asks (non-blocking)

1. **`AddMessageResult::Stored` is returned on the non-MPC-session branch** (`mpc_session.rs`, the "got a message for a non-MPC session, ignoring" path). The message is ignored, not stored — the enum lies for this branch. Harmless today (the caller no-ops on `Stored`), but the first future consumer that branches on `Stored` semantics (e.g., counting stored messages) miscounts. A third `IgnoredNonMpc` variant — or renaming to `NotIgnoredTerminal`-style neutrality — keeps the contract honest.
2. **A completed-burst warn can suppress the first `failed` warn for up to 60s**: `last_warning` is shared across severities, so a `failed` arrival inside the throttle window of a completed-volume warn produces no log at all (the counter moves; the log arrives at the next reminder). Since `failed` is the correctness-adjacent signal and the PR body promises it an "immediate aggregate warning," consider letting the first `failed` of a burst bypass the shared interval (or a per-severity `last_warning`). ≤60s delay, so genuinely minor — but it's the one place the implementation is weaker than the description.

### Infra follow-up on our side (not this PR)

After merge + deploy I'll adopt four of the new series in the dashboards/alerts: `time() - ika_consensus_last_committed_timestamp_seconds` (strictly better than the `changes()`-based consensus-dead detection — restart-proof and freeze-proof), `ika_off_chain_assembly_missing{reason="blob_missing_or_invalid"}` + `_duration_seconds` (the exact visibility the #1881 incident lacked — we diagnosed it from absent advance series and log greps), and `ika_dwallet_mpc_messages_after_terminal_session_total{terminal_status="failed"} > 0` as a warning-tier alert per the playbook's own guidance.

omersadika:
## Review of c85d4f8374: both asks addressed correctly — approve, ready to merge

**1. `IgnoredNonMpcSession` variant** — exactly right: the non-MPC branch now returns an honest variant, the manager treats it as a no-op alongside `Stored`, and the new unit test pins the contract (`message_for_non_mpc_session_is_reported_as_ignored`). The enum can no longer mislead a future consumer.

**2. Failed-warning bypass** — the per-severity throttle is implemented the way I'd hoped: `Failed` gates only on its own `last_failed_warning` interval, so a completed-volume warn can never delay the first failed-session warning; repeated failed arrivals still throttle at one per 60s. Edge cases I checked and am satisfied with:

- A failed warn stamps **both** timestamps, so it defers the next completed-volume warn by ≤60s — correct direction (the failed warn's payload already carries the completed count, so nothing is lost, and completed is the lower-severity signal).
- The completed branch no longer considers `failed > 0` — fine: a throttled failed count is carried in the payload of whichever warn fires next, so counts can't vanish.
- `check_recovery` resets `last_failed_warning` with the rest of the burst state ✓.
- The new test covers the exact scenario from the review: completed-burst warn at the threshold, then a failed arrival 1s later gets an immediate `Warn{completed: 100, failed: 1}`, and a second failed 1s after that is throttled.

No new issues introduced. Implementation now matches the PR description's "immediate" promise precisely.

PR #1919, opened 2026-07-26
fix(upgrade-test): count late byte-matched authorities toward network-key convergence completeness

omersadika:
## Summary

Fixes the two consecutive `v125_churn` failures on `main` ([30174121356](https://github.com/dwallet-labs/ika/actions/runs/30174121356), [30176583158](https://github.com/dwallet-labs/ika/actions/runs/30176583158)): `"has not yet observed a finalizing quorum … observed 3, quorum=4"`.

**The session had actually converged.** From the node logs of the failing runs: production finalized the 5-member reshare with 3 submissions (v0, v3, joiner), and BOTH remaining validators logged the documented legitimate path — `late network-key reconfiguration output matches the quorum-agreed output; session already completed, output not submitted` — a full minute before the 60s window expired. The observer's completeness bar (`observed submissions ≥ n − ⌊(n−1)/3⌋`) tolerated only ONE late validator, but nothing in the protocol bounds how many finish the heavy class-groups computation after quorum forms on a loaded CI host. Not caused by the recent config PRs — the scenario also failed pre-#1908 on its own branch (different timing assertion), and all validators boot/parse/run MPC fine in the failing runs.

## The fix

Completeness now counts **accounted** authorities: an observed submission, **or** the authority's own late-output metrics proving byte-equality with the quorum output (`output_digest == quorum_output_digest` on `ika_dwallet_mpc_session_late_output_info`). Each validator only exports its own late evidence, so the convergence loop sweeps every committee validator's endpoint (best-effort for non-observers — missing evidence only makes the check stricter) and threads the cluster-wide byte-match set into each observer's assessment. The bar stays at `quorum` accounted authorities (one silently-still-computing straggler can't stall the check), and divergence detection is unchanged: canonical-digest equality still hard-fails, and the scenario's `DIVERGES` log assertion still covers late outputs cluster-wide.

New unit test pins production's observed shape both ways: 3 observed + 2 late-matched of 5 → converged; same submissions without the evidence → incomplete.

## Also surfaced (NOT fixed here — separate issue material)

The same runs show a real, unrelated bookkeeping bug: right after each binary-swap restart, validators emit a milliseconds-long burst (~25×) of `completed internal presign output for a key with no resolvable NetworkKeyId; completion counter not advanced` with `should_never_happen=true` (`mpc_manager.rs:3758`) — consensus replay racing the process-global `ObjectID↔NetworkKeyId` mapping re-registration (`network_key_id_mapping.rs` resets to mainnet/testnet seeds on restart; localnet keys re-register later). The counter-drift heal is already tracked as follow-up in `mpc_manager.rs`. It is *not* the cause of this test failure (bursts predate the failing session by ~25 min; the reshare converged), but it deserves its own fix by the network-key owners.

## Test plan

- `cargo test --release -p ika-upgrade-test --lib` — 27 passed, including the new `multiple_stragglers_with_late_byte_match_evidence_converge`.
- `cargo check --release --all-targets -p ika-upgrade-test` — clean.
- `v125_churn` dispatched on this branch to prove the fix against the real scenario (link in comments when it lands).

### DIFF
--- a/crates/ika-upgrade-test/src/cluster.rs
+++ b/crates/ika-upgrade-test/src/cluster.rs
@@ -748,6 +748,38 @@ impl NetworkKeyBoundaryEvidence {
     }
 }
 
+/// (session, authority) pairs from this metrics body's late-output samples
+/// whose locally computed raw digest matches the known quorum raw digest —
+/// byte-match evidence for the completeness accounting in
+/// [`canonical_network_key_outputs`]. Extraction only: non-matching or
+/// not-yet-compared samples simply contribute nothing (they can never make
+/// the completeness check laxer); the full fail-closed validation of late
+/// samples (divergence, malicious counts) runs in the per-observer
+/// assessment.
+fn late_matched_pairs(body: &str) -> Result<BTreeSet<(String, String)>> {
+    let samples = metric_samples_for_protocol(
+        body,
+        "ika_dwallet_mpc_session_late_output_info",
+        NETWORK_KEY_RECONFIGURATION_PROTOCOL,
+    )?;
+    let mut matched = BTreeSet::new();
+    for sample in samples {
+        let label = |name: &str| sample.labels.get(name).cloned();
+        let (Some(session), Some(authority), Some(output_digest), Some(quorum_output_digest)) = (
+            label("session_id"),
+            label("authority"),
+            label("output_digest"),
+            label("quorum_output_digest"),
+        ) else {
+            continue;
+        };
+        if quorum_output_digest != "unknown" && output_digest == quorum_output_digest {
+            matched.insert((session, authority));
+        }
+    }
+    Ok(matched)
+}
+
 fn canonical_network_key_outputs(
     body: &str,
     committee_authorities: &BTreeSet<String>,
@@ -755,6 +787,12 @@ fn canonical_network_key_outputs(
     // depending on its protocol version.
     candidate_authority: &BTreeSet<String>,
     quorum: usize,
+    // Cluster-wide byte-match evidence for authorities that finalized on the
+    // peers' quorum and discarded their own late output: (session, authority)
+    // pairs whose late `output_digest` equals the session's
+    // `quorum_output_digest`, collected from EVERY committee validator's own
+    // metrics endpoint (each validator only exports its own late evidence).
+    cluster_late_matched: &BTreeSet<(String, String)>,
     observer: &str,
 ) -> Result<NetworkKeyOutputConvergence> {
     let output_samples = metric_samples_for_protocol(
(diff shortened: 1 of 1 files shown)


omersadika:
## Proof run: `v125_churn` passes, and the fix was load-bearing

[Run 30199882645](https://github.com/dwallet-labs/ika/actions/runs/30199882645) — **success** (52 min, full scenario).

A green run alone wouldn't prove much (the run might simply have had 4 submissions, which the old code would also have accepted), so I checked the node logs for whether the new accounting actually carried a boundary. It did — this run **reproduced the exact failure condition** and passed only because late evidence now counts:

Session `0x4d2b71a2edfaf959` (the 5-member epoch-4 reshare, quorum = 5 − ⌊4/3⌋ = **4**) had **two** validators late-discard byte-identical outputs:

```
validator-2  12:19:23.349743Z  late … matches the quorum-agreed output  session 0x4d2b71a2edfaf959
validator-3  12:19:24.087239Z  late … matches the quorum-agreed output  session 0x4d2b71a2edfaf959
```

That leaves **3 observed submissions of 5** at that boundary. The old bar (`observed ≥ 4`) fails there — precisely the `observed 3, quorum=4` signature from the two `main` failures. The harness log shows the check clearing that same boundary seconds later:

```
12:19:38  [flow 35/43] <<< expect_network_key_output_converged([0]) done in 10.4s
```

So the passing run is causal evidence, not a coincidence: same shape as the failures, passing because the two late-matched authorities are now counted as accounted.

Divergence detection is unaffected — the run's own assertions confirm it:

```
log-line assertion passed on every expected validator needle="late network-key reconfiguration output DIVERGES" present=false
log-line assertion passed on every expected validator needle="node recognized itself as malicious" present=false
```

(The other late-discard, validator-2 at 12:09:11 on session `0xa795de86`, was a lone straggler — that boundary passed under the old rule too.)

omersadika:
## Rebased onto `main` @ `9a5ec3fc3c`, re-proven — plus a second bug this uncovered

### Re-validation after rebase caught a regression from #1911

The first post-rebase `v125_churn` ([run 30288992801](https://github.com/dwallet-labs/ika/actions/runs/30288992801)) failed — **not** on this PR's accounting, but on a hard failure that fires before it:

```
observed out-of-committee authorities: observed={4 consensus-key-named},
committee={4 consensus-key-named + 5 BLS-named}
```

#1911 made membership compare against the **union** of a validator's BLS and consensus-key namings, but `add_joiner_validator_inner` only appended the BLS one, so the joiner's outputs read as foreign. Filed as **#1937** (including a quieter second consequence: the index-paired candidate lookup at `cluster.rs:1474-1476` silently `.flatten()`s the missing entry, halving the identity basis without erroring). Fixed here in commit `64a1601019` — three lines — because this PR cannot be validated without it. The commit is separate so it can be reviewed on its own.

Affects only joiner scenarios (`v125_churn` today) and only since #1911 merged this afternoon, which is why the branch's earlier run was green.

### Proof re-run: passing, and still load-bearing

[Run 30293080736](https://github.com/dwallet-labs/ika/actions/runs/30293080736) — **success**, on the rebased tree with both fixes.

Checked again that the new accounting actually carried a boundary rather than the run happening to submit a full quorum. It did: session `0xbfd79725…` had **two** validators (v0 and v3) late-discard byte-identical outputs, leaving 3 observed submissions against quorum 4 — the same shape as the original `main` failures, passing only because late-matched authorities now count as accounted. The other two late-discards in the run were lone stragglers on separate sessions, which the old rule already tolerated.

### State

- Rebased, `MERGEABLE`, no conflicts.
- All 27 harness unit tests pass post-rebase, including `multiple_stragglers_with_late_byte_match_evidence_converge`.
- `cargo check --release --all-targets -p ika-upgrade-test` clean.

Conflict resolution note for reviewers: #1911 changed `candidate_authority` from `&str` to `&BTreeSet<String>`, so all 14 test-call conflicts were resolved by keeping this PR's added `cluster_late_matched` argument and adopting main's `candidate()` wrapper. The production merge is semantically coherent — the late-match filter runs over `committee_authorities`, which is in the same identity basis as the metrics `authority` labels that the pre-existing `is_subset` check already relies on.

omersadika:
Reviewed — approve, all green. The key judgement is sound: counting an authority as accounted via its own late-output metrics proving byte-equality with the quorum-agreed output is STRONGER convergence evidence than observing its submission — it demonstrates the authority computed the same bytes, which is the property the check exists to protect, while the old bar (tolerating exactly one late validator) encoded a bound the protocol never promises on loaded CI hosts. Both failing-run receipts show the documented legitimate late path a full minute before the window expired, so this was the observer wrong, not the network. Harness-only, exact-identity semantics retained. Coordinate the merge order with #1938 per its note.

Issue #1829, opened 2026-07-12
mpc_data freeze writes are not crash-atomic: a partial write is latched as frozen forever (divergent frozen set)


omersadika:
Found independently by two review passes over PRs #1823 and #1827 (pre-existing on `main`, not introduced by either).

## The hole

`freeze_mpc_data_if_first` (`crates/ika-core/src/authority/authority_per_epoch_store.rs`) writes the freeze partition as **per-row direct inserts**:

- `frozen_validator_mpc_data_input_set` — one `insert` per frozen member
- `epoch_excluded_validators` — one `insert` per excluded member

Each insert is its own RocksDB write, and the idempotence guard is **table non-emptiness** (`!tables.frozen_validator_mpc_data_input_set.is_empty()`, plus the outer `!self.is_mpc_data_frozen()` re-arm in the consensus handler).

## Failure scenario

The freeze fires at commit N. The node crashes (or an insert returns a transient RocksDB error via `?`) after inserting k of m frozen rows — with `panic = 'abort'` in release, any concurrent panic makes this window real. Commit N was not yet durably recorded (the `ConsensusCommitOutput` batch writes after the boundary block), so on restart commit N **replays** — but the guard sees a non-empty frozen table and returns `Ok(())`.

Result: this validator is permanently frozen at a strict subset of the committee-agreed set. Post-freeze `decide_assembly_inputs` silently skips the missing members, it assembles a smaller committee class-groups map than every peer, and its reconfiguration output byte-diverges → convicted malicious by the output-quorum byte-equality tally (the documented false-malicious class). A crash between the frozen loop and the excluded loop permanently loses the excluded set as well.

The spec (`validator-mpc-data-announcements.md`, frozen-set invariant) claims atomic batch writes via `ConsensusCommitOutput` — true for `mpc_data_ready_quorum_round`, **false for both freeze tables**. Per the spec/code-disagreement rule, one of them must change.

## Fix direction

Route the freeze partition through the commit's `ConsensusCommitOutput` batch (precedent one screen up: `output.set_mpc_data_ready_quorum_round`), so the frozen + excluded rows land atomically with `last_consensus_stats` — a crash then replays the whole commit and the freeze re-fires cleanly. Alternatively a single `WriteBatch` across both tables, keeping the non-empty gate.

Note: PR #1827 closes the *read-side* flavor of the same fork class (cert read error at the freeze → shrunken carry-forward). This issue is the write side.

[...]

PR #1920, opened 2026-07-26
fix(epoch-close): pin the all_voted input at the quorum commit (#1917)

### LINKED ISSUE #1917: Epoch-close `all_voted` fast path diverges between restarted and non-restarted validators (in-memory EOP aggregator caps at quorum, restart hydration does not)
Found while sweeping the off-chain-metadata gates in #1915 (item 2 of #1909). **Pre-existing — not introduced by that PR**, which preserved the behavior verbatim rather than silently changing consensus-critical close timing. Filing so the fix is its own decision.

## The two views of the EndOfPublish vote set disagree

Every sequenced EndOfPublish vote is persisted unconditionally:

`crates/ika-core/src/authority/authority_per_epoch_store.rs:2322` — `record_end_of_publish_vote` inserts into the `end_of_publish` DBMap for **every** vote.

But the in-memory aggregator is only updated while quorum has not yet been reached:

```rust
// authority_per_epoch_store.rs:5068, in process_end_of_publish_vote
if !end_of_publish.has_quorum() {
    end_of_publish.insert_generic(*authority, ());
}
```

`has_quorum()` is evaluated *before* the insert, so the aggregator freezes at exactly the quorum-crossing membership. For a 4-validator equal-stake committee (`quorum_threshold = 3`, `num_members = 4`): votes 1–3 insert, the 3rd crosses quorum, vote 4 is dropped. `keys().count()` stays at 3 forever.

Restart hydration has no such cap — `StakeAggregator::from_iter` (`stake_aggregator.rs:44`) inserts every row it is handed:

```rust
// authority_per_epoch_store.rs:1965
let end_of_publish =
    StakeAggregator::from_iter(committee.clone(), tables.end_of_publish.safe_iter())?;
```

So a restarted validator rebuilds the aggregator from the **table**, which holds all 4 votes.

## Why that matters: it feeds a consensus-critical close decision

```rust
// authority_per_epoch_store.rs:4439
let (has_quorum, voted_count) = {
    let end_of_publish = self.end_of_publish.lock();
    (end_of_publish.has_quorum(), end_of_publish.keys().count())
};
...
let all_voted = voted_count >= self.committee().num_members();   // :4467
let eop_ready = all_voted || grace_elapsed;                      // :4475
```

`eop_ready` goes straight into `decide_deferred_epoch_close`. Two consequences:

**1. Without a restart, `all_voted` is dead.** It requires `voted_count >= num_members`, but `voted_count` is capped at the quorum-crossing count, and `quorum_threshold < num_members` for any committee with f ≥ 1. So the fast path never fires and **every** epoch close waits out the full `end_of_publish_grace_rounds`, even when all N validators voted promptly. (It does fire on a 1-validator localnet, where `quorum_threshold == num_members == 1` — which is likely why this has gone unnoticed in local runs.)

**2. With a restart, `all_voted` becomes true and the validator closes early.** If all N votes land (typical — validators fire EOP off the same epoch-duration signal) and a validator then restarts inside the grace window, it hydrates `voted_count = N`, computes `all_voted = true`, and closes as soon as the handoff-cert quorum holds — while its never-restarted peers keep waiting for `grace_elapsed`. Different close round ⇒ different final checkpoint content ⇒ that validator's final checkpoint diverges from the committee's.

That directly contradicts the comment above the block (`:4431`):

> The anchor round is persisted, so a validator restarting mid-grace closes at the same round as its peers (the final checkpoint must be deterministic).

The persisted `end_of_publish_quorum_round` anchor does make `rounds_since_quorum` restart-stable. `all_voted` is not covered by that persistence and defeats the intent.

## Fix directions (not decided)

- **Make both views the full set** — drop the pre-quorum guard so the live aggregator also accumulates post-quorum votes. Matches restart behavior and revives the fast path, but *changes close timing on every epoch* (closes become earlier once all N vote). Needs a protocol-version gate: it must not flip per-binary during a rolling upgrade.
- **Make both views the capped set** — have the restart path reconstruct the quorum-crossing membership rather than the full table. Preserves today's timing exactly and removes the divergence, but leaves `all_voted` permanently dead, in which case delete it and the fast path outright.
- **Derive `all_voted` from the persisted table on both paths** — sequence-pure and identical everywhere, but again shifts timing and needs a version gate.

Option 2 is the only one that is not a timing change; options 1 and 3 are protocol-visible and belong behind a new protocol version.

## Suggested validation

Whichever direction is taken, this is the class of property the repo's `dev-docs/playbooks/test-testing.md` exists for: a green cluster run is not evidence on its own. The divergence needs a test that restarts a validator inside the grace window after all N votes have landed and asserts the close round matches its peers — then a fault-injection pass proving that test actually reacts.

## Context

- Surfaced during #1915 / #1909 item 2.
- The current guard is documented in-tree as load-bearing (it is — removing it changes close timing), which reads oddly for something that is arguably a defect. Worth updating that comment to point here once a direction is chosen.

omersadika:
Fixes #1917.

A validator that restarted after every EndOfPublish vote had landed could close the epoch at a **different round than its peers** — divergent close round, divergent final checkpoint.

## The bug

Two views of the vote set disagree by construction:

| | source | membership |
|---|---|---|
| Live | `process_end_of_publish_vote` stops inserting once quorum is reached | quorum-crossing prefix of arrival order (3 of 4) |
| After restart | `StakeAggregator::from_iter` over the `end_of_publish` table, which has no cap | the full vote set (4 of 4) |

`record_end_of_publish_vote` persists **every** sequenced vote, so the table is always complete. `all_voted` (`voted_count >= num_members`) fed `eop_ready` and hence the close decision — false on the live path for any committee of 4+, true on the restart-hydrated path. A validator restarting inside the grace window skipped the grace and closed early while its peers waited it out.

This contradicted the comment sitting directly above the block:

> The anchor round is persisted, so a validator restarting mid-grace closes at the same round as its peers (the final checkpoint must be deterministic).

The persisted anchor makes `rounds_since_quorum` restart-stable. Nothing covered `all_voted`.

## The fix

Persist the aggregator size at the quorum-observing commit — in the same batch as the anchor it sits beside — and read `all_voted` back from storage.

This needs its own record rather than a smarter read of the existing table: `end_of_publish` is a `DBMap<AuthorityName, ()>`, so consensus **arrival order is not stored anywhere**, and iteration is by `AuthorityName`. The quorum-crossing subset is a function of that order and is simply not reconstructible after the fact.

The `None` arm also covers an epoch that reached quorum under a binary predating this record: the in-memory count is then exactly what that binary used at the same commit, so close timing stays identical across the upgrade rather than shifting mid-epoch.

## Why not just delete `all_voted`

It looked dead, but it isn't. With per-validator voting power 1 (`ika-sui-client/src/lib.rs:572`) and the chain's `quorum_threshold = 2*(n/3)+1` (`bls_committee.move`), the capped count is exactly `q`, so `all_voted` ⟺ `q >= n`:

| n | 1 | 2 | 3 | 4 | 5 | 6 | 7+ |
|---|---|---|---|---|---|---|---|
| `all_voted` | **true** | false | **true** | false | false | false | false |

Deleting it would make single-validator localnets (`ika start`) and 3-validator committees wait out the full `end_of_publish_grace_rounds` every epoch. Behavior on the live path is unchanged at every committee size; only the restarted validator stops disagreeing.

## Verification

- New regression test drives the real `process_consensus_transactions` path through an **actual store reopen**, not a simulated one.
- **Fault-validated** per `dev-docs/playbooks/test-testing.md`: reverting `all_voted` to the live count makes it fail on the "must NOT close early" assertion. It discriminates the fix rather than passing vacuously — a green run alone would not have been evidence for a property this cheap to fake.
- Committee size 4 chosen deliberately: the smallest size where the test helper's `(2n).div_ceil(3)` and the chain's `2*(n/3)+1` agree (both 3), so the scenario matches production. At n = 3 they differ and quorum is unanimity on chain, where `all_voted` is legitimately true.
- Whole `authority_per_epoch_store` module green (20/20); `cargo clippy --workspace --all-targets --all-features` clean.

This touches epoch-close timing, so the cluster and upgrade suites should go green before merge.

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -2335,13 +2335,6 @@ impl AuthorityPerEpochStore {
             .collect::<Result<Vec<_>, _>>()?)
     }
 
-    pub fn record_end_of_publish_vote(&self, origin_authority: &AuthorityName) -> IkaResult {
-        self.tables()?
-            .end_of_publish
-            .insert(origin_authority, &())?;
-        Ok(())
-    }
-
     /// Whether `authority`'s EndOfPublish vote has been sequenced and
     /// recorded in this epoch's durable table. The handoff signature
     /// sender uses this to confirm its own `EndOfPublishV2` actually
@@ -2351,6 +2344,11 @@ impl AuthorityPerEpochStore {
     /// epoch boundary (exactly when `EndOfPublishV2` fires) or on crash.
     /// Restart-safe — the table is reloaded into the in-memory
     /// aggregator at epoch-store construction.
+    ///
+    /// Votes become visible here at the batch write of the commit that
+    /// sequenced them, not the instant they are processed (they ride
+    /// `ConsensusCommitOutput`), so a `true` here means the vote is durable —
+    /// which is precisely what this check is asking.
     pub fn has_recorded_end_of_publish_vote(&self, authority: &AuthorityName) -> IkaResult<bool> {
         Ok(self.tables()?.end_of_publish.get(authority)?.is_some())
     }
@@ -4881,7 +4879,7 @@ impl AuthorityPerEpochStore {
     #[instrument(level = "trace", skip_all)]
     async fn process_consensus_transaction<C: DWalletCheckpointServiceNotify>(
         &self,
-        _output: &mut ConsensusCommitOutput,
+        output: &mut ConsensusCommitOutput,
         transaction: &VerifiedSequencedConsensusTransaction,
         checkpoint_service: &Option<Arc<C>>,
         system_checkpoint_service: &Option<Arc<SystemCheckpointService>>,
@@ -4996,7 +4994,7 @@ impl AuthorityPerEpochStore {
             SequencedConsensusTransactionKind::External(ConsensusTransaction {
                 kind: ConsensusTransactionKind::EndOfPublish(authority),
                 ..
-            }) => self.process_end_of_publish_vote(authority),
+            }) => self.process_end_of_publish_vote(output, authority),
             SequencedConsensusTransactionKind::External(ConsensusTransaction {
                 kind:
                     ConsensusTransactionKind::EndOfPublishV2 {
@@ -5021,7 +5019,7 @@ impl AuthorityPerEpochStore {
                 // the handoff cert only needs a quorum of valid
                 // signatures, not all of them.
                 self.record_handoff_signature(handoff_signature)?;
-                self.process_end_of_publish_vote(authority)
+                self.process_end_of_publish_vote(output, authority)
             }
         }
     }
@@ -5096,9 +5094,19 @@ impl AuthorityPerEpochStore {
     /// arms.
     fn process_end_of_publish_vote(
         &self,
+        output: &mut ConsensusCommitOutput,
         authority: &AuthorityName,
     ) -> IkaResult<ConsensusCertificateResult> {
-        self.record_end_of_publish_vote(authority)?;
+        // Through the commit batch, NOT an out-of-band table write. Both
+        // durable views of the vote set must land or be lost together: an
+        // immediate insert survives a crash that loses this commit's batch,
+        // and replay would then rebuild the aggregator from a table holding
+        // votes the replayed commit has not re-counted — pinning a larger
+        // count than the sequence implies and reintroducing #1917 through
+        // crash-replay. Batched, the table reflects exactly the pre-commit
+        // state on replay and the votes re-count through the guard below in
+        // consensus order.
+        output.record_end_of_publish_vote(*authority);
         let mut end_of_publish = self.end_of_publish.lock();
         // Duplicate votes can't double-count (the aggregator is a HashMap).
         // The pre-quorum guard is load-bearing, not an optimization: it is
@@ -5426,6 +5434,12 @@ pub(crate) struct ConsensusCommitOutput {
     verified_dwallet_checkpoint_messages: Vec<DWalletCheckpointMessageKind>,
     verified_system_checkpoint_messages: Vec<SystemCheckpointMessageKind>,
 
+    /// EndOfPublish votes sequenced in this commit. Batched rather than
+    /// written on sight so the durable vote set can never run ahead of the
+    /// commit that sequenced it: on crash-replay the table must reflect the
+    /// pre-commit state, or the aggregator rehydrates with votes the replayed
+    /// commit has not re-counted (#1917).
+    end_of_publish_votes: Vec<AuthorityName>,
     /// First commit round at which the EndOfPublish stake quorum was
     /// observed (the grace anchor). Written through this batch so it
     /// commits atomically with the commit that observed it — an
@@ -5471,6 +5485,10 @@ impl ConsensusCommitOutput {
 
     pub(crate) fn set_dwallet_mpc_round_messages(&mut self, new_value: Vec<DWalletMPCMessage>) {
         self.dwallet_mpc_round_messages = new_value;
+    }
+
+    pub(crate) fn record_end_of_publish_vote(&mut self, authority: AuthorityName) {
+        self.end_of_publish_votes.push(authority);
     }
 
     pub(crate) fn set_end_of_publish_quorum_round(&mut self, round: u64) {
@@ -5651,6 +5669,12 @@ impl ConsensusCommitOutput {
             )?;
         }
 
+        batch.insert_batch(
+            &tables.end_of_publish,
+            self.end_of_publish_votes
+                .into_iter()
+                .map(|authority| (authority, ())),
+        )?;
         if let Some(round) = self.end_of_publish_quorum_round {
             batch.insert_batch(&tables.end_of_publish_quorum_round, [(0u64, round)])?;
         }
@@ -6019,12 +6043,19 @@ mod tests {
         }
         assert!(epoch_store.handoff_signatures_meet_quorum().unwrap());
 
-        // STEP 1: reproduce the live path's two views. Every vote is persisted
-        // (what `record_end_of_publish_vote` does), but the in-memory
-        // aggregator stops at the quorum-crossing membership (what the
-        // pre-quorum guard in `process_end_of_publish_vote` enforces).
+        // STEP 1: reproduce the live path's two views as they stand once
+        // EARLIER commits have durably landed every vote — the table holds all
+        // four, while the in-memory aggregator stopped at the quorum-crossing
+        // membership (the pre-quorum guard in `process_end_of_publish_vote`).
+        // Written straight to the table here because these votes belong to
+        // already-committed batches, not the commit this test drives.
         for name in names.iter() {
-            epoch_store.record_end_of_publish_vote(name).unwrap();
+            epoch_store
+                .tables()
+                .unwrap()
+                .end_of_publish
+                .insert(name, &())
+                .unwrap();
         }
         {
             let mut end_of_publish = epoch_store.end_of_publish.lock();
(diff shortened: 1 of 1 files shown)


omersadika:
## Review: the fix is correct for the reported failure and strictly improves every path — but one narrower instance of the same divergence survives it, through crash-replay at the quorum commit

### Verified against the diff and main

- **Both halves of the bug are real as described**: the live path's pre-quorum guard caps the aggregator at the quorum-crossing membership (`process_end_of_publish_vote`, with its own comment calling the guard load-bearing), and restart hydration rebuilds from the uncapped table (`StakeAggregator::from_iter` at store open). `all_voted` fed by whichever view exists ⇒ restart-dependent close round. Confirmed.
- **The pin is batch-atomic with the anchor** — same `ConsensusCommitOutput`, adjacent inserts in `write_to_batch` — so the two can't disagree with each other. ✓
- **The None-arm upgrade semantics hold**: an epoch that crossed quorum under a pre-record binary pins the current in-memory count, which is exactly what that binary was using at the same commit — close timing can't shift mid-epoch across the upgrade. (The pre-upgrade-restarted validator pins the hydrated count, matching the old binary's own post-restart behavior — no regression, same pre-existing exposure.) ✓
- **The `all_voted` small-committee analysis is right**: unit voting power + `2*(n/3)+1` makes the capped count equal `q`, so n=1 and n=3 keep the fast close and n≥4 keeps the grace. Deleting it would indeed slow localnets. ✓
- **Test quality is the house standard**: real `process_consensus_transactions`, real store reopen, discriminating assertions (fault-validated), and the n=4 rationale (smallest size where the test helper's and the chain's quorum formulas agree) is exactly the kind of footnote that prevents a future "simplify to n=3" from silently voiding the test. ✓

### The surviving window: crash at the quorum-observing commit

`record_end_of_publish_vote` is a **direct insert** to the `end_of_publish` table (`authority_per_epoch_store.rs:2322`) — not routed through the commit batch. Sequence:

1. Commit `Cq` carries the quorum-crossing vote **plus at least one more** (votes cluster at epoch end; multi-vote commits are the common shape). Processing inserts ALL of them into the table directly; the in-memory aggregator stops at the crossing prefix (3 of 4).
2. The eop block pins `voted_count = 3` into the **batch**.
3. Crash before `Cq`'s batch writes.
4. Reopen: the direct inserts survived (WAL), the batch didn't. Hydration builds the aggregator from the full table → 4 of 4, over quorum, so the pre-quorum guard blocks nothing on replay. `Cq` replays: anchor re-pins the same round (deterministic ✓), but the count's None arm now pins the **hydrated 4** — `all_voted` flips true, this validator closes at the next commit, peers wait out the grace. The exact #1917 divergence, needing only a crash in the window between processing `Cq` and its batch write.

This is the same out-of-band-write class as #1829, which #1880's layer B fixed for the freeze path by routing the writes through `ConsensusCommitOutput` — and the new record's own doc comment ("an out-of-band write could desync from the commit on crash-replay", inherited from the anchor's) describes precisely this hazard one table over.

### Recommendation

Route the vote insert through `ConsensusCommitOutput` in this PR (or an immediate follow-up): on crash-replay the table then reflects exactly the pre-`Cq` state, votes replay through the capped live path in deterministic consensus order, and the re-pinned count reproduces the original 3 by construction. One consumer to check when moving it: the handoff-signature sender polls `end_of_publish` to confirm its own vote landed before it stops re-submitting (doc at ~2329) — batching delays visibility to the batch write of the same commit, which if anything is *more* honest about durability for that check.

A failpoint regression (`fail_point_async!(\"crash\")` before the batch write, then reopen — the harness #1880's layer C uses) would pin the whole class rather than this instance.

Approve once the residual window is either closed here or explicitly accepted-and-documented with the follow-up filed; everything else is merge-ready.

omersadika:
The residual window is real and is now closed in this PR (`0df9aba9e5`) rather than deferred. Thanks — the original fix was incomplete and its test could not see the gap.

**Verified the mechanism before acting on it.** `record_end_of_publish_vote` wrote straight to the `end_of_publish` table (`:2338`), outside `ConsensusCommitOutput`, so the durable vote set could run ahead of the commit that sequenced it. Replayed the sequence you described and it holds exactly: the direct inserts survive a lost batch, hydration rebuilds an already-over-quorum aggregator, the replayed commit's pre-quorum guard therefore counts nothing, and the `None` arm re-pins the hydrated size. `all_voted` flips true and the validator closes early — #1917 one layer down.

One refinement to the trigger condition: the commit must carry **more than the crossing vote**. If `Cq` carries only the vote that crosses quorum, the table after replay holds exactly the crossing membership and the re-pinned count is correct. It needs the crossing vote plus at least one more — which, as you say, is the common epoch-boundary shape, so the window is real rather than theoretical.

**Fix taken as recommended**: the vote insert rides `ConsensusCommitOutput` alongside the anchor and the count, so all three land or are lost together. On replay the table reflects the pre-commit state, votes re-count through the guard in consensus order, and the count reproduces the quorum-crossing value by construction. Same class and same remedy as #1829.

`has_recorded_end_of_publish_vote` behaves as you predicted — a vote is visible at the batch write of the commit that sequenced it rather than on sight. Documented at the accessor: for a check whose whole purpose is "did my vote actually land", batch-write visibility is the more honest answer, and the delay is within the same commit.

**On the failpoint suggestion** — I went with a simpler model that needs no harness: "crash before the batch write" is faithfully reproduced by processing the commit and never writing its batch, then reopening. The new test does that and is fault-validated: restoring the out-of-band insert makes it fail on the assertion that no vote survived the dropped batch.

The detail worth recording is that **the pre-existing restart test still passes under that same injection**. This window was genuinely invisible to it and needed its own guard — which is the concrete argument for the class-level coverage you were pointing at. If you still want the failpoint harness for the broader class, happy to file it as a follow-up; I did not want to pull #1880's layer-C machinery in here without you asking for it.

Module green 21/21, clippy clean. All three heavy suites re-dispatched on `0df9aba9e5` — the previous green was on `a6144d3f98` and no longer covers HEAD. Note that Upgrade Test only ran at all because it was dispatched by hand: its path filter covers `crates/ika-core/src/dwallet_mpc/**` but not `crates/ika-core/src/authority/**`, so epoch-close changes skip the mixed-binary gate by default. Worth fixing separately.

omersadika:
## Re-review of 0df9aba9e5: the crash window is closed correctly — approve, merge-ready pending the heavy suites

The fix is exactly the right shape, and I walked every angle I could think of:

**Mechanics ✓** — votes accumulate in `ConsensusCommitOutput.end_of_publish_votes` and are inserted in `write_to_batch` alongside the anchor and the pinned count: all three land or are lost together. On replay the table reflects the pre-commit state (asserted), hydration builds a sub-quorum aggregator, votes re-count through the pre-quorum guard in deterministic consensus order, and the re-pin reproduces the quorum-crossing count by construction. Deleting `record_end_of_publish_vote` outright (rather than deprecating) makes the compiler enforce that no future call site can reintroduce the out-of-band write — better than a lint.

**Consumers checked ✓**
- `has_recorded_end_of_publish_vote` (the handoff sender's confirm-my-vote-landed check): now answers at batch-write of the sequencing commit — strictly more honest for a durability question, and the worst case is one extra harmless re-submit inside the same commit window (both the table insert and the aggregator dedup).
- The first test's STEP 1 switch to direct table inserts is right, and the comment justifying it ("votes belong to already-committed batches, not the commit this test drives") keeps it from reading as a contradiction of the fix.
- Empty-vec `insert_batch` on vote-less commits is a no-op; batch atomicity makes insert order irrelevant.
- Duplicate sequenced votes: `Vec` may carry them across commits, but the table insert is idempotent and the aggregator is a HashMap — no double-count path.

**The regression test is the strongest part.** It models the crash faithfully (process the commit, drop the output, never write — no failpoint machinery to maintain), uses the exact hazard shape (all four votes in the quorum-observing commit), and its assertions discriminate: table-empty-after-crash fails if anyone restores an out-of-band insert, and the replay asserts pin=3-not-4 plus no early close. The commit message's test-testing note — that the *pre-existing* restart test still passes under the injected fault, proving this window was invisible to it — is precisely the evidence standard that makes a green suite mean something.

CI: all fast checks green (clippy both profiles, module 21/21 per the message). As the PR body itself says, this touches epoch-close timing — cluster + upgrade suites green before merge, then ship it. Nice catch-turnaround; #1917 is now closed at both depths.

PR #1921, opened 2026-07-26
fix(sui-connector): detect a dwallet package upgrade the binary can't accept events from

omersadika:
Follow-up to #1914 / #1913, from verifying how Sui package upgrades actually
affect type identity.

## The mechanism (verified, not assumed)

A Sui upgrade does **not** move existing types. Sui records type identity
**per datatype** in the package's `TypeOrigin` table, so a type carried
forward keeps the **ORIGINAL** address while a type **first defined in the
upgrade** carries the **UPGRADE** address. Both are live at once. On-chain
today, identical on mainnet and testnet:

| package | types at original | types at upgrade |
|---|---|---|
| dwallet | 79 | **7** |
| ika_system | 36 | 0 |

Five of the seven dwallet upgrade-defined types are **events**:
`DWalletDKGRequestEvent`, `CompletedDWalletDKGEvent`,
`RejectedDWalletDKGEvent`, `SignDuringDKGRequestEvent`,
`UserSecretKeyShareEventType`.

## The problem

`sui_event_into_session_request` admits an event only if its type address is
in the compiled-in set `{ika_dwallet_2pc_mpc_package_id, …_v2}`. So the set
must grow with **every** upgrade that defines a new event type — and if
nobody remembers, those sessions are silently lost. That is the #1908 failure
shape (a fleet deaf to events because a compiled-in package id didn't match
the chain) reached by a different route. The filter is live on **both**
ingestion paths: the legacy JSON-RPC listener and `bag_event_pump` under OCS.

## The fix

Make the omission **impossible to miss**, rather than trying to be clever
about auto-accepting unknown packages — auto-accepting would admit exactly
the wrong package #1908 shipped, defeating #1914's assert.

Each pump tick compares the coordinator's `package_id` (already read — one
comparison) against the accepted set:

| outcome | meaning | signal |
|---|---|---|
| `Executing(id)` | chain runs a package we don't accept — **events dropping now** | warn + `state="executing"` |
| `Pending(id)` | `new_package_id` staged, not yet migrated to — nothing dropped yet | warn + `state="pending"` |

The `Pending` case is the valuable one: it's **lead time**. Ship the id before
the migration epoch and the executing alert never fires — the difference
between a warning and an incident.

**Deliberately not fatal.** An upgrade defining no new event types is
harmless, and refusing to boot the fleet on a benign upgrade would be worse
than the degradation it prevents.

Also collapses the filter's duplicated two-branch address check into the
shared `accepted_dwallet_packages` (one definition of the admitted set) and
names the unrecognised-package case in the rejection log.

## Why not derive the set from `TypeOrigin` at runtime?

It's the theoretically complete answer and I considered it: fetch the current
package, union the distinct `package` values, no constant can ever drift.
I rejected it for now because it adds a package-fetch path to the verified
read surface (nothing handles `Data::Package` today) for a gap that is
currently latent, and auto-widening what the node accepts is the opposite of
what #1913 asked for. This change makes the manual step *loud*, which is the
smaller and safer half. Worth revisiting if upgrade cadence increases.

## Verification

- 7 unit tests: accepted-set shape, both drift kinds, precedence (executing
  outranks pending), and that a staged id already in the set stops being
  reported — so the alert **clears** once the constant ships.
- `cargo clippy -p ika-core --all-targets` clean; `check-metric-names.sh` OK.
- Alert 4 added to `dev-docs/playbooks/production-alerts.md`.

Stacks conceptually on #1914 but touches no shared files — mergeable in
either order.

### DIFF
--- a/crates/ika-core/src/sui_connector/bag_event_pump.rs
+++ b/crates/ika-core/src/sui_connector/bag_event_pump.rs
@@ -44,7 +44,10 @@ use tracing::{debug, error, info, warn};
 use crate::dwallet_session_request::DWalletSessionRequest;
 use crate::sui_connector::metrics::SuiConnectorMetrics;
 use crate::sui_connector::ocs_metrics::OcsMetrics;
-use crate::sui_connector::sui_event_into_request::sui_event_into_session_request;
+use crate::sui_connector::sui_event_into_request::{
+    PackageDrift, accepted_dwallet_packages, detect_dwallet_package_drift,
+    sui_event_into_session_request,
+};
 use crate::sui_connector::verified_reader::{OcsVerifiedReader, VerifiedObject};
 
 pub struct BagEventPump {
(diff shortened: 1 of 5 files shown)


omersadika:
Closing — this was over-engineered for a problem the existing code already handles.

**The premise was wrong.** I justified this by saying an unknown package
address makes the node "silently deaf" to those events. It doesn't. On the
rejection path today, `sui_event_into_session_request` already logs `error!`
naming the module and address, and `bag_event_pump` logs a second `error!`
naming the event type and entry id. Two error-level lines per rejected event
on the live MPC event path — that is not silent, and the address in the log is
exactly the id a fix would need.

**And the accepted set is already right.** It admits both
`ika_dwallet_2pc_mpc_package_id` and `…_v2`, which is what the type model
requires: a type's address is fixed at the version that introduced it, so the
deployed package's types legitimately span both. Nothing needs widening.

So what this PR added — reading the package's `TypeOrigin` table to derive the
exact introducing-version set, a new verified package-read path, a gauge and an
alert — bought early warning for a step (adding a `_v3` constant alongside a
`_v3` contract upgrade) that this team performs deliberately and that already
produces loud errors if missed. That is not worth a new read path in the
verified-state surface.

What the investigation *did* establish is worth keeping, and is recorded in
`dev-docs/specs/ocs-verified-sui-reads.md` via #1914: an upgraded package holds
a **mix** of type addresses (79 at the original and 7 at the upgrade on both
networks, five of the seven being events), which is why the filter takes a set
and why `…_v2` is load-bearing rather than defensive. I had that wrong twice
before checking it on chain.

#1914 (the #1913 anchor-object assert) is unaffected and stands on its own.

PR #1922, opened 2026-07-26
fix(dwallet-mpc): stop flagging pre-adoption internal-presign replays as should-never-happen

omersadika:
Fixes the `should_never_happen` bursts surfaced by the `v125_churn` investigation in #1919.

## The symptom

Every binary-swap restart produced 20–30 of these per validator:

```
ERROR completed internal presign output for a key with no resolvable NetworkKeyId;
completion counter not advanced  should_never_happen=true
```

## Root cause (from the failing runs' node logs)

Validator-1's timeline:

| time | event |
|---|---|
| 22:30:30 | process starts (binary swap) |
| 22:30:52 | **error burst** |
| 22:30:55 | `adopting a network key whose ObjectID has no NetworkKeyId mapping (not seeded, never instantiated here) — deferring` |
| 22:30:56 | `derived and registered the NetworkKeyId from locally-held key data` |

Consensus replays completions of sessions the **previous process** instantiated, while this process's `ObjectID → NetworkKeyId` mapping — a process-global static rebuilt from the compiled-in seeds — hasn't learned the key yet. Every localnet/CI DKG key sits outside those seeds until the background derivation lands with adoption, seconds later. So the key legitimately does not resolve; the code assumed it always would.

## No counter was ever corrupted

Worth stating plainly, because the log message implies data loss. Resolution is **monotone within a process**: `network_key_id_mapping::register` only inserts (never clears) and installed keys are never evicted, so resolution goes `None → Some` but never back. The only writer of `instantiated_internal_presign_sessions` is the top-up loop, which skips unresolvable keys. Therefore "unresolvable now" ⇒ "never resolved in this process" ⇒ this process never instantiated a batch for that id ⇒ `instantiated == 0`, and the saturating `completed < instantiated` guard would have declined the increment **even if the key had resolved**. The defect was the diagnosis, not the bookkeeping — a `should_never_happen` alarm firing on the most routine operation there is, which trains operators to ignore the flag.

## The fix

Classify into `Resolved` / `AdoptedUnresolvable` / `NotAdopted`:

- `error!(should_never_happen)` is now reserved for **AdoptedUnresolvable** — the genuine invariant break, since adoption defers an unmapped key.
- **NotAdopted** logs `info!` **once per key** (deduped, mirroring the existing `warned_cert_digest_mismatches` pattern), `debug!` for the burst that follows. A skip may be correct, but a *silent* skip never is: a key that is never adopted — failed background derivation, or an overlay entry with no DKG output — never enters the top-up loop and **starves that key's presign pool for the epoch**. This is the recurring evidence of that state.
- Counter behavior is byte-for-byte unchanged.

The classification is a returned value rather than inline logging **specifically so it is testable** — the two non-counting arms differ only in log severity, so no counter-based test could distinguish them.

## Test plan

- New Phase 5 in `test_internal_presign_multi_key_install_lag_keeps_identifiers_uniform` pins all three classes on every validator, reusing the existing adopted-but-unresolvable fixture.
- **Test-tested** per `dev-docs/playbooks/test-testing.md`: injecting the pre-fix behavior fails it with `a not-yet-adopted key is an ordinary pre-adoption replay, not an invariant violation`; reverting restores green.
- Counter invariance stays guarded by the pre-existing `test_internal_presign_stale_batch_expiry`, which asserts `instantiated == completed` through the real output path.
- Full `internal_presign` integration suite: **6 passed**. `cargo check --release --all-targets -p ika-core` clean.
- Reviewed by the `consensus-reviewer` agent (determinism, epoch-boundary, silent-skip lenses). Its findings on my first draft — an orphaned doc comment, a safety comment asserting the wrong property, and the silent `debug!` skip — are fixed in this commit. It confirmed the classification feeds only `tracing` macros, so the per-validator variation in `adopted_network_key_data` has no consensus-visible effect.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
+++ b/crates/ika-core/src/dwallet_mpc/integration_tests/internal_presign.rs
@@ -7,7 +7,7 @@ use crate::dwallet_mpc::integration_tests::utils::{
     TEST_PRESIGN_POOL_MINIMUM_SIZE, apply_test_presign_pool_overrides, build_test_state,
     create_test_protocol_config_guard,
 };
-use crate::dwallet_mpc::mpc_manager::ParkedInternalPresignRequest;
+use crate::dwallet_mpc::mpc_manager::{InternalPresignCompletionKey, ParkedInternalPresignRequest};
 use crate::dwallet_mpc::mpc_session::SessionStatus;
 use crate::dwallet_mpc::{NetworkOwnedAddressSignRequest, ValidatorMpcKeysByPartyId};
 use dwallet_mpc_types::dwallet_mpc::{DWalletCurve, DWalletHashScheme, DWalletSignatureAlgorithm};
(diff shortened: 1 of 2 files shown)


omersadika:
Reviewed — approve, all green. This one matters directly for the alerting channel #1925 just created: a 20–30-burst of false `should_never_happen` per validator per restart would have paged us on every rolling upgrade and trained everyone to ignore the flag — the exact failure mode the marker exists to avoid. The monotone-resolution argument (None→Some only, single writer skips unresolvable keys, so the saturating guard would have declined the increment anyway) cleanly establishes the bookkeeping was never wrong. The tri-state split is the right taxonomy: `should_never_happen` reserved for adopted-but-unresolvable (the true invariant), and — the subtle good call — NotAdopted at deduped `info!` rather than silence, because a never-adopted key starves that key's presign pool for the epoch and deserves one visible line. Merge-ready.

PR #1925, opened 2026-07-26
Export invariant violations as a Prometheus metric

### LINKED ISSUE #1923: Export should_never_happen invariant violations as a metric (ika_invariant_violations_total) — logs-only marking is invisible for 40+ external validators
## Motivation

The codebase marks broken-invariant log events with `should_never_happen = true` — 46 sites on current main (defensive guards in `mpc_manager`, `dwallet_mpc_service`, `sign.rs`, `authority_per_epoch_store`, e.g. "off-chain assembly returned EverythingExcluded for a non-empty prior-cert pair set"). The convention is excellent; its reach is not: these exist **only in logs**, and operators' logs are private. dWallet Labs' Loki-based alerting (just deployed: immediate-fire alert + dashboard row + sweep check on the `should_never_happen` line filter) covers exactly our own pods — for the other 40+ committee validators on each network, an invariant violation is invisible to everyone but the operator who happens to read their logs. This month's incident history (#1856/#1879/#1881/#1882/#1892) shows exactly how much signal hides in per-operator observability gaps.

## Ask

A counter every validator exports:

```
ika_invariant_violations_total{site="<short-slug>"}
```

- **One macro, mechanical adoption**: an `invariant_violation!(site, ...)` (or a wrapper around `error!`) that emits the existing `error!(should_never_happen = true, ...)` AND increments the counter — the 46 sites become a find/replace, and the convention stays a single call for future code.
- **`site` must be a bounded compile-time label** (short slug or `module::fn` literal, NOT formatted values) — cardinality stays fixed at the number of marked sites, consistent with the bounded-labels rule dev-docs/conventions/metrics.md already enforces.
- Pre-register nothing: absent series = zero violations is fine here (unlike state gauges, a counter's absence is unambiguous), but listing the metric in metrics.md and the production-alerts playbook keeps it discoverable.
- Alerting guidance for the playbook: `increase(ika_invariant_violations_total[10m]) > 0` per host — any occurrence is investigation-worthy by the marker's own definition; there is no benign threshold.

## Guard

A lint/CI grep asserting no bare `should_never_happen = true` outside the macro would keep the two channels (log + counter) from drifting apart — same shape as the workspace-version and wire-tag guards already in CI.

## Context

- Fleet-wide visibility gap demonstrated repeatedly during the v1.2.x rollout: every wedge class this month was diagnosed from *metrics* (absent advance series, freeze gauges, reconstruction counters) precisely because external validators' logs are unreachable; `should_never_happen` is currently the only severity-max convention in the codebase with no metric shadow.
- dWallet Labs-side (already live): Grafana alert `ika-should-never-happen` (Loki, fires on any occurrence, no for-duration), dashboard panels 1155–1156, and a sweep `lokiChecks` entry — all documented as our-pods-only until this lands.

omersadika:
## Summary

- add `report_invariant_violation!` to emit the existing structured error and increment `ika_invariant_violations_total{site}` with a literal, bounded site label
- migrate all 46 existing `should_never_happen = true` sites and add a CI drift guard
- document metric inventory and immediate per-host alerting guidance

## Verification

- `cargo test --release -p ika-types invariant_violation_macro_increments_only_for_marked_events`
- `cargo check --release -p ika-node`
- `cargo clippy --all-targets --all-features`
- `./scripts/check-invariant-violation-markers.sh`

Fixes #1923

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,6 +96,10 @@ jobs:
         # commit-batched or written directly (and why that is safe).
         # See dev-docs/conventions/epoch-table-write-discipline.md.
         run: ./scripts/check-epoch-table-write-discipline.sh
+      - name: Invariant violation metric coverage
+        # Every should-never-happen log must increment the fleet-visible
+        # invariant violation counter through the shared macro.
+        run: ./scripts/check-invariant-violation-markers.sh
       - name: Sui version consistency
         # The Sui pin lives in several places that don't update together;
         # see dev-docs/conventions/sui-version-bump.md.
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -2011,8 +2011,8 @@ impl AuthorityEpochTables {
         if presigns.is_empty() {
             // This shouldn't happen, but handle it gracefully: remove the
             // corrupted entry and decrement the size counter atomically.
-            error!(
-                should_never_happen = true,
+            ika_types::report_invariant_violation!(
+                "presign_pool_empty_entry",
                 ?signature_algorithm,
                 ?dwallet_network_encryption_key_id,
                 "prepare_pop_presign: found entry with empty presigns vec, removing"
--- a/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
+++ b/crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/sign.rs
@@ -33,7 +33,6 @@ use mpc::{GuaranteedOutputDeliveryRoundResult, Party, Weight, WeightedThresholdA
 use rand_core::SeedableRng;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
-use tracing::error;
 use twopc_mpc::secp256k1::class_groups::NON_FUNDAMENTAL_DISCRIMINANT_LIMBS;
 use twopc_mpc::sign;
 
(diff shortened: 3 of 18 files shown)


omersadika:
## Review: approve — correct registry wiring, bounded labels, two-directional drift guard

Verified against the diff:

- **The make-or-break wiring is right**: `init_invariant_violation_metric(&prometheus_registry)` sits immediately after `start_prometheus_server` in `run_node` — the registry the metrics endpoint actually serves, node-lifetime, shared by all three binaries. (This is also why the macro beat the tracing-layer alternative I'd floated on #1923: the layer would have wanted a counter at tracing-init time, before the registry exists. Right call.)
- **Label discipline**: `$site:literal` enforced by the macro's matcher; all 47 slugs in the diff are unique (attribution stays exact); absent-series-means-zero is the correct choice for a counter, and the test pins exactly that plus single-series increment.
- **The guard is two-directional** — no bare `should_never_happen = true` outside `metrics.rs`, AND the macro must still emit the marker. The second direction protects the Loki-side alerting we deployed last week (which keys on the literal marker) from a silent refactor. Wired into CI.
- `record_invariant_violation` no-ops before init — tools/tests can't panic. One documented consequence worth a line in the macro doc: a violation fired during early startup (before `start_prometheus_server`) logs but doesn't count. Acceptable and inherent to the design; just worth stating.
- Playbook expr matches the #1923 guidance (per-host `increase(...[10m]) > 0`, no for-duration, no benign threshold).

Infra follow-up on our side once this deploys fleet-wide: a Mimir twin of the existing Loki alert on this counter — that's the moment the coverage extends from our pods to all external validators, closing #1923's original gap.

PR #1931, opened 2026-07-27
feat(protocol): protocol v6 — migrate network-key coefficient bound to strict

ycscaly:
## What

Adds **protocol version 6**, which migrates the network-key DKG/reconfiguration equality-of-coefficients discrete-log bound from the `-10` **relaxed** bound (transcribed on-chain since mainnet-v1.1.8) to the **strict** bound.

The bound is transcribed whole into the Fiat–Shamir challenge, so it is consensus-critical: a mixed-bound committee splits its reconfiguration quorum and flags the odd validator malicious. This must therefore be a coordinated, protocol-gated flip — the whole committee switches at the v6 boundary, never per-binary.

## How

- `MAX_PROTOCOL_VERSION = 6` (`MIN` stays 5).
- New feature flag `strict_network_key_coefficient_bound` — off ≤ v5, on at v6 — exposed as `network_key_backward_compatible_coefficient_bound()` and fed to both the DKG and reconfiguration `backward_compatible` bound args in place of the `is_*_version_v3()` predicates.
- The flag is `skip_serializing_if` false, so the **v5 config digest is byte-identical to v1.2.5's** — v5 stays mixed-binary compatible; strict only activates at v6.
- Uses the **current inkrypto** (both bounds still selectable) — no crypto bump here.
- Protocol-config snapshots regenerated (v6 delta is exactly `strict_network_key_coefficient_bound: true`).

## Why this is step 1 of 2

The strict-only inkrypto rev (which removes the relaxed-bound machinery) can only be adopted once the network is already on the strict bound. This PR is the coordinated migration that must ship and roll out first; the inkrypto bump then becomes a wire no-op on top of it (see the chained PR on `chore/inkrypto-bump-remove-nonaggregated`).

## Testing

- `v125_rollout` (step1 binary vs literal v1.2.5) **PASSED** locally — mixed committee converged across two reshares + full swap at v5 (both relaxed), proving v5 compatibility is unbroken. CI upgrade-test dispatched.
- Protocol-config snapshot tests pass.

https://claude.ai/code/session_01NnRjdvthoSVjPG5UtGtj57


omersadika:
## Post-merge record: this PR was bisect-convicted for the `v125_rollout` regression on main; #1949's revert is now fully validated

For whoever re-lands the strict migration, the receipts:

- **Conviction chain** (single-scenario `v125_rollout` dispatches via `candidate_sha`): green at `fe92cc65e0` (#1919) → green at `cb0940ff1a` (#1939) → green at `b55f9b035c` (#1940) → **red at `8e67703fc3` (this PR)** — the only commit in the failing window. Signature: the current-binary observer's `session_output_info{protocol=network_key_reconfiguration}` samples stayed empty for 60s+ in the mixed phase against literal v1.2.5 binaries, at a moving flow point (17 vs 33 across runs) — i.e. output processing/byte-matching against old-rev peers systematically failing, while chain-side completion (carried by the v1.2.5 quorum) still passed.
- **Mechanism consistent with the diff**: this merge carried the feature branch's history onto main — including the full inkrypto rev bump (`6dca4db3` → `3ed95fe7`) and the removal of the `NonAggregatedPublicOutput` / bwd-compat decode paths — the "step 2 cannot run v5" content the branch's own workflow comments fence off.
- **Post-revert validation** (#1949, main at `4e14c042`): all five suites green including the full Upgrade Test matrix — `v125_rollout` passing again.

Suggestion for the re-land: the chained-PR structure the workflow edit describes (step 1 both-bounds to main, step 2 strict-only fenced behind the v6 activation) is right — the failure mode here was the *merge target*, not the design. A `v125_rollout` gate run on the exact main-merge candidate (not just the feature branch) is the one check that would have caught this pre-merge; `candidate_sha` on the dispatch path makes that cheap now.

Issue #1928, opened 2026-07-27
Presign-pool pops write their own batch, not the commit's: crash-replay can re-pop a different presign (possible #1830 substrate, false-malicious risk)


omersadika:
## Mechanism (to be confirmed at the callers — that is step one)

`pop_presign` / `prepare_pop_presign` (`authority_per_epoch_store.rs` ~1553/~1570) mutate the presign-pool tables in a **self-contained write batch**, committed independently of any consensus commit batch. If pool pops are driven by consensus-deterministic processing (a sequenced sign request consumes the pool head), then the pool's durable state can run AHEAD of the commit that consumed it:

1. Commit C carries a sign request; processing pops presign P (pool batch writes immediately) and records the session assignment into C's batch.
2. Crash before C's batch writes.
3. Replay of C: the assignment is gone, but the pool already advanced past P. The replayed pop yields the **next** presign P'.

If the protocol requires every validator to bind the same presign to the same sign session (pool-head determinism — which is the natural reading of a replicated pool with sequence numbers), the crashed validator now computes the sign session over P' while peers use P → byte-divergent outputs → the **false-malicious class** (honest validator flagged for divergent computation). Even where identity doesn't matter, the skew leaks pool entries and desynchronizes the per-pool sequence counters — which is the exact symptom of long-open #1830 ("pool counters never converge after a start skew beyond one batch lifecycle"); this window may be its substrate, reachable by an ordinary crash rather than a start skew.

## Step one: caller inventory

Enumerate every `pop_presign` call site and classify:
- **Consensus-processing-driven** → the pop must ride `ConsensusCommitOutput` (pool mutation lands or dies with the commit that consumed it; replay then re-pops the same head by construction).
- **Local-only scheduling** (e.g. feeding a local computation queue with no cross-validator identity requirement) → document at the call site why divergence is impossible, per the #1926 write-discipline convention.

## Test

Crash-replay in the #1920 pattern: process a sign-request commit, drop its batch, reopen, replay — assert the replayed session binds the SAME presign id and the pool sequence counters match a never-crashed control store. Fault-validate by restoring the out-of-band pool batch and confirming the test catches the P→P' shift.

Related: #1830 (likely the same root), #1920 (the fix template), #1926.

omersadika:
## Step one done — the inventory changes the fix shape

| site | driver | verdict |
|---|---|---|
[...]

Issue #1927, opened 2026-07-27
Handoff-signature quorum is the last non-deterministic input to the epoch-close gate: buffered signatures drain at wall-clock install time


omersadika:
## Mechanism

The close gate reads `handoff_signatures_meet_quorum()` — a stake sum over the `handoff_signatures` table. That table is written from two paths (`record_handoff_signature`, `authority_per_epoch_store.rs` ~3156/~3183):

1. **Consensus arm** — `EndOfPublishV2 { handoff_signature, .. }` processing (~5019): deterministic sequencing, but the row lands as a direct table write at processing time.
2. **Buffered drain at attestation install** (~2640): signatures that arrived (via consensus) *before* the local attestation existed are buffered in memory and written to the table when `install_expected_handoff_attestation` runs — which is **wall-clock local**: install happens when this validator's own hydrate+build pass completes.

So at a given commit C, validator A (installed early) counts buffered signatures in the gate while validator B (installed late) does not. If the quorum threshold is crossed inside that window, A and B observe the close gate satisfied at **different rounds** → different close rounds → divergent final checkpoints — exactly the property #1920 eliminated from the `all_voted`/grace half of the same gate. Timing makes this plausible rather than exotic: attestation install and `EndOfPublishV2` arrivals both cluster at epoch close, and a validator whose hydration lags (slow disk, restart at the boundary) is the one whose install slips past its peers' signature commits.

Secondary (same mechanics as #1920's second commit): the consensus-arm write is out-of-band relative to the commit batch. The pure-function-of-table argument that makes ready-signals safe (#1926 table) holds here *only* for signatures from the consensus arm; the drain breaks it, because drained rows aren't attributable to any commit ≤ the evaluating one.

## Ask

Preferred: **make the gate's input commit-attributable** — route consensus-arm signature recording through `ConsensusCommitOutput`, and re-sequence drained signatures the same way (the drain re-verifies anyway; instead of writing the table directly, re-inject the verified rows into the current commit's output so the table only ever reflects commit-consistent state). The in-memory aggregator (cert minting) can keep its current shape — the cert only needs quorum-eventually, not round-determinism; it is the *close gate's read* that must be deterministic.

Alternative, if the close round provably does not depend on the handoff-quorum observation round: document that proof at the gate and close this.

## Test

[...]

Issue #1926, opened 2026-07-27
Crash-window audit after #1920: two open determinism questions (handoff-signature close gate, presign-pool pops) and a write-discipline convention for the epoch store


omersadika:
## Context

#1920's second commit fixed a crash-replay divergence caused by an out-of-band table write (`record_end_of_publish_vote`) feeding a consensus-visible decision. This audit walked **every direct DBMap write in `authority_per_epoch_store.rs`** asking the same question: can a crash between processing a commit and writing its batch make replayed state diverge from peers?

## Verified clean (with the reasoning, so it stays clean)

| write | why it's safe |
|---|---|
| processed-marker (`consensus_messages_processed`) | rides `ConsensusCommitOutput` — batched with effects, the #1829 discipline done right |
| ready signals (`epoch_mpc_data_ready_signals`, ~3614, from consensus arm ~4953) | direct write, but every consumer (freeze tally, quorum-round anchor) is a **pure function of the full table**, and a crash can only leave the table ahead by the replayed commit's own content — which the original run also counted before deciding. No order-dependent cap anywhere (that cap is exactly what made the EndOfPublish case a bug) |
| capabilities (`authority_capabilities_v1`, ~2324) | same pure-function argument + generation guard makes replay idempotent |
| announcements + mpc_artifact blobs | per-authority idempotent upsert + content-addressed blobs |
| handoff cert persistence | **persist-on-every-mint-path is genuinely done** — the install/hydration path captures a replay-minted cert and writes it to perpetual (the #1736 lesson, implemented) |

## Q1 — handoff-signature quorum is the remaining non-deterministic input to the close gate

`record_handoff_signature` writes the table from two paths: the consensus `EndOfPublishV2` arm, and the **buffered-drain at attestation install**. The buffered signatures are consensus-sequenced, but the drain happens at local install time — wall-clock, per validator. So at a given commit C, validator A (installed before the signatures' commits) counts them in `handoff_signatures_meet_quorum()` while validator B (installed later) still holds them in the buffer. If quorum crosses inside that window, A and B observe the close gate satisfied at **different rounds** — the exact property #1917/#1920 just eliminated from the EndOfPublish half of the same gate. The timing makes this plausible, not exotic: attestation install and signature arrival both cluster at epoch close.

[...]

PR #1933, opened 2026-07-27
docs(epoch-store): declare the write discipline of every AuthorityEpochTables table

### LINKED ISSUE #1929: Declare the write discipline of every AuthorityEpochTables table (commit-batched vs direct-with-stated-reason)
## Problem

`authority_per_epoch_store.rs` now deliberately contains BOTH persistence patterns:

- **commit-batched** (via `ConsensusCommitOutput::write_to_batch`): EndOfPublish votes, quorum anchor, pinned voted-count, freeze partition + ready-quorum round, processed markers, checkpoint outputs;
- **direct table writes**: ready signals, capabilities, announcements, handoff signatures, presign-pool batches, blob store.

Some direct writes are *provably* safe — but each proof is different (pure-function-of-the-full-table consumers for ready signals; a generation guard for capabilities; content-addressed idempotence for blobs) and none of them is written down at the write site or the table definition. The EndOfPublish bug (#1917/#1920) existed precisely because the safety argument lived in one comment (the aggregator cap) while the write discipline lived nowhere — the cap's order-dependence silently invalidated the "rebuild from the table" assumption a screen away.

## Ask

1. **A convention block on `AuthorityEpochTables`**: every table's doc comment states its write discipline —
   `write-discipline: commit-batched` or
   `write-discipline: direct — safe because <pure-function-of-table | idempotent-replay | local-only | content-addressed>`
   — one line each, citing the consumer that the reason protects.
2. **Review rule in dev-docs/conventions** (alongside the logging/metrics conventions): any new direct write to an epoch table must state its reason in the table doc; any new consumer of a direct-written table must check the stated reason still holds (the reader-side is where #1917 was actually born: `all_voted` consumed a capped view of an uncapped table).
3. Optional enforcement: a grep-based CI check that every field in `AuthorityEpochTables` carries a `write-discipline:` line — same lightweight shape as `check-invariant-violation-markers.sh` (#1925).

## Why this is worth a convention rather than case-by-case vigilance

Four incidents/findings this month reduce to reader/writer discipline mismatches on these tables: #1917/#1920 (capped aggregator vs uncapped rebuild), #1829→#1880-layer-B (freeze writes vs processed marker), #1927 (drain-time writes vs commit-time reads), #1928 (own-batch pops vs commit-batch consumers). The class is clearly load-bearing enough to name.

Related: #1926 (audit), #1917, #1920, #1927, #1928.

omersadika:
Resolves #1929 (split from the #1926 audit, item 3). Convention + docs; no behavior change.

## What

`authority_per_epoch_store.rs` deliberately holds two persistence patterns — rows written through `ConsensusCommitOutput::write_to_batch` (atomic with the commit that produced them) and rows written directly — and nothing marked which table was which, or why the direct ones were safe. #1917 existed precisely because the safety argument lived in one comment (the aggregator cap) while the write discipline lived nowhere.

1. **A convention block on `AuthorityEpochTables` + one `write-discipline:` line per field.** All 55 fields. For direct writes the line carries the reason from a fixed vocabulary (`pure-function-of-table`, `idempotent-replay`, `local-only`, `content-addressed`) **and the consumer that reason protects**.
2. **`dev-docs/conventions/epoch-table-write-discipline.md`** — the review rule, including the reader-side half: a new consumer of a direct-written table must re-check the stated reason still holds. That is where #1917 was actually born (`all_voted` folded an arrival-order-capped view of an uncapped table), so a writer-only convention would not have caught it.
3. **`scripts/check-epoch-table-write-discipline.sh`**, wired into the CI conventions job next to `check-metric-names.sh`.

## Classification summary

- **commit-batched (22)**: processed markers, `last_consensus_stats`, the four EndOfPublish/close tables, the mpc_data ready/freeze anchors + freeze partition, the epoch-first-commit timestamp, and every round-keyed MPC service replay stream (messages, outputs, idle/observation/global-presign/NOA streams, verified checkpoint messages, pending system checkpoints).
- **direct, proven (13)**: capabilities (pure-function-of-table + generation guard), ready signals (pure-function-of-table), announcements (local-only + monotonic upsert), the checkpoint builder/pending/signature tables (idempotent-replay / local-only), `noa_assigned_presigns` + `used_presigns` (idempotent-replay), `presign_private_outputs` (local-only secret), the buffer-stake override (local-only operator knob), both network-output digest caches (content-addressed).
- **direct, UNPROVEN (19)**: the 9 internal presign pools + their size counter + the 8 assigned-presign pools (#1928 — pops commit in their own batch, so a replay can bind a different presign than peers), and `handoff_signatures` (#1927 — the buffered drain writes at wall-clock install time while the close gate reads the table). Deliberately marked rather than given an invented proof.
- **no writer (1)**: `pending_consensus_transactions` — the persist in `ConsensusAdapter::submit` is commented out, so `get_all_pending_consensus_transactions` (the restart-resubmit path) always reads empty. Recorded, not changed.

## Verification

- `./scripts/check-epoch-table-write-discipline.sh` → OK, 55/55; `--list` prints the inventory above.
- **Fault-validated** (the check is the new guard, so a green run alone proves nothing): deleted `idle_status_updates`' discipline line → check names that field and exits 1; restored → OK.
- `cargo check --release -p ika-core`, `cargo clippy --release -p ika-core --all-targets` (no new warnings), `cargo fmt --all -- --check`.

The check is deliberately shallow — it verifies the line exists, not that it is true. It catches the failure mode where a table is added and the argument is never made at all.

Related: #1926, #1917, #1920, #1927, #1928, #1829.

https://claude.ai/code/session_01X2XJjaqFLmR9FKXb3djjLA

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,6 +91,11 @@ jobs:
         # is frozen in scripts/metric-name-allowlist.txt.
         # See dev-docs/conventions/metrics.md.
         run: ./scripts/check-metric-names.sh
+      - name: Epoch-table write discipline
+        # Every AuthorityEpochTables field must declare whether it is
+        # commit-batched or written directly (and why that is safe).
+        # See dev-docs/conventions/epoch-table-write-discipline.md.
+        run: ./scripts/check-epoch-table-write-discipline.sh
       - name: Sui version consistency
         # The Sui pin lives in several places that don't update together;
         # see dev-docs/conventions/sui-version-bump.md.
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -1035,6 +1035,39 @@ type PresignPoolTable = DBMap<(ObjectID, u64), (SessionIdentifier, Vec<(u16, Vec
 type PreparedPresignPop = (DBBatch, SessionIdentifier, u16, Vec<u8>);
 
 /// AuthorityEpochTables contains tables that contain data that is only valid within an epoch.
+///
+/// WRITE DISCIPLINE — read before adding a table, a write site, or a
+/// consumer. This store deliberately holds two persistence patterns, and
+/// confusing them has produced real divergence bugs (#1829, #1917/#1920).
+/// Every field below therefore ends its doc comment with one
+/// `write-discipline:` line, in one of these forms:
+///
+/// - `write-discipline: commit-batched` — the only writer is
+///   `ConsensusCommitOutput::write_to_batch`, so the row lands in the same
+///   atomic batch as the processed-markers of the commit that produced it.
+///   A crash before that batch replays the whole commit and re-derives the
+///   row. This is the required discipline for anything a consensus-visible
+///   decision (close round, freeze partition, checkpoint content) reads.
+/// - `write-discipline: direct — safe because <reason>: <consumer>` — the
+///   row is written outside the commit batch, followed by the argument that
+///   makes the table's consumers survive that. Fixed vocabulary:
+///   `pure-function-of-table` (every consumer folds the WHOLE table, with no
+///   arrival-order or size cap), `idempotent-replay` (re-running the
+///   producing work rewrites the same key with the same value),
+///   `local-only` (nothing consensus-visible reads it — only this node's own
+///   scheduling, secrets, or operator overrides), `content-addressed` (the
+///   key is a hash of the value).
+/// - `write-discipline: direct — UNPROVEN (#NNNN)` — a direct write whose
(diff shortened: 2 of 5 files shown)


omersadika:
Reviewed — approve, all green. Matches the #1929 spec on all three asks, and improves it where doing the work exposed the spec's gap: the third `UNPROVEN (#NNNN)` form is the honest state for exactly the tables whose safety arguments don't close (#1927/#1928) — writing fabricated proofs there would have defeated the convention's purpose. The header's insistence on naming the CONSUMER each direct-write reason protects (because #1917 was born reader-side) is the load-bearing detail. Script fault-validated per the house standard. The audit's two by-products are already tracked (#1935 for the writerless table; the assigned-vs-NOA keying note feeding #1928's fix). Merge-ready — docs and a CI check, no behavior.

Issue #1830, opened 2026-07-12
Internal-presign pool counters never converge after a start skew beyond one batch lifecycle (heal: fast-forward from observed completed sequence numbers)


omersadika:
Found by the review pass over PR #1826 — **pre-existing on `main`** (the old shared counter had the same hole with cross-pool blast radius); the per-pool re-keying scopes it to one pool and the spec/comments now document the exposure, but no heal exists yet.

## The residual

Per-pool internal-presign ordinals are start-time-invariant only while a validator's start skew for a pool stays under ONE BATCH LIFECYCLE. A validator whose first top-up of a pool happens only after a full batch quorum-completed among its peers begins that pool's ordinal stream offset and **never converges**:

- Peers instantiate seq s+1..2s while it instantiates seq 1..s — identifiers of already-completed sessions (`try_activate_internal_presign_request` early-returns on them).
- The offset advances in lockstep (s per drain cycle on both sides), so it never closes. The validator contributes nothing to that pool's live sessions for the rest of the epoch.

Reachable via:
- **Mid-epoch restart** (worst case): the service replays ALL committed rounds in its first iteration, racing the overlay watch publish and the adoption pass — replay with an empty adopted set replays zero top-ups; when adoption lands, every pool starts at 1 against a committee at N. A rolling restart of >f stake mid-epoch starves pools below the MPC threshold → global presigns unservable → the epoch-close undershoot wedge.
- Very late install/adoption of a key on one validator.

## The heal (design sketch from the review)

Fast-forward a pool's counters from the **completed sequence numbers observed in consensus outputs** — the completed session's `session_sequence_number` is already carried in the sequenced stream, so it is consensus-anchored data every validator holds. On processing a completed internal-presign output for a pool whose local `next_internal_presign_sequence_number` is ≤ that sequence number, jump the counter past it (and reconcile the instantiated/completed guards). Every validator computing the same rule from the same stream keeps the counters committee-uniform by construction.

Documented in `dev-docs/specs/fast-schnorr-vss.md` ("Scope of the guarantee", added in PR #1826) as tracked follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1934, opened 2026-07-27
fix(presign-pool): make pool fills and global-presign serves replay-idempotent (#1928)

### LINKED ISSUE #1928: Presign-pool pops write their own batch, not the commit's: crash-replay can re-pop a different presign (possible #1830 substrate, false-malicious risk)

omersadika:
Fixes #1928.

## Step one: the caller inventory

| site | driver | verdict |
|---|---|---|
| `dwallet_mpc_service.rs` global presign requests (`pop_presign`) | the replayed consensus round stream | **unsafe** — self-committing pop, re-run on every restart |
| `dwallet_mpc_service.rs` NOA demands (`assign_noa_presign`) | same | already atomic + idempotent by demand digest |
| `assign_presign` | — | no production callers (tests only) |

The issue proposed routing the pop through `ConsensusCommitOutput`. That isn't the right vehicle: the pop does **not** run inside consensus commit processing. It runs in the `DWalletMPCService` loop, whose round cursor (`last_read_consensus_round`) is in-memory and starts at `None` — so a restart replays **every** consensus round of the epoch, not just the one commit that was in flight.

## The bug

The pool is durable per-epoch state, and the replay does not reset it. Re-executing the rounds therefore does not reproduce the original decisions:

1. **Drain side.** A slot filled late in the epoch, still holding presigns at crash time, is visible to drains replayed at rounds *before* its own fill round. When its sequence number is lower than the head at that round, the replayed drain takes it. That condition is routine, not exotic: a top-up batch starts *s* internal presign sessions at once and they reach output quorum in consensus order, not sequence order, so slot *n+1* filling before slot *n* is ordinary. The replayed request is then answered with a **different presign** than the never-crashed peers put in their `RespondDWalletPresign` checkpoint message for the same presign id — byte-divergent checkpoints from an honest validator, the false-malicious class.

2. **Fill side.** `insert_presigns` was unconditional, so the replay absorbed every internal-presign output a second time. That double-counts `internal_presign_pool_sizes` (roughly doubling it per restart), and the count drives the top-up decision — an inflated count reads as a full pool, suppresses top-ups, and starves the pool while its own gauge reports it healthy. It also resurrects presigns that were already served, which can hand one presign to a second on-chain presign id.

Fixing only (1) would have been net-negative: without (2), replay keeps resurrecting served presigns into a pool no longer drained in lockstep.

## The fix

Both operations record an idempotency marker **in the same committed batch** as the state it guards:

| operation | idempotency key | marker table |
|---|---|---|
| `insert_presigns` | `(signature algorithm, network key id, session sequence number)` — the pool slot | `filled_presign_pool_slots` |
| `serve_global_presign` (new) | request's session sequence number | `served_global_presigns` |

A repeated fill is a no-op. A repeated serve returns the presign it returned the first time, without touching the pool. Replay becomes a no-op on the pool by construction. This mirrors the `assign_noa_presign` pattern already in the file. `pop_presign` stays as the bare self-committing pop, now documented as unusable from any replayed path.

Two new per-epoch DBMaps; no wire format changes, no protocol version gate. On a mid-epoch binary upgrade the markers start empty, so the first restart after the upgrade behaves exactly as `main` does today and every restart after it is clean.

## Verification

- New `presign_pool_replay_after_crash_serves_the_same_presigns` drives a round stream (out-of-sequence fill included) through a control store and a crashed-then-replayed store and compares what each round served, the pool counters, and what a post-replay request draws.
- **Fault-validated** per `dev-docs/playbooks/test-testing.md` — each half of the fix reverted independently:
  - serve idempotency off → `left: [Some([161]), None, None]` vs `right: [Some([176]), Some([177]), Some([160])]`. The replay serves `a1` where peers served `b0`, then runs the pool dry: exactly the P→P′ shift the issue predicts.
  - fill idempotency off → pool counter `5` vs `1`, the double-count.
- `authority::authority_per_epoch_store` module 22/22 green; `presign_consensus` 4/4; `validator_restart` 4/4; `cargo clippy --release -p ika-core --all-targets --all-features` clean.

## Scope

This does not address #1830 — the per-pool *session ordinal* counters are a separate in-memory divergence with a separate heal (fast-forwarding from completed sequence numbers observed in consensus outputs). #1928 hypothesized this window might be #1830's substrate; it is a distinct defect, and the spec says so.

Behavioral spec added at `dev-docs/specs/internal-presign-pool.md`.

This touches MPC state that feeds checkpoint content, so the cluster and upgrade suites should go green before merge.

https://claude.ai/code/session_01FKgoPYkWZUQwM5fgxY5ofZ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -276,21 +276,31 @@ already validated, branch names, and in-flight CI run IDs/URLs.
   the commit/round semantics ika's freeze and epoch-close logic rely on
   (leader rounds, commit boundaries) live in Sui's `consensus/core`, not
   in ika (see the upstream reference above)
-- **Protocol v5 ONLY (`MIN_PROTOCOL_VERSION = MAX_PROTOCOL_VERSION = 5`)**:
-  mainnet and testnet both run protocol v5, and the binary supports
-  nothing older — protocol v3/v4 support, the v3→v4 migration
-  scaffolding (issue #1751), the backward-compatible (class-groups-only)
-  network DKG/reconfiguration parties, and pre-aggregation (V3-tagged)
-  output PRODUCTION are all removed. Everything formerly v4/v5-gated
-  (off-chain validator metadata, the cross-epoch handoff, deferred epoch
-  close, aggregated network-key outputs) is unconditionally on. What
-  REMAINS constrained: the networks persist pre-v5 state (V1-tagged
-  chain DKG anchors, V2/V3-tagged outputs), so old `Versioned*` enum
-  variants must stay (BCS variant indices are wire format) and their
-  DECODE paths must keep working; and MPC outputs must reach
-  byte-identical quorum across a mixed-binary committee, so
-  serialization changes to active paths need a new protocol version
-  gate, never a binary-driven flip.
+- **Protocol v5–v6 (`MIN_PROTOCOL_VERSION = 5`, `MAX_PROTOCOL_VERSION = 6`)**:
+  the binary supports nothing older than v5 — protocol v3/v4 support, the
+  v3→v4 migration scaffolding (issue #1751), the backward-compatible
+  (class-groups-only) network DKG/reconfiguration parties, and
+  pre-aggregation (V3-tagged) output PRODUCTION are all removed.
+  Everything formerly v4/v5-gated (off-chain validator metadata, the
+  cross-epoch handoff, deferred epoch close, aggregated network-key
+  outputs) is unconditionally on. **v6 gates one flag,
+  `consensus_key_authority_names`**: `AuthorityName` becomes the Ed25519
+  consensus key (zero-padded into the same 48-byte container, so the wire
+  encoding is unchanged) instead of the BLS protocol key, which stays on
+  chain and on `Committee` for BLS checkpoint-certificate verification.
+  v6 is ADVERTISED, so the capability vote carries a network to it once a
+  quorum upgrades — the flip is live on rollout, not a separate decision.
+  Its known boundary limitation (the next-epoch committee is assembled
+  under the current epoch's version but consumed under its own, so the two
+  disagree at the single activation boundary) is why the first flip
+  attempt wedged: `dev-docs/specs/committee-consensus-keys.md`. v7 is
+  planned for `noa_checkpoints`. What REMAINS constrained: the networks
+  persist pre-v5 state (V1-tagged chain DKG anchors, V2/V3-tagged
+  outputs), so old `Versioned*` enum variants must stay (BCS variant
+  indices are wire format) and their DECODE paths must keep working; and
+  MPC outputs must reach byte-identical quorum across a mixed-binary
+  committee, so serialization changes to active paths need a new protocol
+  version gate, never a binary-driven flip.
 - **NOA not live**: the network-owned-address (NOA) system — both NOA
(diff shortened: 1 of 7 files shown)


omersadika:
Reviewed in full against the #1928 design — approve, conditioned only on the Test Cluster run I just dispatched on this branch ([30301611830](https://github.com/dwallet-labs/ika/actions/runs/30301611830); Upgrade Test is already green twice here). Verified: both idempotency markers ride the SAME batch as the state they guard, with both failure directions documented at the write site (fill-without-marker → re-absorption; marker-without-fill → lost presigns); `serve_global_presign` extends `prepare_pop_presign`'s existing batch rather than writing a second one, so the pop and its record are inseparable by construction; `pop_presign` survives only with the non-replayable contract stated at both the trait and the inherent method; and the service call-site carries the full why. One assumption I chased to ground: `served_global_presigns` is keyed by `session_sequence_number` alone — safe, because that's the coordinator's single chain-assigned per-epoch sequence (`next_session_sequence_number`, coordinator_inner.move), globally unique across algorithms/keys, mirroring `noa_assigned_presigns`' demand-digest key. The replay-harness test encodes the exact lower-sequence-fills-later hazard, and the per-half fault validation from the issue thread (a1-vs-b0 serve shift; 5-vs-1 pool count) meets the standard. The write-discipline annotations flip two UNPROVEN entries to proven — #1933's convention doing its job on its first customer.

Issue #1909, opened 2026-07-25
Post-v5-cleanup follow-ups: reconstruction removal (gauge-gated), always-true flag sweep, v6 upgrade-test resurrection


omersadika:
Follow-ups deferred from #1895 (protocol v3/v4 support removal, `MIN_PROTOCOL_VERSION = 5`), each with its own trigger:

## 1. Remove the network-DKG anchor-reconstruction machinery
`reconstructed_full_network_dkg_output` / the V1|V2→V3→V4 canonical-mirror flip in `mpc_manager.rs`, the mismatch-tolerance arms in `adopt_cert_verified_keys`, and `already_adopted_key_survives_dkg_output_migration`.

**Trigger (verify BEFORE removing):** the `ika_dwallet_mpc_network_encryption_key_canonical_dkg_output_version` and `..._latest_reconfiguration_output_version` gauges read **4 on every validator of both networks**. Until then a validator offline since before the v5 flip still needs the migration on its next instantiation. Keep the `Versioned*` enum variants regardless (BCS indices are wire format; V1 chain anchors persist forever).

## 2. Sweep the always-true v4-era feature-flag gates
`off_chain_validator_metadata_enabled()` and friends are true at every supported version; #1895 collapsed them only inside the network-key flows. Remaining dead-but-present branches: the epoch-close inline (non-deferred) path in `authority_per_epoch_store`, node-level blob-source install gating, handoff trigger gates, and the gates #1899 added. Mechanical; no behavior change on any supported network.

## 3. Resurrect the gradual protocol-upgrade cluster test when v6 is defined
`protocol_version_transition.rs` was deleted (no boundary exists below MAX). Resurrection templates:
- **v4→v5 retargeted flavor** (modernized, per-validator supported-version pins): PR #1895 branch commit `875e7e53d4` — `crates/ika-test-cluster/tests/protocol_version_transition.rs`
- **original v3→v4 flavor**: main commit `d3be82c2b0` (last mainline commit before the merge), same path.

Same trigger applies to the rolling-upgrade rehearsal machinery: `v125_rollout`/`v125_churn` cover binary swaps at v5; a v6 boundary rehearsal reintroduces a version-pinned OLD build (pattern: the pre-#1895 workflow's `old_max_protocol_version` input, main commit `d3be82c2b0`).

## 4. BLS protocol-key cleanup
Tracked in #1751's comment thread — different trigger (checkpoint certs migrating off the BLS aggregate); unaffected by this PR.

[...]

PR #1936, opened 2026-07-27
test(cluster): resurrect the gradual protocol-upgrade test at the v5→v6 boundary (#1909 item 3)

omersadika:
Completes item 3 of #1909 — the half #1911 did not deliver.

#1911 raised `MAX_PROTOCOL_VERSION` to 6, so a protocol boundary below MAX exists again and the gradual-upgrade cluster test (deleted with the v3/v4 removal) comes back, retargeted v5→v6 from the template the issue names (commit `875e7e53d4`). #1911's `v125_v6_upgrade` already covers the out-of-process, real-old-binary half; this is the in-process half — one binary, per-validator pinned advertised ranges — and it runs in the ordinary cluster suite on **every PR**, not just when the release-gated upgrade harness is dispatched.

## What it asserts

Boot at v5 with supporters 1/4 → upgrade in place (stop, mutate `supported_protocol_versions`, restart) → 2/4 → 4/4:

| End-of-epoch vote | supporters | next epoch |
|---|---|---|
| 1 | 1/4 | v5 |
| 2 | 2/4 | v5 |
| 3 | 4/4 | **v6** |
| 4 | 4/4 | v6 |

The two negative assertions are load-bearing: n=4 with the default 5000bps buffer stake makes the effective threshold 4/4, so 1/4 and 2/4 must NOT activate. The 3→4 boundary is the cross-basis reconfiguration (#1911's hard case — committee assembled under BLS names, entered under consensus-key names), and the final epoch confirms a full reconfiguration entirely under consensus-key names.

## Retargeting was not mechanical

Three harness realities changed since the template: nodes boot at epoch 1 (the progression table shifts by one), `trigger_reconfiguration` no longer exists (epoch advance is time-driven), and epochs are 20s so each stop-mutate-restart lands its re-submitted capability notification within the epoch its vote is charged to.

## Validation (both on CI, filtered to this test)

- **Real run green**: [30290428488](https://github.com/dwallet-labs/ika/actions/runs/30290428488) — PASS in 224s.
- **Fault-validated** per `dev-docs/playbooks/test-testing.md`: [30290430401](https://github.com/dwallet-labs/ika/actions/runs/30290430401) withheld validator[3]'s upgrade (3/4 = stake quorum, below the effective threshold) and failed at exactly the activation assertion — `left: ProtocolVersion(5), right: ProtocolVersion(6)` at line 147, identically across nextest's retry. The test can see a missed activation, and the buffer-stake threshold is enforced reality, not documentation. Fault branch deleted after the run.

Also records the in-process counterpart in `dev-docs/specs/cross-binary-upgrade.md`, whose scenario list previously named `v125_v6_upgrade` as the only protocol-boundary coverage.

## Where this leaves #1909

Items 2 and 3 done; item 1 still gauge-gated (needs the two network-key version gauges reading 4 fleet-wide on both networks), item 4 still waiting on checkpoint certs migrating off the BLS aggregate.

### DIFF
--- a/crates/ika-test-cluster/tests/protocol_version_transition.rs
+++ b/crates/ika-test-cluster/tests/protocol_version_transition.rs
@@ -1,0 +1,165 @@
+// Copyright (c) dWallet Labs, Ltd.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+//! Gradual protocol-version transition (v5 → v6) across multiple epochs.
+//!
+//! Models a real rolling upgrade: the network boots at protocol v5 with one
+//! validator supporting `5..=MAX` and the other three pinned at `5..=5`. The
+//! end-of-epoch capability quorum vote
+//! (`choose_highest_protocol_version_and_move_contracts_upgrades_v1`) needs
+//! enough stake to support a higher version before `SetNextConfigVersion`
+//! fires.
+//!
+//! v5 → v6 is the boundary where a validator's `AuthorityName` — its
+//! canonical committee identity — flips from the BLS protocol key to the
+//! Ed25519 consensus key (`consensus_key_authority_names`). The boundary is
+//! the asymmetry the flip's bug history lives in: the next-epoch committee is
+//! assembled mid-epoch under the OUTGOING version's identity basis but
+//! rebuilt by the entering epoch under its own, so the v6-activation
+//! reconfiguration crosses bases mid-flight. This is the in-process
+//! counterpart of the out-of-process `v125_v6_upgrade` rehearsal
+//! (`crates/ika-upgrade-test`): that one swaps REAL old-release binaries and
+//! gates deployed-release compatibility; this one runs a single binary with
+//! per-validator pinned advertised ranges, gating the capability-vote
+//! machinery and the in-process transition — including the epoch-boundary
+//! restart/rebuild paths the out-of-process harness does not exercise.
+//!
+//! (The v4 → v5 flavor of this test was retired when v3/v4 support was
+//! removed and MIN = MAX = 5 left no boundary to cross; this resurrects it
+//! retargeted at the v6 boundary.)
+//!
+//! Effective threshold note: ika inherits Sui's
+//! `buffer_stake_for_protocol_upgrade_bps` (default 5000). For `n=4` (f=1,
+//! quorum=3) the effective threshold is `quorum + ceil(f * 5000/10000) = 4`,
+//! so a protocol upgrade requires **all 4** validators to support the new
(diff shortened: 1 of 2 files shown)


omersadika:
## Verified — code, math, and both CI receipts check out; one claim is false and it's about to be enshrined in the spec doc

### Confirmed against primary evidence

- **The test matches every structural claim**: the 1/4 → 2/4 → 4/4 progression with both load-bearing negative assertions, the cross-basis reconfiguration at the epoch 3→4 boundary, and the full-under-v6 reconfiguration at 4→5 (activation alone would pass a vote-counting bug that broke the flip epoch itself — the extra epoch is what makes this a transition test, not a tally test).
- **Threshold math independently re-derived**: n=4 unit stake → f=1, quorum = 2·(4/3)+1 = 3, buffer 5000bps → effective threshold = 3 + ⌈1·0.5⌉ = 4. Matches the shared #1878 formula, and the fault choice (withhold one → 3/4 = exactly stake quorum, below threshold) discriminates the *buffer* enforcement specifically, not just quorum counting. Well-chosen.
- **Green run [30290428488](https://github.com/dwallet-labs/ika/actions/runs/30290428488)**: real, Test Cluster on the PR branch, success, this exact test visible in the log (SLOW markers consistent with the 224s claim).
- **Fault run [30290430401](https://github.com/dwallet-labs/ika/actions/runs/30290430401)**: real, failure, and the log shows precisely the claimed signature — `protocol_version_transition.rs:147:5`, `left: ProtocolVersion(5) / right: ProtocolVersion(6)` on the "epoch 4 should advance to v6" assertion — **twice, identically**, confirming the deterministic-across-retry claim.

### The correction needed before merge

**"It runs in the ordinary cluster suite on every PR" is false.** `test-cluster.yaml` is `workflow_dispatch`-only — the header's own words are "Manually triggered." The continuous element is the nightly **Scheduled — All Heavy Suites** run plus per-branch manual dispatch (which is how both receipts above were produced). The PR body mis-states this, which is cosmetic — but the same sentence lands in `cross-binary-upgrade.md` ("gated continuously rather than only when this release-gated harness runs"), where it becomes exactly the docs-vs-reality drift the #1901 verification pass existed to remove. One-line fix: "runs in the cluster suite (nightly scheduled + on dispatch)" — or, if per-PR gating is genuinely wanted for this test, that's a `pull_request` trigger discussion for the workflow, not a doc claim.

Approve with that line corrected. Beyond it: 20s epochs with stop-mutate-restart inside a charged epoch is the test's one timing sensitivity — the doc comment says so explicitly and the green run shows margin, so noted rather than objected.

PR #1938, opened 2026-07-27
fix(upgrade-test): register a joiner under both authority namings

### LINKED ISSUE #1937: upgrade-test: a joiner is registered under only one authority naming, breaking v125_churn (regression from #1911)
## Summary

`#1911` (AuthorityName → consensus key at protocol v6) made harness committee-membership checks compare against the **union** of a validator's two identity namings — BLS protocol key and consensus key — but `ClusterOfProcesses::add_joiner_validator_inner` only appends the BLS naming. Any validator that joins mid-scenario is therefore invisible under its consensus-key naming.

Only scenarios that add a joiner are affected. Today that is **`v125_churn`**, which fails deterministically:

```
observed out-of-committee authorities for network-key session 50f21a75…:
  observed={4 consensus-key-named}
  committee={4 consensus-key-named + 5 BLS-named}
```

Four consensus-key entries for a five-member committee — the joiner's is missing, so every output it submits under that naming trips the `is_subset(committee_authorities)` hard failure in `canonical_network_key_outputs`. That check fires *before* any convergence accounting, so it presents as an unrelated convergence failure.

Reproduced on [run 30288992801](https://github.com/dwallet-labs/ika/actions/runs/30288992801) (a branch rebased onto `main` @ `9a5ec3fc3c`). `main` has not run `v125_churn` since #1911 merged (2026-07-27 17:11 UTC), so this is currently latent there rather than observed — the two `main` failures on 07-25 predate #1911 and are the separate convergence-accounting bug fixed by #1919.

## Two consequences, one cause

1. **Hard failure (what we hit).** The joiner's outputs read as foreign → out-of-committee → scenario aborts.
2. **Silent under-coverage (worse, if it ever bites).** `crates/ika-upgrade-test/src/cluster.rs:1474-1476` builds the candidate identity set by **index-paired** lookup across the two `Vec`s:
   ```rust
   let candidate_authority: BTreeSet<String> = [
       self.validator_authorities.get(*index),
       self.validator_consensus_authorities.get(*index),
   ].into_iter().flatten().map(ToString::to_string).collect();
   ```
   For a joiner index the second `get` returns `None`, and `.flatten()` swallows it. The candidate set silently drops to one naming instead of erroring, so a boundary can be judged inconclusive (or miss byte-equality evidence) with no signal that half the identity basis was never considered.

## Root cause

`validator_authorities` and `validator_consensus_authorities` are two parallel `Vec`s whose entries **must** stay index-aligned (the lookup above depends on it), but nothing enforces the invariant and each has its own write path. `add_joiner_validator_inner` (`cluster.rs:1966`) updates one and not the other, and the desynchronization is silent.

## Fix

Three lines registering the joiner under both namings, in [#1919](https://github.com/dwallet-labs/ika/pull/1919) (commit `64a1601019`) because that PR could not otherwise be validated:

```rust
self.validator_consensus_authorities
    .push(AuthorityPublicKeyBytes::from_consensus_key(
        init.consensus_key_pair.public(),
    ));
```

## Suggested follow-up (not in that PR)

The point fix restores correctness but leaves the fragile shape. Worth considering by whoever owns #1911:

- Replace the two index-paired `Vec`s with one `Vec` of a small struct holding both namings, so a validator cannot be half-registered by construction; **or**
- assert `validator_authorities.len() == validator_consensus_authorities.len()` at every mutation and read site, so a future writer that updates one list fails loudly instead of silently under-covering.

Also worth a look: whether any other harness or product path that grew a second identity basis in #1911 has a writer that updates only one of them.

cc @omersadika (author of #1911)

omersadika:
## Summary

Fixes #1937.

#1911 made the harness committee-membership checks compare against the union of a validator's two identity namings — BLS protocol key and consensus key — held in two parallel index-aligned `Vec`s. `add_joiner_validator_inner` appended only the BLS naming, so a mid-scenario joiner was invisible under its consensus-key naming:

1. **Hard failure**: every output the joiner submitted under the consensus-key naming tripped the `is_subset(committee_authorities)` out-of-committee check in `canonical_network_key_outputs`, failing `v125_churn` deterministically (reproduced on [run 30288992801](https://github.com/dwallet-labs/ika/actions/runs/30288992801)).
2. **Silent under-coverage**: the index-paired candidate lookup `.flatten()`ed a `None` from the second `Vec`, so a joiner's candidate set silently dropped to one naming instead of erroring.

## Fix

Structural, per the issue's suggested follow-up: merge `validator_authorities` / `validator_consensus_authorities` into one `Vec<AuthorityNamings>` — a per-validator record holding both namings, derived once from the init config by a helper shared between the builder and the joiner path. A validator can no longer be half-registered by construction, and an out-of-range candidate index now errors instead of shrinking the set.

This subsumes the 3-line point fix carried in #1919 (commit 64a1601019, added there for validation); whichever merges second takes a small conflict in `add_joiner_validator_inner`.

## Validation

- `cargo clippy --release -p ika-upgrade-test --all-targets` clean; all 26 `ika-upgrade-test` unit tests pass.
- `v125_churn` (the deterministic reproducer) dispatched on this branch: [run 30295000573](https://github.com/dwallet-labs/ika/actions/runs/30295000573) — will report the verdict on this PR.

### DIFF
--- a/crates/ika-upgrade-test/src/cluster.rs
+++ b/crates/ika-upgrade-test/src/cluster.rs
@@ -86,17 +86,10 @@ pub struct ClusterOfProcesses {
     /// aligned with `validators` by index; joiners append. A mirrored
     /// validator's `sui_state_mirror_peers` is built from these.
     validator_peer_ids: Vec<String>,
-    /// Protocol authorities aligned with `validators`. The convergence check
-    /// compares exact identities, not just a count that a missing validator
-    /// and an unexpected sender could accidentally satisfy.
-    validator_authorities: Vec<AuthorityPublicKeyBytes>,
-    /// The same validators named under the CONSENSUS-key basis. A
-    /// validator's `AuthorityName` is its BLS protocol key below protocol
-    /// v6 and its consensus key from v6, so a harness that knows only one
-    /// basis compares a correctly-flipped cluster against the wrong
-    /// expectation. Both are derived at construction from the init configs,
-    /// which carry both keypairs.
-    validator_consensus_authorities: Vec<AuthorityPublicKeyBytes>,
+    /// Authority namings aligned with `validators`; joiners append. The
+    /// convergence check compares exact identities, not just a count that a
+    /// missing validator and an unexpected sender could accidentally satisfy.
+    validator_authorities: Vec<AuthorityNamings>,
     /// Bootstrap package state (`ika_supply_id` funds joiner stakes).
     packages: PublishedIkaPackages,
     /// Bootstrap system state (`init_system_shared_version` is needed by
@@ -120,6 +113,36 @@ pub struct ValidatorSlot {
     pub address: SuiAddress,
     pub validator_id: ObjectID,
     pub validator_cap_id: ObjectID,
+}
+
+/// One validator's authority identity under BOTH naming bases. A validator's
+/// `AuthorityName` is its BLS protocol key below protocol v6 and its
+/// consensus key from v6, so a harness that knows only one basis compares a
+/// correctly-flipped cluster against the wrong expectation. One record per
+/// validator — not parallel per-basis lists — so no write path can register
+/// a validator under one naming and miss the other.
+struct AuthorityNamings {
+    /// BLS protocol-key naming (`AuthorityName` below protocol v6).
+    protocol: AuthorityPublicKeyBytes,
+    /// Consensus-key naming (`AuthorityName` from protocol v6).
+    consensus: AuthorityPublicKeyBytes,
+}
+
+impl AuthorityNamings {
+    /// Both namings, derived from the init config that carries both keypairs
+    /// — the one derivation genesis validators and joiners share.
+    fn of(init: &ValidatorInitializationConfig) -> Self {
+        Self {
+            protocol: init.key_pair.public().into(),
+            consensus: AuthorityPublicKeyBytes::from_consensus_key(
+                init.consensus_key_pair.public(),
+            ),
+        }
+    }
+
+    fn both(&self) -> [&AuthorityPublicKeyBytes; 2] {
+        [&self.protocol, &self.consensus]
+    }
 }
 
 /// Either a caller-provided persistent dir or a harness-owned temp dir.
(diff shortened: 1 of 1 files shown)


omersadika:
**`v125_churn` verdict on this branch alone ([run 30295000573](https://github.com/dwallet-labs/ika/actions/runs/30295000573)): the #1937 failure is gone; the run then hits the separate, pre-existing convergence-accounting bug that #1919 fixes.**

- The out-of-committee hard failure from #1937 no longer occurs: the joiner folded into the committee (4→5 assertion passed), and the network-DKG/reconfiguration output-version assertions passed — the scenario progressed well past the old abort point.
- The run then failed at the completeness check with `observed={3 authorities}, quorum=4` for one network-key session after the 60s window — the exact signature of the late-byte-matched-authority accounting gap #1919 addresses (`main`'s 07-25 `v125_churn` failures, which predate #1911, have the same shape). This branch does not carry #1919.

This confirms the coupling stated in #1937: `v125_churn` needs both fixes. Validation run with this PR merged together with #1919: [run 30299153116](https://github.com/dwallet-labs/ika/actions/runs/30299153116) (branch `claude/issue-1937-validate-with-1919`) — will report here.

omersadika:
Reviewed — approve, all green. The structural fix is the right call over the 3-line point fix: one `AuthorityNamings` record per validator, derived by a single shared helper from the init config, makes half-registration unrepresentable — the bug class is gone, not patched. The second (silent) defect is the more valuable catch: the index-paired `.flatten()` shrinking a joiner's candidate set to one naming would have quietly weakened every future convergence check that ran through it. Reproduction receipt on the failing run checks out. Merge note from the PR body stands: this subsumes #1919's 3-line carry — whichever merges second takes the small conflict in `add_joiner_validator_inner`.

omersadika:
**Combined validation GREEN**: `v125_churn` passed on this PR merged with #1919 — [run 30299153116](https://github.com/dwallet-labs/ika/actions/runs/30299153116) (branch `claude/issue-1937-validate-with-1919`).

Full picture:
- This PR alone: #1937's out-of-committee hard failure gone; run then hits #1919's known convergence-accounting gap ([run 30295000573](https://github.com/dwallet-labs/ika/actions/runs/30295000573)).
- This PR + #1919: full `v125_churn` pass — joiner fold-in (4→5), shrink reshare (5→4), all boundary assertions.

The two PRs need to land together (in either order) for `v125_churn` to be green on `main`; whichever merges second takes one small conflict at the joiner-registration site in `add_joiner_validator_inner` — resolve in favor of this PR's `AuthorityNamings` push, which subsumes #1919's 3-line point fix (that resolution is exactly what the green validation branch ran).

PR #1939, opened 2026-07-27
fix(consensus-adapter): delete the writer-less pending_consensus_transactions recovery path (#1935 triage)

### LINKED ISSUE #1935: pending_consensus_transactions has no writer: the restart-resubmission path reads an always-empty table (silently-disabled recovery or dead code — decide and fix)
## Found by the #1929 write-discipline audit (recorded there as `write-discipline: none`)

`ConsensusAdapter::submit_batch` has its persist **commented out** (`consensus_adapter.rs:430`):

```rust
//epoch_store.insert_pending_consensus_transactions(transactions, lock)?;
Ok(self.submit_unchecked(transactions, epoch_store))
```

while the recovery side still runs: `submit_recovered` (`consensus_adapter.rs:297`) iterates `get_all_pending_consensus_transactions()` at startup to re-submit transactions that were in flight when the process died — and that table is **always empty**. The "consensus worker might lose transactions on restart" recovery path is silently disabled; the table, its reader, and the recovery call survive as scaffolding.

## Why it matters (or doesn't) — the triage this issue asks for

Whether this is a latent reliability bug or just dead code depends on which consensus transactions have their own application-level re-submit loops:

- **Covered by app-level retries** (examples): `EndOfPublishV2` re-submits until `has_recorded_end_of_publish_vote` confirms sequencing; MPC messages/outputs are re-driven by the service loop; announcements have the relay/re-send machinery.
- **The question**: enumerate the remaining `ConsensusTransactionKind` variants and identify any whose only redelivery path was this recovery table — e.g. checkpoint signature submissions, capability notifications, pricing votes. A validator restarting with such a tx in flight silently never sends it; quorum machinery usually absorbs one validator's missing contribution, but "usually absorbed" deserves to be a decision, not an accident.

## Fix directions

1. **If any variant depends on it**: restore the persist + the delete-on-processed lifecycle (the reader and table are intact; only the insert is disabled). `git log`/blame on the commented line first — it was presumably disabled for a reason (perf? a bug in the delete path?) that the fix must address rather than resurrect.
2. **If everything has app-level coverage**: delete the table, the reader, `submit_recovered`, and the startup call as dead code, and note in each variant's submitter that restart-loss is handled at its own layer. Half-disabled recovery is worse than either alternative — it reads as covered while covering nothing (#1929's whole thesis).

Either way, the `write-discipline: none` annotation from #1933 should flip to a real state.

omersadika:
Closes #1935.

## The decision the issue asked for

#1935 found `pending_consensus_transactions` half-disabled: the persist in `ConsensusAdapter::submit_batch` commented out (since the initial fork import — it has never run in this repo), while `submit_recovered` still read the always-empty table at startup. The issue asked for a triage: enumerate every `ConsensusTransactionKind` variant, determine which (if any) had this recovery table as its only restart redelivery path, then either restore the persist or delete the scaffolding.

**Verdict: delete.** Every live variant already owns its restart story at the application layer, and for the state-carrying kinds a table replay would be actively wrong, not just redundant.

## Per-variant triage

| Variant | Restart redelivery | Mechanism |
|---|---|---|
| `DWalletMPCMessage` / `DWalletMPCOutput` / `DWalletInternalMPCOutput` | structural | `DWalletMPCService` replays the epoch's full consensus stream from round 0 (`last_read_consensus_round: None`); sessions rebuild in-memory and the advanceable round is recomputed and re-submitted. There is no "already sent" state to get wrong. |
| `EndOfPublishV2` | re-submit loop | `HandoffSignatureSender` 1s tick; trigger is chain-derived (watch re-asserted by the syncer), stop is the durable `end_of_publish` row written through the commit batch. |
| `ValidatorMpcDataAnnouncement` | re-submit loop | 2s tick until `get_validator_mpc_data_announcement` confirms; the cached announcement reuses the stored `timestamp_ms` so post-restart re-submits keep the same consensus key. |
| `GlobalPresignRequest` | demand loop | re-sent every service iteration until the request is seen back **from consensus** (`sent_presign_sequence_numbers` is populated only by our own request arriving in a commit). |
| `CapabilityNotificationV1` | regeneration | one-shot per epoch, but every process start regenerates with a fresh, higher `generation` that `record_capabilities_v1`'s `generation >=` guard prefers. A table-recovered copy would be stale — strictly worse than what restart already does. |
| `IdleStatusUpdate` | regeneration | edge-triggered on flip; `last_sent_idle_status` resets to `None` on restart, so the current status is re-sent with a fresh nonce. A replayed copy would carry a stale `is_idle`. |
| `SuiChainObservationUpdate` | n/a | dead path today (observation source is hardcoded `None`, existing FIXME). |
| `NOAObservation` / `NOAPresignDemand` | re-derivation / demand loop | NOA store is deliberately in-memory and re-derived from consensus replay; demand announcement is retried against the durable `has_noa_assigned_presign` gate and any single validator's announcement suffices (keyed by demand digest). Feature not live. |
| `EndOfPublish` (V1) | n/a | decode-only, never emitted. |
| `DWalletCheckpointSignature` / `SystemCheckpointSignature` | **none — accepted** | see below. |
| `EpochMpcDataReadySignal` | **gap — tracked separately** | see below. |
| `RelayedValidatorMpcDataAnnouncement` | **gap — tracked separately** | see below. |

### The accepted loss: checkpoint signatures

A signature submitted for a built checkpoint is not re-sent after restart (the builder resumes strictly after the last built height and never re-invokes the output). This PR makes that a stated decision rather than an accident, with a doc comment at both submitters: certification needs a stake quorum, not any particular signer, so one validator's lost signature is absorbed; if our own aggregator can't certify a height, the certified checkpoint arrives via state sync. Certification stalls only if in-flight-signature loss is simultaneous and wide enough to break quorum — the same exposure upstream Sui accepted (see precedent below).

### Two real gaps found by the audit (pre-existing, not created by this PR)

Neither is fixed here — each is better fixed at its own layer than by resurrecting a table whose replay semantics are wrong for half the variants. Both carry `KNOWN GAP` comments at the code site, citing their tracking issues:

1. **`EpochMpcDataReadySignal`** (#1942): the re-emit gates (`next_sequence_number`, `last_emitted_validated_peers_count`) are in-memory and reset to 0 on restart. If seq=0 landed and a wider seq=1 died with the process, the restarted sender re-emits the wider set as seq=0 — dropped everywhere by the `consensus_message_processed` dedup — and then goes quiet until the set grows further. Fix direction: initialize the counters from this validator's own recorded ready signal.
2. **`RelayedValidatorMpcDataAnnouncement`** (#1943): the relayer acks the joiner at submit level (one-shot handler, no record, no re-submit), and the joiner's bounded fanout stops on acks it cannot verify against consensus. Correlated relayer restarts inside the ack-to-sequencing window lose the joiner's announcement invisibly. Fix direction: ack only after the tx's processed notification.

## Upstream precedent

Sui deleted the identical mechanism: the recovery path had already been reduced to a state-derived `recover_end_of_publish` check, and the `pending_consensus_transactions` DBMap itself (including the same `optimize_for_large_values_no_scan` options fn) was removed in sui#25319 (April 2026). The pinned `mainnet-v1.73.2` retains only the same stale comment ika carried at `submit_unchecked` — also removed here.

## What changed

- `consensus_adapter.rs`: deleted `submit_recovered`, the commented-out persist, and the stale reconfiguration-lock comment; `submit_batch` now documents the durability contract (in-memory retry until sequenced; restart-loss owned per-kind by each submitter).
- `authority_per_epoch_store.rs`: deleted the `pending_consensus_transactions` DBMap, its options fn, and both `get_all_pending_consensus_transactions` readers. The `write-discipline: none` annotation from #1933 dissolves with the table (54 fields remain, check green).
- `ika-node/src/lib.rs`: deleted the cold-start `submit_recovered` call; documented the capability notification's regeneration-on-restart story at its submitter.
- `messages_consensus.rs`: `ConsensusTransactionKey` DB-format comment now cites only `consensus_message_processed` (the key type itself is unchanged — it remains the processed-marker key, so BCS discipline still applies).
- Checkpoint-signature submitters (both): restart-loss decision documented.
- `mpc_data_announcement_sender.rs` / `announcement_relay.rs`: `KNOWN GAP` notes for the two findings above.

DB compatibility: removing the DBMap field is safe for a mid-epoch restart on an upgraded binary — typed_store's `populate_missing_cfs` opens any on-disk column family not declared in the struct with default options, and per-epoch DBs are pruned at reconfiguration. Downgrade is equally safe (missing CFs are created on open). No wire-format change: the table was local-only and provably always empty.

## Verification

- `cargo check` and `cargo clippy --all-targets` on `ika-types`, `ika-core`, `ika-node`: clean (clippy covers test targets, so no test references the removed symbols)
- `./scripts/check-epoch-table-write-discipline.sh`: OK (54 fields)
- repo-wide grep: zero references to the table, its readers, or `submit_recovered` remain
- consensus-reviewer pass over the diff: safe to merge, all lenses clean; it independently confirmed the CF-name (not position) keying in typed_store's derive, the `generation >=` guard claim, the builder resume bound, and that both KNOWN GAP comments describe real pre-existing behavior

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -1086,15 +1086,6 @@ pub struct AuthorityEpochTables {
     /// transaction whose effects were lost.
     consensus_message_processed: DBMap<SequencedConsensusTransactionKey, bool>,
 
-    /// Map stores pending transactions that this authority submitted to consensus
-    ///
-    /// write-discipline: none — the table has no writer on this branch (the
-    /// persist in `ConsensusAdapter::submit` is commented out), so its only
-    /// consumer, `get_all_pending_consensus_transactions` (consensus-adapter
-    /// resubmit-on-restart), always reads empty.
-    #[default_options_override_fn = "pending_consensus_transactions_table_default_config"]
-    pending_consensus_transactions: DBMap<ConsensusTransactionKey, ConsensusTransaction>,
-
     /// The following table is used to store a single value (the corresponding key is a constant). The value
     /// represents the index of the latest consensus message this authority processed, running hash of
     /// transactions, and accumulated stats of consensus output.
@@ -1655,12 +1646,6 @@ pub struct AuthorityEpochTables {
     pub(crate) network_reconfiguration_output_digests: DBMap<ObjectID, [u8; 32]>,
 }
 
-fn pending_consensus_transactions_table_default_config() -> DBOptions {
-    default_db_options()
-        .optimize_for_write_throughput()
-        .optimize_for_large_values_no_scan(1 << 10)
-}
-
 fn verified_dwallet_checkpoint_messages_table_default_config() -> DBOptions {
     default_db_options()
         .optimize_for_write_throughput()
@@ -1736,14 +1721,6 @@ impl AuthorityEpochTables {
 
     pub fn path(epoch: EpochId, parent_path: &Path) -> PathBuf {
         parent_path.join(format!("{EPOCH_DB_PREFIX}{epoch}"))
-    }
-
-    pub fn get_all_pending_consensus_transactions(&self) -> IkaResult<Vec<ConsensusTransaction>> {
-        Ok(self
-            .pending_consensus_transactions
-            .safe_iter()
-            .map(|item| item.map(|(_k, v)| v))
-            .collect::<Result<Vec<_>, _>>()?)
     }
 
     pub fn get_last_consensus_index(&self) -> IkaResult<Option<ExecutionIndices>> {
@@ -2446,14 +2423,6 @@ impl AuthorityPerEpochStore {
                 })
             }
         }
-    }
-
-    pub fn get_all_pending_consensus_transactions(&self) -> Vec<ConsensusTransaction> {
-        // The except() here is on purpose, because the epoch can't run without it.
-        self.tables()
-            .expect("recovery should not cross epoch boundary")
-            .get_all_pending_consensus_transactions()
-            .expect("failed to get pending consensus transactions")
     }
 
     /// Returns true if all messages with the given keys were processed by consensus.
--- a/crates/ika-core/src/consensus_adapter.rs
+++ b/crates/ika-core/src/consensus_adapter.rs
@@ -293,22 +293,6 @@ impl ConsensusAdapter {
         self.consensus_throughput_profiler.store(Some(profiler))
     }
 
-    // todo - this probably need to hold some kind of lock to make sure epoch does not change while we are recovering
-    pub fn submit_recovered(self: &Arc<Self>, epoch_store: &Arc<AuthorityPerEpochStore>) {
-        // Currently narwhal worker might lose transactions on restart, so we need to resend them
-        // todo - get_all_pending_consensus_transactions is called twice when
-        // initializing AuthorityPerEpochStore and here, should not be a big deal but can be optimized
-        let recovered = epoch_store.get_all_pending_consensus_transactions();
-
-        debug!(
-            "Submitting {:?} recovered pending consensus transactions to consensus",
-            recovered.len()
-        );
-        for transaction in recovered {
-            self.submit_unchecked(&[transaction], epoch_store);
-        }
-    }
-
     fn await_submit_delay(
         &self,
         _committee: &Committee,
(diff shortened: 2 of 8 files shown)


omersadika:
Reviewed — approve, conditioned on the in-flight Upgrade Test run on this branch going green (run 30301493840): its binary-swap scenarios are the empirical answer to the one question this deletion raises — a new binary reopening an old binary's epoch DB that still carries the now-undeclared column family. (The cited Sui precedent deleting this exact DBMap says typed-store handles the extra on-disk CF; the suite makes it evidence.)

The triage itself is the model #1935 asked for: every variant's restart story named with its mechanism, the checkpoint-signature loss converted from accident to stated decision with doc comments at both submitters and the quorum/state-sync absorption argument, and — the best part — the two REAL gaps the audit found (#1942 ready-signal re-emit counters, #1943 relay ack-before-sequencing) tracked at their own layers with KNOWN-GAP comments at the code sites, rather than papered over by resurrecting a table whose replay semantics the per-variant table shows would be wrong for half the kinds (stale generations, stale idle flags). The `messages_consensus.rs` change is comment-only — the wire/DB discipline note now cites the surviving table, append-only rule intact. Deleting scaffolding that reads as coverage while covering nothing is #1929's thesis applied; merge on suite green.

PR #1940, opened 2026-07-27
fix(committee): fail loudly when a committee cannot recover BLS-basis aliases

omersadika:
## Summary

Audit follow-up requested in #1937 (related: #1938, #1911). #1911 split committee construction: `Committee::new_with_protocol_keys` carries the BLS protocol keys from the chain read, while `Committee::new` still derives `expanded_keys` by decoding the names. Under consensus-basis names (protocol v6) nothing decodes, `expanded_keys` is silently empty, and `name_translation()` — the v6 boundary-translation mechanism — degrades to identity with no error: all three callers (`authority_per_epoch_store` mpc_data carry-forward, `mpc_manager` prior-cert key ingest, `peer_blob_fetcher`) fall back to identity on a missing entry, silently reinstating exactly the boundary misses #1911 fixed. Today the safety is wiring-dependent: both production epoch-store committees come from `new_with_protocol_keys`, but nothing asserts that.

## The guard (option b of the audit, sharpened)

`name_translation()` now logs a `should_never_happen` error naming every voting member whose BLS alias is unrecoverable (no `expanded_keys` entry). The predicate is exact:

- BLS-basis names always decode into `expanded_keys` in `load_inner` → silent.
- Consensus-basis committees built with `new_with_protocol_keys` carry the full chain-read map (`build_ika_committee` populates it for every member) → silent.
- A consensus-basis committee mis-built through `Committee::new` → fires, naming the members.

It deliberately checks only BLS-alias coverage, not `consensus_keys` coverage — the freeze tests (and legacy/test committees generally) legitimately carry empty `consensus_keys`, and the boundary items that matter at the v6 flip are BLS-named.

**Why not a construction-time signal (option a):** `Committee::new` legitimately builds consensus-basis committees today — the `SuiSyncer::new_committee` reconfiguration-input committees and the prior-committee handoff fetch (`pubkey_provider_updater.rs`) — so any unconditional signal in `new` cries wolf every sync tick under v6, and distinguishing the cases needs a signature change across ~13 call sites. A restored `debug_assert` is dead weight under this repo's release-mode test convention. The translation site is the choke point where misuse becomes harmful, and erroring there (vs logging) would turn a latent wiring bug into a freeze-path liveness failure — worse than the loud log plus the existing fail-closed behavior.

**Audit of the risky call sites (option c):** neither `sui_syncer.rs` committee (both key maps empty; feeds only the reconfiguration MPC input) nor the `pubkey_provider_updater.rs` prior committee (verifies handoff certs via `consensus_key()` only) reaches `public_key()` or `name_translation()`. `new_at_next_epoch_for_testing` currently has no callers. Recorded in the spec as safe **by audit, not by type**.

## Diagnostics sharpened on the same misuse

- `public_key()` miss now reports "N of M members carry a BLS key" plus a pointer to `new_with_protocol_keys` (previously "committee size 0", implying an empty committee).
- `class_groups_public_key_and_proof()` no longer reports `expanded_keys.len()` — the wrong map — as the committee size.

## Spec

`dev-docs/specs/committee-consensus-keys.md` updated in the same PR: the wiring-dependence, the guard, and the by-audit status of the two `Committee::new` consensus-basis sites. Also fixed a stale claim found while editing: the spec said three sites populate `consensus_keys` including `SuiSyncer::new_committee`, but the syncer deliberately passes an empty map (handoff verification uses the epoch store's committee) — now two sites, with the syncer's empty map documented as deliberate.

## Validation

- New unit tests pin the predicate across all three committee shapes (BLS-basis via `new`, consensus-basis via `new_with_protocol_keys` — including BLS-alias → consensus-name translation — and consensus-basis mis-built via `new`).
- `cargo test --release -p ika-types committee::` 5/5.
- `cargo test --release -p ika-core -- authority_per_epoch_store::tests mpc_manager::tests` 21/21 — the freeze/carry-forward tests exercise `name_translation()` and confirm the guard stays silent on every legitimate committee shape (no false positives).
- `cargo clippy --release -p ika-types --all-targets --all-features` clean; `cargo fmt --all` no-op beyond the change.

### DIFF
--- a/crates/ika-types/src/committee.rs
+++ b/crates/ika-types/src/committee.rs
@@ -335,11 +335,30 @@ impl Committee {
         match self.expanded_keys.get(authority) {
             Some(v) => Ok(v),
             None => Err(IkaError::InvalidCommittee(format!(
-                "Authority #{} not found, committee size {}",
+                "Authority #{} has no BLS protocol key in this committee ({} of {} members \
+                 carry one; a consensus-basis committee that verifies BLS certs must be \
+                 built with new_with_protocol_keys)",
                 authority,
-                self.expanded_keys.len()
+                self.expanded_keys.len(),
+                self.voting_rights.len()
             ))),
         }
+    }
+
+    /// Voting members with no entry in `expanded_keys` — their BLS-basis
+    /// alias is unrecoverable, so [`Committee::name_translation`] cannot map
+    /// prior-epoch items named under the BLS basis back to them. Empty for
+    /// every correctly built committee: BLS-basis names decode into
+    /// `expanded_keys` in [`Committee::load_inner`], and consensus-basis
+    /// committees carry the chain-read map via
+    /// [`Committee::new_with_protocol_keys`]. Non-empty means a
+    /// consensus-basis committee was mis-built through [`Committee::new`].
+    fn members_without_bls_protocol_key(&self) -> Vec<AuthorityName> {
+        self.voting_rights
+            .iter()
+            .filter(|(name, _)| !self.expanded_keys.contains_key(name))
+            .map(|(name, _)| *name)
+            .collect()
     }
 
     /// Maps each member's name under EITHER identity basis to the name this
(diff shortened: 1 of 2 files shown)


omersadika:
Reviewed — approve on content; one merge-sequencing requirement. This is the #1911-review guard delivered at the right layer: detection in `name_translation()` — where the degradation actually bites — naming every affected member and the concrete consequence (carry-forward / prior-cert ingest / blob-repair misses), with the exact predicate argued in the body (BLS-alias coverage only, so legacy/test committees with empty `consensus_keys` stay silent — correct). The `public_key()` error-text upgrade turns a bare not-found into an actionable diagnosis, and the consensus-basis test fixtures close the gap that let the mis-built-committee case go untested.

**The sequencing requirement**: this adds a raw `tracing::error!(should_never_happen = true, ...)` while #1925 (open) adds a CI guard that fails on ANY bare marker outside the macro — the exact merge-ordering collision that broke main's Format Check today (#1933 vs #1934). If this merges before #1925, #1925's guard fails on main the moment it lands; if after, this PR fails its own CI once rebased. Cleanest order: **merge #1925 first, then switch this PR's one site to `ika_types::report_invariant_violation!("committee_missing_bls_alias", ...)`** — which it should use anyway, since a mis-built committee firing only in logs would be invisible for external validators, and this is precisely the class of signal #1923 built the counter for.

PR #1941, opened 2026-07-27
fix(node): derive every node self-identity from the epoch store's name (#1937 follow-up)

omersadika:
## Summary

Audit follow-up requested in #1937 (the "any other product path that grew a second identity basis" sweep). #1911 made `AuthorityName` per-epoch (BLS protocol key below protocol v6, consensus key from v6) and made the epoch store's name basis-aware — but `AuthorityState.name` was still minted once at startup from `protocol_public_key()` (always BLS) and never updated. Two live consequences on a v6 network:

1. **Restart-after-flip boots state-sync in pull mode.** The `is_notifier` argument to `create_p2p_network` compared the BLS startup name against a consensus-basis committee; a validator restarting after the v5→v6 flip found itself "absent" and booted state-sync as a non-committee pull node.
2. **Every capability vote silently dropped — all future upgrades blocked.** The capability notification in `monitor_reconfiguration` carried the BLS `state.name`; `verify_consensus_transaction` drops a capability whose `authority` field disagrees with the consensus-attributed (committee-basis) author. On a v6 committee every validator's vote was discarded **at the receiver** while the sender saw success — no post-v6 protocol or Move-contract upgrade could ever tally.

Latent: `ConsensusAdapter.authority` (its `submission_position` path panics in `get_position_in_list` under a consensus-basis committee) and three log sites printing a name that appears in no v6 committee log.

## Fix

**`AuthorityState.name` is deleted.** Its "fixed, static identity" contract broke at the flip, it had zero load-bearing internal uses, and every consumer wants the per-epoch name. All consumers now read `AuthorityPerEpochStore::name` (re-derived under each epoch's recorded basis at boot and reconfiguration): the state-sync notifier derivation (now a named local with a boot-identity info line), the `ConsensusAdapter` authority, capability notifications, the `IkaNode` Debug name, and both split-brain log sites. The remaining `protocol_public_key()` call sites were audited: the trusted-peer refresher self-excludes under both bases (safe) and the boot banner is config-level identity, not committee membership.

## Coverage

`v125_v6_upgrade` now restarts validator 1 after the flip (same-binary `stop_and_swap`), asserts its post-restart log reports `consensus_key_identity=true is_state_sync_notifier=false` (new `expect_log_line_present_on_validator` scenario step — the whole-cluster log step would demand the line of validators whose last boot predates the flip), then crosses to epoch 5 with all validators local so the restarted member's announcement/freeze/reshare duties are exercised, not just its boot report.

Spec updated in the same PR: `dev-docs/specs/committee-consensus-keys.md` (node-side naming discipline).

## Validation

- `cargo clippy --release --all-targets -p ika-core -p ika-node -p ika-upgrade-test` clean; all `ika-upgrade-test` unit tests pass.
- Fix run (must pass): [run 30301089098](https://github.com/dwallet-labs/ika/actions/runs/30301089098) — `v125_v6_upgrade` on this branch.
- Test-testing fault run (must FAIL at the new log assertion): [run 30301149258](https://github.com/dwallet-labs/ika/actions/runs/30301149258) — branch `claude/bold-taussig-0f7389-fault` reintroduces the BLS-basis derivation with the log line kept, so the restarted validator logs `is_state_sync_notifier=true`.

Verdicts will be reported on this PR.

Fixes the node-side portion of the #1937 audit; the harness-side registration fix is #1938.

### DIFF
--- a/crates/ika-core/src/authority.rs
+++ b/crates/ika-core/src/authority.rs
@@ -40,7 +40,7 @@ use crate::epoch::committee_store::CommitteeStore;
 use ika_config::node::AuthorityOverloadConfig;
 use ika_types::{
     committee::Committee,
-    crypto::{AuthorityName, AuthoritySignature},
+    crypto::AuthoritySignature,
     error::{IkaError, IkaResult},
 };
 use sui_types::base_types::*;
@@ -241,10 +241,14 @@ impl AuthorityStateTrait for AuthorityState {
 }
 
 pub struct AuthorityState {
-    // Fixed size, static, identity of the authority
-    /// The name of this authority.
-    pub name: AuthorityName,
-    /// The signature key of the authority.
+    // Deliberately no `name` field: `AuthorityName` is per-epoch, not static
+    // — its basis is the BLS protocol key below protocol v6 and the consensus
+    // key from v6, so the node's committee identity lives on
+    // `AuthorityPerEpochStore::name` (re-derived for each epoch's basis at
+    // reconfiguration). A name minted once at startup goes stale at the
+    // activation boundary and misreports membership against later committees.
+    /// The BLS signature key of the authority (static; signs aggregate stake
+    /// certificates regardless of which key the committee is named by).
     pub secret: StableSyncAuthoritySigner,
 
     pub(crate) perpetual_tables: Arc<AuthorityPerpetualTables>,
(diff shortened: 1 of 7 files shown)


omersadika:
## Validation verdicts — both conclusive

**Fix run [30301089098](https://github.com/dwallet-labs/ika/actions/runs/30301089098): PASSED (all 38 steps).** The restarted validator's boot-identity assertion matched (`per-validator log-line assertion passed index=1 needle="consensus_key_identity=true is_state_sync_notifier=false"`), the post-restart workload served users (58.8s lifecycle), and the cluster crossed to epoch 5 with all validators local — the restarted member carried its announcement/freeze/reshare duties across the next boundary.

**Fault run [30301149258](https://github.com/dwallet-labs/ika/actions/runs/30301149258): FAILED at exactly the new assertion** (branch reintroduced the BLS-basis derivation, log line kept; branch since deleted — artifacts persist on the run). All 30 prior steps were green — upgrade, boundary reshare, convergence, even the restart itself — and the scenario died only at `[flow 32/38] expect_log_line_present_on_validator(1, ...)`. Node-log evidence from the run's artifacts:

- validator-1 (restarted post-flip, fault active): `state-sync boot identity ... consensus_key_identity=true is_state_sync_notifier=true` — the bug manifested exactly as predicted;
- validators 0/2/3 (last booted pre-flip): `consensus_key_identity=false is_state_sync_notifier=false` — the bug is invisible on pre-flip boots, which is why no existing scenario caught it;
- the real notifier node: `is_state_sync_notifier=true` — true non-members still derive pull mode correctly under the fix.

Per the test-testing methodology: the fault is otherwise symptomless (30 green steps prove it), so the new probe is the only thing standing between this failure mode and a green gate.

PR #1958, opened 2026-07-31
fix(sui): recover pruned network-key registry reads

### LINKED ISSUE #1957: Invariant channel: persistent-condition sites fire per-tick (hundreds/host/hour), and the registry-read-empty condition is persistent under v6 on healthy validators — two fixes
## The specimen that separates the two

`node101` (mainnet, v1.2.7, restarted ~09:50 UTC right after the flip): **fully healthy — 10,979 advances/30m, full MPC participation** — while firing `network_key_registry_read_empty` at ~284/30m ≈ **once per 5s syncer tick, continuously, for 1.6h+**. The consensus-dead pair (StakingCabin, breskulpeak — #1956 cohort 2) churn the same sites at similar rates as a downstream symptom. Two independent problems:

## Fix 1 — reporting defect: invariant sites must be events, not polled conditions

`report_invariant_violation!` at the registry-read site sits inside a retrying loop, so a *persisting* condition re-fires the maximum-severity channel every tick. Consequences: hundreds of firings per host per hour, alert evaluation on `increase() > 0` cannot distinguish a new incident from a known ongoing one, and the channel decays toward the exact alert-fatigue failure #1922 was written to prevent ("trains operators to ignore the flag").

Ask:
- **Convention rule** (metrics.md + the #1925 playbook section): an invariant site must mark an *occurrence*, not a *state*. A condition that can persist across retries fires on **state transition** — once on entry, bounded reminders (e.g. 1/10m), once on recovery — with a paired **state gauge** carrying the ongoing condition for alerting. The codebase already owns this pattern twice over (#1916's `TerminalMessageLogState` / `AssemblyHealthState`); this is its application to the invariant macro's hot callers.
- **Mechanical fix now**: wrap the two known per-tick sites (`network_key_registry_read_empty`, `stranded_network_key_missing_from_registry_read`) in transition-shaped state, keeping their counters incrementing once per *episode* + a `..._condition_active` gauge. A CI-grep guard ("no `report_invariant_violation!` inside the syncer/service tick loops without a transition guard") is cheap and matches the house guard pattern.

## Fix 2 — the underlying condition: the network-keys registry read returns empty persistently under v6 on a subset of healthy validators

This is #1952's transport root cause resurfacing in a worse form: on v1.2.6 it was (apparently) a boot-window transient; under v6 on node101 it is **every tick for hours on an otherwise fully-functional validator**. The #1952 recovery design is working — the validator is healthy *because* boot-time cert-path ingestion covered it, and steady-state MPC does not depend on the periodic re-read — but the read itself is broken, and the first key-state change it exists to catch (a reconfiguration state flip, a new key) will be missed by every affected host.

Ask:
- Root-cause why `get_dwallet_mpc_network_keys` / the registry walk returns empty under v6 for some hosts and not others (node101 yes, our six no — same binary, same network, same epoch; suspect per-host read-path state or the coordinator-object read shape post-flip rather than the object itself, since most of the fleet reads it fine).
- The #1952 dossier's transport-layer step (per-transport reproduction against the live coordinator object) is the starting point; the affected-host set here (node101 + the #1956 pair) gives live reproduction targets.
- Until fixed, the paired state gauge from Fix 1 is also the correct fleet-monitoring surface for "how many hosts currently cannot read the registry" — today that number is invisible inside an undifferentiated counter.

## Why this matters beyond hygiene

The invariant channel just proved its worth in its first real incident — it diagnosed external validators' states with no log access. Fix 1 protects that channel's signal-to-noise before operators learn to ignore it; Fix 2 closes the last known live remnant of #1952.

Related: #1952 (transport root cause + recovery design), #1956 (the incident), #1922/#1916 (the transition-shaped reporting patterns), #1923/#1925 (the channel).

omersadika:
Fixes #1957.

## Root cause

The OCS dynamic-field proof provider drops a live-listed child when the child's last-modifying checkpoint has been pruned. The durable verified-state cache already retains the exact object, proof, and certified summary, but only the local reader used it; mirrored registry walks therefore observed a successful empty page indefinitely. Different upstream pruning/configuration explains the per-host split on the same binary and epoch.

## Fix

- Share the durable verified snapshot cache with the direct proof provider. After a normal proof build returns `NotFound`, reuse a cached proof only when the current object fetched from Sui has the exact same `(id, version, digest)` as the cached object. Direct and mirrored consumers still verify the proof normally.
- Make the two registry invariant sites episode-shaped: one counter increment on entry, one recovery log, and paired active-condition gauges. Transport failures no longer masquerade as successful empty reads.
- Add a CI structural guard preventing those hot-loop invariant sites from losing their transition gate.
- Update the metrics convention, production alert playbook, and OCS verified-read spec.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --release -p ika-network pruned_proof_reuses_only_an_exact_current_snapshot --lib`
- focused release-mode `ika-core` transition, metric-registration, and durable-cache tests
- `./scripts/check-invariant-violation-markers.sh`
- `git diff --check`

Test-testing: temporarily disabled the exact-match recovery predicate; the pruning regression failed at `assert!(cache_hit)` as predicted. The mutation was removed and the clean test passed.

### DIFF
--- a/crates/ika-core/src/sui_connector/metrics.rs
+++ b/crates/ika-core/src/sui_connector/metrics.rs
@@ -51,6 +51,14 @@ pub struct SuiConnectorMetrics {
     /// committee validator stuck non-zero.
     pub(crate) network_key_overlay_incomplete: IntGauge,
 
+    /// 1 while a successful network-key registry read is empty even though the
+    /// coordinator reports existing entries; 0 after an observed recovery.
+    pub(crate) network_key_registry_read_empty_condition_active: IntGauge,
+
+    /// 1 while at least one key awaiting chain-sourced recovery is absent from
+    /// a successful registry read; 0 after an observed recovery.
+    pub(crate) stranded_network_key_missing_from_registry_read_condition_active: IntGauge,
+
     /// Total sync ticks on which the off-chain next-committee
     /// validator-mpc_data assembly was incomplete (benign retry while
     /// announcements/blobs converge; a stall shows as sustained growth).
@@ -224,6 +232,20 @@ impl SuiConnectorMetrics {
                 registry,
             )
             .unwrap(),
+            network_key_registry_read_empty_condition_active:
+                register_int_gauge_with_registry!(
+                    "ika_network_key_registry_read_empty_condition_active",
+                    "1 while a successful network-key registry read is empty despite a non-empty on-chain registry; 0 after recovery",
+                    registry,
+                )
+                .unwrap(),
+            stranded_network_key_missing_from_registry_read_condition_active:
+                register_int_gauge_with_registry!(
+                    "ika_stranded_network_key_missing_from_registry_read_condition_active",
+                    "1 while a stranded network key is absent from a successful registry read; 0 after recovery",
+                    registry,
+                )
+                .unwrap(),
             off_chain_assembly_incomplete_ticks_total: register_int_counter_with_registry!(
                 "ika_off_chain_assembly_incomplete_ticks_total",
                 "Total sync ticks on which the off-chain validator-mpc_data assembly was incomplete",
@@ -362,6 +384,18 @@ mod tests {
         let _mpc_metrics = DWalletMPCMetrics::new(&registry);
 
         assert_eq!(metrics.off_chain_assembly_incomplete.get(), 0);
+        assert_eq!(
+            metrics
+                .network_key_registry_read_empty_condition_active
+                .get(),
+            0
+        );
+        assert_eq!(
+            metrics
+                .stranded_network_key_missing_from_registry_read_condition_active
+                .get(),
+            0
+        );
         assert_eq!(
             metrics
                 .off_chain_assembly_consecutive_incomplete_ticks
--- a/crates/ika-core/src/sui_connector/setup.rs
+++ b/crates/ika-core/src/sui_connector/setup.rs
@@ -195,6 +195,15 @@ pub async fn build_sui_connector_stack(
     metrics: Arc<OcsMetrics>,
     provider_metrics: Arc<ProofProviderMetrics>,
 ) -> Result<SuiConnectorStack, SetupError> {
+    // Durable cache: rehydrates from the perpetual `verified_object_cache`
+    // column on boot and writes through every absorb. The direct proof
+    // provider also reuses a snapshot when Sui still returns the exact same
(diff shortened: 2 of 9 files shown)


omersadika:
## Review: approve — the security-critical predicate is right, the episode semantics are careful, and the root cause explains every observation from the incident

**The proof-reuse predicate — the part that needed the most scrutiny — is sound.** Reuse requires `entry.object.compute_object_reference() == current_object_ref`: the digest is recomputed **from the cached bytes**, so a corrupted cache cannot spoof its way past the check, and same-version-same-digest means the consumer-side high-water/regression machinery is trivially satisfied (nothing moves backward). The residual freshness exposure — a malicious/lagging RPC replaying a once-verified `(id, version, digest)` to hold the node on an old view — reduces to **withholding-equivalence**: RPC control already grants a frozen view (that is literally the pre-fix behavior), and certified-proof verification still forbids fabrication. One doc suggestion, non-blocking: state that withholding-equivalence argument explicitly in the spec's new section, so the next security reader doesn't have to re-derive why bounded staleness here grants no new attacker power.

**Root cause fits every incident observation**: pruned defining-checkpoints varying by upstream configuration explains node101-yes/our-six-no on identical binaries, the per-tick recurrence (each walk re-hits the same pruned child), and why boot-time cert ingestion kept affected validators healthy. The "successful empty page" masquerade — the worst property of the original bug — is now dead twice over: recovered when the cache holds the exact object, and visibly gauged when it cannot.

**Episode shaping matches the #1957 asks with one refinement better than mine**: transport errors preserve the last observed condition state instead of starting or recovering an episode — an error is no-answer, consistent with the adoption code's philosophy, and it prevents a flapping RPC from manufacturing episode churn. Counters now mean "a new episode began," gauges mean "still active," the playbook alerts on the gauges with `for: 1m`, and the CI structural guard keeps the hot sites from silently losing their transition gates.

**Honest residual, correctly documented**: an entry never folded into the cache before upstream pruning (fresh host + aggressive pruning) remains unavailable — but now as a visible gauge condition instead of an invisible empty page. That residual is #1882's retention story, not this PR's scope.

Fault validation meets the standard (predicate disabled → the regression fails at `assert!(cache_hit)`; restored → green). Merge-ready from my side once checks are green; the live proof will be node101's `registry_read_empty` gauge clearing on its first post-upgrade read — worth confirming on the fleet after this ships in v1.2.8.

Issue #1956, opened 2026-07-31
Mainnet v6 flip-boundary incident: 5 validators crashed + 2 exited consensus-dead at the activation boundary; continuous registry-read invariant churn on the casualties


omersadika:
## Active incident, mainnet, opened 2026-07-31 ~11:30 UTC — margin currently safe (49/56 participating vs quorum 38)

Mainnet activated **protocol v6** (the AuthorityName identity flip) at its ~09:40 UTC epoch boundary today. The flip itself succeeded — the network runs v6, epochs advance, overdue 0 — but the boundary took out **7 of 56** validators in two distinct failure shapes, time-correlated to the minute:

## Cohort 1 — crashed at the flip (5 validators, still down 100+ min later)

`3MI Labs ika validator`, `Brightlystake`, `Coinage x DAIC`, `Studio Mirai`, `bitszn | valopers.com` — all last reported metrics at **09:39–09:40 UTC**, within 60 seconds of each other, and none has returned. Five independent operators do not coordinate to the minute: their processes died processing the flip boundary. None returning in 100 minutes suggests either unattended pods or **crash-looping on the entry path** (a restarted process re-entering the v6 epoch and dying again would look exactly like this from outside). Operator contact needed for logs — their panic/error at the boundary is the highest-value diagnostic in this incident.

## Cohort 2 — exited the flip consensus-dead (2 validators, the #1864 signature)

`StakingCabin` and `breskulpeak.com`: alive (uptimes 12.2h / 19.9h — **no restart involved**), v1.2.7, but commit staleness ~5,900s since the flip and `ika_consensus_subscribed_to = 0` with occasional flaps to 1 — resubscription attempts failing continuously. This is #1864's exit-the-handoff-consensus-dead shape, at the highest-stakes boundary possible. StakingCabin additionally reports canonical DKG version 0 and no network-key-loaded epoch — it never entered the v6 epoch's MPC at all. The #1864 remedy (operator restart) should apply; whether a restart lands them in Cohort 1's crash loop instead is the open question — recommend ONE of the two operators restarts first as the probe.

## Cohort 3 — continuous #1952-site invariant churn (overlapping)

[...]

PR #1959, opened 2026-07-31
Drop protocol v5 support (MIN=6), sweep the v6 gates, and add protocol v7 (short AuthorityName encoding)

omersadika:
Two commits. The first is a pure cleanup; the second is a consensus-critical protocol version. Read them separately.

## 1. `chore!:` drop protocol v5 support (MIN=6) and sweep the v6 gates

v6 is live on mainnet and testnet, so `MIN_PROTOCOL_VERSION = 6`. With no supported version boundary below v6, every branch that chose behavior by protocol version becomes unreachable and is removed:

- **`consensus_key_authority_names`** — `AuthorityName` is unconditionally the Ed25519 consensus key. The `consensus_key_identity` bool and its BLS-basis fallbacks are gone, and with them the mitigations that existed *only* to survive the single v5→v6 activation: the per-basis prior-committee candidate list, dual-basis joiner membership matching, and `also_accept_next_committee_pubkeys`.
- **`strict_network_key_coefficient_bound`** — the `backward_compatible` bool threaded into the network-key DKG/reconfiguration constructors is replaced by a constant that is `false` at every supported version.

Safe because **unreachable, not merely unused**: a persisted epoch-start record below v6 now trips `get_for_version`'s MIN assertion in `AuthorityPerEpochStore::new`, before any identity code runs.

`Committee`'s `expanded_keys` alias translation stays — it is structural, not version-gated, and is what keeps BLS-basis committees readable. `ProtocolConfig` keeps both `FeatureFlags` fields and every per-version arm: the flag set is BCS-serialized into the config digest that rides `AuthorityCapabilitiesV1` through consensus, so dropping a field would move every supported version's digest relative to deployed binaries. Snapshots confirm it.

Upgrade tests re-base v1.2.5 → v1.2.7 (the first release supporting v6).

## 2. `feat(protocol):` protocol v7 — emit `AuthorityName` as the raw 32 bytes

`AuthorityName` has been the consensus key since v6, but written zero-padded into the 48-byte container the BLS key occupied through v5. v7 (`short_authority_names`) drops the padding from the wire.

**Why this needs a version gate even though every binary already reads both widths.** Decoding has accepted 32 and 48 bytes since v1.2.7, so nothing rejects the short form. That is necessary but not sufficient: signature and digest bytes are reconstructed by **re-serializing locally** — `verify_handoff_signature` rebuilds the signed payload from its own decoded copy, `hash_next_committee_pubkey_set` BCS-encodes a `Vec<AuthorityName>` before hashing. Two validators emitting different widths compute different digests and reject each other's signatures while parsing each other's messages perfectly. The gate buys **simultaneous emission**, not readability.

**How the width reaches serde.** It cannot be consumed at its call sites — `Serialize` runs inside `bcs::to_bytes` with nothing but the value. It is mirrored into the `AUTHORITY_NAME_SHORT_ENCODING` process global when the epoch store is built. This is the only protocol flag in ika consumed outside its config; the alternative, versioning every message type embedding a name, is a V2 of twelve quorum-critical structs.

**Cross-epoch straddle.** A handoff cert is signed at the end of epoch N and verified in N+1, so the verifier re-serializes at the new width while signers used the old one. `verify_certified_handoff_attestation` retries once at the other width, scoped to the calling thread — mutating the global would corrupt names other threads are serializing concurrently. This does not widen what is accepted: both widths encode the same attestation value.

A BLS-basis name (non-zero tail) is never truncated, so historical committee records keep round-tripping.

**Also fixes a bug this bump introduced.** With `MAX = 7`, the binary-swap rehearsals would have voted themselves to v7 partway through — and failed *open*, since `expect_protocol_version_at_least(6)` passes at 7 — silently conflating deployed-release compatibility with the width flip. `v127_rollout`, `v127_churn`, `malicious_v127` and `workload` now pin their advertised range to `6..=6`.

## Verification

- `clippy --workspace --all-targets --all-features` clean
- `ika-protocol-config` 6/6; three new v7 snapshots, **v5/v6 snapshots unmodified** (no existing config digest moved)
- `ika-types` 32/32, including a vacuity guard that the two widths actually differ and a check that the thread-local override does not leak across threads
- `ika-core` handoff 24/24, including cross-width verification in **both** directions and a negative that a forged signature is still rejected at both widths
- `ika-core` non-integration 333/333
- MPC integration: every failure seen under 16- and 4-way parallelism reproduced **green at `--test-threads=1`** (presign-pool fill-rate preconditions, not regressions)

**Fault-validated:** disabling the dual-width retry fails `handoff_cert_verifies_across_the_authority_name_width_flip` with `InvalidSignature` on a zero-padded name, at exactly the boundary. The green run is evidence, not assumption.

## Reviewer attention — known gaps

1. **Neither transition gate has been executed.** `protocol_version_transition` (in-process, v6→v7) and `v127_v7_upgrade` (out-of-process, real v1.2.7 binaries) are restored and compile; that is all. Both need a real run plus fault injection before this is trusted — the failure mode here is a fleet-wide wedge.
2. **The in-process gate structurally cannot catch the main risk.** The width is process-wide, and that test runs all four validators in one process, so they flip together. Only `v127_v7_upgrade` can surface a genuine per-validator disagreement, and it is opt-in — PR CI will not run it.
3. **`AuthorityName` is still `[u8; 48]` in memory**, carrying 16 zero bytes; only the encoding is narrow at v7. The type split to a real `[u8; 32]` newtype is mapped but not done — its blocker is that pre-v6 `committee_map` records key `voting_rights` by 48-byte BLS names a 32-byte type cannot hold, and `Committee::load_inner` derives `expanded_keys` by decoding the name *as* a BLS key.


omersadika:
## On shipping this in v1.2.8: **recommend splitting — v1.2.8 without it, v1.2.9 with it.** Not a quality objection; a sequencing one.

First, a correction to the PR's own reviewer note: **the transition gates HAVE been executed.** Run [30675628560](https://github.com/dwallet-labs/ika/actions/runs/30675628560) on this branch is green across `v127_v7_upgrade`, `v127_churn`, `malicious_v127`, `v6_rollout`, `restart_spectator`, `legacy_config`, `smoke`, `workload`, and the Test Cluster suite (which carries `protocol_version_transition`) passed too. Known gap #1 in the body is stale — worth updating, because it currently understates the PR's own evidence. The out-of-process gate is precisely the one your note says can surface per-validator width disagreement, and it passed.

So the code case is good. My concern is what shipping it *does* on the day it ships.

**1. `MAX = 7` starts an activation clock the moment the tag exists.** We have direct, week-old evidence of what that costs: v1.2.6 shipped `MAX = 6` with every gate green, and the v6 flip still took out 7 of 56 mainnet validators (#1956). The post-mortem is the important part — **cohort 1, five of the seven, were operators still on v1.2.5**; a MAX=5 binary halts by design at a v6 boundary. That failure mode is not addressable by any gate in this PR. It is addressable only by rollout discipline and lead time, and it argues for v7 getting its own release with its own announcement rather than riding one whose purpose operators will read as "bug fix."

**2. v1.2.8's actual job is urgent and narrow.** #1958 is unreleased, and the condition it fixes is live right now: `network_key_registry_read_empty` at ~700/h on node101 and ~610/h on StakingCabin. Bundling means that fix cannot ship until the protocol work is fully soaked — and it should ship this week.

**3. There is an unfixed network-halting risk on v6 that should land before v7 stacks on top.** Just filed as #1961: `hash_next_committee_pubkey_set` is still name-derived, so a consensus-key rotation straddles the boundary and fail-closed-halts the network. This PR correctly removes `also_accept_next_committee_pubkeys` as dead v5→v6 scaffolding — that does not worsen #1961, but it does mean we would be carrying two protocol transitions' exposure while a known halt risk has only operator guidance behind it. Worth noting the ID-hash fix in #1961 would also **delete the need for this PR's dual-width retry**, since hashing IDs is insensitive to encoding width. That is an argument for doing #1961 first and simplifying v7, not for blocking v7 indefinitely.

## Suggested sequence

- **v1.2.8 now** — #1958 + the two inkrypto commits already on main. No protocol change, no clock, ships the registry fix to a fleet that needs it.
- **#1961** next, ideally landing the ID-hash change (which simplifies this PR).
- **v1.2.9 = this PR**, with: a fault-injection run of `v127_v7_upgrade` (your own ask #1 — the run is green but I did not find an injected-failure control for the gate itself), and an operator announcement that states plainly that upgrading arms v7 at quorum+buffer. The cohort-1 lesson deserves to be the headline of that announcement.

On the MIN=6 half specifically: no objection at any time. It is unreachable-code removal with the `get_for_version` MIN assertion as the proof, and both networks are on v6. If you want to decouple, that commit could ride v1.2.8 safely — it narrows compatibility only against v5 networks, of which we have none.

omersadika:
## Retracting the split recommendation — ship this in v1.2.8. The adoption data says my argument was wrong.

My earlier comment argued for holding v7 to v1.2.9 on the grounds that bundling it with an urgent fix compresses the deadline for slow operators. I went and measured the thing I was reasoning about, and it doesn't support the conclusion.

Mainnet adoption of v1.2.7, from tag to v6 activation:

```
07-28 23:42   5 validators    <- tag lands; only the dWallet Labs canaries
07-29 -> 07-30 11:42   4-6    <- ~1.5 days of essentially nothing
07-30 17:42  38               <- announcement; the fleet moves in ONE burst
07-30 23:42  48
07-31 10:39  v6 activates
```

**The fleet does not adopt gradually across the release window — it sits idle until the announcement, then moves in a single ~6-hour burst.** The release date is not the clock; the announcement date is. The 1.5 days between tag and burst bought nothing.

That reframes the five validators that halted at the flip. They were not operators who ran out of runway — they were operators who didn't act while 48 of their peers did, in the same window, and they all upgraded within hours *after* halting (they are on v1.2.7 now, uptimes 18–27h). The failure mode was attention, not time.

Which kills my argument: **deferring v7 to v1.2.9 does not shrink that risk, it reschedules it.** The same operators will be equally slow for the next release. Delay would buy soak time your gates have already provided — `v127_v7_upgrade` green on this branch (run 30675628560), the dual-width retry fault-validated at unit level — while postponing a registry fix the fleet needs now.

The other two objections I raised don't survive scrutiny either. Rollback protection was illusory: #1958 is a fix, so preserving the ability to revert it protects nothing anyone would want. And attribution is easier than I claimed — a #1958 regression presents as empty registry reads, a v7 regression as handoff-signature rejection; different subsystems, different symptoms.

## What I'd still ask for, in the announcement rather than the release

1. **Lead with the halt condition, not the fix.** "v1.2.8 arms protocol v7; v7 activates at the first epoch boundary after two-thirds of stake upgrades — likely within ~24h of the upgrade wave — and any validator not on v1.2.8 at that moment halts." That sentence's absence is what the v6 flip cost us.
2. **Ping the known tail directly** when it ships: 3MI Labs, Brightlystake, Coinage x DAIC, Studio Mirai, bitszn — the exact five that missed v6.
3. **Use the arming window.** `ika_protocol_upgrade_armed` and dashboard panel 1152 now show supporting stake crossing the threshold, which gives a real window to chase stragglers before the boundary rather than discovering them after it.

Unchanged: #1961 stays open regardless. v7's dual-width retry covers the encoding-width straddle but not consensus-key rotation, so "do not rotate consensus keys" remains guidance-only on both networks until the ID-hash change lands.

Issue #1838, opened 2026-07-14
v1.2.0 release blocker: incomplete v4 epoch-transition safety and liveness gates


omersadika:
# v1.2.0 challenge re-review

Reviewed range: `7d10043b96..d4ae79470c`

Previous release: v1.1.8  
Proposed release: v1.2.0

## Verdict

**Do not publish v1.2.0 yet.**

The new commits materially improve the branch, but the central v4 activation failure remains. Of the original 19 findings:

- 6 are fully resolved.
- 2 are partially resolved.
- 11 remain open.

### [P0] v4 is still advertised automatically despite known-incomplete activation gates

`MAX_PROTOCOL_VERSION` remains 4 in `crates/ika-protocol-config/src/lib.rs`, while the repository itself says the cert-less barrier escape, full startup-path barrier coverage, and sequence-pure close gate were deliberately not built for 1.2.0. The implementation and release posture contradict each other.

### [P0] Epoch close still depends on validator-local state and wall-clock progress

`handoff_signatures_meet_quorum` in `crates/ika-core/src/authority/authority_per_epoch_store.rs` reads a locally populated signature table. The call path then feeds that into `decide_v4_epoch_close`. Validators observing different signature arrival timing can still reach different decisions for the same sequenced state.

### [P0] The certificate-free liveness backstop can still close an epoch whose next-epoch barrier has no escape

At `grace * 4`, close deliberately fires without a handoff certificate. But the next epoch can block forever if no certificate exists anywhere. This is not hypothetical: `dev-docs/plans/handoff-barrier-escape-and-pure-close-gate.md` explicitly describes that failure and says the fix was not included in 1.2.0.

### [P0/P1] “Consensus-agreed NOA key” remains an unjustified claim

Counters are now correctly partitioned per `(NetworkKeyId, curve, algorithm)`, but NOA selection uses the validator's local `adopted_network_key_data` in `crates/ika-core/src/dwallet_mpc/mpc_manager.rs`. Adoption is certificate-gated but explicitly not round-uniform. During installation skew, one validator can see `{K0, K1}` and choose K0 while another sees only `{K1}` and chooses K1. The existing lag test fixes the NOA key everywhere and therefore does not exercise this hole.

The code also explicitly accepts that sufficient start skew leaves a validator permanently ordinal-offset for the epoch. Partitioning the blast radius is useful, but it is not convergence.

### Ready-signal canonicalization

[...]

PR #1964, opened 2026-08-01
fix(epoch-close): route handoff-signature rows through the commit batch (#1927)

### LINKED ISSUE #1926: Crash-window audit after #1920: two open determinism questions (handoff-signature close gate, presign-pool pops) and a write-discipline convention for the epoch store

### LINKED ISSUE #1927: Handoff-signature quorum is the last non-deterministic input to the epoch-close gate: buffered signatures drain at wall-clock install time

omersadika:
Fixes #1927. Also closes #1926 — its Q1 was split into #1927 (this), its Q2 into #1928 (presign pools, still open), and its write-discipline hygiene item already landed.

**Target v1.2.8.** Correcting an earlier note in this body: v1.2.8 carries the *addition* of protocol v7 (`MAX = 7`, advertised), not an activation — the flip is a separate network vote, so there is no protocol transition to collide with. What v1.2.8 does carry alongside this is the strict-only inkrypto bump, which changes v6 runtime behaviour; so if a close-round or reshare anomaly appears on a network, both land in the same tag. `v6_rollout` is the mitigation and it is green: mixed committee against the literal deployed release, which is where an ungated close-timing shift would surface.

## The bug

The close gate reads `handoff_signatures_meet_quorum`, a stake sum over the `handoff_signatures` table. That table had two writers, neither attributable to a commit:

| writer | when it wrote |
|---|---|
| consensus `EndOfPublishV2` arm | sequenced, but inserting on sight, outside `ConsensusCommitOutput` |
| buffered drain in `install_expected_handoff_attestation` | **wall-clock local** — whenever this validator's own hydrate+build pass finishes |

(A third, the stale-row cleanup on an attestation re-install, deleted rows out-of-band on the same wall-clock schedule — moving the gate *down*.)

So the gate could be decided against durable state that no commit produced, and a crash left the table ahead of the last committed commit. Same out-of-band-write class as #1829 and #1917, on the other half of the gate #1920 just fixed.

## The fix

Every writer stages its row mutations on the epoch store; the next commit folds the staged ops into its `ConsensusCommitOutput`, and `write_to_batch` applies them. A row is durable exactly when the commit it was staged under is.

The gate reads the committed table **overlaid with the rows the evaluating commit itself staged**. Both halves are load-bearing:

- Reading only the table would push a signature sequenced by *this* commit into the *next* commit's gate — a one-commit close delay against a binary that writes on sight, i.e. a mixed-rollout close-round split. Avoiding exactly that is why #1920 was careful to leave live-path timing unchanged.
- Reading the un-flushed staging map instead would let rows destined for a later commit decide this one.

One commit rather than two: the split would have a broken middle, since routing the writes *without* the overlay is precisely the close-timing shift above — reproduced below by fault injection.

## What this does not do

It does not make the gate a pure function of the consensus sequence. **Which** commit a drained row lands under still follows the local attestation install, so two validators can still cross the handoff quorum at different commits. Retiring that residue is the sequence-pure tally, item 3 of `dev-docs/plans/handoff-barrier-escape-and-pure-close-gate.md`, which depends on the barrier's cert-less escape and full start-path coverage. Until then close-safety rests where the call-site NOTE already says: buffered-quorum adoption plus the `grace*4` liveness backstop. This is stated at the gate, in the spec, and in the test doc-comments rather than left to be rediscovered.

## What it costs

Rows staged into a commit whose batch is lost to a crash are gone, where a direct write kept them — consensus does not redeliver an already-processed bundle. Bounded by one commit; recovery is the same barrier peer-fetch that already covers a divergent validator. The **certificate** is deliberately not batched: it reaches perpetual storage the moment it is minted, so a quorum formed before such a crash survives. Same trade #1920 made for votes.

Incidental: the gate now propagates a `handoff_signatures` decode error instead of silently skipping the row (`filter_map(Result::ok)` is gone). The install path already propagated on those same rows, so a corrupt row was never benign — and a silently-skipped row is exactly the divergence at issue.

## Docs

`handoff_signatures` moves from `direct — UNPROVEN (#1927)` to `commit-batched`. `dev-docs/specs/handoff.md` and `dev-docs/conventions/epoch-table-write-discipline.md` are updated in the same change. `./scripts/check-epoch-table-write-discipline.sh` passes (56/56).

## Verification

Two new regression tests drive the real `process_consensus_transactions` path with real `EndOfPublishV2` bundles, **fault-validated** per `dev-docs/playbooks/test-testing.md`.

Injection 1 — restore the direct table write:

| test | result |
|---|---|
| `handoff_gate_rows_are_commit_attributable_across_install_skew` | FAILED: `drained rows must not be durable until a commit carries them` (2 vs 0) |
| `handoff_gate_survives_crash_before_signature_commit_batch` | FAILED: `signature rows must ride the commit batch` (3 vs 0) |
| 22 pre-existing tests | all pass |

All pre-existing tests passing under that injection — including #1920's own crash-replay test — is the control: this window was invisible to them and needed its own guards.

Injection 2 — drop the commit-output overlay from the gate: fails all three gate assertions, including the pre-existing stale-row test's `the close gate must not count the superseded rows`. That proves the overlay matters in both directions, not just for inserts: without it the gate would keep counting superseded endorsements at the commit that dropped them.

Both injections reverted; no fault artifacts remain. Whole `authority_per_epoch_store` module green (24/24); `cargo clippy --release --workspace --all-targets --all-features` clean.

This touches epoch-close timing, so the heavy suites and the upgrade matrix ran before merge. All green on `6a0ee7f419`:

| suite | run | result |
|---|---|---|
| CI (build / fmt / clippy / unit) | [30707783773](https://github.com/dwallet-labs/ika/actions/runs/30707783773) | ✅ |
| Test Cluster | [30707787743](https://github.com/dwallet-labs/ika/actions/runs/30707787743) | ✅ |
| Integration Tests CI (`test_threads=4`) | [30707788835](https://github.com/dwallet-labs/ika/actions/runs/30707788835) | ✅ |
| TS Integration Tests | [30707790666](https://github.com/dwallet-labs/ika/actions/runs/30707790666) | ✅ |
| Upgrade Test — PR default (`v6_rollout`) | [30707783774](https://github.com/dwallet-labs/ika/actions/runs/30707783774) | ✅ |
| Upgrade Test — full matrix (`test=all`) | [30707789714](https://github.com/dwallet-labs/ika/actions/runs/30707789714) | ✅ 8/8 |

`v6_rollout` carries the specific weight here: a mixed committee against the literal v1.2.7 release is where a one-commit gate shift relative to an on-sight-writing binary would surface.

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -58,6 +58,7 @@ use crate::system_checkpoints::{
 };
 use commitment::CommitmentSizedNumber;
 use dwallet_mpc_types::dwallet_mpc::DWalletSignatureAlgorithm;
+use fastcrypto::ed25519::Ed25519Signature;
 use group::PartyID;
 use ika_network::mpc_artifacts::mpc_data_blob_hash;
 use ika_protocol_config::{Chain, ProtocolConfig, ProtocolVersion};
@@ -1011,6 +1012,17 @@ pub struct AuthorityPerEpochStore {
     pending_handoff_signatures:
         parking_lot::Mutex<Vec<ika_types::handoff::HandoffSignatureMessage>>,
 
+    /// Pending `handoff_signatures` row mutations, waiting to be folded into
+    /// the next `ConsensusCommitOutput`. `Some(signature)` is an upsert,
+    /// `None` a delete. Every writer of that table stages here instead of
+    /// writing it, so each row lands in exactly one commit's batch and the
+    /// durable table can never run ahead of the last committed commit —
+    /// which is what makes the close gate's read of it commit-attributable
+    /// (#1927). Keyed by signer with last-op-wins, so it stays bounded by
+    /// the committee size no matter how often a peer re-broadcasts.
+    staged_handoff_signature_rows:
+        parking_lot::Mutex<BTreeMap<AuthorityName, Option<Ed25519Signature>>>,
+
     /// Buffer of relayed next-epoch joiner announcements received via
     /// consensus while this validator's `JoinerPubkeyProvider` was
     /// absent or lagged the next-epoch committee (so the joiner's
@@ -1707,19 +1719,24 @@ pub struct AuthorityEpochTables {
     /// `AuthorityPerpetualTables` (perpetual persist lands in step
     /// 7c).
     ///
-    /// write-discipline: direct — UNPROVEN (#1927). Two writers:
-    /// `record_handoff_signature` on the consensus `EndOfPublishV2` arm
-    /// (consensus-sequenced, but written out-of-band relative to the commit
-    /// batch) and the buffered drain in
-    /// `install_expected_handoff_attestation`, which runs at WALL-CLOCK local
-    /// install time. The close gate reads this table
-    /// (`handoff_signatures_meet_quorum`), so a validator whose install lags
-    /// its peers' signature commits can observe the gate satisfied at a
-    /// different round — the property #1920 removed from the `all_voted` half
-    /// of the same gate. The pure-function-of-table argument holds for the
-    /// consensus arm alone; the drain breaks it, because drained rows are not
-    /// attributable to any commit ≤ the evaluating one.
-    pub(crate) handoff_signatures: DBMap<AuthorityName, fastcrypto::ed25519::Ed25519Signature>,
+    /// write-discipline: commit-batched (#1927). Every writer — the consensus
+    /// `EndOfPublishV2` arm, the buffered drain in
+    /// `install_expected_handoff_attestation`, and that same function's
+    /// stale-row cleanup — stages into `staged_handoff_signature_rows`, which
+    /// the next commit folds into its `ConsensusCommitOutput`. So a row is
+    /// durable exactly when the commit it was staged under is, and the close
+    /// gate that sums this table (`handoff_signatures_meet_quorum`) reads a
+    /// state attributable to a commit rather than to whatever had reached
+    /// RocksDB by the instant it looked.
+    ///
+    /// This does NOT make the gate a pure function of the sequence: WHICH
+    /// commit a drained row lands under still follows this validator's local
+    /// attestation install, so peers can still cross quorum at different
+    /// commits. Retiring that residue is the sequence-pure tally in
+    /// dev-docs/plans/handoff-barrier-escape-and-pure-close-gate.md (item 3),
+    /// which has its own prerequisites; see the NOTE at the gate's call site
+    /// for what holds close-safety in the meantime.
+    pub(crate) handoff_signatures: DBMap<AuthorityName, Ed25519Signature>,
 
     /// Local cache of network DKG output digests for this epoch,
     /// keyed by `dwallet_network_encryption_key_id`. Populated by
@@ -2525,6 +2542,7 @@ impl AuthorityPerEpochStore {
             joiner_pubkey_provider: ArcSwapOption::empty(),
             expected_handoff_attestation: ArcSwapOption::empty(),
             pending_handoff_signatures: parking_lot::Mutex::new(Vec::new()),
+            staged_handoff_signature_rows: parking_lot::Mutex::new(BTreeMap::new()),
             pending_relayed_joiner_announcements: parking_lot::Mutex::new(Vec::new()),
             handoff_aggregator: parking_lot::Mutex::new(None),
             perpetual_tables_for_handoff: ArcSwapOption::empty(),
@@ -3153,7 +3171,6 @@ impl AuthorityPerEpochStore {
         // possible (no early-startup fallback needed). Order doesn't
         // matter — the aggregator is stake-weighted.
         let committee = self.committee();
-        let tables = self.tables()?;
         let mut replayed_signatures: usize = 0;
         // Rows that endorse a superseded attestation are dropped from the
         // aggregator AND deleted from the table below. The close-gate quorum
@@ -3161,8 +3178,14 @@ impl AuthorityPerEpochStore {
         // aggregator, so leaving stale rows behind would let a re-install
         // that changed the attestation still count the old endorsements.
         let mut stale_signers: Vec<AuthorityName> = Vec::new();
-        for entry in tables.handoff_signatures.safe_iter() {
-            let (signer, signature) = entry?;
+        // Rows already staged for the next commit count here even though they
+        // are not durable yet. They were verified against an attestation and
+        // consensus will not redeliver them, so an install that read only the
+        // committed table would drop every signature recorded since the last
+        // commit out of the rebuilt aggregator — permanently, since the
+        // aggregator is only ever rebuilt from this read.
+        let staged = self.staged_handoff_signature_rows.lock().clone();
+        for (signer, signature) in self.handoff_signature_rows_with(&staged)? {
             let msg = ika_types::handoff::HandoffSignatureMessage {
                 attestation: attestation.clone(),
                 signer,
@@ -3183,20 +3206,18 @@ impl AuthorityPerEpochStore {
             aggregator.insert_verified(signer, signature);
             replayed_signatures += 1;
         }
-        // Atomically delete the superseded rows so the table matches the
-        // installed attestation (single write batch — no half-deleted state).
-        // Done after the aggregator is fully built, so a delete failure
-        // surfaces as an error without having touched the correct in-memory
-        // aggregator. Idempotent: a crash before the write leaves the rows to
-        // be re-identified and re-deleted on the next install.
+        // Stage the superseded rows for deletion so the table matches the
+        // installed attestation. Staged rather than deleted on sight for the
+        // same reason the inserts are: this moves the close gate DOWN, so the
+        // commit it moves down at has to be a commit, not the wall-clock
+        // instant this install happened to run. Idempotent — a crash before
+        // the batch leaves the rows to be re-identified on the next install.
         if !stale_signers.is_empty() {
-            let mut batch = tables.handoff_signatures.batch();
-            batch.delete_batch(&tables.handoff_signatures, stale_signers.iter())?;
-            batch.write()?;
+            self.stage_handoff_signature_rows(stale_signers.iter().map(|signer| (*signer, None)));
             info!(
                 epoch = attestation.epoch,
                 dropped = stale_signers.len(),
-                "deleted superseded handoff signature rows on attestation re-install"
+                "staged superseded handoff signature rows for deletion on attestation re-install"
             );
         }
         let aggregator_signer_count = aggregator.signer_count();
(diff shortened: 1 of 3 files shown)


omersadika:
## Validation — all suites green on `6a0ee7f419`

| suite | run | result |
|---|---|---|
| CI (build / fmt / clippy / unit) | [30707783773](https://github.com/dwallet-labs/ika/actions/runs/30707783773) | ✅ |
| Test Cluster | [30707787743](https://github.com/dwallet-labs/ika/actions/runs/30707787743) | ✅ |
| Integration Tests CI (`test_threads=4`) | [30707788835](https://github.com/dwallet-labs/ika/actions/runs/30707788835) | ✅ |
| TS Integration Tests | [30707790666](https://github.com/dwallet-labs/ika/actions/runs/30707790666) | ✅ |
| Upgrade Test — PR default (`v6_rollout`) | [30707783774](https://github.com/dwallet-labs/ika/actions/runs/30707783774) | ✅ |
| Upgrade Test — full matrix (`test=all`) | [30707789714](https://github.com/dwallet-labs/ika/actions/runs/30707789714) | ✅ all 8 scenarios |

Full matrix: `smoke`, `workload`, `v6_rollout`, `v127_churn`, `malicious_v127`, `v127_v7_upgrade`, `restart_spectator`, `legacy_config`.

Two of these carry the weight for this change specifically:

- **`v6_rollout`** — the deployed-release gate. Boots the literal v1.2.7 release, upgrades one validator to this build, converges two mixed aggregated reshares with per-authority byte-equality and zero malicious, then swaps the rest. This is the mixed-committee evidence that the close gate did **not** shift by a commit relative to a binary that still writes on sight — the specific hazard the table∪staged overlay exists to avoid.
- **`restart_spectator`** — exercises the restart path where the aggregator is rebuilt from the durable rows, which is the read this PR changed to fold in staged-but-uncommitted rows.

omersadika:
Reviewed the diff line-by-line and re-verified the claims independently rather than from the description. **LGTM — approve** (posting as a comment; formal approval to follow from a maintainer).

**What I checked myself:**
- **No writer escapes the discipline.** Grepped the branch: zero direct `handoff_signatures` writes remain outside tests. All three writers (consensus arm ×2 outcomes, drain, stale sweep) stage; `write_to_batch` is the only path to RocksDB.
- **Fold-before-gate ordering holds**: fold at :5055, gate read at :5163, same function, no early return between them that skips the fold.
- **The only production gate caller is the close decision** — the signature change to `pub(crate)` with the output param breaks nobody.
- **Heavy suites are green**, not just dispatched: Upgrade Test 30707789714, Test Cluster 30707787743, Integration 30707788835, TS Integration 30707790666. Worth pasting these into the body where it says "run links to follow".
- The overlay-in-both-directions argument is the part most reviews would miss, and injection 2 proves it properly — without the overlay the gate keeps counting rows the evaluating commit itself deleted. The pre-existing stale-row test failing under that injection is the right control.

**Three notes, none blocking:**

1. **A pre-existing race worth a one-line comment so it isn't rediscovered as "the fix missed a writer":** a consensus-arm row that verifies against attestation A can be staged concurrently with a re-install to B — the sweep clones the staging map before that stage lands, so the A-row survives into the table under B. This is exactly the window the old code had (an insert landing after the sweep's table iteration), so it is not a regression, and the sequence-pure tally retires it — but a comment at the sweep saying so would save the next reader the same hour it cost me.

2. **"Bounded by one commit" undersells the drain case.** The crash *window* is one commit, but the payload lost in it can be the entire buffered backlog a late install drained (many signers' rows), where a direct write kept them. Recovery is the same peer-fetch either way, so the trade stands — I'd just reword to "bounded to a one-commit window" so nobody reads it as "at most one row".

3. **The crash-replay doc line is conditional in a way worth one word:** "a crash-replay of this commit re-derives the same input" holds once the attestation is installed at replay time; if replay outruns install, the bundles buffer and the close arrives later via drain-or-backstop. That is squarely inside the acknowledged residual ("WHETHER a row exists still depends on off-consensus state") and the crash test models the install-first case explicitly — but the pre-fix rows being (wrongly) durable made replay look unconditionally convergent, so someone comparing behaviours may flag this as a regression unless the comment says "once the expected attestation is reinstalled".

Agree with the release targeting: v1.2.9+, not v1.2.8 — this changes close timing and must not ride the v7 flip. Since it closes #1926 too, note that #1928 (presign pools) is now the last `direct — UNPROVEN` holder, which the convention-doc update already reflects.

omersadika:
All three notes applied in `818c0bb8c4` — comment-only, no behaviour touched. Module still green (24/24).

**1. Concurrent-stage window at the sweep** — confirmed and documented at the sweep. Also confirmed it is genuinely pre-existing rather than taking that on trust: the old code iterated the table then batch-deleted, so an insert landing after that iteration survived the same way. Both versions decide staleness from a snapshot the consensus thread can outrun. The comment says why a wider lock here is the wrong fix (it moves the race to the arm) and that the sequence-pure tally keyed by attestation digest is what actually retires it.

**2. Window vs payload** — reworded. The window is one commit; the payload is not one row, since a late install drains its whole buffered backlog into a single commit. Your reading was the natural one and the original wording invited it.

**3. Crash-replay condition** — added at both sites (the gate doc and the call-site NOTE): replay re-derives the gate's input *once the expected attestation is reinstalled*; replay that outruns the reinstall buffers the redelivered bundles and closes later via drain-or-backstop. Agreed this sits inside the acknowledged residual, but the pre-fix rows being wrongly durable made replay look unconditionally convergent, so it was worth one clause to stop someone reading it as a regression.

Run links are now in the PR body where the placeholder was.

On release targeting — flagging a correction to my own body text rather than leaving it: v1.2.8 carries the *addition* of protocol v7 (`MAX = 7`, advertised), not an activation; the flip is a separate network vote. The real argument for holding is that v1.2.8 already carries the strict-only inkrypto bump, which does change v6 runtime behaviour, and this PR adds an ungated epoch-close write-ordering change on top — so a close-round or reshare anomaly would have to be bisected inside one tag. Whether that outweighs shipping an #1838 P0 sooner is a release-schedule call, not a technical one.

omersadika:
All three notes addressed exactly as asked, comment-only as claimed (verified the diff touches nothing but comments and the spec). The stale-sweep comment is better than what I asked for — naming *why* a wider lock is the wrong closure (it only moves the race to the consensus arm) and pointing at the digest-keyed tally is the part a future reader needs. Checks green on 818c0bb8c4. Ready to merge from my side — target v1.2.9 as the PR states; closes #1927 and #1926 with it.

omersadika:
**Validation record — which commit the suites actually cover.**

The heavy suites in the body ran against `6a0ee7f419`. The current tip is `818c0bb8c4` (the review-notes commit), so to be accurate rather than to imply the suites covered the tip:

`818c0bb8c4` changes 25 lines across `authority_per_epoch_store.rs` and `dev-docs/specs/handoff.md`. Filtering the Rust diff to non-comment lines yields **nothing** — it is comments and spec prose only, zero executable changes:

```bash
git show 818c0bb8c4 -- crates/ | grep -E "^[+-]" | grep -vE "^[+-]{3}" | grep -vE "^[+-]\s*(//|///)" | grep -vE "^[+-]\s*$"
# (empty)
```

So the `6a0ee7f419` suite results carry over unchanged, and I have not re-dispatched ~2h of runners for a comment-only delta. If you would rather the record show green suites on the exact merge SHA, say so and I will re-dispatch — it is cost, not risk.

Still open before merge: an approving **review** (the LGTM above is a comment, so `reviewDecision` is still `REVIEW_REQUIRED` and the PR is `BLOCKED`), plus the two checks still pending on the tip (`v6_rollout`, `Clippy (msim / simtest)`).

The branch is 1 commit behind `main` (#1963) but `MERGEABLE` — no conflict, so a rebase is optional.

Issue #1952, opened 2026-07-28
v1.2.6 release blocker: mid-epoch restart leaves the validator a pure MPC spectator (0 active sessions, completions via reconstruction only) — 6/6 dWallet Labs validators, both networks


omersadika:
## Do-not-ship gate finding: verified on the canary rollout before fleet announcement

All six dWallet Labs validators (5 mainnet, 1 testnet) were upgraded to `v1.2.6 (74a44ed06deb)` via a routine mid-epoch restart ~80 minutes before these measurements. Every one is MPC-dead in a new way that the pre-1.2.6 wedge taxonomy does not match.

## The signature (identical on all six)

| subsystem | state |
|---|---|
| consensus | healthy — `ika_consensus_last_committed_timestamp_seconds` staleness 17–60s on all |
| network key | adopted — `network_key_loaded_epoch` = current epoch (363 mainnet / 366 testnet) |
| canonical anchors | v4 everywhere |
| prior-cert key path | clean — `prior_cert_blobs_missing` = 0, zero cert-path warnings in logs (the #1881 gauge earning its keep) |
| **advances** | **~zero** — testnet host: 2 round-completions in 84 minutes vs a ~22,000/h peer median; mainnet hosts: no post-restart advance series at all vs ~8,500/h median |
| session map | `active: 0`, `waiting_for_session_request: 9`, **`completed: 20,332`**, `computation_completed: 35`, `failed: 0` |
| reconstruction | 1,000–4,100 sessions/30m added with `active: false, session_origin: "reconstructed_from_consensus"` |
| readiness loop | `ready_to_advance_result_total`: 0 ready AND 0 not_ready — never evaluated |

The `completed: 20,332` with `active: 0` is the tell: sessions transit reconstructed → completed on peers' quorum without ever activating locally. The validator follows everything and computes nothing. This is the #1852 shape ("sessions park... re-instantiation never attempted") reappearing on a binary where every previously-diagnosed cause reads healthy.

## Why the gates didn't catch it — the question that matters as much as the bug

[...]

PR #1965, opened 2026-08-01
fix(dwallet-mpc): converge internal-presign ordinals from consensus completions, and export the divergence (#1830)

### LINKED ISSUE #1830: Internal-presign pool counters never converge after a start skew beyond one batch lifecycle (heal: fast-forward from observed completed sequence numbers)

omersadika:
Closes #1830.

## The residual this closes

Per-pool internal-presign ordinal counters are in-memory while a pool's ordinal
stream belongs to the epoch, so a validator can end up minting identifiers the
committee finished long ago. Those mints never produce live work (peers
early-return on an already-resolved identifier), and the offset advances in
lockstep with the live window — constant offset, zero closing speed — so the
validator sits out registry-driven instantiation for that pool for the rest of
the epoch while every event-driven path still looks healthy. Above f stake in
that state starves the pool below the MPC threshold and the epoch cannot close.

**#1952 (#1953) already converges the reachable in-process topologies**: it
seeds a pool's counter on first mint from the persisted
`filled_presign_pool_slots` high-water and walks the mint path past terminal
replayed sessions. I could not construct an organic repro on `main` that fails
to converge — a fill is written by every validator that processes the output,
so the persisted high-water tracks the frontier on any node holding the epoch's
rounds.

What was missing is the rule #1830 itself specifies, and it is not redundant:
both #1952 sources are **local**, and both are consulted only at a pool's FIRST
mint (the seed is read once) or **while a top-up is firing**. A counter that is
already live and has fallen behind is never repaired unless a top-up happens to
run — and then only one ordinal per walk step, bounded at 512/mint with a
`should_never_happen` marker past the budget.

## The heal

Every completed internal-presign output carries its `session_sequence_number`,
so the committee's completion frontier is consensus-anchored data every
validator holds regardless of what it instantiated, what its store persisted,
or whether it was in the pool when the ordinal was minted.
`handle_mpc_internal_output` now records that frontier per
`(NetworkKeyId, curve, algorithm)` and jumps `next_internal_presign_sequence_number`
past it whenever the counter is at-or-below. The same frontier joins the
persisted high-water as a seed source (`max` of the two), covering a seed read
that errored or a store that never held the epoch's fills.

Safe by construction: reads only consensus-agreed data; moves counters FORWARD
only, and only to one past an ordinal the committee has already completed
(never past what peers minted); a validator already at the frontier is
untouched. Batches complete out of sequence order, so a jump can skip a
still-in-flight ordinal below the frontier — that is the same trade the
existing `high_water + 1` seed already makes, deliberately kept identical so a
pool's rejoin point does not depend on which source repaired the counter.

**No protocol gate**: identifiers stay content-derived and no checkpoint bytes
depend on which validator instantiated a session.

**Deviation from the issue's sketch**: it also suggested reconciling the
instantiated/completed guards at the jump. Not done, on purpose — the guard is
pool-aggregated and saturating, so peers' completions reopen it within one
batch, whereas forcing `completed = instantiated` can release it while this
validator's own sub-frontier mints are still live, overlapping batches (the
pool overshoot the guard exists to prevent). Reasoning is in the code.

## The metric (there was none)

The counters are in-memory, so the condition was only ever indirectly visible —
a validator contributing zero advances to one `(NetworkKeyId, curve, algorithm)`
pool while healthy everywhere else. That is why this sat 20 days. Both are
labeled `curve`/`signature_algorithm`/`key_role`, the same bounded reduction as
`internal_presign_pool_size`:

- `ika_dwallet_mpc_internal_presign_ordinal_lag` (gauge) — ordinals between this
  validator's next mint and the committee's completed frontier. 0 in steady
  state (a pool is minted before it completes); positive while a stream sits
  inside already-completed history. Published by the top-up loop from the state
  at the top of a round, so a round that heals still publishes the divergence it
  started with.
- `ika_dwallet_mpc_internal_presign_ordinals_fast_forwarded_total` (counter) —
  ordinals skipped to rejoin. The gauge reads 0 again after a jump, so this is
  what shows the condition happened: one large jump is a mid-epoch rejoin; a
  sustained dribble is a validator not minting for that pool at all, dragged
  along by its peers' completions.

## Validation

- `test_internal_presign_ordinal_stream_rejoins_from_consensus_completions` —
  pins a validator out of minting, drags its counter into already-completed
  history, and requires the peers' completions ALONE to rejoin it, with no mint
  of its own; asserts the gauge reports the exact distance while diverged and
  settles to 0 after, and that the rejoined stream mints live ordinals every
  peer binds to the same session identifier.
- **Fault-validated** (per `dev-docs/playbooks/test-testing.md`): with the
  fast-forward disabled (frontier still recorded), the run fails at the
  predicted assertion — the counter never converges across 60 rounds of peer
  completions — while the phase-2 gauge assertion still passes, proving the
  metric reacts to the condition independently of the heal.
- Whole `internal_presign` module, release: 8/8 pass, including #1952's
  `test_mid_epoch_restart_resumes_internal_presign_ordinals_from_pool_high_water`
  (the seed change does not regress it). Clippy clean.
- Heavy suites + full upgrade matrix (`test=all`, which includes the #1952
  `restart_spectator` gate) dispatched on this branch; conclusions recorded
  below as they land.

## Docs

`dev-docs/specs/internal-presign-pool.md` gains an "Ordinal-stream convergence"
section (the three convergence sources and what each needs locally, the safety
argument, the two metrics), `fast-schnorr-vss.md`'s "Scope of the guarantee" no
longer describes this as tracked follow-up, and the metric-name inventory is
updated.


omersadika:
Reviewed the diff line-by-line and re-verified the claims against the branch. **LGTM — approve, with one merge gate: the full upgrade matrix (run 30708631777, `test=all`) is still in progress.** The green Upgrade Test on the PR is the PR-triggered subset; the dispatched full matrix is the one that carries `restart_spectator`, which is precisely the gate this change must not regress (it reworks the #1952 seed path). Wait for it before merging.

**What I checked myself:**

- **The heal is correctly consensus-anchored.** `observe_completed_internal_presign_ordinal` hangs off the `Resolved` arm of the completion handler, i.e. it reads the `session_sequence_number` of quorum-agreed outputs that every validator processes regardless of participation — the one source with no local precondition, exactly as #1830 specified. Forward-only, idempotent on replayed completions, and a participating validator is provably untouched (it minted every ordinal it later sees complete, so `next > frontier` always holds in steady state).
- **The seed `max()` is correct including the `Option` semantics** (`None < Some(_)`), and the two sources' blind spots genuinely complement: the persisted high-water covers pre-adoption replay fills (no `NetworkKeyId` to record a frontier under), the frontier covers a store that never held the fills. The error arm falling back to the frontier instead of a bare `1` is a real improvement over main.
- **The guard non-reconciliation is the right call**, and the reasoning comment is exactly where the next person will need it. Forcing `completed = instantiated` at the jump would overlap batches — the overshoot the guard exists to prevent — while peer completions reopen it within one batch anyway.
- **The test earns its length.** Phase 2's manual counter pin is the honest way to do this given the PR's own finding that no organic repro exists on main (#1952 covers the reachable topologies — worth keeping that admission in the body, it is evidence, not weakness). The discriminating assertions are right: the heal must come from the output stream with zero local mints, and phase 4's every-peer-binds-the-same-identifier check is the property that actually matters. Driving the export site directly to sample the gauge mid-divergence is legitimate given the heal and the observation share a call.
- **Metrics follow the house rules**: bounded labels (`curve`/`algorithm`/`key_role`), inventory updated, and promoting the `key_role` strings to shared constants kills the drift risk between the two sites.

**Three notes, none blocking:**

1. **The gauge's label reduction hides per-key resolution**: several network keys with `key_role="other"` collapse onto one series, last-write-wins per export. Same reduction as `internal_presign_pool_size`, so it is consistent — but dashboards should treat the gauge as per-role, and the log line (which carries the full key id) as the drill-down.
2. **`internal_presign_restart_seeded_pools` semantics widened silently**: a frontier-only seed (fresh store, no persisted fills) now also lands in the set, so whatever its drain emits should say "mid-epoch rejoin" rather than "restart". Cosmetic.
3. **Ops follow-up on our side, not this PR's**: infra should grow a sweep query + alert on sustained `internal_presign_ordinal_lag > 0` and on the fast-forward counter's dribble shape — the PR gives us the signal; nothing pages on it yet.

Release note: unlike #1964 this one is NOT close-timing sensitive — the heal only moves a counter a wedged validator was burning on dead mints anyway — but it lands in `mpc_manager`, so I would still put it in v1.2.9 with #1964 rather than slipping it into v1.2.8 at this stage.

omersadika:
Thanks — actioned notes 1 and 2 in `3d9e55739c`; agreed on the merge gate and note 3.

**Note 2 (the `restart` misnomer).** Fixed properly rather than cosmetically: `internal_presign_restart_seeded_pools` → `internal_presign_mid_epoch_seeded_pools`, with its doc rewritten to say what the set now means (seeded mid-stream from the persisted high-water OR the consensus frontier, however this process got there — restart, late key install, or a store holding none of the epoch's fills). The resume line now reads "…after a mid-epoch restart or late rejoin".

That line is the `restart_spectator` needle, so the widening is deliberately a **suffix** — `wait_for_log_line_on_validator` is a substring match and the gate's literal is still an exact prefix. A comment at the log site says so, so nobody shortens it later without noticing. Since I changed a string a CI gate asserts on, I re-dispatched the gate on its own against the new SHA rather than trusting the argument: run [30715237671](https://github.com/dwallet-labs/ika/actions/runs/30715237671) (`test=restart_spectator`).

**Note 1 (label reduction).** Kept the reduction — consistency with `internal_presign_pool_size` is worth more than per-key cardinality here — and made your drill-down framing actually hold: the per-completion advance log now carries `dwallet_network_encryption_key_id` on **both** arms. Previously only the multi-ordinal jump did, so the `+1` dribble shape — precisely the "not minting for this pool at all" case where you would want to know *which* `other` key — had no key in it.

**Note 3.** Agreed, ops-side. The two shapes worth wiring: sustained `internal_presign_ordinal_lag > 0` (per validator × pool, with a `for` clause — a single scrape can catch a round mid-heal, since the gauge is published at the top of a round and settles on the next one), and a nonzero `rate(internal_presign_ordinals_fast_forwarded_total)` sustained over an epoch, which is the dribble.

**Merge gate.** The full matrix ([30708631777](https://github.com/dwallet-labs/ika/actions/runs/30708631777)) is down to `v6_rollout`; `restart_spectator`, `v127_v7_upgrade`, `malicious_v127`, `v127_churn`, `legacy_config`, `smoke` and `workload` are all green on it. Note the matrix ran against `54e96ebb66` — the three commits after it were docs/comments only, and `3d9e55739c` is the first to touch code again (a field rename plus two log strings), which is exactly why the gate is re-running rather than being inherited. Not merging until both land.

Release note: v1.2.9 with #1964 works for me.

omersadika:
Merge gate closed, and the follow-up commit is covered.

**Full matrix, `54e96ebb66`** — [30708631777](https://github.com/dwallet-labs/ika/actions/runs/30708631777) **success**, all eight scenarios: `smoke`, `workload`, `restart_spectator`, `v6_rollout`, `v127_v7_upgrade`, `v127_churn`, `malicious_v127`, `legacy_config`.

**`restart_spectator` re-run on `3d9e55739c`** — [30715928150](https://github.com/dwallet-labs/ika/actions/runs/30715928150) **success**. The first attempt ([30715237671](https://github.com/dwallet-labs/ika/actions/runs/30715237671)) failed in `Install Rust 1.94` — `rustup-init` panicked on the runner before checkout-build or any test, no artifacts — so it was infra, re-dispatched once per the playbook, not evidence about the string change.

The re-run confirms the suffix argument rather than just asserting it. From its log, both needles matched against the widened line:

```
[flow 11/19] wait_for_log_line_on_validator(0, "seeded internal-presign ordinal stream from the persisted pool-slot high-water") ... done in 8.0s
[flow 12/19] wait_for_log_line_on_validator(0, "internal presign top-up resumed live instantiation after a mid-epoch restart") ... done in 0.0s
restart_spectator PASSED: the restarted validator seeded its internal-presign ordinal stream from persisted state and resumed live registry-driven instantiation mid-epoch
```

Worth noting flow 11 passed too: that needle is a prefix of the seed line I rewrote in the original commit (it now names both sources — persisted high-water and consensus frontier), so both of this PR's log edits are gate-verified, not just the one from the review.

**Also on `3d9e55739c`**: PR checks (CI, Labeler, Move Formatter) green, and the `internal_presign` module re-run locally in release — 8/8, including `test_mid_epoch_restart_resumes_internal_presign_ordinals_from_pool_high_water`, so the `internal_presign_mid_epoch_seeded_pools` rename did not disturb the #1952 path. The PR-triggered `v6_rollout` on this SHA is still running; it is a strict subset of the matrix scenario already green, so I do not consider it gating — flagging it only so nobody reads the in-progress badge as an open question.

Ready to merge from my side.

PR #1966, opened 2026-08-01
fix(ci): restore the v127_rollout scenario name and the two guards it silently disabled

omersadika:
An incomplete rename of the upgrade-test rollout gate (`v127_rollout` → `v6_rollout`) left the old name in **nine** places. Two of them are **live control flow**, so both protections on the PR-default deployed-release gate quietly stopped covering it.

Found while reviewing #1964's evidence — that PR leaned on a green `v6_rollout` run, which prompted the question of whether the scenario name still described what it tests.

## The two live defects

**1. The real-crypto guard did not cover the gate.**

```bash
if { [ "$TEST_NAME" = "v127_rollout" ] || [ "$TEST_NAME" = "v127_churn" ] || [ "$TEST_NAME" = "malicious_v127" ]; } && [ "$USE_MOCKS" = "true" ]; then
  echo "$TEST_NAME is a rollout gate and must use real cryptography" >&2
  exit 1
fi
```

The first arm named a scenario that no longer existed, so `-f test=v6_rollout -f use_mocks=true` fell through and built with `--features dwallet-mpc-unsafe-mock`. The one job whose entire purpose is real cross-binary crypto evidence could be run on insecure deterministic mocks and pass.

**2. It ran with the reduced presign pools.**

`IKA_ENABLE_SMALL_PRESIGN_POOLS: "1"` is set unconditionally; the only `unset` was gated on the same dead name. The comment directly above that `unset` says the reduced pools are *"a resource concession for the lighter scenarios, not compatibility evidence"* — and the gate has been getting them.

## The fix, and why this direction

Renamed the scenario **back** to `v127_rollout` rather than repointing the guards at `v6_rollout`. That repairs both defects mechanically — the guards were already written for this name, which is itself the evidence the rename was the accident.

It also restores the convention every sibling follows: these gates are named for the **old binary's release** (`v127_churn`, `malicious_v127`, and the retired `v125_rollout` / `v121_rollout`), not for a protocol version. `v6_rollout` described something already finished — both networks run v6 — and it hid the maintenance: when the deployed release moves to v1.2.8 the gate must boot v1.2.8, at which point a protocol-version name stays superficially correct while the thing it tests changes underneath. That is exactly how nine references rotted unnoticed.

## Also fixed, same find/replace

- The header documented `v127_rollout` **and** `v6_rollout` as separate scenarios with near-identical topologies, though there is one matrix entry and one harness target. Merged, keeping the second's distinct claim (the reshares are strict-bound) rather than dropping it.
- `[ "$TEST_NAME" != "v6_rollout" ] && [ "$TEST_NAME" != "v6_rollout" ]`, with the message *"only supported by v6_rollout and v6_rollout"* — the same condition twice. Now a single check. `malicious_v127` is deliberately **not** a second supported scenario: it builds its own faulty validator rather than taking it from this input, and it is expected to **pass** (the honest committee convicts the fault) where this input must make its run **fail closed**. Opposite expected outcomes should not share an input.

## Standing rule, not a one-off

`dev-docs/playbooks/ci-suites.md` gains: retarget the whole `v12X_*` family whenever the deployed release moves — v1.2.8 → `v128_rollout`/`v128_churn`/`malicious_v128` with `old_ref=release/mainnet-v1.2.8` — with this incident recorded as why the names track the binary.

## Verification

- Workflow YAML parses; `cargo check -p ika-upgrade-test --tests` clean; no `v6_rollout` / `RUN_V6_ROLLOUT` reference remains anywhere in the repo.
- **Fail-closed check dispatched**: [run 30715468146](https://github.com/dwallet-labs/ika/actions/runs/30715468146) — `-f test=v127_rollout -f use_mocks=true`, which must now abort with *"is a rollout gate and must use real cryptography"* instead of building mocks. Result posted below.

No behaviour change to any scenario that was actually running. Independent of #1964 and does not block it; #1964's own runs used the default `use_mocks=false`, so its evidence stands.

### DIFF
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -9,9 +9,9 @@ name: Upgrade Test
 # dev-docs/specs/cross-binary-upgrade.md).
 #
 # Manual dispatch can run any scenario. Relevant pull requests run only
-# `v6_rollout`, avoiding the full expensive matrix on unrelated changes.
+# `v127_rollout`, avoiding the full expensive matrix on unrelated changes.
 # (The v3/v4-era rehearsals — cross_binary, v118_*, v121_rollout — were
-# retired when protocol v3/v4 support was removed; v6_rollout is their
+# retired when protocol v3/v4 support was removed; v127_rollout is their
 # successor against the current v1.2.7 release.) Pick the scenario with
 # the `test` input:
 #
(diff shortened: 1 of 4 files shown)


omersadika:
## Fail-closed verification: confirmed

[Run 30715468146](https://github.com/dwallet-labs/ika/actions/runs/30715468146) — `-f test=v127_rollout -f use_mocks=true` — **failed, for the right reason.** The exit code alone would not have been evidence, so the log line is the assertion:

```
v127_rollout is a rollout gate and must use real cryptography
```

Two details that matter:

- It aborted in **Build current binaries**, i.e. *before* `--features dwallet-mpc-unsafe-mock` was applied — the guard fails closed ahead of producing a mocked binary at all, not after.
- `setup: success`, `v127_rollout: failure`. Targeted, not collateral breakage.

Pre-fix, this exact dispatch would have built mocks and run to a green compatibility gate.

## Clean control

The specificity control comes free: this PR touches `.github/workflows/upgrade-test.yaml`, which is in the workflow's own `paths` filter, so it auto-dispatched `v127_rollout` at the default `use_mocks=false` ([run 30715486263](https://github.com/dwallet-labs/ika/actions/runs/30715486263)). That confirms two things the rename could plausibly have broken:

1. The guard does **not** reject legitimate runs.
2. The scenario still actually executes rather than silently skipping. This is the failure mode worth naming: the harness gates on `RUN_V127_ROLLOUT` and returns early with a "skipping" message when it is unset, so if the workflow's `RUN_FLAG` and the test's env-var name had drifted apart in the rename, the job would have gone **green while running nothing**. Both were renamed together; the control run is what proves it.

Result posted when it lands.

omersadika:
## Clean control: green, and verified non-vacuous

[Run 30715486263](https://github.com/dwallet-labs/ika/actions/runs/30715486263) — auto-dispatched by this PR's own `paths` filter at the default `use_mocks=false`. **All checks on this PR are now green.**

The green tick is not the evidence I care about here. The renamed gate could have passed while running *nothing*, since the harness returns early when `RUN_V127_ROLLOUT` is unset — so if the workflow's `RUN_FLAG` and the test's env-var name had drifted apart in the rename, the job would go green having skipped. Checked directly in the log:

| evidence | result |
|---|---|
| `skipping: set RUN_V127_ROLLOUT` lines | **0** |
| cargo output in *Run v127_rollout* | `test result: ok. 1 passed; 0 failed` |
| scenario body | reached `[flow 62/62]` |

So the scenario really executed end to end under its new name, with the real-crypto guard active and the production presign-pool sizing restored.

Worth noting for the record: the workflow already carries its own vacuity guard at line 592 — it errors with `exited 0 but did not report 'test result: ok. 1 passed' (gate misfire / skipped)` if the job succeeds without the test having run. That protection was in place and is exactly the right shape; it just could not help with the two defects this PR fixes, because those were about *which configuration* the gate ran in, not *whether* it ran.

## Remaining

Checks: ✅ all green. Branch is up to date with `main` (0 behind).

Blocked only on review — `mergeStateStatus: BLOCKED`, `reviewDecision: REVIEW_REQUIRED`, **0 of the 2 required approvals**.

PR #1969, opened 2026-08-02
Remove deprecated Sui JSON-RPC and notifier gas modes

omersadika:
## Description

Remove the two deprecated node compatibility paths:

- make notifier gas payment unconditionally use the Sui address balance and remove `notifier-gas-from-address-balance` plus the gas-coin state machine
- remove `sui-rpc-url`, the legacy Sui JSON-RPC backend, JSON-RPC event polling, and direct `sui-json-rpc-api` / `sui-json-rpc-types` dependencies from every Rust crate and example
- require `sui-data-source` and use the current Sui gRPC/protobuf client throughout
- replace legacy transaction response types with `sui_rpc_api::client::ExecutedTransaction` or native `sui_types` values, following the pinned Sui implementation
- update Rust tests, integration scripts, operator documentation, and behavioral specs; TypeScript and JavaScript files are unchanged

`sui-json-rpc-*` packages can still appear transitively in `Cargo.lock` through pinned upstream Sui crates, but Ika no longer directly depends on or calls them.

## Test plan

- `cargo check --release --workspace --all-targets`
- `cargo clippy --release --workspace --all-targets --all-features`
- `cargo clippy --release -p ika-sui-client -p ika-swarm-config --all-targets --all-features`
- `cargo test --release -p ika-config`
- `cargo test --release -p ika-sui-client`
- `cargo test --release -p ika-core sui_connector::tests::`
- `git diff --check`
- shell syntax checks for `crates/ika/tests/integration/helpers.sh` and `run_all.sh`

---

## Release notes

- [ ] Protocol:
- [x] Nodes (Validators and Full nodes): Node configuration now requires `sui-data-source`; `sui-rpc-url` and `notifier-gas-from-address-balance` have been removed. Notifier funds must be available in the Sui address balance.
- [ ] Indexer:
- [ ] JSON-RPC:
- [ ] GraphQL:
- [x] CLI: Sui-facing commands now use the current gRPC client and native transaction responses.
- [x] Rust SDK: Ika's Sui client no longer exposes the legacy Sui JSON-RPC backend or response types.
- [ ] REST API:


omersadika:
# Review — code is sound; the review is really about the rollout

**Verdict: mergeable on the code, but do not ship it in the release after v1.2.9 without a config-migration campaign first.** Evidence below.

## Code verification (against today's main, not the PR's 8-commit-old base)

- **Merged-tree compile**: merged this branch with current main (`#1972` authenticated gRPC, `#1973` MIN=7, `#1983`–`#1988`) — zero textual conflicts, and `cargo check --workspace --all-targets` on the merged tree is clean. The scary class for a −3.6k-line removal (a deleted field with a caller added by a later merge) does not occur.
- **Tests on the merged tree**: `ika-config` (26) and `ika-sui-client` (19) suites green.
- The event-ingestion consequence — JSON-RPC `query_events` was the only legacy event source, now removed — is fine in practice: the anchor-verified `BagEventPump` path this forces is what ~85 validators across both networks already run in production since the state-path migration.
- Nit: of the four fix-up commits, only "fix gRPC Move number decoding" carries a test; the two coin-lookup fixes and the Move-JSON one are untested CLI/test-harness tweaks. Acceptable for CLI paths, but the coin-lookup logic is the kind that regresses silently.

## The breaking-change blast radius, measured

Post-PR, `sui_data_source` is a required field (`SuiConnectorConfig` stays permissive otherwise, so stale `sui-rpc-url` / `notifier-gas-from-address-balance` keys are silently ignored — good). A config **without** the section fails at boot.

Measured against the live fleet (a node on the new-style config exports `ika_ocs_role_info`; absence = legacy config):

- **10 validators would fail to boot on upgrade**: testnet — 3MI Labs, DTEAM; mainnet — 3MI Labs, Allnodes, DTEAM, Perfect-nodes, Ruby Nodes, Staking4All, Tecnodes, remrerli.
- **Our own infra templates** (`config-defaults.ts`) still emit `sui-rpc-url` unconditionally and `sui-data-source` only behind a ternary — the infra side needs migrating before any build carrying this deploys.

The saving property, worth putting at the top of the operator comms: **the new-style config already works on v1.2.8 binaries.** Operators can migrate today, ahead of the release — this converts a flag-day break into a rolling migration, and `ika_ocs_role_info` coverage is a live migration tracker (currently 40/42 testnet, 47/55 mainnet). Suggested sequencing: announce now, ship in the minor after v1.2.9, and gate the tag on the tracker reaching full coverage (or on explicitly accepting the stragglers' downtime).

## One small change I'd make before merge

The deleted `select_sui_transport` legacy arms carried operator-grade error strings ("set `sui-data-source` (gRPC; the supported path)…"). What replaces them is serde's bare `missing field \`sui-data-source\``. For the exact 10 operators who will hit this at boot, wrap the config-load error with the migration hint — one error-mapping site, and it turns the failure mode from a support ticket into a self-service fix.

Also: rebase onto current main before merging so CI runs against what it will actually land on (the branch is 8 commits behind; my merged-tree check covers compile + two test suites, not the full matrix).

PR #1971, opened 2026-08-03
feat(metrics): make the AuthorityName wire width and its handoff retry observable

omersadika:
Protocol v7 changes how `AuthorityName` is serialized — 48-byte zero-padded to raw 32 bytes — and ships with **no instrumentation of its own**.

Verifying the testnet activation this morning, the only available evidence that the encoding worked was that the epoch advanced. That is sound reasoning: the handoff cert is signed at the end of epoch N under the old width and verified in N+1 under the new one, and the epoch cannot close without a cert quorum, so crossing the boundary proves the dual-width retry did its job. But it is *inference*, and it cannot answer the two questions that matter next:

- **Which width is any individual validator actually emitting?**
- **Did the retry fire, or was it simply never needed?**

Neither is answerable after the fact, and both matter more on mainnet: 56 validators, real value, and a partial-width fleet that presents as unexplained signature rejection with nothing in it naming an encoding. This is the failure mode #1959's own review notes flagged as the main risk.

## `ika_authority_name_encoding_width_bytes` (gauge: 48 or 32)

Published from `set_authority_name_short_encoding` **itself** rather than its caller, so every path is covered — including the test-only `FAULT_INVERT_AUTHORITY_NAME_WIDTH` inversion. That last part is the point: a gate that injects a width straggler should be able to *see* the straggler it injected.

Two distinct values across hosts is now a visible split rather than something to deduce. Also logged on transition, which is what dates a flip when reconstructing a boundary afterwards.

## `ika_handoff_attestation_width_retry_total{outcome}`

- **`recovered`** — the other width verified. Expected **exactly once per activation boundary**. Anywhere else it means two validators disagree about the width mid-epoch.
- **`exhausted`** — neither width verified, i.e. a bad certificate rather than an encoding problem. Kept on its own label so the retry is not blamed for it.

Counting the recovery is the load-bearing part: a successful retry is otherwise indistinguishable from a verification that never needed one, so "the boundary worked" cannot tell you which happened.

Both follow the established `OnceCell` pattern from `ika_invariant_violations_total` and register alongside it at boot.

## Verification

Two tests: the gauge must report **bytes** and follow the setter (not the protocol version — a validator on the wrong width is precisely where those disagree), and the two retry outcomes must land on separate series.

**Fault-validated** per `dev-docs/playbooks/test-testing.md`:

| injection | result |
|---|---|
| setter no longer publishes the gauge | `authority_name_width_gauge_reports_the_wire_width` FAILS at `pre-v7 emits the zero-padded 48-byte form` |
| retry outcomes collapse onto one label | `handoff_width_retry_counter_separates_recovery_from_a_bad_cert` FAILS (left `3.0`, right `2.0`) |

Both reverted; no artifacts remain. `ika-types` metrics 3/3, `ika-core` handoff 4/4, `cargo clippy` clean across `ika-types`/`ika-core`/`ika-node`, `check-metric-names.sh` passes with both names added to the inventory.

One implementation note worth flagging for review: the metrics live behind a process-global `OnceCell`, so tests in the same binary must share the registry it was initialised with — a second `init_*` with a fresh registry is silently a no-op. The test helper makes that explicit rather than leaving it as a trap.

## Convention

`dev-docs/conventions/metrics.md` gains the general rule this is an instance of: **a setting the whole committee must agree on, held in process memory, needs a fleet-scrapeable gauge rather than only a log** — and it should be published from the setter, so no path escapes it.

## Not in this PR

No alerting. The infra side (dashboard panel + an alert on two distinct widths across the fleet) is a separate change in `dwallet-labs/infra`, and I would land it before the mainnet v7 activation rather than after.

### DIFF
--- a/crates/ika-core/src/handoff_cert.rs
+++ b/crates/ika-core/src/handoff_cert.rs
@@ -520,10 +520,35 @@ pub fn verify_certified_handoff_attestation(
     match verify_certified_handoff_attestation_at_current_width(cert, committee, provider) {
         Err(first @ IkaError::InvalidSignature { .. }) => {
             let other_width = !ika_types::crypto::authority_name_short_encoding();
-            ika_types::crypto::with_authority_name_short_encoding(other_width, || {
-                verify_certified_handoff_attestation_at_current_width(cert, committee, provider)
-            })
-            .map_err(|_| {
+            let retried =
+                ika_types::crypto::with_authority_name_short_encoding(other_width, || {
+                    verify_certified_handoff_attestation_at_current_width(cert, committee, provider)
+                });
+            // Counted, because this path is otherwise invisible: a successful
+            // retry looks exactly like a verification that never needed one,
+            // so "the boundary worked" cannot distinguish the two. Exactly one
+            // `recovered` is the expected shape at an activation boundary —
+            // the cert is signed at the end of epoch N under the old width and
+            // verified in N+1 under the new one. Anywhere else means two
+            // validators disagree about the width mid-epoch. `exhausted` is a
+            // bad cert rather than an encoding problem, kept on its own label
+            // so the retry is not blamed for it.
+            match &retried {
+                Ok(()) => {
+                    ika_types::metrics::record_handoff_attestation_width_retry(
+                        ika_types::metrics::WIDTH_RETRY_RECOVERED,
+                    );
+                    tracing::info!(
+                        epoch = cert.attestation.epoch,
+                        retried_at_width_bytes = if other_width { 32 } else { 48 },
+                        "handoff attestation verified at the other AuthorityName width"
+                    );
+                }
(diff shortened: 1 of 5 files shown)


omersadika:
Closing without merge: this PR existed to make the v6→v7 AuthorityName width flip observable, and both networks completed that flip (testnet 2026-08-03, mainnet 2026-08-04) before it merged — verified instead by inference (the flip boundary is load-bearing, so crossing it proved the dual-width retry worked). Its commit rode #1973's branch, which then removed the width machinery *and* these metrics at MIN=7; the squash-merge of #1973 nets the two out, so merging this separately would only add code that the very next commit deletes. What survives is deliberate: the convention the metrics encoded — a committee-wide wire setting held in process memory needs a fleet-scrapeable gauge for the duration of its transition, published from the setter — is recorded in dev-docs/conventions/metrics.md for the next wire change that rides a version boundary.

Issue #1961, opened 2026-08-01
v6+ consensus-key rotation halts the network at the epoch boundary: next_committee_pubkey_set_hash is name-derived, so a rotation straddles the producer/consumer boundary


omersadika:
## Live exposure on both networks today. Agreed in the #1911 review thread, never filed until this issue, still unfixed.

`hash_next_committee_pubkey_set` (`crates/ika-core/src/handoff_cert.rs:82`) hashes a `Vec<AuthorityName>`. Since protocol v6, `AuthorityName` **is** the validator's Ed25519 consensus key — and consensus-key rotation is a live, cap-gated operator action (`system::set_next_epoch_consensus_pubkey_bytes`, `system.move:477`) that effectuates **at the epoch boundary** (`validator_info.move:295`, `rotate_next_epoch_info`).

## The straddle (mechanism, verified against the code)

The critical detail is that **consensus keys are not in the on-chain committee at all**. `BlsCommitteeMember` carries `{ validator_id, protocol_pubkey }` (`contracts/ika_common/sources/bls_committee.move:26`), and the next-epoch committee is built from `validator.validator_info().protocol_pubkey()` (`validator_set.move:838`). So to build the name set, every validator separately fetches each member's validator record and reads a **mutable** field — at whatever wall-clock moment its own sync loop happens to run.

Concretely, for a validator rotating `K_old` → `K_new`:

1. **During epoch E** the rotation is submitted. On chain `next_epoch_consensus_pubkey_bytes = K_new`; the current `consensus_pubkey_bytes` stays `K_old`.
2. **Mid-epoch E**, everyone assembles E+1's committee, reads the *current* field → `K_old`, hashes a set containing `K_old`, and signs the handoff attestation over it.
3. **At the boundary**, `rotate_next_epoch_info` applies the pending value → the record now holds `K_new`.
4. **In E+1**, validators verify the cert, read → `K_new`, recompute a set containing `K_new`. **It does not match.**

Result: every entering node rejects the cert → **fail-closed halt, network-wide**. Nobody misbehaved; the two sides read the same field at two different times and it changed in between. This is the halt #1911's fault injection reproduced on all four test validators.

The old `also_accept_next_committee_pubkeys` tolerance never covered this (it accepted the same membership under the *other identity basis*, BLS vs consensus, whereas a rotation is consensus-vs-consensus), and #1959 has since removed it as dead scaffolding — correct on its own terms, and it leaves this gap unchanged.

[...]

PR #1973, opened 2026-08-03
chore!: drop protocol v6 support (MIN=7) and remove the 48-byte AuthorityName encoding

omersadika:
> ## ⚠️ DO NOT MERGE until protocol v7 is active on **mainnet**
>
> Testnet flipped to v7 on 2026-08-03. **Mainnet is still on v6.** Merging before mainnet activates produces a binary that cannot run there at all: `AuthorityName` is a field of every signed consensus payload, so a 32-byte-only emitter computes different digests than the entire v6 committee and is rejected by all 54 validators — while parsing their messages perfectly.
>
> Stacked on #1971, which should merge first (it is what makes the mainnet flip observable).

## What this removes

`MIN_PROTOCOL_VERSION` 6 → 7, and `AuthorityName` becomes the raw 32-byte consensus key unconditionally. The machinery that carried the second width is deleted rather than left as dead scaffolding for a future reader to reason about:

- the `AUTHORITY_NAME_SHORT_ENCODING` process global, the thread-local override, and `set`/`get`/`with_authority_name_short_encoding`
- the **dual-width retry** in `verify_certified_handoff_attestation` — it existed solely for the one payload that straddled the flip
- the per-epoch width mirror in the epoch store, and its `FAULT_INVERT_AUTHORITY_NAME_WIDTH` injection point
- `short_authority_names` as a readable flag

The **field** stays, per the convention #1959 established: the flag set is part of the BCS-serialized `ProtocolConfig` whose digest validators compare, so removing it would change every supported version's digest.

## What this deliberately keeps

**Decoding stays lenient.** Rows and archives written before the v7 boundary hold the 48-byte form, and a node must keep reading its own history — dropping that tolerance would turn every pre-v7 persisted name into a deserialization failure on restart. `the_pre_v7_padded_form_still_decodes` pins it, and also asserts that decoding the padded form is *not* the same as emitting it.

This is the one asymmetry worth reviewing carefully: emit narrow, accept both.

## Metrics from #1971 are removed with the scaffolding

The width gauge could now only ever report one value, and the retry counter measures a retry that no longer exists. That is their intended lifecycle — they earn their keep during the mainnet flip window and retire with it. The *convention* they encoded is kept in `metrics.md` for the next wire change that needs a version boundary.

## Test infrastructure

- **`v127_v7_upgrade` and `protocol_version_transition` are retired.** Both cross a v6→v7 boundary that no supported version has. Unlike the v5→v6 flavor these are dropped rather than retargeted, because MIN = MAX = 7 leaves nothing to transition between. `ci-suites.md` now records that **whichever change introduces v8 must resurrect a transition gate, and that this is a release blocker for v8, not a follow-up** — the version boundary is the one thing a pure binary-swap gate cannot cover.
- **`v127_*` → `v128_*`** with `old_ref=release/mainnet-v1.2.8`. This is exactly the retarget the #1966 rule mandates once the deployed release moves, and it is now forced: v1.2.7 is MAX = 6 and cannot serve as the old binary against a v7-only current.
- The v7 width tests in `validator_metadata` go with the second width they exercised, leaving a pointer to what survives them.

## Verification

Workspace compiles, clippy clean across all targets, `ika-types` 25/25 including both surviving encoding tests, `check-metric-names.sh` passes.

**Fault-validated** per `dev-docs/playbooks/test-testing.md`: making the decoder reject the 48-byte form fails `the_pre_v7_padded_form_still_decodes` at its assertion. Reverted; no artifacts remain.

**Not yet run, and required before merge:** the heavy suites and the retargeted upgrade matrix. They cannot be trusted until they are re-run against a v1.2.8 old binary, which is itself part of what this PR changes — so I would dispatch them once mainnet v7 is live and this is rebased, rather than now.

## Sequencing

1. Mainnet activates v7 (pending the v1.2.8 rollout)
2. #1971 merges
3. Rebase this, dispatch `test=all`, merge on green

Worth noting this gets simpler if #1961 lands first: hashing validator IDs makes the handoff hash width-insensitive, so some of what is removed here would already be gone.


omersadika:
Correction to the verification claim in the commit message and PR body: **the `ika-core` library test suite was not actually run locally.** Three attempts were still in the compile phase when I tore down the worktree, so they produced no result — I should not have left "verification" reading as though the whole picture was covered.

What genuinely was verified locally: `cargo check --workspace --all-targets` clean, `cargo clippy` clean, `ika-types` 25/25 (including both surviving encoding tests), `check-metric-names.sh` passing, and the decode-tolerance test fault-validated — rejecting the 48-byte form fails `the_pre_v7_padded_form_still_decodes` at its assertion, then reverted.

`ika-core` is covered by this PR's own `Cargo Test Check` and `Clippy (msim / simtest)`, which are running now, so the gap closes on its own — but the claim should be accurate in the record rather than corrected only by CI going green.

Note also that `v128_rollout` is currently running against this branch, and its result will be **meaningless until mainnet is on v7**: the scenario now boots `release/mainnet-v1.2.8` as the old binary against a MIN=7 current, which is the configuration this PR creates. Whatever it reports today, the matrix has to be re-run after the rebase.

Issue #1978, opened 2026-08-04
node_cr: MPC round processing froze permanently at the moment of malicious conviction (consensus healthy, no restart) — live capture


omersadika:
## Live now on testnet — captured while still in the state, not reconstructed afterwards

`node_cr` (testnet, v1.2.8-fff400c29eb2, uptime 57.8h with **no restart**) stopped processing consensus rounds for MPC at the exact moment it was convicted malicious, and has not resumed in 3.5+ hours.

## The trace

`ika_last_process_mpc_consensus_round`, node_cr vs fleet max, 20-minute steps (epoch 373 began ~12:51, hence the counter reset):

```
15:31  node_cr=193,029  fleet=193,399  behind=370     Δ=+23,050
15:51  node_cr=216,069  fleet=216,453  behind=384     Δ=+23,040
       << conviction at 15:57:16 UTC >>
16:11  node_cr=222,986  fleet=239,362  behind=16,376  Δ=+6,917
16:31  node_cr=222,986  fleet=262,275  behind=39,289  Δ=0
...
19:31  node_cr=222,986  fleet=470,020  behind=247,034 Δ=0
```

Perfect tracking (~380 rounds behind, i.e. normal) until 15:51. Partial advance through the conviction window. Then **frozen at exactly 222,986**, permanently, while the fleet advanced another 247,000 rounds.

## What it is NOT

- **Not #1864.** Consensus is entirely healthy: `consensus_subscribed_to=1`, `consensus_subscribed_by=1`, commit staleness **26s** (Rhei 19s, StakingCabin 36s), and it commits ~34,600 subdags/30m — indistinguishable from healthy peers. #1864 is zero subscriptions and growing commit staleness. I initially suspected #1864 and was wrong.
- **Not a crash or restart.** 57.8h continuous uptime.
- **Not simple exclusion.** A validator whose contributions peers ignore would still *consume* consensus rounds. This one stopped consuming.
- **Not a stale metric.** It still exports: 15 active sessions, 16,605 completed, presign pool 1,039. The process is alive and the MPC subsystem is loaded — it just does not advance.

## What it is

Consensus-alive, MPC-dead — the shape of the v1.2.6 spectator wedge (#1952), but on **v1.2.8, which carries that fix**, and triggered by conviction rather than by a mid-epoch restart. Corroborating: `ready_to_advance_result_total` = 0 over 30m and `advance_completions` = 0 over the whole window.

[...]

PR #1985, opened 2026-08-06
fix(noa): bound chain-submitter calls with a timeout

omersadika:
## Summary

From the [#1978](https://github.com/dwallet-labs/ika/issues/1978) stall audit. The NOA chain-submitter calls (`submit_tx` / `check_tx_status`) run **inline in the MPC service loop** — `poll_chain_status` and the sign-output path are steps of `run_service_loop_iteration` — so a call that never returns freezes all MPC processing on the validator: the consensus-alive/MPC-dead silent-stall shape.

Today this is latent: the only submitter is `LogOnlyChainSubmitter`, which cannot hang. This PR is armor for the real chain submitter that replaces it when v8 enables NOA checkpoints, where a black-holed connection would otherwise become a service-wide wedge that no reviewer of the future submitter would think to prevent (the trait signature gives no hint its callers can't tolerate an unbounded await).

All four call sites now go through `call_with_timeout` (15s ceiling). A timeout surfaces as an ordinary `anyhow` error, which every call site already handles by warning and keeping the item for retry on a later tick — so the failure mode degrades from "validator MPC wedged" to "one NOA item retried".

## Testing
- New unit tests: a never-resolving call becomes a timeout error (under `start_paused` so the 15s ceiling elapses instantly); fast successes and callee errors pass through unrewrapped.
- Fault-validated per `dev-docs/playbooks/test-testing.md`: bypassing the timeout inside `call_with_timeout` makes the hung-call test hang (detected; killed by the harness bound), restoring it goes green.
- `cargo test -p ika-core --lib chain_submitter` green.

### DIFF
--- a/crates/ika-core/src/noa_checkpoints/chain_submitter.rs
+++ b/crates/ika-core/src/noa_checkpoints/chain_submitter.rs
@@ -18,6 +18,32 @@ pub enum TxExecutionStatus {
     Failed(String),
 }
 
+/// Ceiling on any single chain-submitter RPC.
+///
+/// These calls run inline in the MPC service loop (`poll_chain_status` and
+/// the sign-output path are steps of `run_service_loop_iteration`), so a call
+/// that never returns freezes ALL MPC processing on this validator — the
+/// consensus-alive/MPC-dead silent-stall class from the #1978 audit. The
+/// placeholder `LogOnlyChainSubmitter` cannot hang; this bound is armor for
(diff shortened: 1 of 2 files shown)


omersadika:
Follow-up on the local-test caveat in the PR body: verified against the merge parent (pre-this-PR main) — the 5 local `noa_checkpoint` test failures reproduce identically **without** this change, on macOS only, while Linux CI passes the full suite on both commits. This PR is exonerated; the platform flakiness is now tracked in #1989.

Issue #1989, opened 2026-08-06
NOA checkpoint integration tests fail (and run 20-40 min each) on macOS


omersadika:
Five tests in `dwallet_mpc::integration_tests::noa_checkpoint` fail on macOS (arm64, macOS 15) while passing in Linux CI:

- `test_noa_checkpoint_chain_failure_retry`
- `test_noa_checkpoint_multiple_sequential`
- `test_noa_checkpoint_dwallet_e2e`
- `test_noa_checkpoint_buffered_context`
- `test_noa_checkpoint_both_handlers`

Evidence it is platform-flakiness, not a recent regression:
- Reproduced identically on `main` at `23d7c27fb7` **and** at its merge parent (pre-#1985), so the chain-submitter timeout change is not implicated.
- `cargo test -p ika-core --lib` in the same environment: 27 of the 32 `noa`-filter tests pass; the 5 above fail after burning **20–40 minutes each** (a single test ran 2,298s solo), which looks like an internal wait that only resolves by timing out on this platform.
- CI (`Cargo Test Check`, Linux) passes the full suite on the same commits.

Impact: local development on macOS — a full `--lib` run takes 90+ minutes and exits red, so contributors can't use the suite as a pre-push gate. Diagnosis start: whatever these five wait on (channel delivery timing? localnet ports? tokio timing assumptions) behaves differently on macOS; instrumenting one with `RUST_LOG=ika_core::dwallet_mpc=debug` on a Mac should show where it parks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

omersadika:
Closing as invalid — the conclusion was wrong, and the error was mine in how the tests were invoked.

The five NOA tests (and ~50 other `dwallet_mpc::integration_tests`) are not macOS-broken. The ika workspace has no `[profile.dev]`/`[profile.test]` opt-level overrides, so a plain `cargo test -p ika-core` compiles the real class-groups crypto at **opt-level 0** — each 4-party step then takes minutes, and the harness's fixed ~60s advancement budget ("Party advancement did not complete after 600 iterations") fails everything that touches MPC, nondeterministically, on any hardware. All my failing runs — including the ones this issue was filed on — were debug-profile.

[...]

PR #1992, opened 2026-08-06
chore: upgrade Sui to mainnet-v1.76.1 and Rust to 1.97.1

omersadika:
## Summary

Dependency + toolchain refresh ahead of the next release.

- **Sui: `mainnet-v1.73.2` → `mainnet-v1.76.1`** (all 47 git-dep pins), with the coupled Mysten git pins aligned to what sui itself uses at that tag: fastcrypto `3273734`, anemo `68adc31`, sui-rust-sdk `d8643443` — the last one fixes a latent wrong pin (the manifest comment already claimed the 1.76.1 rev while pinning `5b41bc70`).
- **Rust: 1.94 → 1.97.1** — `rust-toolchain.toml` plus all 8 CI workflow pins, matching (slightly ahead of) sui 1.76.1's own 1.96.1.

## Lockfile strategy — why not a blanket `cargo update`

A full `cargo update` breaks the build twice over: it moves `allocative` past what `starlark_map` supports, and — worse — it splits the inkrypto **RustCrypto release-candidate family** (`ecdsa 0.17.0-rc.7`, `elliptic-curve 0.14.0-rc.14`, `sec1 0.8.0-rc.9`, `signature 3.0.0-rc.4`): some members have since graduated to real releases, and cargo mixes graduated + RC members into an uncompilable set. Those crates only move in matched sets and belong to an **inkrypto** upgrade, not this one.

Instead: re-resolve only the Sui dependency cone on top of the known-good lock. Shared deps still came forward to sui's own picks (tokio 1.52, serde 1.0.228, http 1.4, prost 0.14.3, hyper 1.11 — a new `sui-http` minimum), plus targeted bumps: `allocative` 0.3.4 (sui's pin), `ethnum` (pre-1.97 versions fail const-eval on the new rustc). Net lock churn: +586/−353 instead of ±3,700.

## Code adaptations (4)

1. **`sui-protocol-config-macros` now generates `ProtocolConfig` accessors** for every bool feature flag: deleted our 14 now-duplicate trivial getters/setters. One exception kept manual: `mysticeti_fastpath()` consults an env override before the flag, so the field carries `#[skip_protocol_config_accessor]` and the generated plain-forward getter is suppressed (the generated setter replaces our manual one).
2. **`TransactionClient::submit` gained a `Priority` lane** — `Normal` everywhere, preserving pre-1.76 single-lane behavior. `High` is a 128-slot reserved lane for low-volume critical submissions; ika's dominant traffic on this path is bulk MPC messages. Routing select kinds (EndOfPublish, checkpoint signatures) as `High` is a possible follow-up tune.
3. **`ConsensusAuthority::start` gained an optional randomness-signature handler** — `None`; ika runs no randomness protocol.
4. **`build_object_store` gained a headers argument** — empty, matching sui's own call sites.

## Testing

Local (macOS): full `cargo check --workspace --all-targets` clean on 1.97.1; suites green: `ika-protocol-config` 6/6 (the riskiest crate — macro change), `ika-config` 26/26, `ika-sui-client` 19/19, `ika-types` 28/28. The long `ika-core` suite is deferred to CI per its runtime (and the 5 pre-existing macOS-broken NOA tests, #1989). CI is the adjudicator here — simtest + the dispatched Upgrade Test matter more than usual since the consensus runtime, transport stack, and toolchain all changed underneath.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ env:
   RUSTDOCFLAGS: -Dwarnings
   RUST_BACKTRACE: 1
   # Change to specific Rust release to pin or `stable` for the latest stable version.
-  rust_stable: 1.94
+  rust_stable: 1.97.1
   CARGO_NET_GIT_FETCH_WITH_CLI: true
   GH_DEPLOY_KEY: ${{ secrets.GH_DEPLOY_KEY }}
   CARGO_INCREMENTAL: 0
--- a/.github/workflows/integration-tests-ci.yaml
+++ b/.github/workflows/integration-tests-ci.yaml
@@ -43,7 +43,7 @@ concurrency:
 env:
   RUST_BACKTRACE: 1
   # Change to specific Rust release to pin or `stable` for the latest stable version.
-  rust_stable: 1.94
+  rust_stable: 1.97.1
   rust_nightly: nightly
   CARGO_NET_GIT_FETCH_WITH_CLI: true
   GH_DEPLOY_KEY: ${{ secrets.GH_DEPLOY_KEY }}
--- a/.github/workflows/publish-typescript-sdk.yml
+++ b/.github/workflows/publish-typescript-sdk.yml
@@ -18,7 +18,7 @@ permissions:
 env:
   RUSTDOCFLAGS: -Dwarnings
   RUST_BACKTRACE: 1
-  rust_stable: 1.94
+  rust_stable: 1.97.1
   SUI_VERSION: 'sui@mainnet'
 
 jobs:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -230,7 +230,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: "1.94"
+          toolchain: "1.97.1"
           targets: ${{ matrix.target }}
 
       - name: Install dependencies (macOS)
--- a/.github/workflows/simtest.yaml
+++ b/.github/workflows/simtest.yaml
@@ -64,7 +64,7 @@ env:
   # long sub-quorum windows and slow recovery tails; virtual seconds cost
   # ~nothing in wall time.
   SUI_SIM_TEST_TIMEOUT_SECS: "3600"
-  rust_stable: "1.94"
+  rust_stable: "1.97.1"
 
 jobs:
   simtest:
--- a/.github/workflows/simtest.yaml
+++ b/.github/workflows/simtest.yaml
@@ -126,6 +126,11 @@ jobs:
         env:
           MSIM_TEST_SEED: ${{ inputs.seed }}
           MSIM_TEST_NUM: ${{ inputs.test_num }}
+          # The fault-matrix sim tests run real class-groups crypto whose
+          # synchronous steps can exceed msim's 5s no-progress watchdog;
+          # 60s matches scripts/simtest/codecov.sh. Without it every test
+          # aborts with "deadlock detected" while progressing normally.
+          MSIM_WATCHDOG_TIMEOUT_MS: "60000"
         run: |
           # --no-tests=pass is a nextest option, so it must come BEFORE the
           # `--` separator: the workspace currently has zero #[sim_test]
--- a/.github/workflows/test-cluster.yaml
+++ b/.github/workflows/test-cluster.yaml
@@ -58,7 +58,7 @@ env:
   RUSTUP_MAX_RETRIES: 10
   RUST_BACKTRACE: 1
   RUST_LOG: ${{ inputs.rust_log || 'error' }}
-  rust_stable: "1.94"
+  rust_stable: "1.97.1"
 
 jobs:
   test-cluster:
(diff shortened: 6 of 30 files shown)


omersadika:
## Upgrade Test: reproducible consensus-core panic on the post-swap catch-up (blocking)

Two independent runs, same signature, so this is deterministic rather than a flake:

- run [31178087213](https://github.com/dwallet-labs/ika/actions/runs/31178087213) — `v128_rollout` failed
- run [31183762790](https://github.com/dwallet-labs/ika/actions/runs/31183762790) — `v128_rollout` **and** `malicious_v128` failed

Both fail the same way, and the harness-level `Connection refused` on `/node-config` is a consequence, not the cause: the node is already dead.

### What the node logs show

`validator-0` boots the candidate binary at 14:05:22 (`Ika Node version: 1.2.8-341236aa4ba5`), replays certified commits, and **28 seconds into its first post-swap process** panics inside Sui's consensus-core:

```
add_certified_commits: panicked at consensus/core/src/linearizer.rs:144:22:
We should have all blocks in dag state.
  3: <consensus_core::linearizer::Linearizer>::handle_commit
  4: <consensus_core::commit_observer::CommitObserver>::handle_commit
```

The panicking call is the median-timestamp computation added in this Sui range:

```rust
// linearizer.rs, calculate_commit_timestamp
let block_refs = leader_block.ancestors().iter()
    .filter(|r| r.round == leader_block.round() - 1).cloned().collect();
let block_opts = dag_state.get_blocks(&block_refs);
let blocks = block_opts.iter().map(|b| b.as_ref()
    .expect("We should have all blocks in dag state."));   // <-- here
median_timestamp_by_stake(context, blocks)
```

It runs on the `add_certified_commits` path (commit-sync), which by its own doc comment does **not** accept blocks up front — acceptance happens inside `try_commit`, only for blocks belonging to the certified commits being committed. A leader's round−1 ancestor that sits on an uncommitted branch is therefore legitimately absent from dag state, and the `expect` turns that into a process kill. A catching-up node replaying ~100 leaders at once (exactly what the log shows: `Committing 100 leaders`) is the shape that hits it.

Ruled out on our side:
- ika's consensus flags and `gc_depth` (60, non-msim) match Sui's own production values exactly.
- The harness preserves `consensus_db`/`authorities_db` across the swap (`swap_binary` = stop, replace binary, start), so this is not a wiped-state artifact.
- Not caused by our `transaction_voting_enabled=true` fix: the first failing run predates it.
- No post-`mainnet-v1.76.1` upstream commit touches `linearizer.rs` (last change is the v3 FlexCommitter work from 2026-07-06, which is in our pin).

### Why it matters beyond the test

The upgrade test is doing its job: this is the rolling-upgrade catch-up path. The same shape — a validator restarting and syncing a backlog of certified commits — happens on every operator restart in production. With crash-on-panic (#1984) the failure mode is a crash-restart loop rather than a silent wedge, which is the right posture but still an outage for that validator.

### Options

1. **Report upstream and hold the bump** until Mysten fixes the `expect` (make the missing-ancestor case fall back rather than panic). Safest, unblocks nothing today.
2. **Pin consensus-core past a fix** if one lands on their release branch — none exists as of this writing.
3. **Ship 1.76.1 anyway** on the argument that Sui runs this code on mainnet at scale — I do not recommend this: our upgrade test reproduces it 2/2, and we would be shipping a known crash on the restart path.

Recommendation: option 1, and keep this PR open. Everything else on the branch is green — CI, Simtest, Test Cluster, TS Integration, TypeScript CI, Integration Tests CI, plus a full local battery (workspace tests 437/437, both clippy variants, sim fault matrix 11/11 on the final commit). The Sui bump is otherwise ready the moment the consensus panic is resolved.

omersadika:
## Correction: the Upgrade Test failure was self-inflicted, not a Sui regression

My previous comment blamed consensus-core. That was wrong, and the diff between the two tags proves it:

```
$ diff 1f6e1e6/consensus/config/src/consensus_protocol_config.rs \
       433212f/consensus/config/src/consensus_protocol_config.rs
(no output — byte-identical, 199 lines at both tags)
```

Position 7 of `ConsensusProtocolConfig::new` is `transaction_voting_enabled` at **both** mainnet-v1.73.2 and mainnet-v1.76.1. The rename I claimed happened in this bump range did not happen here — Sui renamed that parameter out from under ika's `config.mysticeti_fastpath()` call at some *earlier* tag.

The consequence is the interesting part: **ika has been running consensus with transaction voting OFF this whole time**, on 1.73.2, on both live networks, with a green upgrade test. `mysticeti_fastpath` is false at every supported protocol version, so what reads like a fastpath toggle has actually been ika's transaction-voting switch for as long as the networks have existed.

So hardcoding `true` did not restore parity with upstream — it **enabled transaction voting for the first time**. That changes the DAG consensus builds, and a node replaying a backlog of certified commits then hits the `linearizer::calculate_commit_timestamp` `expect` (the commit-sync path doesn't accept a leader's uncommitted round−1 ancestors). Hence 2/2 reproduction, and hence why main — which passes the flag — has always been green.

Fixed by restoring the flag-sourced value, with a comment explaining the trap so the next reader doesn't 'fix' it again. Enabling voting is a deliberate, version-gated protocol change (needs its own ika feature flag and a rollout plan), never a side effect of a dependency bump.

Upgrade Test re-dispatched on the corrected head: https://github.com/dwallet-labs/ika/actions/runs/31243525108

Also withdrawing the recommendation in my previous comment: there is no upstream issue to file, and nothing here blocks the Sui bump.

PR #2020, opened 2026-08-10
fix(ocs): recover pruned pending-gap checkpoints from the checkpoint archive

### LINKED ISSUE #2018: Pruning-loss residual of #1736: permanently dropped user request event wedges the epoch close (CI run 31353003990)
## Summary

In CI run [31353003990](https://github.com/dwallet-labs/ika/actions/runs/31353003990/job/93347193918) (TS Integration Tests, `workflow_dispatch` on `main` @ `4dca1a23`), a user DKG request event was **permanently lost by all four validators** because the fullnode pruned the defining checkpoint before any recovery path could fetch it — both shipped recovery mechanisms from the #1736 fix series (the checkpoint pusher's pending-gap repair and the verified-cache resolution of pruned dynamic-field children, PR #1809) ran and still lost the event. The lost session then sat inside the epoch's locked close target, making the epoch permanently unclosable: `all_epoch_sessions_finished` could never become true, and the run ended with the network wedged in epoch 5.

This is the residual tail of #1736 (closed as substantially resolved): the recovery paths narrow the pruning race but do not close it. When the race is lost, the failure is not a slow retry — it is an unrecoverable, epoch-wedging event loss.

Two of 21 cases in `all-combinations-future-sign.test.ts` failed with `Timeout waiting for DWallet … Last error: Object … does not exist`; the rest of the suite passed.

## Mechanism 1 — request event permanently dropped (session seq 214)

The Secp256r1/Shared test submitted its DKG request at ~04:43:45, landing on the epoch 4→5 boundary (epoch 4 closed 04:43:42–46). At 04:44:21–22 **each of the four validators** logged:

```
WARN ika_core::sui_connector::verified_reader: listed dynamic-field child with a pruned
defining checkpoint could not be resolved from the verified cache either
id=0xcba32026… error=Transport(NotFound("Transaction C3DU…"))
```

i.e. the fallback added for #1736 fired and failed: the fullnode had pruned the defining checkpoint, and the committee-verified cache could not serve the defining transaction either. No validator ever admitted the session — there is no `Adding a new MPC session` line for any user session between 04:43:30 and 04:54:21. The dwallet object never materialized and the SDK timed out after 600s.

Environmental driver: the runner was heavily throttled (cgroup sampler shows ~0.5 effective CPUs; `nr_throttled` non-zero), and the OCS checkpoint pusher was persistently losing the pruning race all run long:

- 743 × `checkpoint never materialized within the gap retry deadline — dropping seq=N` (561 in a single burst at 04:54)
- continuous `BagEventPump tick failed … NotFound` warns with recoveries (`after_failures=1`)

## Mechanism 2 — the lost session wedges the epoch close

- 04:58:46 — `Successfully locked last session in current epoch` (epoch 5). The frozen lock target (~234) **includes the dead session 214**, so on-chain `completed_sessions_count` can never reach it.
- From ~04:59 to the end of the run, all validators warn every 60s: `end-of-publish gate STUCK after the epoch locked to close` with `session_locked=true all_epoch_sessions_finished=false checkpoint_writer_lag=0` — the checkpoint writer is explicitly not the blocker; a locked session is.
- Epoch 5 never closed; the log ends 10 minutes later still wedged. Postmortem checks ruled out: malicious conviction (0 hits), epoch-entry key-gap warns (0), writer lag (0).

The consequence is severe out of proportion to the trigger: **one lost user request permanently wedges the epoch**, because the close predicate requires every session in the locked set to complete and there is no path (timeout, cancellation, or forced rejection) for a session whose request event no validator ever received.

## Open sub-question — session seq 234 admitted but never scheduled

A second, possibly separate anomaly in the same run: the Ristretto/Imported-Key test's session (seq 234) was admitted by all four validators at 04:58:50 (`active=true`, session id `0xf4a329e8fbbad0c8efe192d9874ed5a534596590c7c1c3db9099463fc099bd3f`) — 4 seconds after the epoch lock — and then nothing:

- It is the only session in the whole run admitted but never quorumed (501 of 502 admitted sessions completed).
- Zero `dwallet_mpc` log lines on any validator after 04:58:51.
- CPU sampler shows the machine near idle throughout — round 1 never spawned (a "not spawning" shape, not a hung computation).
- All four validators' metrics report `ika_dwallet_mpc_user_session_local_output_rejected{session_seq="234"} -1` (no local output ever observed).

Why an admitted, active session never scheduled a computation cannot be determined from this run: the load-bearing lines (`Starting cryptographic computation`, `Advancing session`) are debug-level and the run logged at info. Candidates: the same pruning loss hitting a secondary input fetch, or a scheduling gap for sessions admitted just after the epoch-close lock. Needs a debug-level (`ika_core::dwallet_mpc=debug`) reproduction or live chain counters to distinguish.

## Evidence

- Failing job: https://github.com/dwallet-labs/ika/actions/runs/31353003990/job/93347193918
- Artifacts on the run: `localnet-logs` (validator log with all lines quoted above), `cpu-sampler-ts-integration-1` (throttling/idle evidence)
- Diagnosis followed `dev-docs/playbooks/mpc-stall-postmortem.md` (ordered checks 0–6)

## Suggested directions

1. **Close the pruning race for good** rather than narrowing it further — e.g. have validators treat "defining checkpoint pruned AND verified-cache miss" as a hard signal to fetch the bag child via a committee-verified point read of the live object (it is still live and listable — only its *defining* checkpoint is gone), instead of dropping it.
2. **Bound the blast radius**: an epoch-close design that can eventually exclude a locked session that no validator ever admitted (with quorum agreement), so a single lost event cannot wedge the epoch permanently.
3. Instrument the session-234 shape: an info-level line when an active session sits >N rounds with no computation attempt would make the open sub-question diagnosable from CI artifacts.

Refs: #1736 (closed — this is its residual tail recurring), #1809 (the recovery paths that ran and were insufficient here).

omersadika:
Closes #2018.

## Problem

CI run 31353003990 wedged epoch 5 permanently: the Sui localnet fullnode (aggressive pruning, throttled runner) pruned a checkpoint before the validators' OCS checkpoint pusher could fetch it. The checkpoint carried a user DKG-request `session_events` bag entry; both existing recovery layers (pending-gap retry against the fullnode, verified-cache resolution of pruned dynamic-field children) ran and still failed, so all four validators permanently lost the request event. The lost session sat inside the epoch's locked close target, `all_epoch_sessions_finished` could never become true, and the epoch never closed — the residual tail of #1736.

## Fix: a third, verified recovery layer — the checkpoint archive

The pusher's pending-gap repair now falls back to the resolved **checkpoint archive** (the same object store the committee ratchet already uses as its verified fallback):

- A gap the fullnode keeps refusing past a 5s grace is fetched from the archive (paced at 5s per gap, capped at 4 fetches/tick, 5s per-fetch timeout — the object-store client otherwise retries internally up to 60s, and an unbounded stall would hold the scan cursor back while the fullnode prunes more).
- An archive-served checkpoint must pass verification **strictly beyond** what the fullnode fold path applies before it may clear a gap (`verify_archive_checkpoint`): the blob's committee-signed `sequence_number` must equal the requested seq (file naming is untrusted), committee BLS on the summary, and the artifacts-digest binding over the blob's **entire** object set — both unconditionally. Without this, a blob with no Ika-typed outputs folds vacuously (`build_entries` verifies nothing on that path) and a mislabeled or emptied blob would silently "repair" the gap while suppressing the dropped-gap alarm. The archive stays an availability source, never a trust source.
- Only a gap that BOTH sources fail to serve until the 600s deadline is dropped, still with the loud warn.
- The end-of-epoch retained store is now written only **after** the fold's committee verification passes (previously `persist_end_of_epoch` ran first, letting an unverified blob overwrite a genuine retained entry — availability-only exposure, since mirrored peers re-verify, but needless).
- The public Sui checkpoint stores (mainnet/testnet defaults) retain *ordinary* checkpoints ~30 days — far beyond any fullnode pruning window — so on the public chains this closes the pruning race outright.

The seq-binding/vacuous-fold hazard, the persist-before-verify ordering, the fetch-timeout stall amplification, and the fast-forward-sacrificed gaps not counting as drops were all found by an adversarial consensus review of the first draft and are fixed + regression-tested in this PR.

New plumbing: `CheckpointArchive::fetch_checkpoint_data` (full checkpoint with transactions + objects), the archive shared from `SuiConnectorStack` into `IkaCheckpointPusher`.

## Localnet/CI coverage for the same race

Localnets resolve no default archive, so the race stayed open exactly where it fired. Now:

- `sui start --data-ingestion-dir <dir>` makes the localnet fullnode dump every executed checkpoint as `{seq}.binpb.zst` — the archive layout.
- New `ika start --sui-checkpoint-archive-url file://<dir>` threads that into every validator's `sui_checkpoint_archive` (CLI → SwarmBuilder → ConfigBuilder → ValidatorConfigBuilder).
- The TS-integration workflow wires both, replacing the step comment that claimed the OCS path was "self-sufficient against this localnet's aggressive pruning" (disproved by the incident).

## Observability

- `ika_ocs_pusher_gap_archive_repairs_total` — gaps recovered from the archive (the fallback doing real work).
- `ika_ocs_pusher_gap_dropped_total` — gaps lost by both sources; each is a potential wedged epoch. Alert on any increase (previously drops were only a warn line).
- `ika_dwallet_mpc_user_sessions_active_without_local_output` + a once-a-minute warn ("active without a local output past the stall threshold") for the issue's open sub-question: a user session admitted by all validators that never computes (session seq 234 in the incident was invisible in logs from admission to test timeout). This makes the shape diagnosable from CI artifacts at info-level logging.

## Not addressed here (deliberately)

The epoch-close blast radius — a protocol-level path to exclude a locked session that no validator can ever complete — is direction 2 in #2018 and needs a consensus/Move design discussion; this PR closes the event-loss root cause and makes any recurrence loudly visible instead.

## Testing

- New unit tests: `pruned_gap_checkpoint_is_recovered_from_the_archive` (archive repair: fold verified, committee captured, end-of-epoch retained, metric counted; archive NOT consulted within the fullnode grace), `archive_blob_for_a_different_seq_is_refused` (a genuine, correctly-signed blob for the WRONG seq must not clear the gap nor install its committee transition), and `gap_missing_from_fullnode_and_archive_is_dropped_at_the_deadline` (both-sources failure: expired gap dropped + counted, younger gap kept).
- Full `sui_connector` module run in release (includes the ratchet `MockArchive` trait-extension change).
- Docs updated in the same PR: `ocs-verified-sui-reads.md` (pending-gap repair section), `mpc-stall-postmortem.md` (new grep signals + #2018 shape), `localnet.md` (archive-wired startup).

### DIFF
--- a/.github/workflows/ts-integration-tests.yaml
+++ b/.github/workflows/ts-integration-tests.yaml
@@ -165,13 +165,18 @@ jobs:
 
       - name: Start Sui localnet
         run: |
-          # Native checkpoint cadence + epoch length. The OCS path is
-          # self-sufficient against this localnet's aggressive pruning
-          # (num_epochs_to_retain=0): committees come from the live
-          # SubscribeCheckpoints follower, and object reads fall back to the
-          # committee-verified local cache when an anchor's defining tx is
-          # pruned upstream — so no Sui-side slowdown is needed.
-          RUST_LOG=error sui start --with-faucet --force-regenesis > sui-localnet.log 2>&1 &
+          # Native checkpoint cadence + epoch length. This localnet prunes
+          # aggressively (num_epochs_to_retain=0), and under a throttled
+          # runner the pruning watermark can outrun the validators' checkpoint
+          # pusher — losing a checkpoint (and any session-request event on it)
+          # permanently, which wedges the epoch close (issue #2018). The
+          # data-ingestion dir keeps every executed checkpoint as a file; the
+          # ika validators are pointed at it (--sui-checkpoint-archive-url
+          # below) as their verified fallback for pruned-before-fetched
+          # checkpoints. Committees still come from the live
+          # SubscribeCheckpoints follower.
+          RUST_LOG=error sui start --with-faucet --force-regenesis \
+            --data-ingestion-dir "$PWD/sui-ingestion" > sui-localnet.log 2>&1 &
           echo $! > sui-localnet.pid
           for i in $(seq 1 60); do
             if ! kill -0 "$(cat sui-localnet.pid)" 2>/dev/null; then
@@ -197,7 +202,8 @@ jobs:
           rm -rf ~/.ika Pub.localnet.toml
           RUST_LOG="${LOCALNET_RUST_LOG:-warn,ika=info,ika_node=info}" RUST_MIN_STACK=67108864 \
             ./target/release/ika start --force-reinitiation \
-            --epoch-duration-ms "$EPOCH_DURATION_MS" > ika-localnet.log 2>&1 &
+            --epoch-duration-ms "$EPOCH_DURATION_MS" \
+            --sui-checkpoint-archive-url "file://$PWD/sui-ingestion" > ika-localnet.log 2>&1 &
           echo $! > ika-localnet.pid
 
       - name: Wait for ika localnet readiness
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_metrics.rs
@@ -423,6 +423,14 @@ pub struct DWalletMPCMetrics {
     /// nobody (us included) has seen quorum. Per-tick gauge.
     pub(crate) sessions_with_self_output_no_quorum: IntGauge,
 
+    /// The complement: user sessions Active well past any normal computation
+    /// latency with NO local output submitted — admitted but possibly never
+    /// computing. One such session inside the epoch's locked close set wedges
+    /// the epoch (issue #2018), so any sustained non-zero value is a stall
+    /// signal. Per-tick gauge; the manager also warns once a minute while
+    /// non-zero.
+    pub(crate) user_sessions_active_without_local_output: IntGauge,
+
     /// Per-completion: number of consensus rounds elapsed between this validator submitting
     /// its own output and 2/3 quorum being reached. Wide tails ≡ slow consensus, lots of
     /// retries. Observed once per session at the moment quorum is reached.
@@ -868,6 +876,12 @@ impl DWalletMPCMetrics {
             sessions_with_self_output_no_quorum: register_int_gauge_with_registry!(
                 "ika_dwallet_mpc_sessions_with_self_output_no_quorum",
                 "User sessions where this validator submitted an output but no quorum has been observed.",
+                registry,
+            )
+            .unwrap(),
+            user_sessions_active_without_local_output: register_int_gauge_with_registry!(
+                "ika_dwallet_mpc_user_sessions_active_without_local_output",
+                "User sessions Active past the stall threshold with no local output submitted (admitted but possibly never computing); sustained non-zero can wedge the epoch close — alert on it.",
                 registry,
             )
             .unwrap(),
--- a/crates/ika-core/src/dwallet_mpc/mpc_manager.rs
+++ b/crates/ika-core/src/dwallet_mpc/mpc_manager.rs
@@ -671,6 +671,9 @@ pub(crate) struct DWalletMPCManager {
     /// rate-limited because the service loop ticks every ~20ms and the
     /// refresh iterates every tracked session.
     last_observability_refresh: Option<Instant>,
+    /// When the stalled-user-session warn was last emitted, so a wedged
+    /// session logs once a minute instead of once per refresh.
+    last_stalled_session_log: Option<Instant>,
 
     /// NOA finalization observation votes: tx_ref → set of party IDs that observed finalization.
     noa_finalization_observations: HashMap<NOACheckpointTxRef, HashSet<PartyID>>,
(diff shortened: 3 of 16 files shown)


omersadika:
Validation dispatch: TS integration suite on this branch (the workflow from the incident, now with the data-ingestion-dir archive wired): https://github.com/dwallet-labs/ika/actions/runs/31365951187 — watch for `recovered from the archive and folded` / `ika_ocs_pusher_gap_archive_repairs_total` in the artifacts as evidence the fallback engages under CI pruning.

omersadika:
Validation run https://github.com/dwallet-labs/ika/actions/runs/31365951187 is green: **9/9 suites passed**, including `all-combinations-future-sign` at 17m59s (the incident run failed it at 35m51s with two 600s timeouts). Notes on evidence quality: this run had **zero** pusher gap-drops (the incident run had 743 — its runner was far more throttled), so the archive fallback never needed to engage end-to-end; the engagement paths are pinned by the unit tests instead (`pruned_gap_checkpoint_is_recovered_from_the_archive`, `archive_blob_for_a_different_seq_is_refused`, `gap_missing_from_fullnode_and_archive_is_dropped_at_the_deadline`). The new wiring is confirmed live in the run: `sui start --data-ingestion-dir` and `ika start --sui-checkpoint-archive-url file://…` both executed and the localnet booted and ran the full suite on top.

PR #2028, opened 2026-08-11
fix(metrics): move the vendored consensus registry out of ika's namespace (#2022)

### LINKED ISSUE #2022: duplicate metrics: ika_consensus_last_committed_leader_round is published twice
There are two gauges with the same name `ika_consensus_last_committed_leader_round` published in metrics.

```
# curl -s http://127.0.0.1:9184/metrics|grep ika_consensus_last_committed_leader_round
# HELP ika_consensus_last_committed_leader_round The last round where a leader was committed to store and sent to commit consumer.
# TYPE ika_consensus_last_committed_leader_round gauge
ika_consensus_last_committed_leader_round 772138
# HELP ika_consensus_last_committed_leader_round Leader round of the latest consensus commit processed at the commit boundary (the freeze-grace round domain); -1 before the epoch's first commit
# TYPE ika_consensus_last_committed_leader_round gauge
ika_consensus_last_committed_leader_round 772138
```

One is the metric `last_committed_leader_round` defined in sui core which is re-exported with the prefix `ika_consensus_`. The other is `consensus_last_committed_leader_round` defined in ika core epoch metrics published with the prefix `ika_`.

prometheus logs duplicate sample of timestamp errors such as the following each time it scrapes ika:
```
Aug 10 22:52:29 HOSTNAME prometheus[1333094]: time=2026-08-10T22:52:29.415+02:00 level=DEBUG source=scrape.go:1998 msg="Duplicate sample for timestamp" component="scrape manager" scrape_pool=ika target=http://X.X.X.X:9184/metrics series=ika_consensus_last_committed_leader_round
Aug 10 22:52:29 HOSTNAME prometheus[1333094]: time=2026-08-10T22:52:29.425+02:00 level=WARN source=scrape_append_v2.go:402 msg="Error on ingesting samples with different value but same timestamp" component="scrape manager" scrape_pool=ika target=http://X.X.X.X:9184/metrics num_dropped=1
```

The `consensus_last_committed_leader_round` metric defined in ika core should either be removed if it is simply sampling the same data that `last_committed_leader_round` in sui core is sampling or otherwise it should be renamed.

omersadika:
## Problem

Fixes #2022, reported by @svenski123 from production Prometheus scrape logs.

The node published the metric family `ika_consensus_last_committed_leader_round` **twice** on `/metrics`:

- upstream `consensus-core`'s `last_committed_leader_round`, exported through the namespaced registry ika starts consensus with — `Registry::new_custom(Some("ika_consensus"))` in `consensus_manager/mod.rs`;
- ika's own `EpochMetrics` gauge, registered with that literal name into the node's default registry.

`RegistryService` merges those registries at `/metrics`, and prometheus's duplicate check is **per-registry**, so it never fired — the endpoint just served the same family name twice. Every scrape logged `Duplicate sample for timestamp` and dropped a sample, silently making one of the two series unreliable.

## The fix

Two parts: keep the gauge but move it, and make the rule that prevents a recurrence structural rather than incidental.

### 1. Rename, not remove — the gauges measure different things

The issue offered both options. The code shows they are not redundant:

| | consensus-core gauge | ika gauge |
|---|---|---|
| HELP | "The last round where a leader was committed to store and **sent to commit consumer**" | "Leader round of the latest consensus commit **processed at the commit boundary** (the freeze-grace round domain)" |
| Set in | `consensus/core/src/commit_observer.rs` | `AuthorityPerEpochStore::process_consensus_transactions`, from `consensus_commit_info.round` |
| Side | producer — handed *to* the consumer | consumer — the consumer actually *processed* it |
| Restart | consensus-core's own recovery | re-seeded from persisted consensus stats at epoch-store open |

They coincide on a healthy node and **diverge exactly when ika's commit handler stalls while Mysticeti keeps committing** — the failure class this gauge was added to expose during the v1.2.x observability campaign (cf. #1978/#1980, where a validator kept serving consensus while its MPC side was dead). Removing it would delete the consumer-side view and leave only the producer-side one, which by construction cannot show that divergence. It is also the round domain the mpc_data freeze / epoch-close grace countdowns are measured in: `dev-docs/specs/validator-mpc-data-announcements.md` documents the freeze ETA as `clamp_min(ready_quorum_round + grace_rounds - <this gauge>, 0)`, a subtraction that is only meaningful in the commit-boundary domain.

### 2. Out of the namespace entirely — `ika_last_committed_leader_consensus_round`

The first version of this PR renamed the gauge to `ika_consensus_last_committed_leader_consensus_round`. That was wrong, and the reason is the actual lesson of #2022.

`ika_consensus_*` is not a shared namespace ika may pick unused names from — it is the namespace of the registry ika **hands to consensus-core**, so upstream can add a metric of any name there at any time. A name inside it that upstream merely does not use *today* has exactly the property the broken name had until upstream grew into it. And validating ika's names against a snapshot of upstream's current names cannot help: such a snapshot is stale the moment upstream adds a metric, which is how #2022 happened in the first place.

So the invariant is structural — **ika registers nothing under `ika_consensus_*`** — and the gauge moves out: `ika_last_committed_leader_consensus_round`. HELP text unchanged. The upstream-name snapshot introduced in the first version is deleted, along with every code path reading it; there is now zero dependency on upstream metric names.

Enforced in two layers, both keyed on a list of the 13 pre-existing violators:

- `LEGACY_IKA_CONSENSUS_NAMESPACE_METRICS` in `crates/ika-core/src/epoch/epoch_metrics.rs` (runtime, over the composed registry)
- the same list in `scripts/check-metric-names.sh` (static, over every registered literal)

The crucial property: **these lists enumerate our own names** — stable, ours to control, and they can only shrink. There is nothing upstream to mirror and nothing to regenerate on a Sui bump. Both carry a comment saying so, because the natural instinct on seeing a checked-in list of metric names is to "helpfully" regenerate it. A new violating name fails the build; removing one requires touching nothing but the list.

The 13 legacy names are **left alone here** — moving them means downstream dashboard and alert churn mid-release. Tracked in #2029, which lists all 13 with their files and what each move involves. When the last one moves, both lists go empty and the rule is simply "no ika metric starts with `ika_consensus_`".

## Impact sweep

**Code — `crates/ika-core/`**
- `src/epoch/epoch_metrics.rs` — registered name → `ika_last_committed_leader_consensus_round`; struct field `consensus_last_committed_leader_round` → `last_committed_leader_consensus_round`; field doc now records why the name is deliberately outside `ika_consensus_*`.
- `src/epoch/epoch_metrics.rs` — the `dwallet_mpc_data_ready_quorum_round` doc comment's freeze-ETA formula referenced the old name.
- `src/authority/authority_per_epoch_store.rs` — both write sites (commit-path set, epoch-store-open re-seed) and 5 existing test assertions.

**Docs**
- `dev-docs/specs/validator-mpc-data-announcements.md` — freeze-progress metric list; now also states explicitly that the old name is consensus-core's producer-side gauge, so nobody wires the freeze delta back to it.
- `dev-docs/conventions/metrics.md` — inventory updated, and the convention section now carries the namespace rule and points at both enforcement layers.

**Checked, no references to update:** `crates/ika-upgrade-test/` (asserts on `ika_current_epoch`, `ika_dwallet_mpc_*`; never this name), `dev-docs/playbooks/production-alerts.md` (uses the *timestamp* metric), `dashboards/` in this repo (README + logo only, no Grafana JSON).

### Downstream (the separate infra repo — cannot be fixed here)

The infra repo queries the old name in committed Grafana JSON and a runbook. **These do not break loudly** — the name still resolves, because consensus-core keeps it. What changes is the *meaning*: it is now unambiguously the producer-side gauge. Panels that wanted the commit-boundary domain must migrate to `ika_last_committed_leader_consensus_round`:

**Must migrate** (freeze-grace domain — the subtraction is only correct against the commit-boundary gauge):
- `ika-k8s-cluter/grafana-dashboards/ika.json:311` — freeze-ETA stat expression (two occurrences: the grace delta and the `deriv(...)` rate)
- `ika-k8s-cluter/grafana-dashboards/ika.json:6604` — freeze-ETA panel (two occurrences, same formula)
- `ika-k8s-cluter/grafana-dashboards/ika.json:6516` — `max(ika_consensus_last_committed_leader_round{network="$network"})`, the commit-boundary progress readout paired with the freeze panels

**Can keep the old name** (genuine consensus/Mysticeti progress — producer side is the right series):
- `ika-k8s-cluter/grafana-dashboards/ika-sui-from-trustedpoint.json:113, 940, 2440, 2957`
- `ika-k8s-cluter/grafana-dashboards/ika-validators-mysticeti.json:5648`
- `apps/ika-observability-agent/runbooks/30-consensus-checkpoints.md:18` — worth a clarifying word that this is the producer-side gauge

**Continuity note for operators:** the renamed series starts fresh — anything scraping `ika_consensus_last_committed_leader_round` for ika's commit-boundary semantics loses history at the upgrade. That is the deliberate trade-off in renaming the *ika* side: the consensus-core-prefixed metric keeps the old name and remains the continuity series, and it is also the one upstream Mysticeti dashboards expect. The old series' pre-upgrade history was in any case a coin-flip between the two gauges on every scrape, so it was already not trustworthy.

## Testing

In `crates/ika-core/src/epoch/epoch_metrics.rs`:

1. **`ika_metrics_stay_out_of_the_consensus_core_namespace`** — composes the ika-side registry the way `IkaNode::start_async` does (all 10 ika-core metric structs), and fails on any `ika_consensus_*` family name not in the legacy list.
2. **`legacy_consensus_namespace_allowlist_is_sorted_and_unique`** — keeps the list sorted, duplicate-free and on-topic, so deleting a migrated entry stays a clean one-line diff.
3. **`ika_side_registry_has_no_duplicate_family_names`** — intra-registry duplicate guard, kept as a cheap independent check.
4. **`commit_boundary_gauge_is_outside_the_consensus_core_namespace`** — pinned #2022 regression.
5. **`scripts/check-metric-names.sh`** — the same namespace rule plus a stale-entry check, over every registered literal in `crates/**`.

**The two layers are complementary, not redundant, and the difference is load-bearing.** `Registry::gather()` returns only families that have samples, so a `*_vec` with no label children yet is invisible to it — 6 of the 13 legacy names are vecs and do not appear in the runtime test's view. The runtime test therefore covers what the node actually *serves* (the right domain: an unobserved family is never written to the endpoint and cannot collide), while the static script covers everything ika *registers*, vec or not. This is documented at the helper in `epoch_metrics.rs` so neither layer looks redundant to a future reader. I found this because my first draft included a runtime stale-entry check and it failed on exactly those 6 vecs; the stale check now lives in the script, which can see them.

### Fault validation

**(a) Gauge renamed back into the namespace** (`ika_consensus_last_committed_leader_consensus_round` — a name upstream does not currently use, i.e. what the first version of this PR shipped):
- `ika_metrics_stay_out_of_the_consensus_core_namespace` — **FAILED**: `these ika metrics are registered inside ika_consensus_*, ... (ika #2022): ["ika_consensus_last_committed_leader_consensus_round"]`
- `commit_boundary_gauge_is_outside_the_consensus_core_namespace` — **FAILED**
- `scripts/check-metric-names.sh` — **exit 1**, same name reported

**(b) Throwaway `ika_consensus_zzz_probe` added, not in the allowlist:**
- `ika_metrics_stay_out_of_the_consensus_core_namespace` — **FAILED**: `... ["ika_consensus_zzz_probe"]`
- `scripts/check-metric-names.sh` — **exit 1**, same name reported

Note (a) is the case a snapshot-based check would have *passed*.

Both reverted; everything green again.

### Results

- `cargo test --release -p ika-core` — the 10 tests covering this change pass: the 4 above plus the 6 `freeze_metrics_*` tests in `authority_per_epoch_store.rs` holding the 5 updated gauge assertions.
- `cargo clippy --release -p ika-core --all-targets` — clean, no warnings
- `cargo fmt --all --check` — clean
- `./scripts/check-metric-names.sh` — `metric names OK (310 literal, 3 dynamic sites skipped; 13 legacy ika_consensus_* names remaining to migrate)`

A full `-p ika-core` run showed 20 `dwallet_mpc::integration_tests::*` failures, which I chased down rather than assumed away: they were contention from my running other tests out of the same binary concurrently. Re-run in isolation on this branch, the representative case (`encrypt_secret_share`) passes in 185s. No failing test referenced the renamed metric.

Rebased onto current `main` (picks up #2027); no conflicts.

Unrelated pre-existing drift, deliberately left alone: the generated inventory block in `dev-docs/conventions/metrics.md` is missing 22 metrics added since it was last regenerated and lists one (`ika_ocs_unverified_committee_fallback_total`) that no longer exists. Regenerating it here would bury this fix in an unrelated diff.


omersadika:
Design note on where this PR sits, after discussion of the longer-term namespace question.

The permanent fix has changed direction: rather than migrating ika's remaining names *out* of `ika_consensus_*`, the plan is to move the **vendored** registry to `ika_vendor_consensus_` — see the rewritten #2029. That makes ika's own `ika_consensus_*` names correct rather than grandfathered (they sit naturally beside `ika_dwallet_*`, `ika_ocs_*`, `ika_validator_*`), empties the allowlist with zero renames on our side, and reduces the standing invariant to "every `new_custom` prefix starts with `ika_vendor_`" — enforceable at the one line where vendored registries are created, and self-extending to namespaces added later.

**This PR is unchanged and should merge as-is.** Two reasons it is not being adjusted to anticipate that plan:

1. `ika_commit_boundary_leader_round` is correct under both futures. If the vendor move lands, the name is valid (just not prefix-matched to its siblings); if it stalls, the name is the only thing keeping this gauge out of a namespace upstream can still grow into. Reverting it to `ika_consensus_commit_boundary_leader_round` only pays off in the first case and actively costs us in the second. There is no double-rename to avoid either — the current name never *needs* changing again.
2. Reverting the name while this PR simultaneously adds a rule forbidding new `ika_consensus_*` registrations would mean adding a brand-new entry to a *legacy* allowlist, which contradicts the rule the PR exists to establish.

The enforcement here stays useful until whichever path lands: it is what prevents a *new* violation appearing in the meantime, and it is cheap to delete afterward.

omersadika:
Renamed in `0343ae6b`: the gauge is now **`ika_last_committed_leader_consensus_round`** (was `ika_commit_boundary_leader_round`), with the Rust field matching.

It measures the consensus round of the last committed leader as observed by ika's commit handler, so the name now says that rather than describing where it happens to be read. It reads as a deliberate near-homonym of consensus-core's `last_committed_leader_round` — same subject, different observation point — which is the honest relationship between the two.

Safe under both outcomes of #2029: it sits outside `ika_consensus_*` whether or not the vendored registry moves to `consensus_ika_`, so it needs no follow-up to stay correct either way.

**Downstream (infra repo)** — the migration target changes accordingly. `ika-k8s-cluter/grafana-dashboards/ika.json` lines 311 (×2), 6516, 6604 (×2) migrate to `ika_last_committed_leader_consensus_round`; the other references listed earlier still legitimately track consensus-core's producer-side metric and keep the old name.

Verified after the rename: `epoch_metrics` 4/4 pass, `scripts/check-metric-names.sh` clean (310 literals, 13 legacy remaining), `cargo fmt --check` clean.

omersadika:
Reworked in `ea46e3af` to resolve #2029 here rather than defer it.

**The vendored registry moved instead of our names.** `consensus_manager` now starts consensus-core with `Registry::new_custom(Some("consensus_ika"))`. ika's metrics all start with `ika_` and the vendored ones now never do, so the namespace rule is a bipartition that needs no list on either side.

**The 13 legacy names stay and become correct.** `ika_consensus_handler_processed` is the right name for ika's consensus-handler metric, sitting beside `ika_dwallet_*` and `ika_ocs_*`. They were never bad names — they had a squatter in their namespace. `LEGACY_IKA_CONSENSUS_NAMESPACE_METRICS` is deleted from both the runtime module and the script; there is no longer a list to shrink.

**Enforcement is now two source-level checks**, both in `scripts/check-metric-names.sh`: every registered literal starts with `ika_`, and no `Registry::new_custom` prefix starts with `ika`. The second is self-extending — a vendored registry added later is caught with nobody updating anything, which was the failure mode of every list-based design we tried.

**One finding worth recording.** A runtime prefix check is deliberately *not* duplicated in `epoch_metrics.rs`. Adding one surfaced four families that `gather()` returns but ika's source never names — `verifier_runtime_per_{module,ptb}_{success,timeout}_latency` — registered inside an upstream type's constructor that ika merely instantiates. They are harmless here (outside both `ika_*` and `consensus_ika_*`, so nothing can collide with them) and not ika's to rename, but asserting on them would have meant maintaining an exception list of names we do not own — exactly what this design removes. The rationale is in the module so nobody re-adds it.

**Fault-validated both halves**: putting the vendored prefix back under `ika_` fails the script (exit 1); registering an ika metric without the `ika_` prefix fails the test. Plus `epoch_metrics` 2/2, guard script clean (`310 literal, 1 prefixed registry, none under `ika``), fmt clean.

**Downstream (infra repo)** — this now changes the ~127 consensus-core re-exports from `ika_consensus_*` to `consensus_ika_*`, on top of the gauge rename. Ship the recording-rule aliases described in #2029 with the rollout so operator dashboards keep working through the transition.

PR #2031, opened 2026-08-11
chore: drop vestigial narwhal naming from comments and test fixtures

omersadika:
## What narwhal was, and why ika never ran it

Narwhal was Mysten's DAG mempool/consensus layer, used by sui before it migrated to Mysticeti. Ika forked sui's node/core scaffolding *after* that migration and has only ever run **Mysticeti via `consensus-core`**. There is no narwhal dependency in any `Cargo.toml`, no `narwhal/` directory, and no narwhal code path anywhere in the tree.

So every remaining `narwhal` reference is a fork artifact: a name that survived the copy but never described anything ika does. Two of them were actively misleading — one comment justified live code with mechanics (worker processes, port reuse) that do not exist under Mysticeti, and a test fixture named a registry `"narwhal"` in a way that read as evidence of a real second metrics namespace and misled a review.

**This PR changes no runtime behaviour.** It touches comments, one test fixture, and dead fork tooling.

Each reference was classified rather than find-and-replaced, and cross-checked against the vendored sui checkout ika builds against (`433212f`), plus sui `main` where noted.

### Per-reference table

| file:line | case | what I did | what sui does today |
|---|---|---|---|
| `crates/ika-protocol-config/src/lib.rs:327` | 1 — stale name, behaviour still true | "one Narwhal consensus output" → "one consensus commit". Still accurate: ordering applies within one commit. | Unchanged at our pin **and** on `main` — still says "Narwhal consensus output". We're ahead. |
| `crates/ika-core/src/consensus_adapter.rs:464-465` | **2 — no longer true** | Old text justified the `within_alive_epoch` barrier by "narwhal workers reuse same ports when narwhal restarts". There are no workers in ika — `worker` appeared nowhere in the consensus adapter or manager except this comment. Rewrote to the reason the barrier actually holds today. | **sui already rewrote this post-Mysticeti**, and the rewrite is in our pin. I adopted their applicable rationale (see below). `narwhal` count on sui `main` for this file: 0. |
| `crates/ika-core/src/system_checkpoints/mod.rs:944` | 1 | `signature.authority == narwhal_cert.author` → `signature.authority == the consensus block author`; also disambiguated which "author" the second clause means. | Unchanged at our pin **and** on `main` (`checkpoints/mod.rs:2551`). We're ahead. |
| `crates/ika-core/src/dwallet_checkpoints/mod.rs:1015` | 1 | Same rewrite — this is ika's own dwallet-checkpoint aggregator, carrying a copy of sui's system-checkpoint comment. | No sui equivalent (dwallet checkpoints are ika-only); the ancestor comment is unchanged upstream. |
| `crates/ika-node/src/metrics.rs:138,159-161,176-178` | 3 — inert | Second registry prefix `"narwhal"` → `"vendored"`, and `narwhal_counter_1` → `vendored_counter_1`. The test only needs *a* second prefix to prove registry composition/removal; the name now says that instead of implying a namespace ika doesn't have. No production metric names touched. | Unchanged at our pin **and** on `main` — still `"narwhal"`. We're ahead. |
| `scripts/update_narwhal.sh` (whole file, 73 lines) | 3 — dead | **Deleted.** Its entire purpose was bumping a `github.com/MystenLabs/narwhal` git pointer that ika does not have; it clones the narwhal repo and greps Cargo.tomls for a dep with zero matches. Nothing to rename — the file only exists for narwhal. | **sui already deleted it.** Absent from our pin. |
| `scripts/dependency.py:67,97,108-109,136` | 3 — dead | Removed the `narwhal` project mode (path map, subpath map, branch, and the `--project` assert). `--project narwhal` previously would have rewritten Cargo.tomls to point at a nonexistent `narwhal/` dir; it now fails fast. `move` mode untouched. | **sui already removed narwhal from this script** — 0 hits at our pin. |
| `scripts/release_notes.py:45` | 3 — dead | Dropped `"narwhal"` from `INTERESTING_DIRECTORIES`; the directory does not exist, so it contributed nothing. | **sui already removed it** — 0 hits at our pin. |
| `scripts/generate-release-notes.sh:17` | 3 — dead | Dropped `narwhal` from the `git log` pathspec (nonexistent directory). | **sui already removed it** — 0 hits at our pin. |
| `scripts/update_fastcrypto.sh:67` | 1 — wrong name, behaviour true | Copy-paste bug in a **live** script (fastcrypto *is* a real git dep): it announced "as the Narwhal pointer in Ika" while updating the fastcrypto pointer. → "fastcrypto pointer". | sui **still has this bug** ("as the Narwhal pointer in Sui") at our pin and on `main`. We're ahead. |

### Case 2 — the one item wanting an owner's eye

`consensus_adapter.rs` `submit_and_wait`. The old comment justified the `within_alive_epoch` barrier with narwhal worker/port-reuse semantics. That rationale is definitively dead — ika has no workers and no per-worker ports.

**The code is still correct, but for a different reason.** I did not change it. The replacement rationale is the one sui documented when they rewrote the same block post-Mysticeti, adapted to what I could verify in ika:

- A validator may leave the committee at an epoch change and run as a fullnode, after which it cannot submit to consensus (`ika-node/src/lib.rs` reconfig path gates validator components on `is_committee_member`).
- The consensus instance is shut down and restarted across the boundary (`ConsensusManager::start`/`shutdown`), so an in-flight submission could otherwise land in the new epoch's instance.

I deliberately did **not** copy sui's second paragraph about user vs. system transactions: ika has no user transactions — every `ConsensusTransactionKind` is a validator/system message — so that sentence would have been false here.

Worth a maintainer confirming the substituted rationale matches their understanding of why the barrier is there.

### Deliberately left alone

`crates/ika-config/src/test_gateway.yml` still contains 60 `narwhal-primary-address` / `narwhal-worker-address` / `narwhal-consensus-address` keys, and is the only remaining `narwhal` match in the repo.

I left it because it is **not narwhal-only**: it is a validator-set fixture whose other keys (`account-key`, `protocol-key`, `network-address`, `stake`, …) are unrelated, so deleting it exceeds "drop narwhal naming", and renaming keys in a file **nothing parses** would make a dead file look maintained. Confirmed dead: no reference to `test_gateway` exists anywhere in the repo, in any filetype, and there is no `include_str!` in `ika-config`. sui ships the identical dead file at our pin.

Recommend deleting the whole file in a follow-up — that is a file-removal call for the owner, not a naming fix.

### Verification

- `cargo test -p ika-node --release --lib metrics` — renamed fixture passes.
- `cargo clippy` and `cargo fmt --check` clean on `ika-node`, `ika-core`, `ika-protocol-config`.
- Edited Python parses; edited shell passes `bash -n`.
- Repo-wide case-insensitive grep for `narwhal` returns only `test_gateway.yml`, justified above.

### Adjacent fork-artifact naming, NOT fixed here

Noted while sweeping, listed for a possible follow-up:

- `scripts/generate-release-notes.sh` queries the **`MystenLabs/sui`** GitHub API for PR bodies and prints `https://github.com/MystenLabs/sui/pull/N` links — it generates sui's release notes, not ika's.
- `release_notes.py` / `generate-release-notes.sh` still list directories that do not exist in ika: `doc`, `external-crates`, `kiosk`, `nre`, `ika-execution`.
- `scripts/dependency.py` is now move-only, but `move/` does not exist in ika either, so the whole script is dead.
- Several touched files still carry `Copyright (c) Mysten Labs, Inc.` headers.


omersadika:
Added `338ee2bf`: deleted `crates/ika-config/src/test_gateway.yml` rather than editing its narwhal keys out.

Investigated before removing, because the file turned out to be foreign in its entirety rather than an ika fixture with some stale keys:

- **No ika type can parse it.** There is no `validator_set` type and no gateway config anywhere in the repo, and `NodeConfig` knows none of `worker-key`, `delegation`, or `gas-price`. `worker-key` and the three `narwhal-*` addresses are narwhal's primary/worker architecture, which ika never ran.
- **It is a sui *gateway client* fixture** — its last line is `db_folder_path: /data/gateway_client_db`. Sui removed the gateway component itself long ago.
- **Nothing loads it**: no reference anywhere in the repo in any filetype, and `ika-config` has no `include_str!`. Its only commits are a global `fmt` and an unrelated dependency bump.

Removing only the narwhal keys would have left a file that is still 100% foreign and still parsed by nothing, but now looks curated — which is worse for the next person who finds it.

Verified: `cargo test -p ika-config --release` → 26 passed / 0 failed, and the repo-wide case-insensitive `narwhal` grep is now **completely empty**. This PR now removes the last narwhal reference in the tree.

PR #2037, opened 2026-08-13
fix(ocs): read the latest-checkpoint watermark through a pruning-immune probe

omersadika:
## Summary

`get_latest_checkpoint` fetches a summary through the fullnode's availability window (`lowest_available..=latest`). A fullnode pruning **at head** can empty that window, and it then serves NotFound for its **own latest checkpoint** — persistently, for as long as the window stays empty. A localnet has been measured holding that state for ~36 minutes.

Two call sites need only the checkpoint **height**, but took it from that summary with `?`, so both fail inside such a window:

| Site | Effect |
|---|---|
| `IkaCheckpointPusher::advance` | Aborts at the top of **every tick**. The cursor stops moving and the verified state cache freezes behind it, so cache-first reads fall through to the network for the duration. |
| `IkaCheckpointPusher::new` | Fails outright — a node whose pusher first-starts inside the window does not come up. |

This is the same class as the boot artifacts-digest probe fixed in #2025: fetching *content* to learn a *number* the pruner can take away.

## The fix

Both sites now read `get_latest_checkpoint_sequence` — a new `SuiTransport` method backed by `GetServiceInfo.checkpoint_height`, the store's own watermark, read **without** the availability check, so it still answers while the window is empty. `get_latest_epoch` gets the same treatment for consumers that need only the epoch.

Both methods default to the existing summary fetch, so transports with no lighter path are unaffected. The three wrapper transports (`fallback`, `retained`, `verified`) forward them **explicitly** — inheriting the default behind a wrapper would silently reopen the hole for exactly the stacks that run in production.

Deliberately **not** included: an availability-keyed cursor fast-forward. The pusher's scan loop already advances the cursor past unfetchable checkpoints (queuing them as pending gaps, now archive-recoverable via #2020), so whether that additional mechanism is still needed here is unproven — better evaluated separately with evidence than bundled into a fix that stands on its own.

## Tests

A regression test drives the pusher against a transport that NotFounds its own latest and asserts the tick survives and still folds what **is** available. Fault-validated: restoring the summary fetch in `advance` reproduces the abort.

### DIFF
--- a/crates/ika-core/src/sui_connector/fallback_transport.rs
+++ b/crates/ika-core/src/sui_connector/fallback_transport.rs
@@ -64,13 +64,12 @@ impl SuiTransport for FallbackTransport {
     async fn get_latest_checkpoint_sequence(
         &self,
     ) -> Result<CheckpointSequenceNumber, TransportError> {
-        // Forward explicitly: the trait default would fall back to the
-        // summary fetch, reopening the pruned-at-head abort behind this
-        // wrapper.
+        // Forward explicitly so a DIRECT primary keeps its pruning-immune
+        // probe through this wrapper; the trait default would substitute the
+        // window-bound summary fetch even when the primary implements one.
+        // (Behind a mirror relay there is no watermark RPC to reach, so the
+        // default is what runs — see the trait method's doc.)
         self.primary.get_latest_checkpoint_sequence().await
-    }
-    async fn get_latest_epoch(&self) -> Result<u64, TransportError> {
-        self.primary.get_latest_epoch().await
     }
     async fn get_full_checkpoint(
         &self,
@@ -183,6 +182,11 @@ mod tests {
         ) -> Result<CertifiedCheckpointSummary, TransportError> {
             self.record("get_latest_checkpoint")
         }
+        async fn get_latest_checkpoint_sequence(
+            &self,
+        ) -> Result<CheckpointSequenceNumber, TransportError> {
+            self.record("get_latest_checkpoint_sequence")
+        }
         async fn get_full_checkpoint(
             &self,
             _seq: CheckpointSequenceNumber,
@@ -247,6 +251,10 @@ mod tests {
         let _ = transport.get_chain_identifier().await;
         let _ = transport.get_current_epoch().await;
         let _ = transport.get_latest_checkpoint().await;
+        // Explicitly forwarded, not inherited: a dropped forward would fall
+        // through to the trait default and record `get_latest_checkpoint`
+        // instead, which this ordered log catches.
+        let _ = transport.get_latest_checkpoint_sequence().await;
         let _ = transport.get_full_checkpoint(7).await;
         let _ = transport
             .get_checkpoint_summary_by_digest(CheckpointDigest::random())
@@ -265,6 +273,7 @@ mod tests {
             "get_chain_identifier",
             "get_current_epoch",
             "get_latest_checkpoint",
+            "get_latest_checkpoint_sequence",
             "get_full_checkpoint",
             "get_checkpoint_summary_by_digest",
             "last_checkpoint_of_epoch",
--- a/crates/ika-core/src/sui_connector/fallback_transport.rs
+++ b/crates/ika-core/src/sui_connector/fallback_transport.rs
@@ -61,6 +61,17 @@ impl SuiTransport for FallbackTransport {
     async fn get_latest_checkpoint(&self) -> Result<CertifiedCheckpointSummary, TransportError> {
         self.primary.get_latest_checkpoint().await
     }
+    async fn get_latest_checkpoint_sequence(
+        &self,
+    ) -> Result<CheckpointSequenceNumber, TransportError> {
+        // Forward explicitly: the trait default would fall back to the
+        // summary fetch, reopening the pruned-at-head abort behind this
+        // wrapper.
+        self.primary.get_latest_checkpoint_sequence().await
+    }
+    async fn get_latest_epoch(&self) -> Result<u64, TransportError> {
+        self.primary.get_latest_epoch().await
+    }
     async fn get_full_checkpoint(
         &self,
         seq: CheckpointSequenceNumber,
--- a/crates/ika-core/src/sui_connector/push_worker.rs
+++ b/crates/ika-core/src/sui_connector/push_worker.rs
@@ -125,8 +125,10 @@ impl IkaCheckpointPusher {
                 persisted
             }
             None => {
-                let latest = transport.get_latest_checkpoint().await?;
-                let cursor = *latest.sequence_number();
+                // Watermark probe, not a summary fetch: a fullnode pruning at
+                // head NotFounds its own latest, which would fail node boot
+                // here.
+                let cursor = transport.get_latest_checkpoint_sequence().await?;
                 info!(
                     cursor,
                     "checkpoint pusher first start — initializing at upstream latest"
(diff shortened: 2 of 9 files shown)


omersadika:
## Review

The probe's premise checks out against the pinned Sui sources: `GetServiceInfo` and `GetCheckpoint(None)` read the identical store value (`service.reader.inner().get_latest_checkpoint()`), differing only in the availability-window check — so the probe returns exactly the number the old path returned when it succeeded, minus the pruning abort. The two pusher call sites are correctly fixed, the trust model does not regress (the watermark was already an unverified hint: neither pusher site BLS-verified the summary it took `sequence_number()` from), and `pusher_tick_survives_a_fullnode_pruning_at_head` models the real server split faithfully.

That said, the fix as it stands does not deliver its headline scenario, and there are two more consumers of the same bug class left live.

### Blocking

**1. Node boot still dies inside the pruned-at-head window — before the pusher ever runs.**
`crates/ika-sui-client/src/grpc_backend.rs:227` — `get_latest_checkpoint_sequence_number` still derives the height from the window-bound `get_latest_checkpoint()`. It is reached via `describe()` → both `SuiClient` constructors (`ika-sui-client/src/lib.rs:132,154`) → node boot (`ika-node/src/lib.rs:633-639, 655-663`), all `?` with no retry, for **every role**. Inside the ~36-minute window this PR itself measured, `IkaCheckpointPusher::new` is unreachable because `SuiClient` construction failed ~450 lines earlier — so "a node whose pusher first-starts inside the window does not come up" is still true, just with the failure moved upstream. One-line fix: route it through the new `get_latest_checkpoint_sequence()`.

**2. `LocalProofProvider::current_head_seq` has the same bug with a wider blast radius.**
`crates/ika-network/src/proof_provider.rs:579` fetches the full latest summary just to stamp `claimed_latest_checkpoint_seq`, and it is `?`-fatal in **every verified-read response** (`verified_object` at 630, `batch_verified_objects` at 649, `verified_dynamic_fields_page` at 693) — including proof-snapshot-cache reads that exist specifically to survive pruning, and for every mirrored peer this node serves, for the whole window. The field is untrusted-by-design (the reader folds it through `fetch_max`), so the pruning-immune probe is semantically valid here too.

**3. No spec update in the PR.**
`dev-docs/specs/ocs-verified-sui-reads.md` still describes the summary-fetch head probe and the abort-on-pruned tick semantics this PR changes, and the deliberately-unverified `GetServiceInfo` watermark read needs its trust posture stated against the spec's "the relay's claimed head is never trusted directly" invariant. CLAUDE.md requires the spec change in the same PR.

### Should fix

**4. The survived tick opens an uncapped per-tick retry storm.**
Pre-PR the `?` abort froze the cursor during a window; now the scan traverses it, queueing one pending gap per checkpoint — at mainnet rates ~2,400–3,000 by the 600s `GAP_RETRY_DEADLINE`. `retry_pending_gaps` (`push_worker.rs:347-422`) then retries **every** gap against the already-degraded fullnode serially each 250ms tick — only *archive* attempts are capped (`ARCHIVE_ATTEMPTS_PER_TICK = 4`). That is the self-amplification spiral the archive-cap comment itself warns about, now reachable through the uncapped half of the loop; and on expiry the window drips thousands into `pusher_gap_dropped_total`, the alert-on-any-increase metric. A per-tick cap on fullnode gap retries, mirroring the archive cap, closes it.

**5. The mirrored-stack forwards promise something the stack cannot deliver.**
`fallback_transport.rs:64-71` and `verified_transport.rs:108-114` forward the new methods with comments saying this avoids the pruned-at-head abort — but `SuiMirrorTransport` has no override and the relay wire has no watermark RPC (`ika-network/build.rs` registers only `get_latest_checkpoint`), so the trait default fires one layer down: a relayed summary fetch, bounded by the *serving peer's* availability window. No mirrored caller exists today, so this is latent — but the comments are wrong where they sit, and the first mirrored consumer silently loses the property the method's doc promises. Either add a relay watermark RPC + mirror override, or document the probe as direct-transports-only and fix the comments. The `RecordingTransport` routing test (`fallback_transport.rs:241`) was not extended to the new methods, so a dropped forward also passes the suite.

**6. The first-start probe arm has no regression coverage.**
`pusher_tick_survives_a_fullnode_pruning_at_head` pre-seeds the cursor (`put_sui_pusher_last_seq(99)`), so `IkaCheckpointPusher::new` takes the `Some(persisted)` arm; the only test hitting the `None` arm uses `latest_summary_pruned: false`. Reverting the `None` arm to `get_latest_checkpoint()` passes the whole suite while re-breaking exactly the first-start-in-window case the PR description names.

**7. Drop `get_latest_epoch`.**
Zero callers (trait default + gRPC impl + three wrapper forwards, ~40 lines across 5 files). It also silently diverges from the pre-existing `get_current_epoch` at every epoch boundary — `GetServiceInfo.epoch` is `latest_checkpoint.epoch()`, one behind between the end-of-epoch checkpoint and the next epoch's first, while `get_current_epoch` (already pruning-immune, natively relayed) reads live state — and its impl duplicates the `GetServiceInfo` fetch verbatim. Land it with its first consumer, pointed at the right semantics.

### Nits / hardening

- `crates/ika-sui-client/src/grpc.rs:273,295` — function-local `use sui_rpc_api::proto::sui::rpc::v2::GetServiceInfoRequest;` in both new methods; the file already imports `v2 as proto` at the top, so `proto::GetServiceInfoRequest` needs no new import (CLAUDE.md: imports at file top).
- Pre-existing but widened by this PR: one inflated watermark value permanently bricks the pusher — the fast-forward persists `new_cursor` ahead of any fetch, `note_processed`'s `fetch_max` disarms the reader staleness tripwire forever, and a restart resumes the poisoned cursor. Pre-PR that required a forged `CertifiedCheckpointSummary` (though never actually signature-checked at these sites); now a bare unauthenticated integer suffices. A clamp on watermarks implausibly far ahead of the cursor is cheap; fine as a follow-up issue if not here.

### Clean lenses

Consensus determinism (watermark feeds only the scan cursor, a metric, and the fast-forward; cache content stays gated by `verify_before_fold`/`verify_archive_checkpoint`), epoch-close lock gating, batch-loop per-item error handling, parameter-set agreement, and silent-skip discipline (both new gRPC error paths return loud `TransportError`s) all came back clean.

---
*Adversarial two-pass review (general + consensus/OCS-invariant lens); findings 1 and 2 verified mechanically against the working tree and the pinned sui-rpc-api sources.*

omersadika:
Re-reviewed `c746c53f` against the review: **all findings addressed** — the review has no remaining blockers from my side.

1. ✅ Boot path: `get_latest_checkpoint_sequence_number` now takes the probe; `describe()` survives the window.
2. ✅ `current_head_seq`: probe, with the trust-posture rationale (`fetch_max`-folded, untrusted by construction) stated at the site.
3. ✅ Spec: the new "Reading the head" section covers all four consumers, the not-a-trust-input argument against the "claimed head is never trusted directly" invariant, and the direct/relayed split; the retry-cap paragraph landed too.
4. ✅ `FULLNODE_ATTEMPTS_PER_TICK = 64`, oldest-first. One observation, not a change request: the `break` at the cap also stops *archive* attempts for gaps beyond position 64 that tick, so during a long window younger gaps wait behind the oldest 64 for their archive shot. Bounded and self-clearing — oldest-first means the front drains (by repair at ~4 archive/tick + fullnode successes, or at worst by the 600s deadline), so the wait is finite and the priority order is the right one.
5. ✅ Wrapper comments now state the truth; trait doc carries the split; the routing test's dropped-forward detection (default records `get_latest_checkpoint` instead) is a nice mechanism.
6. ✅ First-start arm has its own regression test with `latest_summary_pruned: true` and no pre-seeded cursor, fault-validated per the commit message.
7. ✅ `get_latest_epoch` dropped everywhere.
8. ✅ Function-local imports gone.

The one deliberately-deferred item is the inflated-watermark hardening (a bare unauthenticated integer can now poison the persisted cursor via the fast-forward and disarm the staleness tripwire) — that was a pre-existing gap this PR widens at the margin, and it deserves its own issue rather than scope creep here. Worth filing so it doesn't evaporate.

CI is still running (clippy/cargo-test pending at time of writing); modulo green checks, LGTM.

omersadika:
Filed the deferred inflated-watermark hardening as #2041.

omersadika:
Heavy-suite validation on this branch is green:

- **Test Cluster** (dispatched): success — https://github.com/dwallet-labs/ika/actions/runs/31752063830
- **TS Integration Tests** (dispatched): success — https://github.com/dwallet-labs/ika/actions/runs/31752065365 — the highest-signal suite for this change: its localnet prunes aggressively (`num_epochs_to_retain=0`), which is exactly the environment the probe and the traversed-window path run against.

The auto-triggered `v128_rollout` upgrade gate is still in the queue; everything else (PR checks, both dispatched suites, review) is green. Merge-ready from the evidence side once that lands — the remaining gate is the review approval.

Issue #2023, opened 2026-08-10
A validator re-entering MPC processing >~2k rounds behind tip falls into a stable non-contributing equilibrium; restart re-creates the entry condition instead of healing it


omersadika:
## Live now on mainnet (our own validator, v1.3.0, full instrumentation) — captured across a two-restart natural experiment

`dWallet Labs 4` (mainnet, `1.3.0-c86bad38be59`) has contributed **zero MPC advances for 2+ hours** across two process generations, while following consensus normally the whole time (subscriptions 53/53, commit processing at fleet rate). Its four siblings upgraded in the same rolling window recovered fully within minutes. The difference is the entry gap, and the state is self-sustaining once entered.

## The equilibrium (mechanism, from raw endpoint data)

At ~4-5k rounds behind the tip (~4 min at the fleet's ~19 rounds/s), every internal-presign session the validator starts computing has **already completed network-wide** by the time its contribution would matter:

- Its metrics endpoint shows computation series for **rounds 1-2 only** — no round 3/4 has ever run since the restart (a healthy sibling shows all four). It starts sessions, then observes their terminal state and abandons them.
- ~1.3 CPU cores burn continuously on this dead-on-arrival crypto (healthy siblings: ~3.3 cores on useful work).
- `mpc_manager` warns `sustained MPC messages received after terminal sessions` at the 100-cap every 60s — the after-terminal counter running hot is the wasted-work signature.
- The wasted work holds MPC-side round processing at ~18.3 rounds/s against a 19.5 rounds/s tip — **the backlog never drains**. Observed gap oscillates 545–7,346 for 30+ minutes (it touches near-tip, then the live-session crypto burst pulls it back).
- Active session set balloons: 6,577 → 8,571 in-flight internal presigns. Advance completions decayed 292 → 147 → 98 → 49 → 0 per 10m across the first generation and never registered above 0 in the second.

## Why restart makes it worse, not better

Restart #2 (clean pod, fresh process) replayed the epoch's 765k consensus rounds in ~13 min at ~40× tip rate — and necessarily re-entered live processing ~4k rounds behind (replay duration × tip rate), i.e. **inside the trap radius**. It then produced literally zero advances (first generation at least managed a decaying trickle). Replay-acquired terminal knowledge does not help: the wasted work is on *live* sessions near the tip, not replayed ones.

[...]

PR #2038, opened 2026-08-13
fix(mpc): hold the stopped-contributing alarm for an actively-reported catch-up

### LINKED ISSUE #2036: MPC lag alarm fires on every mid-epoch restart: the #1983 stopped-contributing detector does not know about the #2024 catch-up gate
The stopped-contributing detector added in #1983 does not know about the catch-up gate added in #2024, so the gate's normal, healthy operation trips it. Every mid-epoch restart logs, at `ERROR`, that the validator "will not recover on its own" and should be restarted — roughly ten seconds after the MPC service has already logged, at `INFO`, that the very same gap is a backlog it is deliberately draining.

We hit this on all six of our own validators during the v1.3.1 rollout. It is not intermittent: at production round rates it fires on essentially every restart.

## Observed

Six validators (5 mainnet + 1 testnet), restarted onto `1.3.1-d0bb6ac9`. Every one produced the same four-line sequence:

```
18:57:35 INFO  MPC service entered catch-up mode: suppressing new internal-presign and
               user-session computations until the consensus-round backlog drains (issue #2023)
               gap_rounds=644594
18:57:45 ERROR MPC subsystem has stopped keeping up with consensus: this validator is
               following consensus normally but is no longer contributing MPC work. It will
               not recover on its own in every known case — check for an earlier fatal in the
               dWallet MPC service, and restart the node if none explains it
               lag_rounds=637879
19:04:56 INFO  MPC subsystem has caught back up with consensus  lag_rounds=49969
19:05:13 INFO  MPC service exited catch-up mode: the consensus-round backlog drained,
               resuming all cryptographic computations  gap_rounds=71
```

Entry gaps were 643,022 / 644,233 / 644,594 / 645,362 / 645,741 rounds (mainnet) and 421,066 (testnet); the `ERROR` followed each within ~10s with a `lag_rounds` 6–7k lower — the same quantity, sampled a moment later by the other subsystem. All six drained and cleared in 5–10 minutes. Nothing was wrong with any of them.

The `ERROR`'s own remediation advice is the worst available action here: a validator told to restart mid-drain discards its progress and starts the same replay over.

## Why it always fires

Two subsystems measure the same distance with unrelated thresholds and no shared state:

| | where | threshold |
|---|---|---|
| catch-up gate | `dwallet_mpc/catchup_gate.rs:36,45` | enter at `CATCH_UP_ENTER_GAP_ROUNDS = 5_000`, exit at `500` |
| lag alarm | `authority/authority_per_epoch_store.rs:114` | raise at `MPC_LAG_ALARM_ROUNDS = 50_000` |

Consensus rounds restart at 0 each epoch, and a restart builds a fresh `AuthorityPerEpochStore` with `mpc_consumed_consensus_round = 0`, so the gap a restarting node reports is simply *the number of rounds elapsed so far in the current epoch*. At the observed ~19.5 rounds/s that is 50,000 rounds about **43 minutes into a 24-hour epoch**. Past that point — 97% of the epoch — the alarm is not merely likely but arithmetically guaranteed on any restart.

`MPC_LAG_ALARM_ROUNDS` was chosen against exactly this concern; the doc comment says so:

> The threshold is deliberately generous. Genuine catch-up after a restart transiently produces a large gap and then closes it, and we would rather miss the first minutes of a real stall than teach operators to ignore this line. `MPC_LAG_ALARM_ROUNDS` is roughly an hour of production round production at observed testnet rates

The calibration is sound and simply outrun: it sizes the threshold against *an hour of rounds*, but a mid-epoch restart replays *all rounds since the epoch began*, which reaches 1.7M by the end of an epoch. No constant that stays useful as a stall detector can also clear a full-epoch replay — which is why the fix below is a state check rather than a bigger number.

## The fix must not silence the detector

The obvious change — suppress the alarm while `catchup_mode == 1` — reintroduces the failure the detector exists for, and the reason is written at its own definition site:

> The MPC service cannot report its own stall: every path that stops it also stops the code that would notice — most starkly the deliberate `break` on self-recognised maliciousness, which ends the service loop for the life of the process while consensus keeps running normally. So the MPC side only publishes its progress here, and the CONSENSUS commit path — which is still alive in exactly the cases that matter — does the comparing.

`catchup_mode` is published by the MPC service loop. If that loop dies *while catch-up is engaged* — the self-malicious `break` of #1978, taken during the post-restart replay window, which is a real and reachable ordering — the flag latches at `1` and a naive suppression hides the stall permanently. That is strictly worse than today's false positive.

So the suppression has to be one the MPC service must keep actively renewing.

## Suggested shape

Publish catch-up state through the existing progress channel rather than reading the gauge. `record_mpc_consumed_consensus_round` already crosses this boundary for precisely this purpose (`authority_per_epoch_store.rs:291`); extend it, or add a sibling, so the service reports *round consumed + whether the gate is engaged*, stamped with the observation time. Then in `report_mpc_consensus_round_lag`:

- gate engaged **and** the observation is fresh (say, within a small multiple of the service-loop interval) → the gap is explained; hold the alarm;
- gate engaged but the observation has gone stale → the service stopped reporting, which is the stall case → raise, as today;
- gate not engaged → unchanged.

A dead service stops refreshing, the suppression expires on its own, and the detector keeps the property that makes it worth having.

Two smaller things worth folding in:

- **A stuck drain is currently unalarmed.** #2023's trap was a node that sat at a large gap without draining. Once the alarm is suppressed during catch-up, that case has no loud signal at all — the `catchup_gap_rounds` gauge's slope carries it, but only for someone already watching. Consider raising when catch-up has been engaged beyond some bound without the gap decreasing; that is the genuinely actionable version of this alarm, and it is a condition today's threshold cannot express.
- **Say what to do.** Whatever remains loud during a legitimate catch-up should point at `ika_dwallet_mpc_catchup_gap_rounds` and say "draining is normal; act only if it stops falling", rather than "restart the node".

## Impact

Cosmetic in effect, corrosive in practice — this is the release whose headline fix is that restarts are now safe, and every operator performing the upgrade restart will meet an `ERROR` telling them it did not work and to restart again. It teaches the fleet to ignore the one line that means an MPC subsystem is genuinely dead.

Not a v1.3.1 blocker: the underlying behaviour is correct and the drain completes. Worth landing before the alarm's credibility is spent. The v1.3.1 release notes call the line out as expected so operators are not misled in the meantime.

omersadika:
Closes #2036.

## Problem

The #1983 stopped-contributing detector and the #2024 catch-up gate measure the same distance with unrelated thresholds and no shared state. Consensus rounds restart at zero each epoch and a restarted node's fresh epoch store starts its MPC cursor there too, so a mid-epoch restart's replay gap is simply the rounds elapsed this epoch — past ~43 minutes of a 24h epoch it exceeds `MPC_LAG_ALARM_ROUNDS` and the ERROR ("will not recover on its own … restart the node") was arithmetically guaranteed on every restart. All six production validators hit it during the v1.3.1 rollout, ~10s after the MPC service logged that the same gap is a backlog it is deliberately draining; the advice — restart — would have discarded the drain and replayed it.

## Fix: the suppression is a lease, not a flag

Naive suppression (read the `catchup_mode` gauge) would hide a service that dies *mid-catch-up* — the gauge is published by the service loop and latches at 1 exactly when that loop dies, and the self-recognised-malicious `break` is reachable during the post-restart replay window. So instead, per the issue's prescribed shape:

- The MPC service publishes **(consumed round, catch-up gate engaged, observation time)** as one immutable `ArcSwapOption` snapshot on every consumed round — one coherent triple, because split atomics could pair a fresh timestamp with a stale gate flag, the exact bit that decides suppression. The gate state is sampled from the `CatchUpGate` itself, never from its metric.
- The consensus commit path holds the stopped-contributing alarm only while the reported catch-up is **fresh** (`MPC_CATCH_UP_REPORT_FRESHNESS = 60s` — ~3,000 service-loop ticks, and a draining validator republishes per consumed round at ~1-2k rounds/s, so nothing short of the service actually stopping expires it; still 1/60th of the alarm's own time scale, so a service that dies mid-catch-up is reported within a minute rather than never).
- Report stale → raise exactly as today. Gate not engaged → behavior unchanged (verified by the pre-existing tests passing untouched, and by fault-injection: reverting the hold breaks only the four new tests).

## The hole the hold creates gets its own alarm

While a live catch-up holds the alarm, a drain that *stopped draining* would have no loud signal — the genuinely actionable case from #2023. New transition-latched alarm: a reported catch-up whose lag has not reached a **new low** for `MPC_CATCH_UP_STUCK_DRAIN = 15min` raises "stuck drain". Low-water tracking rather than sample-to-sample comparison, because the tip advances while the cursor chases it — a healthy drain's samples legitimately tick upward. Calibration: the largest production backlog (~640k rounds) drained completely in 5–10 minutes at ~1-2k rounds/s; a drain at a hundredth of that still sets new lows every few commits. Its message says the opposite of the old advice: **do not restart** — a restart discards the drain's progress and replays it.

## Alerting moves off the raw lag

Two new gauges — `ika_mpc_stopped_contributing_condition_active` and `ika_mpc_catch_up_stuck_condition_active` — because any operator alert on `ika_mpc_consensus_round_lag > threshold` still pages on every healthy restart; the condition gauges already account for the reported catch-up. The no-report sentinel branch resets both so a condition raised at the end of one epoch cannot export stale through the next epoch's pre-report window (found in review). `production-alerts.md`, `mpc-stall-postmortem.md`, and the metrics convention doc updated in the same change, including the general rule this bug teaches: an outside-the-subsystem detector must be told what the subsystem is deliberately doing, but must never learn it from a flag that outlives the loop that set it.

## Review

Adversarially reviewed for consensus-determinism and detector soundness before commit: no invariant violations; every service-stop mode (self-malicious break, panic, channel error, death while catching up at small or large lag) traced to a bounded loud alarm; everything here is node-local observability — the detector's return value is discarded at the call site and nothing feeds votes, checkpoints, or MPC inputs. Review-found fixes folded in: epoch-boundary gauge reset, drain-rate figure reconciliation across comments.

Known accepted scopes (pre-existing, unchanged by this PR, called out in review): a service that never reports at all after restart is still detectable only via the `-1` lag sentinel, and a gate stuck in the 500–5,000 hysteresis band with a trickle of new lows holds both alarms — both predate this change.

## Testing

All release mode, full modules:

- `authority_per_epoch_store`: 31 passed — including four new deterministic (`start_paused` + `tokio::time::advance`, no sleeps) tests: held at 10x threshold through a 20-minute drain; raises when the report goes stale at exactly the freshness bound; stuck-drain raises once at the bound and clears on a new low while the stopped alarm stays held; five hours of jittery-but-progressing drain stays silent.
- `dwallet_mpc::catchup_gate`: 7 passed; `epoch::epoch_metrics`: 1 passed.
- `dwallet_mpc::integration_tests::catchup`: 2 passed (~218s, real crypto) — the end-to-end gate test now also asserts the service publishes its catch-up state to the epoch store on every drained round.
- Fault-injection (test-testing playbook): reverting the hold → exactly the 4 new tests fail; dropping only the freshness bound → exactly the stale-report test fails, proving the lease is load-bearing. Both injections reverted.
- `cargo fmt`, `clippy -p ika-core --all-targets` clean, `scripts/check-metric-names.sh` OK.

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -17,6 +17,7 @@ use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::time::Duration;
 use sui_types::base_types::{EpochId, ObjectID};
 use tracing::{debug, error, info, instrument, trace, warn};
 use typed_store::rocks::{DBBatch, DBMap, DBOptions, MetricConf, default_db_options};
@@ -113,6 +114,70 @@ pub const EPOCH_DB_PREFIX: &str = "epoch_";
 /// catching up closes the gap; a stopped one only grows it.
 const MPC_LAG_ALARM_ROUNDS: u64 = 50_000;
 
+/// How recent an MPC progress report must be for its catch-up flag to hold the
+/// stopped-contributing alarm.
+///
+/// The hold is a LEASE the MPC service renews, never a flag it can leave set:
+/// the service publishes a report for every consensus round it consumes, so a
+/// service genuinely draining a backlog renews this hundreds to thousands of
+/// times a second, and one that stopped renews never. An expiry-free flag would
+/// hand the stall its perfect hiding place — the deliberate service-loop
+/// `break` on self-recognised maliciousness is reachable during the
+/// post-restart replay window, and would otherwise latch the suppression for
+/// the life of the process.
+///
+/// A minute is ~3,000 service-loop iterations (the loop sleeps 20ms between
+/// them) and orders of magnitude beyond the per-round drain interval a
+/// catching-up validator sustains (~1-2k rounds a second in production), so
+/// nothing short of the service actually stopping expires it — the bound is
+/// generous in the direction where being wrong reintroduces the false alarm
+/// this exists to remove. It is still 1/60th of the time scale
+/// `MPC_LAG_ALARM_ROUNDS` encodes (~an hour of round production), so a service
+/// that dies mid-catch-up is reported within a minute rather than never.
+const MPC_CATCH_UP_REPORT_FRESHNESS: Duration = Duration::from_secs(60);
+
+/// How long a reported catch-up may run without its lag reaching a new low
+/// before the drain is called stuck.
+///
+/// While the service keeps reporting a catch-up the stopped-contributing alarm
+/// is held, so this is the only line left that can say the validator is not on
+/// its way back. The bound is generous on purpose: the largest backlog seen in
+/// production (~640k rounds, a mid-epoch restart replaying the epoch) drained
+/// COMPLETELY in five to ten minutes — with computation withheld the drain runs
+/// at ~1-2k rounds/s against a ~19.5 rounds/s tip. A drain running at a
+/// hundredth of that still reaches a new low every few commits, so a quarter
+/// hour without one is not a slow drain — it is a drain that stopped.
+const MPC_CATCH_UP_STUCK_DRAIN: Duration = Duration::from_secs(15 * 60);
+
+/// What the MPC service publishes for the consensus-path stall detector: the
+/// round it has finished consuming, whether its catch-up gate was engaged when
+/// it finished, and when it said so.
+///
+/// One immutable value behind a single `ArcSwapOption` rather than three
+/// separate atomics, so the consensus path always weighs a round against the
+/// gate state and the timestamp that were true *together*. Torn across atomics
+/// these would pair a fresh timestamp with a stale gate flag, which is the
+/// exact combination that decides whether the alarm is held.
+#[derive(Debug, Clone, Copy)]
+struct MpcServiceProgress {
+    consumed_round: Round,
+    catching_up: bool,
+    observed_at: Instant,
+}
+
+/// Consensus-side view of the catch-up currently being reported: the smallest
+/// lag seen since it started and when that low was reached.
+///
+/// The MINIMUM rather than the previous sample, because the tip keeps advancing
+/// while the cursor chases it: a drain closing the gap fast still produces
+/// individual samples that tick upwards, and comparing consecutive samples
+/// would read those as a stall.
+#[derive(Debug, Clone, Copy)]
+struct CatchUpDrainProgress {
+    lowest_lag_rounds: u64,
+    lowest_lag_at: Instant,
+}
+
 /// What a single MPC-lag sample changed, so the loud log fires on transition
 /// rather than on every consensus commit. Returned rather than kept private so
 /// the latching is directly testable — the state flag alone cannot distinguish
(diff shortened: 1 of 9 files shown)


omersadika:
Validation complete — all dispatched evidence is green:

- **Integration Tests CI** (dispatched, test_threads=4): success — https://github.com/dwallet-labs/ika/actions/runs/31744882390
- **Test Cluster** (dispatched): success — https://github.com/dwallet-labs/ika/actions/runs/31744884489
- **v128_rollout upgrade gate** (auto-triggered on this PR): success (1h9m) — the detector change survives an actual cross-binary swap scenario.
- PR check suite (Cargo Test, Clippy, Clippy-msim, formatting): all green.

Together with the pre-commit validation in the description (module suites, fault-injection test-testing, adversarial consensus review with zero invariant violations), this is merge-ready from the evidence side; the only remaining gate is the required review approval.

Issue #2019, opened 2026-08-10
NOA presign demand: the payload fields an announcer can still choose


omersadika:
Successor to #2015, which is closed: its `signature_algorithm` half was fixed in #2017 by DERIVING the algorithm from the demand identity instead of trusting the announcement. Two things that issue raised are not fixed, and they share its root cause — the consensus dedup key for a presign-demand announcement is the demand-id digest alone, so the first announcement sequenced supplies any payload field the drain uses, and the honest duplicates are dropped behind it.

**Neither is reachable on a live network.** Announcement is gated behind `noa_checkpoints()`, planned for protocol v8 and enabled on neither mainnet nor testnet.

---

## 1. `network_encryption_key_id` is still announcer-supplied

Unlike the algorithm, this one cannot be derived at the drain: it is frozen at announce time on purpose, so the assignment does not rest on the announce-time and instantiate-time key resolutions agreeing.

The consumer-side options are both unattractive, and the reason is worth keeping in front of whoever picks this up:

- **Reject on mismatch** strands the demand. The honest announcements were already deduplicated away, so no correct one can follow; its sign never happens and the epoch cannot finalize its NOA checkpoints. This is the same argument that made deriving the right shape in #2017.
- **Check against the local selection** is only safe if two honest validators always agree on the current NOA key at the moment they drain.

### The open question: can honest validators disagree, transiently?

Evidence so far says yes. `network_owned_address_signing_network_encryption_key_id` takes the minimum over the ADOPTED key set, and adoption is not a consensus-synchronised step — `adopt_cert_verified_keys` is a per-tick local pass gated on the prior epoch's handoff certificate. Which keys are adoptable is network-uniform; WHEN a validator adopts is not:

- a validator whose certificate read ERRORS skips adoption entirely for that tick (deliberately — an error is not the same answer as an absent certificate);
- individual keys can be skipped on a digest mismatch while others are adopted, so the minimum can differ.

Note that the selector's own comment asserts validators "agree on the NOA key by construction", but the justification it gives is about the internal-presign top-up loop being consistent, which does not establish cross-validator uniformity of the adopted set. **That claim is worth settling on its own merits** — other decisions rest on it.

### Likely shape

[...]

PR #2039, opened 2026-08-13
fix(noa): commit the signature algorithm in a gRPC-attestation demand id

omersadika:
Closes half of #2019 — the half that needs no design decision. The `network_encryption_key_id` half stays open there, since it is blocked on settling whether honest validators can transiently hold different adopted key sets.

## The problem

The consensus dedup key for a presign-demand announcement is the demand-id digest alone. That is deliberate — several validators announcing the same demand collapse into one transaction — but the consequence is that the **first** announcement sequenced supplies any payload field the drain uses, and the honest duplicates are dropped behind it.

#2017 removed the announcer's say over `signature_algorithm` for **checkpoint** demands by deriving it from the demand identity: the id carries `kind_name`, and the kind fixes the algorithm.

The **attestation** arm could not derive. Its identity was `keccak256(message)` alone, which says nothing about the algorithm, so `expected_signature_algorithm` returned `None` there and the drain fell back to the announced value:

```rust
GrpcAttestation([u8; 32]),
...
Self::GrpcAttestation(_) => None,
```

That is inert on main today — nothing announces attestation demands here — but it means #2017's fix does **not** cover gRPC-attestation signing when that lands. A byzantine committee member could pre-announce the id for a message it expects, with an algorithm (and so a presign pool) of its choosing.

## The fix

Make the algorithm part of the demand's **identity** rather than payload beside it:

```rust
GrpcAttestation {
    session_identifier: [u8; 32],
    signature_algorithm: DWalletSignatureAlgorithm,
},
```

A substituted algorithm now yields a **different digest** — a demand no honest consumer ever looks up — instead of a competing answer for the same consensus dedup key. This closes the hole at the source rather than policing it at the consumer, which also avoids the reject-strands-the-demand trap that shaped #2017: there is nothing left to reject.

`expected_signature_algorithm` becomes total, so the `None` arm and the drain's `unwrap_or` both disappear.

## Why now, not after

A demand-id preimage cannot be versioned once demands are flowing. Changing it today costs nothing: the digest keys the consensus dedup and the per-epoch presign-assignment table, both per-epoch, and `noa_checkpoints()` is enabled on no network. #2019 flags this as a prerequisite for enabling gRPC-attestation signing rather than a follow-up to it, and that sequencing is the whole point.

## Tests

- every demand identity determines its own algorithm (both arms);
- a substituted algorithm is a **distinct** demand — the property the fix rests on.

Fault-validated: keying `digest()` on the session identifier alone, as the pre-fix shape did, fails the second test. The existing drain end-to-end test from #2017 still passes.

### DIFF
--- a/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
+++ b/crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs
@@ -650,15 +650,20 @@ impl DWalletMPCService {
     /// Drains the channel into a pending buffer, then instantiates sessions
     /// for requests whose network key is already available.
     fn process_network_owned_address_sign_requests(&mut self) {
-        // Drain the receiver into the shared pending buffer, deduplicating by message.
+        // Drain the receiver into the shared pending buffer, deduplicating by
+        // the demand IDENTITY rather than by the message: the same message
+        // under two algorithms is two distinct legitimate demands (the
+        // algorithm is part of the identity), and keying on the message alone
+        // would drop the second at intake while a peer's announcement of it
+        // still consumed a presign in the drain.
         while let Ok(request) = self.network_owned_address_sign_requests_receiver.try_recv() {
-            let message_hash: [u8; 32] = DefaultHash::digest(&request.message).into();
-            if self.submitted_noa_sign_messages.contains(&message_hash) {
+            let demand_digest = request.demand_id.digest();
+            if self.submitted_noa_sign_messages.contains(&demand_digest) {
                 ika_types::report_invariant_violation!(
                     "duplicate_noa_sign_request",
                     message_len = request.message.len(),
                     curve = ?request.curve,
-                    algorithm = ?request.signature_algorithm,
+                    algorithm = ?request.demand_id.expected_signature_algorithm(),
                     "Skipping duplicate network-owned-address sign request"
                 );
                 continue;
@@ -666,7 +671,7 @@ impl DWalletMPCService {
             debug!(
                 message_len = request.message.len(),
                 curve = ?request.curve,
-                algorithm = ?request.signature_algorithm,
+                algorithm = ?request.demand_id.expected_signature_algorithm(),
                 "Received network-owned-address sign request"
             );
             self.pending_network_owned_address_sign_requests
@@ -720,7 +725,6 @@ impl DWalletMPCService {
                     .push(ConsensusNOAPresignDemand {
                         authority: self.name,
                         demand_id: request.demand_id.clone(),
-                        signature_algorithm: request.signature_algorithm,
                         network_encryption_key_id,
                     });
                 self.announced_noa_demand_digests.insert(digest);
@@ -758,7 +762,7 @@ impl DWalletMPCService {
                             .instantiate_network_owned_address_sign_session(
                                 request.message.clone(),
                                 request.curve,
-                                request.signature_algorithm,
+                                request.demand_id.expected_signature_algorithm(),
                                 request.hash_scheme,
                                 presign_session_id,
                                 presign_blending_index,
@@ -921,7 +925,6 @@ impl DWalletMPCService {
                 let tx = ConsensusTransaction::new_noa_presign_demand(
                     demand.authority,
                     demand.demand_id.clone(),
-                    demand.signature_algorithm,
                     demand.network_encryption_key_id,
                 );
                 if let Err(e) = self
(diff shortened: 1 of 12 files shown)


omersadika:
## Review

Two-pass review (general + consensus/wire-format lens). **Nothing blocks merge.** The fix does exactly what it claims: the substitution hole is genuinely closed at the digest — a byzantine announcer who front-ran an honest demand with a different algorithm used to supply the payload for the whole network (the dedup key is `demand_id.digest()` alone, so every validator would pop from the wrong pool and the attestation wedged); now a different algorithm is a different identity that no honest consumer looks up. The derivation is a pure function of the consensus-agreed id (no local inputs, no iteration-order dependence), and the wire audit is clean: the only `messages_consensus.rs` change is inside the wire-format *test* module, `NOAPresignDemand` stays at tag 15, no live variant moves, and the layout change rides exclusively inside the `noa_checkpoints`-gated demand — free to change under the NOA-not-live rule. The digest-separation unit test pins the collision case. Epoch boundaries, dedup/LRU interaction (the #2014 fix is unaffected), and batch-loop discipline all came back clean.

The findings that survived verification are all one theme: **the identity is now authoritative, but its sibling copies of the algorithm are still alive downstream** — and the window to remove them for free closes when NOA ships.

### The one change worth making before this pattern calcifies

**1. Instantiation still trusts the sibling field, not the identity** (`dwallet_mpc_service.rs:761`, announce copy at `:723`). The drain pops the presign for `demand_id.expected_signature_algorithm()`, but the sign session is instantiated with `request.signature_algorithm` — and the `noa_assigned_presigns` record stores no algorithm, so nothing can cross-check. A future gRPC producer that lets the two copies drift gets: every validator pops from the identity's pool, wraps per the *field's* `is_vss()` rule, runs the wrong session, `instantiate` returns true on all paths, the presign is permanently burned, and the demand is never signed — deterministic network-wide, with only the mismatch log as signal. Pre-PR the `unwrap_or(demand.signature_algorithm)` made the two values equal by construction on this arm; the PR makes the drift representable.

**2. The fix is to delete the redundancy, not guard it** (`messages_dwallet_mpc.rs:218` + `NetworkOwnedAddressSignRequest.signature_algorithm`). With `expected_signature_algorithm()` now total, both sibling fields are fully derivable state — their only reads are the wire re-emit, log lines, the mismatch-compare, and the instantiate arg above. Delete both fields, drop the parameter from `new_noa_presign_demand`, derive at announce and instantiate: the mismatch class becomes unrepresentable, and the compare block plus finding 3's alarm site are deleted outright. NOA carries zero compat constraints today — this is the one moment the deletion is free.

### Should fix (individually small; several vanish if you take #2)

**3. The mismatch invariant-violation re-fires unbounded** (`dwallet_mpc_service.rs:1785`). It sits inside the `retain` closure whose pool-empty arm KEEPS the demand, so one byzantine mismatched announcement on a starved pool re-fires `report_invariant_violation!` every drain pass (~20ms) until the pool fills — an attacker-paced paging flood, against the metrics convention's fire-on-transition rule. Pre-existing shape on the Checkpoint arm (#2017); this PR makes the attestation arm reachable too.

**4. Intake dedup contradicts the new identity model** (`dwallet_mpc_service.rs:656`). Local sign requests dedup on `digest(message)` alone, but the PR's own model says the same message under two algorithms is two distinct legitimate demands — the second is dropped at intake with a `should_never_happen` report and never signed, while a peer's announcement of it still burns a presign here. Key the dedup by `request.demand_id.digest()` (already computed two lines away).

**5. No spec update** — CLAUDE.md requires the decision rule land in dev-docs in the same PR, and `internal-presign-pool.md` is the spec that owns this drain. Nothing there became *false* (the old shape was never documented — #2017 skipped it too), but the substitution defense now lives only in code comments; the spec should say which pool a demand pops from and why the announced value is ignored.

**6. Test-plumbing duplication** — the `GrpcAttestation` construction is repeated at 12 sites across three test files, each now writing the algorithm literal twice; a drift between the copies is no longer a reported discrepancy but a silently different identity. A `noa_sign_request(...)`/`noa_grpc_demand_id(...)` helper collapses them, matching the module's existing `send_*_to_all_parties` convention.

**7. One comment ask**: the drain's "the announced value is never consulted" framing is now true for the algorithm but NOT for `network_encryption_key_id`, which stays announcer-chosen behind the same dedup — the deferred half of #2019, whose permanent per-demand wedge (bogus key → `Ok(None)` forever → honest re-announcement blocked) is real. One `TODO(#2019)` on the field so an auditor finds the known hole before NOA ships.

Also noted: the end-to-end derived-over-announced integration test still exercises only the Checkpoint arm; if the field deletion (#2) isn't taken, an attestation-arm variant is the guard that the announced value stays inert.

### Killed during verification

Fabricated-demand presign-burn DoS (byte-for-byte pre-existing on main), stale-comment claims (accurate in context), fully-qualified-path style flag (preserves pre-existing file style), tautological-getter-test claim (the non-default-algorithm pin has real regression value).

---
*Findings 1–4 were found independently by both passes; wire-safety and determinism verified against the PR branch and the dedup-key/LRU call sites.*

PR #2042, opened 2026-08-14
refactor(presign): assign by demand identity; take the bare pop off the trait

omersadika:
## The hazard

Every presign consumer replays its demand stream from a per-epoch table after a restart. A bare pool pop on that replay binds a **different** presign than the never-crashed peers bound — fills complete out of sequence order, so the pool head moves between the original pop and the replayed one — and the checkpoint or attestation built from it is byte-divergent from an honest validator.

Both live consumers already defend against this, **separately**:

| Consumer | Idempotency key |
|---|---|
| `serve_global_presign` | session sequence number |
| `assign_noa_presign` | demand-id digest |

The two implementations were near-identical. Meanwhile `pop_presign` — the bare, self-committing pop with **no** idempotency key — sat on the same trait beside them, with no production caller.

## Why now

That is a latent defect rather than a live one: nothing pops bare today, so this PR fixes no bug. The exposure is entirely in the *next* consumer, which can reach for the bare pop exactly as easily as for the safe path, and whose author has to know this hazard to choose correctly.

So the fix is to remove the choice. The two implementations collapse into one `assign_presign_for_demand`, keyed by a `PresignDemand` whose identity **is** the idempotency key, and the bare pop comes off the trait. A consumer can now only obtain a presign by naming what it is for. Tests that legitimately need a raw pop for pool inspection get an explicit `pop_presign_for_testing`.

**No behavior change.** The merged bodies keep each arm's exact semantics, including the NOA arm's used-presign marking inside the same committed batch as the pop.

## Scope note

The user-facing `assign_presign` is deliberately left alone. It is *also* uncalled here — as is the entire assigned-pool path it writes — but it carries real test coverage, and removing it is a separate question that belongs with the user-presign flow that will eventually consume it. Flagging it because a reader may notice the same dead-code smell and wonder why only half of it moved.

## Tests

- New: assignment is idempotent in the demand identity, for **both** demand kinds — re-seeing a demand returns the same presign and does not pop the pool again. Fault-validated by removing the short-circuit, which fails it.
- Existing: the crashed-vs-control replay test (a replayed round must bind the presign it bound before the crash) now exercises the unified API.
- Store tests 28/28; `internal_presign` + `validator_restart` 13/13; clippy clean under `-D warnings`.

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -353,6 +353,26 @@ pub struct ExecutionIndicesWithStats {
 /// `(presign session id, blending index, raw presign bytes, network key id)`.
 pub(crate) type NoaAssignedPresign = (SessionIdentifier, u16, Vec<u8>, ObjectID);
 
+/// What a presign is being assigned TO — and, because assignment is keyed by
+/// it, the idempotency key.
+///
+/// Every consumer's demand stream is replayed from a per-epoch table after a
+/// restart, so a bare pool pop on replay binds a DIFFERENT presign than the
+/// never-crashed peers bound (fills complete out of sequence order, so the
+/// pool head moves between the original pop and the replayed one). That is a
+/// byte-divergent checkpoint or attestation from an honest validator. Routing
+/// every assignment through a demand identity makes the replay a no-op
+/// instead: a re-seen demand returns what it was already given, without
+/// popping.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PresignDemand {
+    /// A global presign request, identified by its consensus sequence number.
+    GlobalRequest { session_sequence_number: u64 },
+    /// A network-owned-address sign demand, identified by its demand-id
+    /// digest.
+    Noa { digest: [u8; 32] },
+}
+
 /// Trait for the AuthorityPerEpochStore, which gets recreated at the beginning of each epoch.
 pub trait AuthorityPerEpochStoreTrait: Sync + Send + 'static {
     fn insert_pending_dwallet_checkpoint(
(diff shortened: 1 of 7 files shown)


omersadika:
## Review

Two-pass review (general + consensus/replay-determinism lens). **No correctness regression found** — and this was checked hard, not assumed: both passes independently verified the unified `assign_presign_for_demand` byte-preserves every guard of the two deleted paths (idempotent short-circuit re-read *before* any pop; pop + record + NOA `used_presigns` marker in one committed batch), the crash windows all close (crash-after-record replay rebuilds the identical `RespondDWalletPresign` from recorded bytes; no path deletes assignment records, so even a stale record replays deterministically on every validator), the unified key is consensus-agreed data only (request sequence number / demand-id digest — no pool-head position, arrival order, or clock), the per-epoch tables can't leak across a boundary, and the epoch-close lock lens confirms serving stays deliberately ungated per `epoch-close-session-lock.md`. The refactor genuinely removes the hazard it names.

What must be fixed is the documentation layer — which for this subsystem is load-bearing, because the docs are what the next consumer's author reads:

### Must fix before merge

1. **The owning spec was not updated** — `dev-docs/specs/internal-presign-pool.md:79-100` still specifies the replay contract in terms of `serve_global_presign`, `assign_noa_presign`, and a live `pop_presign`, all deleted/renamed here; the PR touches no dev-docs file (CLAUDE.md: the spec is part of the change). The next author greps for those names, finds zero code hits, and lands in exactly the "spec and code disagree" trap. The table should describe the single `assign_presign_for_demand` entry point keyed by `PresignDemand`.
2. **Garbled rustdoc** — `authority_per_epoch_store.rs:2097-2115`: the deleted method's 19-line doc block ("atomically and idempotently… records the serve in the SAME committed batch… Mirrors `assign_noa_presign`") now heads `pop_presign_for_testing` — the one function whose entire hazard is being NOT idempotent now carries a doc promising idempotency. Delete the block; its content is restated correctly on the unified method.
3. **Factually inverted atomicity comment** — `:2146-2149` says "`prepare_pop_presign` self-commits its half", which is backwards: it returns an *uncommitted* batch (its own doc at :2221 says so), and that is the entire mechanism making the shared-batch design work. A safety comment that misstates which half commits is precisely the text a future editor "fixes" the code to match, opening the crash window it warns about.
4. **Eight stale name references** across `dwallet_mpc_service.rs` (:1611, :1614, :1728, :1753), `mpc_manager.rs:3264`, and the write-discipline audit annotations at `authority_per_epoch_store.rs:1511/:1658/:1673` — these are the justification trail reviewers use to accept the drain loops' replay-safety, and post-merge their cited symbols won't exist.

### Should fix

5. **NOA-arm unit assertions**: the new idempotency test asserts the returned triple and pool size but not the two merge-fragile writes — the `used_presigns` marker and the network-key id in the 4-tuple. A refactor dropping either passes every unit test today and fails only at signing time. Two cheap asserts close it.
6. **The mock pool pops LIFO where the real store pops FIFO** (`integration_tests/utils.rs:178`, pre-existing) — the replay-binding tests therefore prove idempotency against an ordering the real store never produces, and structurally cannot reproduce the exact "pool head moved between original and replayed pop" scenario the PR's own body names. Make the mock FIFO.
7. **`assign_presign` remains on the trait** — self-committing, non-idempotent, zero production callers — while the new docs claim "Production has no bare pop" absolutely. Either demote it to an inherent method until the user-presign flow lands, or scope the claims to "no bare pop on any replayed path"; as written, it's the one door left open for the identical divergence.
8. **The unification is the natural point to pin the signature algorithm**: the NOA record stores the network-key id precisely so replay binds consistently, but not the algorithm — for `GrpcAttestation` the pool is chosen by the first-sequenced announcement while the sign session takes its algorithm from the local request. This is the same root as the sibling-field finding in the #2039 review; deleting the redundant field there (deriving from identity) resolves both at once. Coordinate the two rather than patching independently.
9. **`PresignDemand::Noa([u8; 32])`** discards the `NOAPresignDemandId` identity type whose `digest()` is documented as exactly this table key — any 32-byte value type-checks at a future call site and silently keys the assignment wrong. Carry the identity type (or a digest newtype) so the mix-up is a compile error.

### Nit

10. The headline idempotency test's assert message contains an 18-space run from a broken string wrap — it will render as corrupted output at the exact moment it needs to be trusted.

### Verified clean

Trait-surface removal complete (no production bare-pop caller; test helpers `#[cfg(test)]`-gated); the widened return tuple absorbed at all call sites; no interaction with the consensus-output dedup-LRU hardening; batch loops per-item; liveness symmetric (pool-empty keeps the demand queued with unchanged observability); base already contains the derived-algorithm change, so no semantic conflict with it — the demand-id digest commits the algorithm where identity determines it, and replay re-derivation is deterministic.

---
*Findings 1–4 verified line-by-line in the PR-branch source by both passes independently.*

PR #2049, opened 2026-08-15
chore: upgrade Sui to mainnet-v1.77.2

omersadika:
## Description

Upgrade the Sui pin from `mainnet-v1.76.1` to `mainnet-v1.77.2`, everywhere it
lives in one PR:

- **`Cargo.toml`** — 45 Sui git-dep tags -> `mainnet-v1.77.2`.
- **`sui-rust-sdk` rev** — `sui-sdk-types`/`sui-crypto`/`sui-rpc` rev bumped to
  `5781b718`, the rev Sui 1.77.2 itself pins. Leaving the old rev would put two
  `sui-sdk-types` in the graph (one per git source) and fail type unification.
- **msim simtest rev** — `2c2ec3ff` -> `6b8e1e8c` (the rev Sui 1.77.2's own
  simtest uses). The `tokio` exact pin stays `=1.52.1`: verified both Sui 1.77.2
  and that mysten-sim rev still ship `msim-tokio 1.52.1`, so the
  `[patch.crates-io.tokio]` stays effective.
- **`consensus_manager`** — Sui 1.77.2 added a required
  `transaction_pool: Option<Arc<dyn TransactionPool>>` param to
  `ConsensusAuthority::start`; ika has no in-process pool (it submits via
  `TransactionClient`), so pass `None`. Verified 1.77.2's upstream
  `to_consensus_protocol_config` still hardcodes `enable_v3=false` and
  `300/12`, so ika's mirrored constants hold.
- **CI sui-binary URLs** (`ts-ci`, `ts-integration-tests`,
  `system-tests-ci`), **`CLAUDE.md`** gotcha, **`Cargo.lock`** (surgical — only
  the Sui git revs and their sui-rust-sdk transitives move; no unrelated
  registry drift).

> [!important]
> Developers running localnets must update the local `sui` binary to
> `mainnet-v1.77.2` — a mismatched binary completes DKG but silently stalls
> reconfiguration (see `dev-docs/playbooks/localnet.md`).

## Test plan

- `./scripts/check-sui-version-consistency.sh` passes.
- `cargo build --release` passes.
- Dispatched `integration-tests-ci`, `test-cluster`, and
  `ts-integration-tests` on the branch for real-crypto / epoch-boundary
  validation (watching to verdict).

### DIFF
--- a/.github/workflows/system-tests-ci.yaml
+++ b/.github/workflows/system-tests-ci.yaml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Install Sui from Github
         run: |
-          curl -L https://github.com/MystenLabs/sui/releases/download/mainnet-v1.76.1/sui-mainnet-v1.76.1-ubuntu-x86_64.tgz > sui.tgz
+          curl -L https://github.com/MystenLabs/sui/releases/download/mainnet-v1.77.2/sui-mainnet-v1.77.2-ubuntu-x86_64.tgz > sui.tgz
           tar -xvzf sui.tgz
           sudo cp ./sui /usr/local/bin/sui
 
--- a/.github/workflows/ts-ci.yaml
+++ b/.github/workflows/ts-ci.yaml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Install Sui from Github
         run: |
-          curl -L https://github.com/MystenLabs/sui/releases/download/mainnet-v1.76.1/sui-mainnet-v1.76.1-ubuntu-x86_64.tgz > sui.tgz
+          curl -L https://github.com/MystenLabs/sui/releases/download/mainnet-v1.77.2/sui-mainnet-v1.77.2-ubuntu-x86_64.tgz > sui.tgz
           tar -xvzf sui.tgz
           sudo cp ./sui /usr/local/bin/sui
 
--- a/.github/workflows/ts-integration-tests.yaml
+++ b/.github/workflows/ts-integration-tests.yaml
@@ -122,11 +122,11 @@ jobs:
         with:
           version: "latest"
 
-      # The Cargo workspace pins Sui to mainnet-v1.76.1; a mismatched local
+      # The Cargo workspace pins Sui to mainnet-v1.77.2; a mismatched local
       # sui completes DKG but stalls reconfiguration, so pin the binary too.
-      - name: Install Sui mainnet-v1.76.1
+      - name: Install Sui mainnet-v1.77.2
         run: |
-          curl -L https://github.com/MystenLabs/sui/releases/download/mainnet-v1.76.1/sui-mainnet-v1.76.1-ubuntu-x86_64.tgz > sui.tgz
+          curl -L https://github.com/MystenLabs/sui/releases/download/mainnet-v1.77.2/sui-mainnet-v1.77.2-ubuntu-x86_64.tgz > sui.tgz
           tar -xzf sui.tgz
           mkdir -p "$HOME/.local/bin"
           cp ./sui "$HOME/.local/bin/sui"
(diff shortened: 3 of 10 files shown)


omersadika:
## Review: verified against upstream at both tags — no correctness findings, one checklist gap

This bump was reviewed verification-first: every load-bearing rev, parameter position, and mirrored constant was checked against Sui at `mainnet-v1.76.1` and `mainnet-v1.77.2` (blobless-clone diff of the two tags), not taken from the diff on faith.

### Verified correct

- **Pins & locks.** Root `Cargo.toml` and `Cargo.lock` resolve to a single Sui rev (`51d177ad` at the new tag); single `sui-sdk-types` (`5781b718`, exactly what the tag depends on); single `fastcrypto`; **zero collateral crates.io churn** — only Sui-family crates moved. No stale `1.76.1` references anywhere (the one remaining string is a historically-accurate "since mainnet-v1.76.1" comment). `sdk/ika-wasm/Cargo.lock` contains no Sui/Mysten deps, so its non-regeneration is vacuously fine. `scripts/check-sui-version-consistency.sh` passes on the PR head.
- **simtest.** `cargo-simtest` and `config-patch` both pin mysten-sim `6b8e1e8c`, which equals Sui's own `scripts/simtest/cargo-simtest` rev at the tag AND Sui's workspace `msim` pin — the "two mysten-sim copies / no reactor running" trap is closed. `msim-tokio` at that rev is `1.52.1`, equal to the root `tokio = "=1.52.1"` pin, so the crates-io patch cannot no-op.
- **The `None` argument** in `consensus_manager/mod.rs` corresponds to 1.77.2's new `transaction_pool: Option<Arc<dyn TransactionPool>>` parameter (after `transaction_verifier`, before `commit_consumer`); upstream itself passes `None` there, and `None` selects the `TransactionConsumerPool` wrapper around the `TransactionClient` channel — behaviorally identical to 1.76.1 for ika's submit path. The added comment is accurate.
- **Fork-mirror coverage.** Upstream's entire `sui-core/src/consensus_manager/mod.rs` delta between the tags is that single line; `to_consensus_protocol_config` is byte-identical (still hardcodes `enable_v3 = false`, `300`, `12`), so the comment refreshes are correct and complete. Note the Mysticeti params ika passes come from **ika's own** protocol config (v7), so Sui adding protocol versions 131–133 cannot drift them.
- **Commit/round semantics ika relies on** (freeze + epoch-close): the `consensus/core` delta is `TransactionPool` plumbing plus graceful-shutdown hygiene (`CommitFinalizerHandle::stop`, drain-before-exit). `block.rs`, `commit_consumer.rs`, and the whole `consensus/config` crate are untouched between tags; `TransactionVerifier` is unchanged; every symbol ika imports (`CommittedSubDag`, `CommitConsumerMonitor`, `BlockStatus`, `BlockAPI`, `VerifiedBlock`, `CommitDigest`) has an unchanged definition. The sui-core `consensus_handler`/`consensus_adapter`/`consensus_validator` deltas are Sui-execution-specific with no ika-fork obligation.

### One finding (maintenance, in this PR's scope)

`dev-docs/conventions/sui-version-bump.md`, checklist item 3 lists only `ts-integration-tests.yaml` and `ts-ci.yaml` as sui-binary-download locations — but this PR also had to bump `.github/workflows/system-tests-ci.yaml`. The checklist governing exactly this change type omits one of its own locations; a future bumper following it discovers the gap only when `check-sui-version-consistency.sh` fails CI (or worse, if that workflow ever leaves the script's list, ships a mismatched binary — the completes-DKG-but-stalls-reconfiguration trap). Suggest adding the missing line here, since docs are part of the change.

### Merge gate

The mixed-binary `v131_rollout` upgrade job (plus the other heavy suites) should be green before merge — for a bump that swaps the consensus crates, that run is the actual wire-compat evidence.

Issue #1719, opened 2026-05-24
IkaClient.initialize() fails on mainnet — BCS decode "Invalid typed array length: 32"


t3hrache1nn:
# Title: IkaClient.initialize() fails on mainnet — BCS decode "Invalid typed array length: 32"

**Repo to file at:** https://github.com/dwallet-labs/ika (verified via `@ika.xyz/sdk@0.4.1` `package.json` field `repository.url = "https://github.com/dwallet-labs/ika"`, directory `sdk/typescript`)

**SDK version:** `@ika.xyz/sdk@0.4.1` (latest on npm — 12 versions published, this is `latest` dist-tag)
**Sui mainnet checkpoint at repro:** 279223505
**Node version:** v24.15.0
**OS:** Windows 11
**`@mysten/sui` resolved version (in same install tree):** 1.45.2
**`@mysten/bcs`:** resolved transitively via `@ika.xyz/sdk` dep `^2.0.2`

## Symptom

`IkaClient.initialize()` against Sui mainnet (with the network config returned by `getNetworkConfig('mainnet')`) throws `NetworkError: Network error: Failed to fetch objects`. The underlying cause, surfaced via `error.cause.message`, is `Invalid typed array length: 32` — a `RangeError` from `new Uint8Array(32)` being passed a corrupted length argument during BCS deserialization of the Ika system objects.

The same package IDs and object IDs returned by `getNetworkConfig('mainnet')` resolve cleanly via the underlying `SuiClient` — both `suiClient.getObject({ id, options: { showContent: true } })` and `suiClient.core.getObjects({ objectIds, include: { content: true } })` return the live objects without error. The failure is isolated to the SDK's BCS-decode step inside `fetchObjectsFromNetwork_fn`.

## Stack trace

```
NetworkError: Network error: Failed to fetch objects
    at IkaClient.fetchObjectsFromNetwork_fn (file:///C:/inn/phase-22-dwallet/node_modules/@ika.xyz/sdk/dist/esm/client/ika-client.js:707:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async IkaClient.getObjects_fn (file:///C:/inn/phase-22-dwallet/node_modules/@ika.xyz/sdk/dist/esm/client/ika-client.js:634:20)
    at async IkaClient.ensureInitialized (file:///C:/inn/phase-22-dwallet/node_modules/@ika.xyz/sdk/dist/esm/client/ika-client.js:138:5)
[...]

PR #2053, opened 2026-08-15
fix(sdk): document the gRPC client and probe the live fullnodes in CI (#2052)

### LINKED ISSUE #2052: SDK: IkaClient.initialize() fails on public Sui fullnodes — JSON-RPC transport deprecated, migrate to gRPC/GraphQL
Supersedes #1719, whose reported symptom is now unreachable behind a harder failure.

## The out-of-box SDK is broken against public Sui mainnet fullnodes

Sui has deprecated JSON-RPC on public fullnodes. `IkaClient.initialize()` with the published SDK and its documented setup now fails before doing anything useful:

```
INITIALIZE: FAIL: Network error: Failed to fetch objects
cause: Method not found. JSON-RPC on public fullnodes has been deprecated.
       Please migrate to gRPC or GraphQL endpoints.
       See https://docs.sui.io/develop/accessing-data/json-rpc-migration
```

Reproduced 2026-08-15 with `@ika.xyz/sdk@0.4.1` (the npm `latest`, unchanged since #1719 was filed in May) + `@mysten/sui@1.45.2` (the version pinned in #1719's environment) against `getFullnodeUrl('mainnet')`.

Two compounding problems:

1. **Transport**: the SDK's object-fetch path speaks JSON-RPC; every user following the README against a public fullnode gets `Method not found`. Users with private JSON-RPC endpoints still work, which will make this look intermittent in reports.
2. **Dependency drift**: with a *current* `@mysten/sui` install, the classic import (`import { SuiClient, getFullnodeUrl } from '@mysten/sui/client'`) no longer resolves — the module layout changed — so "just upgrade the peer dep" is not a fix either; the SDK needs adapting to the new client surface (which is also where the gRPC transport lives).

## What #1719 reported, for continuity

A BCS decode error (`Invalid typed array length: 32`) inside `fetchObjectsFromNetwork` while the same objects fetched fine via raw `SuiClient` — i.e. an SDK-side deserialization bug. That may well still exist underneath; it is simply unverifiable until the transport works again. When migrating, re-test `initialize()` end-to-end on mainnet and either confirm the decode bug fixed or re-file it with the new transport.

## Suggested scope

- Migrate `IkaClient`'s reads to the gRPC (or GraphQL) client surface of current `@mysten/sui`; update the peer-dependency range accordingly.
- CI: an integration probe that runs `initialize()` against live mainnet + testnet public endpoints, so transport rot is caught at PR time rather than by users (this exact class has now shipped broken for 3+ months unnoticed).
- Publish a new npm release; 0.4.1 has been `latest` since before May.

Iamknownasfesal:
Closes #2052.

## What was actually broken

The issue reports that `IkaClient.initialize()` fails against public Sui fullnodes and concludes the SDK's object-fetch path speaks JSON-RPC. It does not. `IkaClient` takes a `ClientWithCoreApi` and has been transport-agnostic since #1655, with a gRPC-specific fix already landed in #1704.

Verified against the **published** `@ika.xyz/sdk@0.4.1` + `@mysten/sui@2.20.1`:

| Setup | Result |
| --- | --- |
| `SuiGrpcClient` | `initialize()` OK — mainnet epoch 381, testnet epoch 384 |
| `SuiJsonRpcClient` (what our docs told users to build) | `Method not found. JSON-RPC on public fullnodes has been deprecated` |

So the out-of-box breakage is real, but it lives in the documentation: every setup snippet we ship — docs site, SDK README, and both the `ika-sdk` and `ika-move` skills — constructed a JSON-RPC client. Those are switched to `SuiGrpcClient`, with a note that JSON-RPC remains valid for a localnet or a private fullnode.

The reporter's second point (the `SuiClient`/`getFullnodeUrl` import failing to resolve) is the v1 → v2 rename; they tested with `@mysten/sui@1.45.2` while the published SDK already requires `^2.5.0`.

## The failure the issue could not see past

Probing further with the published SDK over a working transport surfaces a second, worse break on **both** networks:

```
Error: invalid value: integer `3`, expected variant index 0 <= i < 2
  at reconfiguration_public_output_to_protocol_pp
```

`@ika.xyz/ika-wasm@0.2.1` (Nov 2025, still the newest on npm) knows two variants of the reconfiguration output; live protocol v7 emits V4. Object reads succeed, so a dApp only discovers this when it reaches a real dWallet operation.

This is purely a release gap — `crates/dwallet-mpc-centralized-party/src/lib.rs:1211` handles V4 correctly. Rebuilding the WASM from this revision and re-running the probe against live mainnet and testnet passes end to end, protocol public parameters included. **The fix therefore requires publishing a release; no further code change.** The pending changeset already bumps both packages, so a tag is all that is missing.

## Preventing the next silent rot

Both breakages shipped for months because every existing suite runs against a localnet pinned to what the repo already expects. `sdk/typescript/test/live/public-endpoints.test.ts` probes the real thing — `initialize()`, an epoch read, and protocol-public-parameter derivation against public mainnet and testnet over gRPC. No funds, no keypair, no localnet; it only reads. It runs daily and on PRs touching the SDK, WASM, or centralized party.

An unreachable fullnode **skips** the assertion rather than failing it, so a Sui outage does not red-build an unrelated PR, while anything the SDK itself raises is a hard failure.

### The probe was fault-injected rather than trusted green

| Injected fault | Probe behavior |
| --- | --- |
| Published 0.2.1 WASM swapped in | **Fails** — `invalid value: integer 3`, the exact user-reported error |
| `SuiJsonRpcClient` instead of gRPC | **Fails** — `Method not found` |
| Unresolvable host | **Skips** with a note; the other network still asserts |

## Also

- The pending changeset described the WASM decode failure as a risk to prepare for before protocol v4 activates. It is live breakage against both networks today, and the variant that trips it is V4, not V3. Corrected, since that text becomes the published changelog entry for exactly this bug.
- #1719's `Invalid typed array length: 32` decode error did not reproduce anywhere in these runs against either network.

## Left for a maintainer

Publishing. `0.4.1` has been `latest` since 8 May 2026 and is unusable against mainnet regardless of transport.


Iamknownasfesal:
## Release prep added — and the reason nothing shipped since May

Attempting `pnpm changeset-version` failed outright:

```
🦋  error The package or glob expression "ika-wallet" is specified in the `ignore` option
    but it is not found in the project.
    ... (the same error for all nine entries)
```

Every entry in `.changeset/config.json`'s `ignore` list is a package inherited from the Sui fork — `ika-wallet`, `@mysten/core`, `kiosk-demo`, `mev-bot`, and so on — and changesets treats an unknown ignore entry as a hard error, not a warning. So the release staged in the spring was **unappliable, not merely unapplied**. That is the mechanical reason npm `latest` has been `0.4.1` since 8 May while `main` moved on to protocol v7.

Emptied the list and applied the release it was holding:

| Package | From | To |
| --- | --- | --- |
| `@ika.xyz/sdk` | 0.4.1 | **0.5.0** |
| `@ika.xyz/ika-wasm` | 0.2.1 | **0.3.0** |

The WASM bump is the one that unblocks users: `0.2.1` cannot decode the V4-tagged reconfiguration outputs both networks now write, so every dApp on `0.4.x` fails at its first dWallet operation regardless of transport.

Verified at these versions: `pnpm build` clean, and the live-endpoint probe passes against public mainnet and testnet (4/4, none skipped).

## What publishing now takes

Merge, then push the tag — the existing workflow does the rest, and its guard skips any version already on npm:

```
git tag sdk/typescript-0.5.0 && git push origin sdk/typescript-0.5.0
```

I have deliberately not pushed that tag: it publishes publicly to npm under a version number that cannot be reused. That call is yours.

omersadika:
## Review: docs/release half is right and verified — but the probe's skip/fail oracle is miscalibrated in both directions

The review ran verification-first: the classifier findings below were reproduced **empirically** against the real `@mysten/sui@2.20.1` gRPC transport using fault-injecting local HTTP servers, not inferred from reading.

### Verified correct

- The diagnosis holds end to end: public fullnodes are gRPC-only, every shipped snippet (docs site, README, both skills) built a JSON-RPC client, and the V4-decode failure (`invalid value: integer 3, expected variant index 0 <= i < 2`) matches live protocol v7 emitting V4-tagged reconfiguration outputs against a published WASM that knows V1/V2. Placing the fix in *publishing* rather than code is correct — `dwallet-mpc-centralized-party` already handles V4.
- Release prep is coherent: changeset consumed, sdk `0.5.0` / ika-wasm `0.3.0`, and the changelogs corrected the stale changeset's "V3" language to the accurate V4 story.
- `vitest.config.ts` properly fences `test/live` out of every localnet-backed run, with a comment.
- Refuted along the way (so nobody re-litigates them): the Sui binary in the workflow IS needed (`pnpm install` → `prepare` → `sui client … && sui-ts-codegen generate`); the release build fits the 45-min cap (~3m15s observed on this branch); the changeset `config.json` change is benign; the vitest `--dir`/exclude interaction works.

### The probe's classifier fails BOTH ways it promises not to (empirically reproduced)

`isEndpointUnavailable` does raw `includes()` over joined error text. Two consequences, each defeating one of the workflow's two stated guarantees:

1. **It skips real SDK defects.** Bare-digit markers (`502`/`503`/`504`) and the broad word `unavailable` match inside genuine SDK errors: `errors.ts` builds messages embedding 64-hex object IDs (`${objectType} object with ID ${objectId} not found`; `Failed to read table vector as raw bytes: ${tableID}`), and an embedded hex ID contains one of those digit-runs with ~4.4% probability. A real decoder/state regression then classifies as a Sui outage: the test skips, the daily job stays green, and the months-long blind spot this PR was written to close is recreated.
2. **It hard-fails routine transport outages.** Against the real `GrpcWebFetchTransport`: HTTP 502 surfaces as `RpcError('Bad Gateway', UNKNOWN)` — no marker matches → hard fail. A 504's rescuing `DEADLINE_EXCEEDED` code name lives only in `RpcError.toString()`, which the classifier never sees because `String(error)` runs on `IkaClient`'s outer `NetworkError` wrapper (every probe path wraps once: ika-client.ts:1044/1151/1240/1297) → wrapped 504 hard-fails. A 503 with an empty HTTP/2 reason phrase yields an empty message → hard fail. Mid-stream disconnects (`terminated`, `premature EOF`, `missing trailers`) and `RESOURCE_EXHAUSTED` rate-limits match nothing → hard fail. A load-balancer blip on `fullnode.testnet.sui.io` reds every PR touching `sdk/typescript/src/**`.

**One fix addresses both:** classify on structured codes — `RpcError.code` (walking the `cause` chain through the `NetworkError` wrapper) and syscall codes on the cause — never message substrings.

### Supporting findings

3. **No client-side deadline at any layer** (`SuiGrpcClient` built without a timeout, `@mysten/sui` sets none, vitest's testTimeout is 100 min). A fullnode that accepts connections but never sends headers is bounded only by undici's ~300s header timeout; with 3 retries + backoff ≈ 15 min per test × 4 sequential tests ≈ 61 min, killed at `timeout-minutes: 45` — an outage-driven red job, exactly what the comments promise to avoid. Pass a transport timeout (`GrpcWebOptions.timeout`) or a per-file vitest timeout well under the job cap.
4. **The new `mainnet-v1.76.1` pin is registered nowhere**: `scripts/check-sui-version-consistency.sh`'s `check_file` list ends at `system-tests-ci.yaml`, and `dev-docs/conventions/sui-version-bump.md` doesn't list this workflow. Since the binary is genuinely used, the next Sui bump (#2049, in flight at v1.77.2) leaves this workflow silently on the old tag with CI green. Add the file to the script AND the checklist in this PR; whoever merges second reconciles the pin, enforced.
5. **The `pull_request` paths filter omits the probe's own files** (`sdk/typescript/test/live/**`, `vitest.config.ts`, `package.json`). A PR editing `public-endpoints.test.ts` — e.g. loosening the skip oracle itself — merges with zero signal from the only job that runs it.
6. **The probe never exercises the published artifact.** Both scheduled and PR runs build the WASM from HEAD, but the motivating incident lived in npm's `0.2.1` — HEAD handled V4 while the published package did not, which is precisely the state a HEAD-built probe reports green. The scheduled run should additionally install `@ika.xyz/sdk@latest` into a scratch project and run the same probe; that also covers publish-pipeline defects (files whitelist, exports map, dependency ranges) that exist only in the tarball.
7. **`examples/` still ships both documented breakages**: keyspring pins `@ika.xyz/sdk ^0.2.3` and builds a JSON-RPC `SuiClient` against hard-coded public endpoints (`backend/src/dkg-executor.ts:103`); multisig-bitcoin pins `^0.2.2` resolving the V4-blind WASM. A user cloning an example hits, unmodified, exactly the two failures this PR's changelog documents as broken-today.
8. **ika-wasm's changelog talks about a different package**: under `## 0.3.0` it opens "Release 0.5.0" and warns "0.4.x is broken", but ika-wasm went 0.2.1 → 0.3.0 and never had a 0.4.x; it also documents SDK-only APIs. A direct consumer reading it concludes the mandatory-upgrade notice doesn't apply to their 0.2.x install. Split the shared changeset text per package.
9. **The gRPC URL is a hand-copied literal in 7 files** (probe, README, 2 docs pages, 3 skill files), and each snippet now carries two literals (`network` + `baseUrl`) that must agree — a mismatch (mainnet network, testnet baseUrl) produces object-not-found errors that look like SDK bugs, a misconfiguration the old single-token `getFullnodeUrl('testnet')` made impossible. Export a `getGrpcFullnodeUrl(network)` helper (or a field on `getNetworkConfig`) and consume it everywhere, probe included.
10. **The probe's `T | undefined` return channel is dead code**: vitest's `ctx.skip()` throws (typed `never`, verified on 4.1.9), so `return undefined` after `skip()` never executes; the `=== undefined` guards in both tests and `withRetries`' trailing `throw` are unreachable. Type `skip` as `(note?: string) => never`, return `Promise<T>`, delete the three dead lines — otherwise every future test copies a guard that guards nothing.

### Verdict

The documentation/release half can ship as-is. The probe needs its classifier (structured codes), a deadline, self-coverage in the paths filter, and the consistency-script registration before the guard can be trusted — as written it can both recreate the blind spot it exists to close (finding 1) and red unrelated PRs on routine outages (findings 2–3).

Iamknownasfesal:
Thanks — the classifier findings are correct, and both directions were reproducible. Fixed in 4148919, 847fc66, c51693e, e1d0652. Two places where I went somewhere different from the suggested fix, and two I'm declining, below.

## Findings 1–2: the classifier — fixed, but not by enumerating codes

Both directions confirmed. I took the structural route instead of `RpcError.code`, because code enumeration still requires getting the list right and 502→`UNKNOWN` shows how easily that goes wrong. On failure the probe now makes an independent `getServiceInfo` call and only skips if *that* fails. No error shape needs naming, and an outage cannot mask a defect.

Fault-injected both ways:

| Injected | Before | Now |
| --- | --- | --- |
| SDK error `object with ID 0x503ab not found: service unavailable (503)` | would skip (silent green) | **fails in ~2s** |
| Published 0.2.1 WASM | fails | **fails** (unchanged) |
| Hung endpoint (accepts TCP, never answers) | ~61m, killed at cap | **skips**, file completes in 8m19s |

## Finding 3: deadline — fixed, though the suggested mechanism wouldn't have worked

`GrpcWebOptions.timeout` is not a client-side deadline: it only sets the `grpc-timeout` *request header* (`grpc-web-format.js:41`), which a wedged fullnode is free to ignore. And it never reaches the transport anyway — `SuiGrpcClient` forwards only `baseUrl`/`fetchInit` (`grpc/client.ts:82`), dropping everything else. So the bound is imposed in the probe: 60s per attempt, 10s per reachability check. Worst case with both networks hung is ~13.6 min against the 45-min cap, measured rather than estimated.

## Findings 4–6: taken as written

- **Sui pin** registered in `check-sui-version-consistency.sh` and the bump checklist (which was also missing `system-tests-ci.yaml`). Verified the script now reports it.
- **Paths filter** widened to `test/live/**`, `vitest.config.ts`, `package.json`.
- **Published artifact**: new scheduled-only `published` job installs `@ika.xyz/sdk@latest` and runs the same reads. It **fails today** — `latest` resolves 0.4.1 with the 0.2.1 WASM — which is the honest signal; it goes green when the release publishes. Good catch: it demonstrates the blind spot live, since the source probe is green in the same run.
- **ika-wasm changelog** rewritten in terms of that package.

## Declining two

**The `getGrpcFullnodeUrl` helper.** The footgun is real, but a new public export belongs in its own change, not in a release whose changelog is already cut — and Mysten deliberately ships no gRPC URL helper, so we'd be taking on an endpoint list that changes on their schedule. Happy to do it as a follow-up if you'd still like it.

**`withRetries`' trailing `throw`** was called unreachable along with the `T | undefined` channel. The channel is gone — `skip()` throws `PendingError`, verified — but a `for` loop's trailing throw is required for TypeScript's control-flow analysis; deleting it fails the build. The loop is now `for (;;)`, so the question doesn't arise.

## examples/ — split out to #2055

Correct that they're broken, with one correction: the universal breakage is the V4-blind WASM, not the transport. Verified by direct request that `sui-testnet-rpc.publicnode.com` (keyspring testnet) **still answers JSON-RPC**, while keyspring's mainnet endpoint answers nothing and multisig-bitcoin inherits its client from dapp-kit. Transport breakage is per-endpoint; the WASM breakage is unconditional.

Not fixed here because it isn't a pin bump: `0.2.x → 0.5.0` drags `@mysten/sui` v1 → v2 across 16 files in two apps, and it can't even be verified until 0.5.0 is on npm.

Iamknownasfesal:
## Merged main — and the pin registration paid off immediately

The PR went `CONFLICTING` while the review fixes were being pushed: #2049 (Sui → mainnet-v1.77.2) landed on main first. Worth noting, since it's the review's finding 4 playing out live rather than hypothetically:

```
ERROR: .github/workflows/sdk-live-endpoints.yaml references Sui tag(s)
       [mainnet-v1.76.1 ] != mainnet-v1.77.2
```

That's `check-sui-version-consistency.sh` failing on the first merge attempt, exactly as predicted for whichever side merged second. Had the workflow not been registered in c51693e, this branch would have merged carrying a v1.76.1 binary with every check green. Pin updated; the script passes.

Also of note: main independently made the *same* changesets fix (`"ignore": []`), so that conflict resolved to identical content either way. main did not cut the release — it's still 0.4.1/0.2.1 there — so there's no double bump.

Conflict resolution in the bump checklist keeps both sides: main's expanded simtest entry, and this branch's workflow list plus the enforced-source-of-truth note.

Re-verified on the merged tree with the WASM rebuilt against the new Sui revision: live probe 4/4 against public mainnet and testnet, lint and format clean. Checks are green or running on `14e4c69`, with `published` correctly skipping on pull requests.

I used a merge rather than a rebase deliberately, so the commit SHAs cited in the review reply above still resolve.

PR #2060, opened 2026-08-16
Use standalone Sui Rust SDK for client and transaction tooling

omersadika:
## Summary

- migrate Sui RPC, execution waiting, faucet calls, signing, and wallet configuration to `MystenLabs/sui-rust-sdk`
- finalize existing core PTBs through the standalone transaction builder
- retain Sui-main only for Move publication metadata, test wallets, consensus, and deployed core types
- document the remaining dependency boundaries

## Testing

- `cargo test --offline --release -p ika-sui-client --lib`
- `cargo check --offline -p ika-sui-client -p ika-swarm-config -p ika -p ika-upgrade-test --all-features`
- strict Clippy with `-D warnings`
- `./scripts/check-sui-version-consistency.sh`


omersadika:
## Review

Multi-pass review: six finder angles (correctness ×2, removed-behavior, efficiency, reuse/simplification, conventions) plus a deep consensus/trust-model pass, with SDK behavior verified against the pinned sui-rust-sdk rev (`5781b718`) and old behavior against the pinned sui checkout. The migration's *read* side is solid — see "verified clean" below. The *write* side is not mergeable as it stands.

### Blocking

**1. The wallet signs whatever transaction bytes the endpoint echoes back** — `transaction_builder.rs:28-38` → SDK `build()`. The SDK sends the resolved PTB in a `SimulateTransactionRequest` and returns the finalized transaction **deserialized from the response**; `construct_unsigned_txn` signs it with no comparison against the locally-built PTB. The old path built `TransactionData` locally — a malicious endpoint could at worst supply bad object refs that fail at execution; it could not redirect asset flow. Now a compromised third-party gRPC provider (a configuration the OCS spec explicitly supports with authenticated endpoints) can return a transaction that transfers the `ProtocolCap`/`ValidatorCap` or drains the address, and the CLI signs and submits it. This inverts the premise the whole OCS system is built on — reads are committee-verified precisely because "trusting a fullnode's word means trusting whoever serves the read"; signing on that word is strictly worse. Affects every constructor in `ika_validator/protocol/dwallet_transactions.rs` plus `create_sui_transaction` in swarm-config. **Fix**: assert the echoed PTB's commands and non-gas inputs equal the locally replayed ones (only gas metadata and object-input resolution may differ), or keep building `TransactionData` locally and use simulation for validation/gas only.

**2. The notifier's submission deadline was deleted, and the replacement comment claims a bound that does not exist** — `grpc.rs:1048-1059`. Old code wrapped submission in `tokio::time::timeout(SUBMIT_TIMEOUT, …)` with a comment naming the exact hazard (accepted-but-never-answered request, keepalives keep the channel "healthy"). The new comment says "Bounded by `execute_transaction_and_wait`" — verified false in the SDK source: `SUBMIT_TIMEOUT` bounds only the post-execution checkpoint-scan `select!`; the execution RPC arm has **no deadline** (the SDK's own docs: "No default bounds the total duration of a call"; the response-headers timeout is opt-in and not set). The notifier holds its global `notifier_tx_lock` across this call, and epoch close depends on notifier submissions — a never-answering upstream silently wedges every Sui write and the epoch boundary with no timeout, no metric, no log. This is the #1892 outage class, minus the logging it had. **Fix**: restore an outer timeout (sized above the checkpoint wait) and/or set a per-request/response-headers deadline on the writer client.

**3. Dropped gas-budget estimation breaks validator MPC-data flows at the CLI default** — `ika_validator_transactions.rs:673`. Old: `estimate_gas_budget(...).unwrap_or(gas_budget)` — the 0.2 SUI CLI default never bound for real transactions. New: the fixed budget is authoritative. `become-candidate` / `set-next-epoch-mpc-data` store hundreds of KB of class-groups data (~7,600 MIST/byte → 0.2 SUI is exceeded past roughly 26 KB): these now fail unless the operator hand-passes `--gas-budget`. CI is structurally blind to it — cluster tests route through the swarm path whose default is 5 SUI. Restore estimation (the builder's simulation already computes what's needed) or raise the defaults with a justification.

**4. Validation is below the repo's own bar, and the new pins are documented as CI-enforced but aren't** — the PR's evidence is offline lib tests + checks; CLAUDE.md's minimal-verification table requires the cluster suite for this layer, and a brand-new signing path (`SdkTransactionContext`) has never touched a chain in any test here (a digest mismatch bricks every CLI transaction — loudly, but unproven). Also verified directly: `Cargo.toml` pins four sui-rust-sdk crates at one rev and both modified/added convention docs claim `check-sui-version-consistency.sh` enforces the pin set — the script contains **zero** sui-rust-sdk references. That's the silent-partial-bump class the script exists to prevent (and the docs' own warning text: "can compile while silently desynchronizing"). Extend the script in this PR; dispatch cluster + TS integration + one localnet CLI transaction before merge.

### Should fix

- **Checkpoint-inclusion coupling on the serial writer** (`grpc.rs:189-201, 1048`): even healthy-case, every notifier write now opens a per-tx `SubscribeCheckpoints` stream and blocks the lock until checkpoint inclusion (~1/(execution+inclusion) throughput); and `CheckpointTimeout`/`CheckpointStreamError` — which *carry the successfully executed response* — are collapsed into `TransportError::Network`, so a committed transaction is reported failed, its effects discarded, and the caller retries/rebuilds. Return-at-effects for the notifier (keep wait-for-checkpoint only where a caller reads `checkpoint`/`timestamp_ms`), and surface the executed effects on checkpoint-wait failures.
- **Every build error is now `IkaError::DryRunFailed`** (`ika_validator_transactions.rs:683`) — and `set_next_epoch_mpc_data_bytes:1334` branches on that error *semantically* (switches to the `destroy_none` PTB variant). A transient connect failure now silently triggers the wrong rebuild and masks the real fault. Narrow the mapping to actual simulation failures.
- **`basic_auth` support silently dropped**: migrated connect sites send empty headers and the new `ClientConfig` deserializer ignores the field — authenticated fullnodes get 401 on every CLI transaction command that previously worked.
- **Wallet-loading narrowed, breaking read-only commands** (`transaction_context.rs:107,146`): construction requires a File keystore, aborts on the *first* unparseable keystore entry (`collect::<Result>` over all keys), and hard-errors when `active_address` has no local key — external-signer/ledger/multisig/InMem configs that `WalletContext` served fine now fail at CLI startup even for `dwallet get`. The file's own docstring claims those cases "remain on the WalletContext adapter", but `CliContext` no longer constructs one — the claimed fallback doesn't exist. Defer key lookup to signing time.
- **Faucet regressions** (`faucet.rs:27`, `workload.rs:429`): `normalize_faucet_url` turns a `/v1/gas` URL into `<root>/v1/` + `/v2/gas` = `/v1/v2/gas`; and the workload's documented tolerance of the localnet faucet's error-shaped 200s was dropped, so benign flakes now abort upgrade-test runs.
- **Swarm gas-smashing removed** (`sui_client.rs:1911`): the deleted pay-with-all-coins path carried a comment documenting the exact dust-fragmentation failure it fixed; server-side gas selection may reintroduce it in long scenarios, and the dust now never re-aggregates. Test-tooling only, but the tombstoned flake was real.
- **`ObjectInput` kind left server-resolved** (`transaction_builder.rs:128-131`): `kind: None` lets the endpoint pick — compounds blocking finding 1; `ObjectInput::owned(id, version, digest)` is a one-liner.
- **`simulate_transaction` is dead and broken** (`grpc.rs:207`): its read mask uses `ExecutedTransaction`-relative paths against a response-root-relative mask, so any future caller gets a guaranteed decode error. Fix the paths or delete the method.

### Cleanups (condensed from the efficiency/reuse/simplification angles)

- `SuiGrpcClient::connect` is inlined at **49 sites**; single logical operations open 3–5 fresh channels (and `dwallet_commands.rs`'s `create_grpc_client` uses the *probing* constructor, adding a `GetServiceInfo` round-trip per helper call). One provided `grpc_client()` on `TransactionContext`, threaded through, fixes latency and gives header/timeout changes a single home.
- Hand-rolled duplicates of things that already exist: `proto_value_to_json_value` (renders every number as f64 — `7` becomes `7.0`, diverging from the SDK's safe-integer serializer in test comparisons); the hardcoded `UpgradeCap` type string vs `UpgradeCap::type_()`; a private serde mirror of `client.yaml` while the same repo *writes* those files with `SuiClientConfig` (reader/writer schema drift); string-`parse()` conversions where the sui-types↔sdk `From` impls the PR itself uses elsewhere exist; `workload.rs`'s second hand-rolled builder pipeline beside `build_transaction_data`; `list_owned_gas_coins`/`list_dynamic_fields` re-inlining their own helpers; `get_object_ref` fetching full object BCS for three scalars (inherited, flagged as such).
- `&mut impl TransactionContext` on ~50 signatures for an entirely-`&self` trait; a dead `timestamp_ms()` getter; a vestigial always-empty `gas_payment` parameter enforced by runtime `ensure!`.
- CLAUDE.md conventions: function-local `use` statements and inline fully-qualified paths in `transport.rs`, `grpc.rs`, `transaction_context.rs`, `transaction_builder.rs`, `workload.rs` — new files have no legacy excuse.

### Verified clean

The read-side contract survived intact, checked site-by-site: the #2037 pruning-immune watermark probe; **every** `TransportError::NotFound` production site (each reproduces the old client exactly — including the epoch-record mapping whose regression would recreate the 2026-07 MissingCommittee spin, still covered by its test); the 128 MiB checkpoint decode limit; the genesis chain-identifier trust root (#1875) untouched; signing-intent byte-compatibility at the pinned rev with keystore round-trip tests; PTB replay fidelity (input order, command order, shared-object mutability carried explicitly, `NonExclusiveWrite` rejected loudly); all four SDK crates at a single rev with a single copy in Cargo.lock; and zero consensus/`dwallet_mpc`/checkpoint-content code touched.

**Bottom line: do not merge as-is.** Findings 1 and 2 are the write-path trust model and the notifier wedge-with-a-false-comment; 3 breaks documented validator operations at defaults CI can't see; 4 is the repo's own evidence bar. The read-side migration underneath is good work and the dependency-boundary doc is genuinely useful — the fixes are localized.

omersadika:
**Addendum — verification pass completed after the review above; two corrections and three escalations.**

Corrections (claims above now refuted by dedicated verification):
- **Swarm gas-smashing (should-fix #7): retracted.** The multi-coin regression claim did not survive verification; treat the `create_sui_transaction` note (vestigial `gas_payment` param) as the only remaining item there.
- **Keystore parse-abort detail: retracted.** The confirmed failure is narrower but still real: `from_sui_client_config` drops upstream `SuiClientConfig`'s `external_keys` support (the field is silently ignored) and hard-requires the active address to derive from a local file-keystore key — so external-signer/multisig configs that `WalletContext` accepted now fail at CLI startup for every command including pure reads. The 'aborts on first unparseable entry' mechanism was wrong; the read-only-commands breakage stands.

Escalations (now CONFIRMED server-side, stronger than stated above):
- **Gas budget (blocking #3)**: verified in Sui's simulate handler — a caller-supplied budget is taken as-is (`budget_was_estimated` only when none is set). The ~32 KB class-groups payload costs **~0.24–0.27 SUI in storage alone** vs the 0.2 SUI CLI default: `become-candidate`/`set-next-epoch-mpc-data` fail at defaults, deterministically.
- **Checkpoint-lag misreport (should-fix #1)**: the consequence chain ends harder than 'retry churn' — `sui_executor` retries the falsely-failed (but committed) transaction for up to an hour and then **panics** the validator.
- **Faucet `/v1/gas` (should-fix #6)**: confirmed the pinned sui-faucet actually serves `/v1/gas`, so the mangled `/v1/v2/gas` 404 hits a documented real endpoint form, not a hypothetical. Two now-dead '200 OK' error guards also survive in `ika-swarm-config/src/sui_client.rs` (~124-141, ~319-328) and should be deleted.

The four blocking findings are unchanged (echoed-bytes signing, missing execution deadline, gas defaults, evidence/pin-enforcement gap).

Issue #2058, opened 2026-08-15
Event-source the epoch: rebuild all derived state by batched replay of the consensus store on every restart (drop the watermark and the per-round stream)


omersadika:
> **v2 — full event-sourcing.** This issue originally proposed keeping ika's per-round stream as a reconciled materialized view and keeping the exactly-once watermark (see edit history). Rewritten after design discussion (Omer): the watermark and the stream are both *second truths*, and the honest completion of "one source of truth" is to drop them. Delivery remains **one change**, not stages.

## Goal

**The consensus store is the only truth for epoch-scoped state. Everything derived from it is rebuilt by replay on every restart.**

Today a hard kill can leave three persistence layers disagreeing: Mysticeti's commit store, ika's per-round MPC stream, and the per-epoch watermark — and `dWallet Labs 5` was bricked on 2026-08-15 by exactly such a disagreement (#2057). Reconciling the copies (this issue's v1) shrinks the disagreement window; deleting the copies removes it. The consistency argument is structural: **deterministic derivation from the commit stream is already a hard cross-node requirement** — all validators must reach identical checkpoints and convictions from the same commits — so relying on the same property within one node across restarts adds no new assumption.

## The model

**On every restart:**
1. Wipe all *derived* per-epoch state.
2. Replay the current epoch's commits from the consensus store, **in bounded batches** (read a batch → fold it through the handler → release it → next; precedent: consensus-core's own recovery batches at 250 commits, `COMMIT_RECOVERY_BATCH_SIZE`). Memory stays flat regardless of epoch age; no full-epoch buffering, no unbounded channel.
3. Reach the store head, switch to live tail-follow.

**No watermark.** The handler processes every commit of the epoch on every boot. Exactly-once is replaced by *deterministic fold over wiped state* — there is no surviving partial state to double-apply against (the #1920/#1927 failure shape requires partial state; wipe-and-rebuild removes the category).

[...]

Issue #2057, opened 2026-08-15
Consensus store losing its unsynced tail crash-loops the validator until the next epoch boundary — clamp replay_after instead of asserting


omersadika:
## A node-level storage incident bricked a validator until the next epoch boundary — live on our own fleet right now

`dWallet Labs 5` (mainnet, v1.3.1) is crash-looping with:

```
panicked at consensus/core/src/commit_observer.rs:158:9:
assertion failed: last_commit_index > replay_after_commit_index
```

exit code 12, ~28s per attempt, since 20:23 UTC 2026-08-15. It will stay down until the epoch 381→382 boundary (~13h from onset).

## Incident chain (all timestamps UTC, from k8s events + Loki)

1. ~20:15–20:22 — the OpenEBS LVM CSI node plugin on the worker restarted (8 container restarts); kubelet reported `MountVolume.MountDevice failed ... driver name local.csi.openebs.io not found`, the pod sandbox was killed and re-created mid-write. Node recovered; **the volume was never replaced** (same PVC, 229 days old).
2. 20:23:48 — first post-incident boot, before any human intervention: `Node has not participated in previous epoch consensus. Boot counter (0) will not increment.` followed by the assertion panic. Every subsequent boot identical.

## Mechanism: one logical watermark, two databases, no crash-consistency story

Both stores live on the same volume but are separate RocksDB instances:

- the **authority per-epoch store** survived the kill with its `last_processed_commit_index` watermark intact (the node also still has its handoff cert — it logs re-anchoring on it);
- the **consensus store** lost its unsynced tail: it still has *a* last commit, but at an index **behind** the watermark ika passes as `replay_after`.

`CommitObserver::recover_and_send_commits` (pinned sui, `consensus/core/src/commit_observer.rs:140-158`) then has no tolerant path: store-empty asserts `replay_after == 0`, store-behind asserts `last_commit_index > replay_after_commit_index`. Either way: panic, and since panics abort, a crash loop.

## Why it self-heals only at the boundary

Each boot reads the current epoch from Sui chain state, so once the fleet enters epoch 382 the node builds a fresh per-epoch store (`replay_after = 0`) and a fresh consensus epoch-dir, and starts clean. Until then every boot replays the same mismatch. Cost: a validator is down for up to a full epoch (24h) after any storage incident that loses the consensus store's tail — a class of incident (node crash, power loss, CSI failure, disk-full kill) that every operator will eventually hit.

## Why clamping is safe (proposed fix)

[...]

PR #2061, opened 2026-08-16
feat(consensus): event-source the epoch — wipe derived state and rebuild it by replaying the consensus store

omersadika:
Part 1 of #2058. The consensus store becomes the only truth for epoch-scoped state: on every boot the derived per-epoch tables are wiped and rebuilt by a bounded-batch replay of the epoch's commits. The processed-commit watermark is deleted outright.

**Why this is the fix for #2057.** That incident had one logical watermark living in two databases: a storage failure cost the consensus store its unsynced tail while the per-epoch store kept a watermark ahead of it, and Mysticeti's `recover_and_send_commits` asserted rather than tolerating the disagreement — crash-looping a mainnet validator until the epoch boundary. `replay_after` now comes from `read_last_finalized_commit` on the store consensus is about to open, which makes both upstream assertions structurally unreachable. There is no second record left to disagree.

**The audit is the load-bearing part.** Every per-epoch table is classified derived (wiped and rebuilt) or preserved (not re-derivable: private MPC material, presign pools and their idempotency markers, peer-fetched outputs, operator overrides) in `dev-docs/derived-epoch-state-audit.md`, declared once in a registry that generates the wipe, the post-wipe emptiness check, and the test fixtures — one list, no second list to drift. The enforcement test that matters takes truth from behaviour: every table the real commit boundary writes must be classified derived. Fault injection confirmed the two obvious tests stay green on a misclassification while that one fails.

**Emission stays silent for settled work.** Replaying an epoch re-invokes the checkpoint signature output for checkpoints the committee settled long ago; suppression keys on the greater of the state-sync-verified watermark and this node's own aggregated certificate head — both mean "a stake quorum signed this", both converge across validators, neither records local progress. The pre-existing guard keyed on state sync alone, so a node that certified a checkpoint itself but had not synced past it re-signed; that was a real gap, not a hypothetical.

Upstream framing worth keeping: Sui is already event-sourced — it keeps a durable snapshot at boundaries it has proven safe and re-replays above it through a side-effect-free fold. This is that architecture with the snapshot removed and the replay horizon set to the epoch.

## Evidence

- Determinism (fold the same commits twice from wiped state, byte-identical), the #2057 shapes in process (tail-truncated store, wiped store), emission silence counted at peers, watchdog armed during boot replay, audit enforcement — each fault-validated: ten injections, each producing its predicted failure, reverted.
- Rust integration suite, cluster suite, TS suite, and the full-matrix upgrade test (all seven scenarios) green on CI.
- Consensus review clean across six lenses; its two findings (a stale `handoff.md`, an incomplete non-determinism audit) are fixed here.

## CI-only remainders, tracked in the spec

Correlated restarts across the close window, kill-storm including mid-replay, memory flatness over a >1M-round epoch, restart-to-live latency at ≥90% epoch age, and mixed-binary rollback.

Refs #2058, #2057.


omersadika:
## Review (Fable 5, against the #2058 spec) — approve

Focused on whether the load-bearing spec items survived implementation. They did:

- **Boot replay is the right shape end to end**: `boot_replay.rs` reads the consensus store directly in 250-commit batches (matching consensus-core's own recovery batch), folds and drops, and only then starts consensus — with `replay_after` **read from the store it is about to open** rather than remembered beside it. Both #2057 assertions become structurally unreachable, and memory stays flat on an old epoch by construction, not by hope.
- **The audit registry is the ratchet the spec asked for, done better than specified**: one static list generating the wipe, the post-wipe emptiness invariant, and the test fixtures — and the enforcement test takes its truth from *behaviour* (`every_table_the_consensus_fold_writes_is_classified_derived`), which the PR fault-validated against a misclassification the two naive tests missed. That last detail is the difference between a checklist and a guarantee.
- **Emission suppression keys on the right thing**: quorum-settled state (state-sync watermark ⊔ own aggregated certificate head) — convergent, unforgeable-by-progress, and pointedly not a watermark reintroduced through the back door. The pre-existing gap it closes (a node that certified but hadn't synced past its own certificate would re-sign an epoch of checkpoints) is a real find.
- **The #2058 watchdog clause held**: replayed commits feed the commit-liveness sink, fault-validated, so the boot-into-isolation gap #2054 documented stays closed.

The CI-only remainders (kill-storm mid-replay, memory flatness >1M rounds, latency at ≥90% epoch age, mixed-binary rollback) are correctly tracked in the spec; of these, **mixed-binary rollback** deserves a release-notes line when this ships — an old binary rolled back after this ran will recreate the deleted tables empty and recover reconstruction-heavy, which is survivable but must not surprise an operator mid-incident.

omersadika:
Superseded by #2074, which merged the complete #2058 change as one commit — this layer's content is byte-identical inside it (equivalence verified mechanically in #2074's review). Kept closed-not-merged as the permanent review record for the event-sourcing layer: the wipe-model design, the derived/preserved classification registry and its behaviour-derived enforcement (later made structural and deleted in the memory layer), the #2057 structural fix, the ten fault validations (A–J), and the approving review. Refs #2058.

PR #2062, opened 2026-08-16
feat(consensus): carry rounds to the MPC drain over a bounded blocking channel

omersadika:
Part 2 of #2058, stacked on #2061. The ten per-round projection tables are deleted; the consensus fold hands each round's inputs to the MPC drain over a bounded channel and blocks when the drain is behind.

**This was decided by measurement, not argument.** The alternative — keep the tables as a boot-rebuilt materialized view — carried the same safety content, so the question was live cost. Benchmarks on identical 50k-commit fixtures with the drain throttled 5× slower than the producer:

| | tables | channel |
|---|---|---|
| epoch DB written | 10.35 MiB | 7.25 MiB (**−30%**) |
| total drain time | 123.1s | 123.1s (parity) |
| fold to head | 79.8s | 120.6s (+51%) |
| peak upstream queue | 1 commit | 15,446 commits |

Restart recovery came out at parity — the measurement the decision was originally sought for — and the channel's savings are understated there, since that run still wrote the two fold-output tables this branch also deletes. The coupling is real: under burst the fold runs at drain pace, which is checkpoint-signing lag. The issue author took that trade knowingly.

**The memory story rests on an upstream guard.** A blocked fold stalls `highest_handled`, and consensus-core's commit syncer stops scheduling fetches once the consumer lags by `commit_sync_batch_size × commit_sync_batches_ahead` = 3,200 commits — so deep backlogs stay on peers' disks rather than in RAM. That makes `CommitConsumerMonitor` reporting POST-fold load-bearing, and it must not be confused with the #2054 commit-liveness sink, which is fed at ARRIVAL so a blocked fold never reads as isolation. Two sinks, opposite placements, both deliberate.

**What review caught, in order of severity.** A first pass found seven confirmed findings, the worst a boot-replay deadlock: the drain waited for the replay signal while the replay's folds filled the channel — any restart with more than a minute of commits would hang, silently, forever. Also: the integration harness had never been migrated (its drain was a no-op, so ~100 tests exercised none of this); `fold_blocked` could latch across an aborted park and disable the isolation watchdog for the process lifetime; the catch-up gate could never engage from the live path; the NOA streams lost their protocol gates, which would have been a binary-decided divergence on a mixed committee. All seven are fixed and re-reviewed CONFIRMED-FIXED, and the fix round found one more of its own — the blocked-time counters were gauges, so a per-epoch reset would have read as recovery on the one signal that has no second opinion.

`dev-docs/conventions/consensus-output-consumption.md` records the resulting rules as the standing pattern for any future consumer, and the superseded per-round mechanics are swept out of dev-docs rather than left beside the new text.

## Evidence

Every load-bearing behaviour is fault-validated — aborted park, boot replay deeper than the channel through the real wiring, a permanently-parked wedge (the superseded assertion passes green on that, demonstrated), the watchdog hold from a real full channel, the NOA gate off-flag, the tombstone under a real replay. Full integration suite green on CI at `test_threads=4`.

The last CI round is worth noting: seven failures, all one class the migration itself created — rounds were idempotent table writes and are now sequenced messages. Fixed at the class level, and `deliver_round` now asserts strict ascent so the next instance fails at the call site instead of as an opaque drain panic.

## CI-only remainders

Real-payload upstream queue bytes (the harness used ~1 KiB synthetic commits, so its peak understates production RAM by an unmeasured factor), a wedged drain end to end, correlated restarts, mid-epoch rollback — all tracked in the spec.

Refs #2058.


omersadika:
## Review (Fable 5, against the #2058 spec) — approve, with two items promoted to release blockers

This PR reverses two of my recorded recommendations (the spec rejected the channel; #2054 deliberately fed its sink post-boundary). I reviewed both reversals as the places most likely to hide a regression. Neither does — both are implemented with exactly the failure-mode handling I would have demanded:

- **The channel decision is properly re-decided by measurement**, with the dial table for capacity and honest parity/regression numbers. The two objections that originally killed it are answered in code: boot replay never rides the channel's memory story (it reads the store directly, batched — see #2061), and **`mark_drain_gone` restores the decoupling for the case that mattered most**: a *dead* drain (the #1978 self-malicious break) does not park the fold — one warning, fold continues, consensus participation survives MPC death exactly as in the table design. Only a slow-but-alive drain parks.
- **The watchdog rewiring keeps deadlock coverage**: "arrival" is the fold's own `recv`, so a fold wedged for any reason other than the one named hold still goes silent and still breaches. The hold is a single shared flag with a written considered-and-rejected for the channel-depth alternative — and the aborted-park latch bug this PR's own review caught is precisely the failure I would have hunted first. Blocked-time as counters (not gauges) so an epoch reset cannot impersonate recovery: correct, and the kind of correction that only comes from actually thinking about the consumer.

**The two blockers before the fleet sees a release with this:**

1. **Real-payload upstream queue bytes.** The remaining queue is consensus-core's *unbounded* commit channel, bounded in practice by the syncer's 3,200-commit lag cap — measured here with ~1KiB synthetic commits. A production `CommittedSubDag` carries every validator's blocks including class-groups payloads; 3,200 × real-sized commits plausibly reaches gigabytes. The module doc flags this honestly; I'm asking for the cluster measurement (or a defensive bytes bound) to be a **condition of the release**, not a follow-up.
2. **Wedged-but-alive drain, end to end.** Drain death is handled; a drain that is alive-but-stuck parks the fold forever with the watchdog correctly held — detection then rests entirely on the park counters and the #2038 lag alarm. That end-to-end test is currently a CI remainder, and the monitoring side (alerts on the park counters and condition gauges) rolls with the release in infra. Both halves should exist before an operator can hit this state.

With those two conditions attached to the release rather than the merge: ship it.

omersadika:
Both release conditions addressed as far as this repo reaches, in `03df712001`. Neither is closed by it — the parts that need a cluster stay open in #2064 — but the repo-side halves are no longer missing.

**1. The ceiling is ours now.** ika was inheriting `commit_sync_batch_size × commit_sync_batches_ahead` from upstream defaults, which are tuned for a consumer that never blocks; ours blocks deliberately. Pinned in ika's `Parameters` with the math and the reason, so an upstream retune cannot move our memory ceiling without a diff. **Held at 100 × 32 = 3,200 on purpose:** lowering it also slows a legitimately-behind node's catch-up, and that is a liveness trade that should follow your measurement rather than precede it. The comment names it as the lever if the measurement comes back badly — which is the honest state, since the number that decides it (bytes per real `CommittedSubDag`) is exactly what nobody has yet.

**2. The wedge has a deployable rule.** The playbook had the reading but no expression, so the one failure mode with no second opinion had nothing infra could implement. Alert 7 now carries both expressions (endless park; and the corroborating channel-pinned-with-consumed-round-flat shape), the stopped-versus-slow discrimination, why the commit-liveness watchdog correctly stays quiet through it, and why a restart is safe here — boot replay rebuilds the epoch's derived state, so nothing the committee agreed is lost.

What remains genuinely outside this repo, tracked in #2064: the real-payload byte measurement, the end-to-end wedged-drain cluster scenario, and deploying the alerts.

Also: the four heavy suites are running on this head — the earlier integration pass was two commits stale, and cluster, TS, and the full-matrix upgrade had never been dispatched against this branch at all. Verdicts to follow before I'd call it merge-ready.

omersadika:
Superseded by #2074 (equivalence verified mechanically in its review). Kept closed-not-merged as the permanent record for the blocking-transport layer: the tables-vs-channel benchmark tables and the capacity dial sweep (the measured basis of the design decision), the seven consensus-review findings and their confirmed fixes — including the boot-replay deadlock and its phase-split — the two-sinks rule (liveness at arrival, consumer monitor post-fold), the syncer lag-guard memory bound, and the release conditions that became #2064. The experiment branch `experiment/2058-blocking-capped-channel` holds the benchmark harness itself. Refs #2058.

omersadika:
## Benchmark record — the measurements that settled the tables-vs-channel decision

Recorded here in full so the numbers survive with the decision they produced. Harness: `experiment/2058-blocking-capped-channel` (`consensus_manager::stream_experiment`, three `#[ignore]`d benchmarks). Method notes that matter: both arms ran behind an IDENTICAL unbounded producer topology mirroring consensus-core's real shape (`commit_consumer.rs:35`, non-blocking finalizer sends) — a producer that blocked with the fold would have shown flat memory and hidden the question; arms differ only in channel-vs-tables.

### 1. Restart recovery (unthrottled drain, identical fixtures)

| commits | tables: fold done → drain done | channel: fold done → drain done | peak RSS (t/c) | epoch DB (t/c) |
|---|---|---|---|---|
| 1k | 105ms → 105ms | 104ms → 104ms | 51 / 53 MiB | 0.57 / 0.51 MiB |
| 10k | 206ms → 310ms | 210ms → 210ms | 170 / 170 MiB | 2.36 / 1.74 MiB |
| 50k | 828ms → 1,036ms | 626ms → 626ms | 695 / 697 MiB | 10.29 / 7.19 MiB |

Near-parity; recovery speed does not separate the designs. (Absolute times are not restart-latency predictions — empty commits give the fold nothing to verify.)

### 2. Live burst — 50k commits, producer 200µs/round, drain throttled to 1ms/round (5×)

| | tables | channel |
|---|---|---|
| fold reaches head | 79.8s | 120.6s (**+50.2%**) |
| drain reaches head (total work) | 122.6s | 123.1s (**parity, +0.4%**) |
| fold sends blocked | 0 | 46,391 / 50,000 (93%) |
| peak upstream queue | **1 commit** | **15,446 commits** |
| epoch DB written | 10.35 MiB | 7.25 MiB (**−30.0%**) |

The decision trade, in numbers: −30% write volume and total-throughput parity, bought with +50% fold-to-head under sustained burst (checkpoint-signing lag window) and the backlog relocating to the upstream in-memory queue.

### 3. Capacity sweep — 10k commits, same 5× throttle

| cap | fold done | blocked sends | peak upstream |
|---|---|---|---|
| 64 | 24.6s | 9,752 (98%) | 3,343 |
| 256 | 24.1s | 9,130 (91%) | 3,104 |
| 1024 | 21.7s | 6,376 (64%) | 2,153 |
| 4096 | 15.7s | **0** | 1 |

Monotone dial, not a step function: the cap trades fold blocking against in-channel memory, and cap-large-enough-to-never-block is by definition cap-large-enough-to-hold-the-burst.

### Limitations (they gated what the decision could claim)

- Synthetic ~1 KiB commits: RSS deltas understate production; **queue DEPTH is the honest metric**, and the 15,446-commit peak at real `CommittedSubDag` sizes is the unmeasured term that became a #2064 release condition.
- Sequential arms in one process for benchmark 2's RSS — cumulative RSS is confounded there; depth carried the conclusion.
- The harness models consensus-core's topology; it does not contain consensus-core. Post-decision, the author's verification found the commit-sync lag guard (fetch pause at 3,200 unhandled commits) bounds the deep-lag case the never-pausing producer overstates — recorded in the merged spec's memory section.
- First-run vacuity was possible and was guarded against after the fact in the v3 benchmark (see #2074's record); this harness pre-dated that lesson.

PR #2066, opened 2026-08-17
Migrate the examples to Sui SDK v2 / dApp Kit 2, vendor ikavery, on a shared dependency set

### LINKED ISSUE #2055: examples/: keyspring and multisig-bitcoin are pinned to a V4-blind SDK and @mysten/sui v1
Split out of #2052 / #2053, which fixed the docs, README, and skills but deliberately did not touch `examples/` — see scope note at the bottom.

## What is broken

Both examples pin an SDK whose bundled WASM cannot decode what mainnet and testnet emit today:

| Example | `@ika.xyz/sdk` | `@mysten/sui` |
| --- | --- | --- |
| `examples/keyspring/backend` | `^0.2.3` | `^1.45.2` |
| `examples/keyspring/frontend` | `^0.2.3` | `^1.45.0` |
| `examples/multisig-bitcoin/frontend` | `^0.2.2` | `^1.43.1` |

`^0.2.x` resolves `@ika.xyz/ika-wasm@0.2.1`, which understands only V1/V2-tagged reconfiguration outputs. Both networks now write V4, so anything reaching `getProtocolPublicParameters` — every dWallet operation — dies with:

```
invalid value: integer `3`, expected variant index 0 <= i < 2
```

This is unconditional, and it is the failure a user cloning an example actually hits.

## Transport is a smaller, conditional problem

Contrary to the first read, the examples do **not** all point at the deprecated public fullnodes. Verified by direct request:

- `sui-testnet-rpc.publicnode.com` (keyspring testnet) still answers JSON-RPC — `{"result":"4c78adac"}`
- `ikafn-on-sui-2-mainnet.ika-network.net` (keyspring mainnet, `backend/src/dkg-executor.ts:103`) returns nothing to a JSON-RPC request
- `multisig-bitcoin` takes its client from dapp-kit's `useSuiClient()`, so its transport follows the app's network config

So transport breakage is per-endpoint; the WASM breakage is universal.

## Why this is its own issue

The fix is not a version bump. Moving to `@ika.xyz/sdk@0.5.0` also means `@mysten/sui` v1 → v2, which renames the client surface and moves dapp-kit integration. That is 16 source files across two apps:

```
examples/keyspring/{frontend,backend}        3 files
examples/multisig-bitcoin/frontend          13 files
```

Doing it inside the docs/release PR would have buried a fix users are waiting on behind an unrelated migration.

## Order of work

Wait for `@ika.xyz/sdk@0.5.0` to publish, then migrate both examples and point them at `SuiGrpcClient`. Until then the examples are broken no matter what is changed in them, since the working SDK is not on npm yet.

Iamknownasfesal:
Closes #2055.

Both example apps spoke JSON-RPC, which public fullnodes no longer serve, and pinned an SDK whose bundled WASM cannot decode what the networks emit today. Migrated per the official [Sui 2.0 guides](https://sdk.mystenlabs.com/sui/migrations/sui-2.0/llms.txt).

## One dependency set instead of three

The examples were standalone bun projects with their own pins, which is exactly how they drifted a full major behind without anything noticing. They are now pnpm workspace members, and every shared library is declared once in a `catalog:` in `pnpm-workspace.yaml`.

They take the SDK as `workspace:*`, so **they build against this checkout, not npm** — which also means this PR did not have to wait on the 0.5.0 release.

Verified `catalog:` is publish-safe before relying on it: `pnpm pack` of `sdk/typescript` resolves the protocol away, emitting `@mysten/sui: ^2.26.1` and `@ika.xyz/ika-wasm: 0.3.0`.

**LTS:** Node 24 (Krypton) is the Active LTS line, so `@types/node` tracks 24 rather than the newer Current 26, and the apps declare `engines.node >=24`. The SDK's own floor stays permissive — a published library should not pin consumers to our runtime.

## Transport

Everything reads and executes over `SuiGrpcClient`. For multisig-bitcoin that forced a package choice: the legacy `@mysten/dapp-kit` is JSON-RPC only and receives no further updates — its provider is typed to `SuiJsonRpcClient`, and its sign path *resolves the transaction through that client*, so wallet signing would break on public fullnodes. It moves to `@mysten/dapp-kit-react`, whose `createDAppKit({ createClient })` takes a gRPC client natively.

## Read/write API

`multiGetObjects` → `getObjects`, `getOwnedObjects` → `listOwnedObjects`, `getDynamicFields` → `listDynamicFields`, `showBcs`/`showContent` → `include: { content: true }`, `getBalance` → `{ balance }`, `devInspectTransactionBlock` → `simulateTransaction`. Execution returns a discriminated union, so an on-chain failure arrives as `FailedTransaction` rather than a throw.

One place got materially simpler rather than merely translated: the Core API returns a dynamic field's value inline, so listing a requests table with `include: { value: true }` replaces list-fields → fetch-each-object → hand-unwrap `Field<u64, Request>`. The two copies of that logic in `useMultisigRequests` collapsed into one helper (−328 lines there).

## Framework bumps

| | before | after |
| --- | --- | --- |
| multisig-bitcoin | Next 16.1, dapp-kit 0.19 | Next 16.3, dapp-kit-react 2.1 |
| keyspring | Next 15.5, React 18, Tailwind 3 | Next 16.3, React 19, Tailwind 4 |

keyspring's webpack shim existed for Web Worker support it never used, so it is gone and the build runs on Turbopack. Tailwind 4 moves the theme into `@theme` in CSS; only `text-primary`, `bg-primary` and `font-mono` are used anywhere in the app and all three keep their names, with the remaining tokens carried over so nothing can silently stop resolving.

## Verification

- `tsc --noEmit` clean: multisig frontend, keyspring frontend, keyspring backend
- `next build` succeeds for both apps
- SDK unit tests (91) and the live mainnet/testnet probe still pass on the shared dependency set

## Two things left deliberately

**Pre-existing lint errors.** `eslint` reports 28 errors in multisig-bitcoin — 22 `no-explicit-any` plus React hooks rules — in components this migration does not touch (`RequestCard`, `MultisigDetailsView`, generated code, `use-mobile`). Confirmed pre-existing rather than introduced: the rules already existed in the `eslint-plugin-react-hooks@7.0.1` the app's own bun lockfile pinned, and my diff adds zero `any`. Happy to clean them up, but they are example-app quality debt rather than migration fallout.

**A cosmetic gap.** dApp Kit's account menu resolves SuiNS names through the JSON-RPC-only `resolveNameServiceNames`, which fails harmlessly against a gRPC client, so the menu shows the raw address.

Neither app has been exercised in a browser — the verification here is typecheck plus build.


Iamknownasfesal:
Parking this for now — closing rather than merging. The branch `chore/migrate-examples-sdk-0.5` stays pushed, so reopening picks up exactly where this left off.

State at the time of parking: both apps typecheck clean and build, the SDK's 91 unit tests and the live mainnet/testnet probe pass on the shared dependency set. Nothing here depends on the 0.5.0 release, since the examples consume the SDK as `workspace:*`.

Iamknownasfesal:
Reopened, and the branch has grown since it was parked. Two additions beyond the original description:

## ikavery is now vendored here (0756580)

`examples/ikavery/` — 47k lines across shared core/UI packages, a Sui half (Move package, SDK, Next app) and a Solana half, imported from [Iamknownasfesal/ikavery](https://github.com/Iamknownasfesal/ikavery) and licensed here under the repository's Clear BSD License, which upstream carried none of.

Its bun/turbo/biome/lefthook toolchain is dropped in favour of the workspace, and its Sui half is migrated to gRPC + dApp Kit 2 — 49 `SuiJsonRpcClient` sites and 20 `getJsonRpcFullnodeUrl` calls. The Solana half is deliberately untouched, which is also why the `@fesal-packages/*` package names stay: `solana/packages/*` import them.

Docs: a README in the house style, an entry in `examples/README.md`, and a page under Move integration examples.

## A real runtime bug, found by a deployment (9d4787e)

keyspring took its IKA fee coin from an `IKA_COIN_ID` env var and passed it as `tx.object(...)`. A coin object ID is only valid until that coin is spent, merged or split — after which every operation fails with:

```
The following input objects are invalid: Object 0x8552... not found
```

Replaced with `coinWithBalance`, which resolves a coin from the admin address at build time and cannot go stale. Budget moves to an optional `IKA_FEE_BUDGET` (default 1 IKA), applied at all three fee sites: DKG, presign, sign.

That commit also carries two `@noble/curves` v2 fixes — `randomPrivateKey` → `randomSecretKey`, `ed25519.ExtendedPoint` → `Point`, and `p256.sign()` returning bytes rather than a Signature object. **Nine ikavery tests were failing on those while typecheck and build both passed**, which is worth knowing: the compile-time gates in this PR do not cover that class of change.

## Verification status

| | |
| --- | --- |
| multisig-bitcoin, keyspring | typecheck + build clean (verified earlier on this branch) |
| ikavery | build 6/6, typecheck 10/10, tests 9/9 — via the standalone repo, whose sources are identical |
| SDK | 91 unit tests + live mainnet/testnet probe pass on the shared dependency set |
| The fee fix | verified in the standalone key-spring repo; **not** re-verified in this monorepo copy, which needs a workspace install |

Both examples have matching PRs open against their own repos: [ikavery#1](https://github.com/Iamknownasfesal/ikavery/pull/1) and [key-spring#2](https://github.com/Iamknownasfesal/key-spring/pull/2).

Neither app has been exercised in a browser, and no end-to-end DKG, wallet or passkey flow has been run against a funded key.

Issue #2072, opened 2026-08-19
Release the checkpoint builders during the boot replay (removes the replay-window memory peak)


omersadika:
Follow-up carved out of #2058 v3 (memory-only derived state), deliberately NOT taken in that change.

## The peak

During a boot replay the MPC drain fills `pending_dwallet_checkpoints` / `pending_system_checkpoints` throughout, while the checkpoint builders wait for the replay signal — so nothing consumes the queues for the whole replay, and with derived state now in memory the peak is an in-memory term: one entry per replayed round with MPC output, each carrying that round's checkpoint content. It is transient (drains once the builders release) and bounded by the epoch's MPC-output rounds; `ika_epoch_pending_dwallet_checkpoints` / `ika_epoch_pending_system_checkpoints` measure it (sampled at mutation, because the builder-sampled gauge reports nothing during exactly this window).

Prune-on-arrival was investigated and is WRONG, not merely hard: pending checkpoints are keyed by consensus height, the certified watermark by sequence number, and nothing maps them at arrival time during a replay; worse, skipping already-certified heights would shift the builder's sequence numbering — the node must BUILD every checkpoint of the epoch to reproduce the committee's numbering and merely must not re-SIGN, which the settled-state suppression already handles.

## The mechanism

Release the builders during the replay instead of after it. The spec's stated reason for the gate is volume, not safety — "a builder that drained an epoch's rebuilt queue against an unset consensus client would park one submission task per checkpoint and trip the adapter's in-flight limit" — and the settled-state emission suppression now drops everything below the certified watermark before it reaches submission, which covers that argument for the certified prefix. What needs deliberate treatment:

- the uncertified tail still parks submissions against the unset client — bound it (the tail is small by definition, but "small" needs the adapter's in-flight limit checked against it);
- the builder gate section of `dev-docs/specs/event-sourced-epoch.md` was written for the current ordering and must be REWRITTEN, not patched (it also carries the release-ordering semantics note from `76ac48469e`);
- the `ReplayWaiter` plumbing and the drain-only phase-split (`drain_while_replaying`) interact with builder release — the boot-deadlock analysis that produced the phase-split must be re-run with builders live.

## Acceptance

The replay-peak gauges flat (bounded by the uncertified tail, not the epoch) across a full-epoch replay on a loaded cluster, with the boot-deadlock test family still green and no re-submission burst at peers (the emission-silence criterion re-asserted under the new ordering).

[...]

Issue #2064, opened 2026-08-17
Release conditions for the event-sourced epoch and the blocking round transport (#2058)


omersadika:
Two conditions attached to the RELEASE that ships #2058 (PRs #2061 and #2062), raised in review of #2062 and recorded here so they cannot be lost when those PRs merge. Neither blocks the merges; both block shipping to the fleet.

Both exist because the blocking round transport moves the deep-lag backlog from ika's disk into consensus-core's in-memory commit channel, and because a drain that is alive-but-stuck is a state no in-process test can reach.

## 1. Real-payload upstream queue bytes, measured or bounded

The queue that absorbs a burst is consensus-core's **unbounded** commit channel, bounded in practice only by the commit syncer's lag cap — fetch scheduling stops once the consumer trails by `commit_sync_batch_size × commit_sync_batches_ahead` = 3,200 commits (`consensus/core/src/commit_syncer.rs`, thresholds from `consensus/config/src/parameters.rs`, at the pinned rev).

The benchmarks behind the design decision used synthetic commits of roughly 1 KiB, where a 15,446-commit peak cost about 15 MiB. A production `CommittedSubDag` carries every validator's blocks including class-groups MPC payloads, so 3,200 real commits plausibly reaches gigabytes. `round_transport.rs` documents this as unmeasured rather than claiming the small number.

Required before release, either:
- a cluster measurement of resident bytes with a throttled drain under real MPC load, at the 3,200-commit lag cap; or
- a defensive bytes bound on the ika side, so the ceiling is ours rather than inherited.

## 2. Wedged-but-alive drain, end to end, plus its monitoring

A drain that **dies** is handled and tested: the fold detaches, warns once, and keeps folding, so consensus participation survives MPC death exactly as it did with the per-round tables. A drain that is **alive but stuck** is different — it parks the fold indefinitely, and the commit-liveness watchdog deliberately holds (a parked fold is holding a commit it received, which is the opposite of the isolation the watchdog exists to catch).

Detection therefore rests entirely on:
- `ika_consensus_fold_blocked_seconds_total` climbing while `ika_dwallet_mpc_consumed_round` stays flat — one endless park;
- both blocked counters climbing together — a merely slow drain;
- the MPC round-lag alarm.

Required before release:
- the end-to-end scenario on a cluster (drain stuck, fold parked, watchdog held, counters climbing), currently a CI remainder in `dev-docs/specs/event-sourced-epoch.md`; and
[...]

PR #2073, opened 2026-08-19
feat(epoch): hold the epoch's derived state in memory, never on disk

omersadika:
Part 3 of #2058, stacked on #2062, at the issue author's direction: derived epoch state is never written to the database. On restart the node starts the consensus output and the round channel over from the epoch's first commit, and that replay is the only source of derived state.

**The point is what gets deleted.** With derived state in memory, the derived/preserved split is structural — a `DBMap` field survives restarts by definition, an in-memory field is rebuilt by definition — so the classification registry, the boot wipe, the post-wipe emptiness invariant, and the three tests that enforced the classification are gone rather than adapted (net −34 lines over 23 files, with `derived_epoch_state.rs` deleted whole). One test remains, pinning the 25 tables that stay on disk, so adding a durable table is a deliberate act. `AuthorityEpochTables` is down from 46 fields to 25.

**The conversion preserves the fold's semantics literally, not by argument.** The eleven pieces the fold committed in one RocksDB batch live behind one lock taken exactly where `batch.write()` was — a `WriteBatch` is atomic to concurrent readers, so one guard acquisition reproduces its visibility, and the freeze partition still cannot be observed without its round. Consensus review confirmed all of it against code: one acquisition per commit, pre-commit snapshots for the fold's own reads, no production double-hold across all 26 sites, fold purity, and BCS digest determinism across every key variant of the processed set.

**Two retention bugs the conversion surfaced would have shipped multi-GiB epochs invisibly on disk; in memory they were fatal, so they are fixed:**
- Checkpoint signatures and builder outputs were never deleted within an epoch, and every signature row carries a full copy of the checkpoint content — one copy per signer, per checkpoint, forever. Both families now prune below the certified watermark (strictly below — the sequence being certified is retained, fault-validated), with `retain_from = min(next_to_certify, highest_built)` because the highest built message is the sequence-number cursor: state sync certifying past local build progress must not restart the epoch's checkpoint numbering (fault-validated; the case was invisible at design time).
- The prune is what also keeps a boot replay flat: the aggregator is deliberately NOT replay-gated, so signatures the replay re-collects below the durable certified watermark are dropped within one aggregator interval. That asymmetry is now a dependency of the memory bound, pinned three ways — spawn-site comments, a spec paragraph, and a compile-time signature assertion whose error lands on the test that explains why (fault-validated by performing the exact symmetrizing refactor it guards against).

**The honest costs, stated rather than buried:**
- The submission-dedup set (`Blake2b256(bcs(key))`, measured 37.7–69.2 B/entry) is unbounded in the epoch's traffic. The same cardinality existed on disk — it now occupies a resource that aborts instead of degrading, and because the set is a deterministic function of the epoch, an OOM during replay re-accumulates identically and crash-loops until the boundary. That determinism is why the loaded-epoch RSS measurement in #2064 is a release gate, and the constraint on any future LRU lever is recorded: a threshold that fires on one node's replay must fire on every node's, or the eviction is itself the divergence.
- The builder input queues grow unfilled through a boot replay (builders wait for the replay; the drain does not). Prune-on-arrival was investigated and is wrong, not hard — nothing maps consensus heights to certified sequence numbers at arrival, and skipping would shift the committee's checkpoint numbering. Accepted and gauged (sampled at mutation, because the builder-sampled gauge is blind during exactly this window); releasing the builders during replay is #2072, decided after the measurement.
- Rollback: an old binary rolled back mid-epoch finds empty fold-side tables and no watermark row, and re-derives the whole epoch through its own table-writing path with none of this stack's emission suppression. The old compat argument is deleted from the spec, not softened; the upgrade matrix is the required evidence.

**Every guard is fault-validated** (prune boundary, numbering cursor, processed-before-waiter-wakes with two vacuity guards that also fired, `ensure_epoch_alive` — readers must FAIL after release, not answer from cleared state, which the MPC drain's graceful stop depends on — release actually freeing memory, and the epoch clock anchored on the commit timestamp, caught by both determinism tests naming the diverged field).

Consensus review: eight lenses, all clean, no blockers. Full integration suite green on CI at the exact head (and once more one commit back). Local: 424 lib tests, six integration families (restart, ordinals, mid-epoch keys, NOA sign, NOA checkpoint, catch-up, replay-drain), clippy clean, zero lint suppressions.

Specs rewritten around the structural split: `event-sourced-epoch.md`, `preserved-epoch-state-audit.md` (renamed — only what stays on disk needs an audit now), `consensus-output-consumption.md` rule 8, the write-discipline convention, alerts playbook (including the operator reading: a sawtooth returning to its floor is healthy; a floor that climbs is the certification stall).

Refs #2058, #2064, #2072.


omersadika:
## Review (Fable 5, against the #2058 spec) — approve; this is the best-argued PR of the stack

Verified in code, not from the body: `derived_epoch_state.rs` deleted whole and `AuthorityEpochTables` at exactly 25 fields (led by the presign pools and private outputs — the #1952 lesson holding its ground); the prune's `retain_from = min(next_to_certify, highest_built)` with `split_off` keeping the numbering cursor and dropping strictly below; `ensure_epoch_alive` failing closed (`EpochEnded`) rather than answering from cleared state; and the aggregator asymmetry pinned by the function-pointer coercion test — a compile-time guard whose failure lands the reader on the paragraph explaining why the "tidying" refactor it blocks would silently make boot replay the worst case for signature retention. That test is the best artifact in the PR.

**The structural-split thesis is the right call, and the deletion count proves it.** Part 1's classification registry was the correct defence for disk-backed derived state; making the split structural (a `DBMap` survives by definition, a field is rebuilt by definition) deletes the defence *and its three enforcement tests* rather than adapting them, net −34 lines for an architecture change. When the third pass over a design removes machinery instead of adding it, the abstraction is converging.

**The two retention bugs are the review's real treasure — and they are live in v1.3.1 today.** Checkpoint signature rows carrying a full checkpoint copy per signer, never pruned within an epoch, is a multi-GiB-per-epoch disk leak production currently *has*, bounded only by the per-epoch DB drop at the boundary. Moving the state to a resource that aborts instead of degrading is exactly what surfaced it. That is the memory-first design paying for itself before it ships.

**The sharpest risk is named honestly and gated correctly, but I want it sharpened one notch in #2064**: the dedup set is unbounded in the epoch's traffic, and a deterministic OOM during replay is #2057 reborn as a memory property — crash-loop until the boundary, unfixable by restart. The loaded-epoch RSS measurement must therefore be sized against the *busiest observed epoch's* traffic with margin, not a typical one; and `ika_epoch_processed_consensus_messages` should get a budget alert on the infra side (queued, with the four `ika_epoch_pending_*` gauges) so the fleet sees the set approaching any future budget long before an abort. The recorded LRU constraint — eviction must be consensus-uniform or it is itself the divergence — is exactly right and should stop anyone reaching for the obvious wrong lever.

**Rollback is now a release-notes obligation, not a compat property.** An old binary rolled back mid-epoch re-derives the epoch through its table-writing path with none of this stack's emission suppression. Deleting the compat argument from the spec instead of softening it was correct; the upgrade-matrix scenario in #2064 is the required evidence, and the release notes for whatever ships this must say "mid-epoch rollback re-emits; roll back at a boundary or expect DAG noise."

**Merge mechanics**: bottom-up only (#2061 → #2062 → #2073), CI re-run after each retarget. The stack is now three deep on unmerged bases — every day it waits is a day of main drifting under 6k+ lines of stacked change; I'd land it promptly after #2061's admin merge rather than let it season.

CI green at the head. With #2064 carrying the RSS gate and the wedged-drain end-to-end: ship it.

omersadika:
Superseded by #2074 (equivalence verified mechanically in its review — byte-identical outside the footprint of main's five intervening commits). Kept closed-not-merged as the permanent record for the memory-only layer: the design-pause inventory and blocker measurements (the dedup-set per-entry costs, the retention-bug discovery), the certified-watermark prune with both boundary fault validations, the aggregator-ungated-start pin (compile-time signature assertion), the eight-lens consensus review with the OOM-determinism finding, and the v2-vs-v3 head-to-head benchmark (fold +5–28%, DB 72–3,602× smaller, reopen 12ms flat vs superlinear wipe, RSS lower at all measured sizes — with the shape caveat that gates the epoch-scale answer on #2064). Refs #2058, #2064, #2072.

PR #2074, opened 2026-08-19
feat(epoch): event-source the epoch — consensus store as the only truth, derived state in memory

### LINKED ISSUE #2058: Event-source the epoch: rebuild all derived state by batched replay of the consensus store on every restart (drop the watermark and the per-round stream)

omersadika:
The complete #2058 change as ONE PR against current main — the alternative presentation to the three-PR stack (#2061 → #2062 → #2073), which stays open for comparison. Same end state, one commit, already rebased over #2070/#2071.

**Why one PR is arguably the truer shape:** the issue said "delivery remains one change, not stages," and the stack contains scaffolding its own later layers delete — part 1 builds the classification registry that part 3 removes; part 1 classifies ten tables that part 2 deletes. A stack reviewer reviews machinery that does not survive; this diff contains only the end state (49 files, +7,175/−2,791 against main). The full evolution, its review rounds, and every fault-validation record remain readable in the three PRs and the branch history.

**What it is** (short form — each layer's PR body carries the full argument):
1. **No watermark**: `replay_after` reads from the consensus store itself; #2057's two-databases-one-fact crash-loop is unbuildable.
2. **No per-round tables**: the fold hands rounds to the MPC drain over a bounded blocking channel (benchmarked decision); deep-backlog RAM bounded by consensus-core's 3,200-commit sync lag guard; liveness at arrival, consumer-monitor post-fold.
3. **No derived tables at all**: 21 tables become in-memory state rebuilt by replay; the split is structural, so the registry/wipe/enforcement apparatus is deleted (net −34 lines for the architecture change); `AuthorityEpochTables` = exactly the 25 tables no replay reproduces, pinned by one test.
4. **Two live retention bugs fixed** (checkpoint signatures/builder outputs never pruned within an epoch — a multi-GiB leak production has today), with the certified-watermark prune fault-validated at both boundaries and the aggregator's ungated start pinned three ways including a compile-time signature assertion.

**Evidence**: all suites green per layer (integration head-exact on the v3, cluster green on the v3 head, TS + upgrade matrix in flight at PR time); three consensus reviews across the layers, zero unresolved findings; every load-bearing behavior fault-validated with captured evidence; v2-vs-v3 benchmark: fold +5–28%, epoch DB 72–3,602× smaller, restart reopen 12ms flat vs superlinear wipe, RSS 2.7–6.3× lower at measured sizes (epoch-scale crossover gated by #2064).

**Release gates unchanged**: #2064 (five-series loaded-cluster measurement — sized against the busiest observed epoch with margin, per review — plus wedged-drain end-to-end and alerts incl. a budget alert on `ika_epoch_processed_consensus_messages`); #2072 (builder release during replay, post-measurement). Release notes must carry: mid-epoch rollback re-emits — roll back at a boundary or expect DAG noise.

Fixes #2058. Refs #2057, #2064, #2072.


omersadika:
## Equivalence verification (Fable 5) — this IS the reviewed stack; all three approvals carry over

Checked mechanically, not by re-reading 7k lines:

- **Zero unexplained drift.** 34 files differ between this head and #2073's head; every one sits inside the footprint of the five commits main gained between the two bases (#2059, #2065, #2067, #2070, #2071). Outside that footprint, this PR is **byte-identical** to the stack the three reviews approved.
- **Conflict resolutions verified where main and the stack touched the same code**: #2070's `error_chain` logging survives in both rewritten checkpoint modules (ref-count matches main exactly); #2071's rate-limit gate wiring in `ika-node/src/lib.rs` is fully preserved (5/5 references) alongside this stack's boot-replay and watchdog wiring; `rate_limit.rs` is blob-identical to main.
- CI green at the head; one commit against current main.

So the review state transfers: approved, with the same conditions — #2064's loaded-RSS gate sized against the busiest observed epoch, the wedged-drain end-to-end, and the release-notes line that mid-epoch rollback re-emits (roll back at a boundary).

One request for the merge: the granular history — the channel benchmarks, the capacity dial table, the fault-validation records — lives in the three stacked PRs and their branches. Close #2061/#2062/#2073 as superseded-by-#2074 with a comment saying so (rather than silently), so the evolution stays findable from here.

omersadika:
## Benchmark record — v2 (tables+wipe) vs v3 (memory-only), the comparison behind the final shape

Recorded on the merged PR so the numbers live with the code. Method: one byte-identical benchmark patch applied UNCOMMITTED to both `2e2ed2b94c` (v2) and `84be4b0bcf` (v3) — the two APIs it needs (`AuthorityPerEpochStore::new(EpochStoreParams)`, `process_consensus_transactions_and_commit_boundary`) have identical signatures on both; v3-only gauges read from the prometheus registry by name (print `unavailable` on v2). One process per (arm, size) so cumulative RSS cannot flatter either side. Fixtures: 4 tx/commit, 16 KiB MPC message bodies (parameter choice, not a mainnet measurement — v2's DB cost scales linearly in it; rescale freely).

| 1,000 commits (4k tx) | v2 | v3 | delta |
|---|---|---|---|
| fold | 0.053s / 18,854 c/s | 0.051s / 19,755 c/s | v3 +4.8% |
| epoch DB after fold | 17.4 MB | 240 KB | v2 **72.6×** larger |
| reopen (restart cost) | 0.061s | 0.012s | v2 5.1× slower |
| RSS delta | 34.7 MB | 12.7 MB | v2 2.7× higher |

| 10,000 commits (40k tx) | v2 | v3 | delta |
|---|---|---|---|
| fold | 0.460s / 21,741 c/s | 0.366s / 27,326 c/s | v3 +25.7% |
| epoch DB after fold | 173.3 MB | 240 KB | v2 **722×** larger |
| reopen | 0.627s | 0.012s | v2 52× slower |
| RSS delta | 104.6 MB | 16.6 MB | v2 6.3× higher |

| 50,000 commits (200k tx) | v2 | v3 | delta |
|---|---|---|---|
| fold | 2.258s / 22,145 c/s | 1.765s / 28,335 c/s | v3 +27.9% |
| epoch DB after fold | 864.4 MB | 240 KB | v2 **3,602×** larger |
| reopen | 3.974s | **0.012s (flat at every size)** | v2 331× slower |
| RSS delta | 126.1 MB | 29.4 MB | v2 4.3× higher |

### Readings

- **v2's reopen cost is the wipe, and it grows superlinearly** (10× commits → 10.3× time; 5× → 6.3×), consistent with the tombstone-reseek behaviour the old spec flagged as unmeasured. v3's reopen is flat at 12ms because there is nothing to delete.
- **v2 used more RAM at every measured size** — not the "disk for RAM" trade anyone assumed: RocksDB memtables hold the full payload-embedding dedup keys that v3 hashes to 32 bytes (~85 B/entry all-in vs the bare-set 69.2 B/entry, allocator-measured separately).
- **The fence that must travel with the headline**: the two RSS curves differ in SHAPE — v2's is bounded by RocksDB buffers and returns after flush; v3's grows until the epoch boundary. 50k commits ≈ 3% of a mainnet epoch, so the crossover (if any) is past everything measured. The epoch-scale answer is #2064's loaded-cluster measurement, sized against the busiest observed epoch; these numbers cannot answer it and do not claim to.
- Fixture sanity held: v2's post-fold DB = 4,322 B/tx vs 4,096 predicted from one payload-carrying key in four; v3's processed entries equalled folded transactions exactly at all sizes.

### Limitations

Synthetic content (fixed byte patterns, no class-groups material, fold handed pre-verified transactions, as every test of this path is); **no checkpoint-signature transactions, so the certified-watermark prune contributes to neither arm** — the one v3 mechanism whose correctness was fault-validated is unexercised here; replay measured as reopen-after-real-fold rather than through `replay_epoch_commits` (the store fixture cannot carry ika transactions — documented on both branches), which isolates exactly the part that differs; fixture construction inside the timed loop (inflates both arms equally; deltas hold, absolutes don't); single machine, single run per cell; RSS via `ps` (process total, ~12 MB fixed floor visible in v3's 1k row).

### The meta-lesson, recorded in `dev-docs/learnings/pitfalls.md`

The first run produced a clean-looking table measuring almost nothing — the fixture's key space wrapped at 256 and the dedup set plateaued (769 entries for 4,000 transactions). **The change's own processed-set gauge caught it**, not harness review. Rule now in pitfalls: every benchmark needs an internal consistency check tying what it measured to what it claims to have driven, before the table is trusted.

Channel-vs-tables benchmarks (the #2062-layer decision) are recorded on #2062.

PR #2076, opened 2026-08-19
docs: correct a phantom metric name in the wedged-drain alert and its prose

omersadika:
The wedged-drain corroboration expression in `production-alerts.md` — and four doc-comment/playbook sites echoing it — named `ika_dwallet_mpc_consumed_round`, which was never registered. The exported metric is `ika_last_process_mpc_consensus_round` (same value, set in the drain the line before `record_mpc_consumed_consensus_round`). An alert written from the playbook as-is would silently never fire — the exact failure class it exists to catch, one level up. Found by infra's `check-dashboard-queries.py` while implementing #2064's alerts; the infra rules are being written with the correct name and a comment noting the substitution. Five prose sites corrected, no code paths touched. PR CI's compile jobs cover the doc-comment edits.

Refs #2064.

### DIFF
--- a/crates/ika-core/src/authority/round_transport.rs
+++ b/crates/ika-core/src/authority/round_transport.rs
@@ -251,7 +251,7 @@ impl RoundTransportSender {
     /// (a parked fold is not an isolated node), so a drain that stops
     /// consuming entirely holds the watchdog indefinitely and correctly —
     /// isolation is not what has gone wrong. This climbing while
-    /// `ika_dwallet_mpc_consumed_round` is flat is that failure, and it is
+    /// `ika_last_process_mpc_consensus_round` is flat is that failure, and it is
     /// the only place it shows.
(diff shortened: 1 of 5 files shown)


omersadika:
## Review (Fable 5) — approve

Verified: `ika_dwallet_mpc_consumed_round` appears zero times in the metric registry (`check-metric-names.sh --list`) and `ika_last_process_mpc_consensus_round` exactly once — the phantom is real, the substitution is right, and the five corrected sites are prose/doc-comments only.

Worth savoring what happened here: an alert expression that would have deployed and **silently never fired** — the precise failure class the wedged-drain alert exists to catch — was caught by infra's name ratchet before it ever reached Grafana. The checker was built after this campaign shipped four dashboards-referencing-dead-names bugs by hand; this is it paying rent on the producer side. Merge at will.

PR #2077, opened 2026-08-19
test(upgrade): mid-epoch rollback of v1.3.1 onto event-sourced stores

omersadika:
Adds the one direction nothing in the upgrade matrix exercises: the deployed **v1.3.1 release rolled back onto stores an event-sourcing binary (#2074) has been writing, mid-epoch**. This is the required evidence named by #2064 — the rollback section of `dev-docs/specs/event-sourced-epoch.md` was an argument with no measurement behind it.

## The scenario (`tests/mid_epoch_rollback.rs`, opt-in via `RUN_MID_EPOCH_ROLLBACK=1`)

Candidate runs a 4-validator network first → epoch 2, past a completed network-key reconfiguration → a **control window** measures what an ordinary full user lifecycle does to the peers' `ika_skipped_consensus_txns` (the noise floor) → validator 0 is swapped to the literal `release/mainnet-v1.3.1` binary on its existing stores → three assertions, all witnessed at the **peers**, never self-reported:

1. **It re-derives the epoch.** v1.3.1 finds empty fold-side tables and no watermark, so `replay_after` reads 0 and Mysticeti re-sends the epoch from its first commit; its own table-writing fold rebuilds the per-round tables its MPC drain reads. Witness: `ika_last_process_mpc_consensus_round` on the rolled-back node climbing to the round its peers had consumed (target snapshotted once — no moving finish line).
2. **It re-emits already-settled work, and the volume is measured.** Delta of `ika_skipped_consensus_txns` summed over the three peers, required to beat the control window (a bare "nonzero" proves nothing on a cumulative counter). Per-kind composition logged via `ika_consensus_handler_processed{kind}`. The measured number is the "mid-epoch rollback re-emits" figure for the operator notes.
3. **Full participation returns.** After catch-up, a session-snapshot is taken, a fresh workload is driven, and a peer must record an MPC output *authored by the rolled-back validator* for a session outside that snapshot — re-emitted outputs are inside it by construction, so the spray cannot masquerade as participation.

The whole experiment is bracketed by `expect_epoch_at_most(2)`: an epoch boundary hands every validator a fresh epoch store and would make all three assertions trivially true, so drifting across one **fails** the run instead of silently voiding it.

## Spec correction, in the same PR

Reading both binaries' source while building this falsified two claims in the spec's rollback section:

- v1.3.1 is **not** free of settled-state suppression: its checkpoint-signature gate (`get_highest_verified_dwallet_checkpoint`) reads the perpetual checkpoint store, which a rollback does not touch, so it re-signs only the verified→certified band — not an epoch of checkpoint signatures. The spray's dominant term is MPC messages/outputs replayed over the rebuilt rounds.
- The old binary's drain reads the tables its own fold writes, so MPC participation returns once the re-derivation catches up — not "sits out until the next epoch boundary". The release note is rewritten accordingly (a re-derivation paid once, scaling with epoch age).

The section now explicitly names the two numbers it still owes to this scenario's first green run: the re-emission count and the catch-up duration. If the run contradicts the participation branch, one paragraph and the release note flip back.

## Wiring & verification

- `mid_epoch_rollback` added to the workflow matrix (`ALL`, `RUN_MID_EPOCH_ROLLBACK`, OLD-binary build from `release/mainnet-v1.3.1`, real-crypto refusal list). Off the PR trigger, matching `restart_spectator`. Dispatch entry in `dev-docs/playbooks/ci-suites.md`.
- `cargo test --release -p ika-upgrade-test --lib`: 33 passed (6 new — they pin the two places a matcher could pass on the wrong evidence: author-filtering of output sessions, unset-gauge handling, dual authority naming, and the three-way re-submission verdict).
- clippy (both CI configs, incl. `cargo simtest clippy`) exit 0; fmt clean; zero `#[allow]`.
- Also fixes a stale doc comment in `restart_spectator.rs` (logs append across restarts; they are not truncated).

Fault-validation patches (dead subject, blind observer, boundary rollback, no-fresh-sessions) are prepared and `git apply --check`-verified; they run as separate dispatches after the first green run, per `dev-docs/playbooks/test-testing.md`.

Refs #2064.

### DIFF
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -48,6 +48,16 @@ name: Upgrade Test
 #                  not merely stay consensus-healthy while peers absorb the
 #                  sessions. Current build only; production presign-pool
 #                  sizing (the sustained-volume ingredient).
+#   mid_epoch_rollback — the only BACKWARD direction in the matrix: the
+#                  candidate runs the network, then the literal v1.3.1 release
+#                  is put back on one validator mid-epoch, on the same stores.
+#                  Asserts the old binary re-derives the epoch through its own
+#                  table-writing fold against empty fold-side tables, COUNTS
+#                  the already-settled work it re-sends (measured at the peers,
+#                  against a control window of ordinary traffic), and that a
+#                  peer witnesses it contributing MPC output again before the
+#                  boundary. Same OLD defaults as v131_rollout; needs a long
+#                  epoch so the whole experiment fits inside one.
 #
 # Each scenario binds the fixed Sui localnet ports (9000/9123) and chdirs the
 # process during publish, so only ONE runs per job — `--test-threads=1`, one
@@ -237,7 +247,7 @@ jobs:
           TEST: ${{ inputs.test || 'v131_rollout' }}
         run: |
           set -euo pipefail
-          ALL="smoke workload v131_rollout v131_churn malicious_v131 restart_spectator"
+          ALL="smoke workload v131_rollout v131_churn malicious_v131 restart_spectator mid_epoch_rollback"
           if [ "$TEST" = "all" ]; then
             SELECTED="$ALL"
           else
@@ -353,6 +363,7 @@ jobs:
             v131_churn)   RUN_FLAG=RUN_V131_CHURN ;;
             malicious_v131) RUN_FLAG=RUN_MALICIOUS_V131 ;;
             restart_spectator) RUN_FLAG=RUN_RESTART_SPECTATOR ;;
+            mid_epoch_rollback) RUN_FLAG=RUN_MID_EPOCH_ROLLBACK ;;
             v131_rollout)   RUN_FLAG=RUN_V131_ROLLOUT ;;
             *) echo "unknown test: $TEST" >&2; exit 1 ;;
           esac
@@ -362,6 +373,12 @@ jobs:
           OLD_BIN_NAME="$IN_OLD_BIN_NAME"
           case "$TEST" in
             v131_churn | malicious_v131)
+              : "${OLD_REF:=release/mainnet-v1.3.1}"
+              : "${OLD_BIN_NAME:=ika-validator}"
+              ;;
+            mid_epoch_rollback)
+              # Same release, opposite direction: OLD is what the node is
+              # rolled BACK to after the candidate has been writing its stores.
               : "${OLD_REF:=release/mainnet-v1.3.1}"
               : "${OLD_BIN_NAME:=ika-validator}"
               ;;
@@ -405,7 +422,7 @@ jobs:
           # exists again — it is the run standing behind a version flip on a
           # live network, which is where mocked agreement is most misleading.
           case "$TEST_NAME" in
-            v131_rollout|v131_churn|malicious_v131)
+            v131_rollout|v131_churn|malicious_v131|mid_epoch_rollback)
               if [ "$USE_MOCKS" = "true" ]; then
                 echo "$TEST_NAME is a cross-binary gate and must use real cryptography" >&2
                 exit 1
@@ -456,7 +473,7 @@ jobs:
           cargo test --no-run --release -p ika-upgrade-test --test "$TEST_NAME"
 
       - name: Build OLD binary from ${{ env.OLD_REF }}
-        if: ${{ matrix.test == 'v131_rollout' || matrix.test == 'v131_churn' || matrix.test == 'malicious_v131' }}
+        if: ${{ matrix.test == 'v131_rollout' || matrix.test == 'v131_churn' || matrix.test == 'malicious_v131' || matrix.test == 'mid_epoch_rollback' }}
         run: |
           set -euo pipefail
           WT="$GITHUB_WORKSPACE/old-build"
--- a/crates/ika-upgrade-test/src/cluster.rs
+++ b/crates/ika-upgrade-test/src/cluster.rs
@@ -43,6 +43,7 @@ use ika_types::sui::{DWalletCoordinatorInner, SystemInner};
 use ika_types::supported_protocol_versions::SupportedProtocolVersions;
 use rand::rngs::OsRng;
 use sui_sdk::wallet_context::WalletContext;
+use sui_types::base_types::ConciseableName;
 use sui_types::base_types::{ObjectID, SuiAddress};
 use sui_types::crypto::SuiKeyPair;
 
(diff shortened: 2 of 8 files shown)


omersadika:
First green run of the scenario: https://github.com/dwallet-labs/ika/actions/runs/32318375717 (1h24m57s, dispatched with `-f test=mid_epoch_rollback -f epoch_duration_ms=1800000`).

Measured, on a 4-validator localnet (3-quorum, 30-min epoch, rollback at ~0.59 E over a 19,330-round backlog):

- **Re-emission attributable to the rollback: zero within noise** — 144 already-folded consensus transactions in the rollback window vs 150 in a control window of identical composition (difference −6). The DAG spray the spec originally predicted does not exist; v1.3.1's #2023 catch-up mode reconstructs replayed sessions without recomputing them.
- **Re-derivation drain: ~35s** for the 19.3k-round backlog; catch-up mode entered and exited (both pinned as fresh log occurrences).
- **MPC participation returned**: a peer witnessed subject-authored outputs for two post-snapshot sessions with 0s witness latency after the post-rollback workload, and the subject's completions counter rose. Instruments were calibrated in-run against the healthy subject during the control phase.

Caveat: one run on a lightly-loaded localnet — the loaded-cluster measurement remains #2064 gate 1. Fault-validation dispatches (per `dev-docs/playbooks/test-testing.md`) follow on throwaway branches.

omersadika:
Fault validation complete, per `dev-docs/playbooks/test-testing.md` — every verdict below was read from the run's artifact evidence, not its exit code. Fault branches are deliberate injections marked never-merge.

## Green path: two independent passing samples

| run | re-emission (rollback vs control) | recovery | witness latency |
|---|---|---|---|
| [32318375717](https://github.com/dwallet-labs/ika/actions/runs/32318375717) | 144 vs 150 (**−6**) | ~35s, 25.1s of it process boot | 0s |
| [32346054927](https://github.com/dwallet-labs/ika/actions/runs/32346054927) / [32336593160](https://github.com/dwallet-labs/ika/actions/runs/32336593160) | 144 vs 144 (**0**) | 25.1s boot again | 0s |

Both rollback windows landed on the same 144 with identical restart re-announcements (3 capability notifications, 3 idle-status updates): the level is steady, not just the difference. Two runs of the same shape agree with each other; they do not constitute a range — the loaded-cluster measurement remains #2064 gate 1.

## Faults — all five caught, with the named evidence

| fault | injected defect | evidence in artifact |
|---|---|---|
| F1 [32324957657](https://github.com/dwallet-labs/ika/actions/runs/32324957657) | subject stopped, never restarted | fails on the dead node's unreachable `/metrics` (connection refused) — cannot pass on absence |
| F2a [32346059611](https://github.com/dwallet-labs/ika/actions/runs/32346059611) | re-emission measured on the rolled-back node itself | `validator(s) [0] restarted while this measurement window was open… the window's subtraction is not a delta` — the restart-generation guard, at the rollback window close |
| F3 [32324964434](https://github.com/dwallet-labs/ika/actions/runs/32324964434) | participation snapshot taken after the workload | witness timeout at 300s, `while still inside epoch 2` — the fault's failure, not a boundary artifact |
| F4 [32336600329](https://github.com/dwallet-labs/ika/actions/runs/32336600329) | epoch ceiling set below the experiment | `the network is in epoch 2, past the expected ceiling of 1` |
| F5 [32324971665](https://github.com/dwallet-labs/ika/actions/runs/32324971665) | run-2's real harness defect re-injected (full-name matching against concise labels) | fails **during calibration**, before the rollback — the in-run calibration provably catches the historical defect |

Two escapes during the campaign are part of the record, both closed:

- An earlier F2a attempt ([32336596634](https://github.com/dwallet-labs/ika/actions/runs/32336596634)) went **green with a corrupted measurement** after the assertion-2 verdict was demoted from a control-gate to a report — demoting a verdict silently disarmed the fault whose detection relied on it. Closed by the structural restart-generation guard (a window refuses to close over an observer that restarted inside it, before any number is computed), pinned by unit tests, then re-validated above.
- The first calibration design flaked ~1-in-3 on a read race (a peer can witness the subject's output milliseconds before the subject's own drain counts the completion); two fault runs died there instead of at their faults. Closed by polling to the bounded deadline; re-proven by two subsequent green runs.

## Boundary experiment (X1) — the spec's "a rollback at a boundary is unaffected", measured

- [Attempt 1](https://github.com/dwallet-labs/ika/actions/runs/32324974973): the subject came back to a ~300-round backlog — three orders of magnitude under the 5,000-round catch-up `enter_threshold` — so #2023 catch-up mode never engages. There is nothing to re-derive.
- [Attempt 2](https://github.com/dwallet-labs/ika/actions/runs/32336604518): re-emission 150 vs control 159 (**−9**, zero within noise), participation witnessed.

A boundary rollback is indistinguishable from not rolling back at all.

omersadika:
## Review (Fable 5, against the #2064 brief) — approve; the methodology exceeds the brief

Verified in the diff, not the body: the control-window noise floor on `ika_skipped_consensus_txns` (delta over peers, not a meaningless nonzero on a cumulative counter); `expect_epoch_at_most(2)` bracketing **both** phases so an epoch drift fails the run instead of silently voiding all three assertions; the participation proof requiring a peer-recorded MPC output *authored by the rolled-back validator* for a session **outside** the pre-workload snapshot — re-derived sessions are inside it by construction, so the spray cannot impersonate participation. Workflow wiring confirmed (`ALL` list, dispatch flag mapping, old-binary build from the literal release tag). All three assertions witnessed at peers, never self-reported. This is the strongest test design in the campaign.

**The spec correction is the more important half of this PR, and I'm one of its corrections.** I verified the falsifying claim against the v1.3.1 tag: `get_highest_verified_dwallet_checkpoint` reads the perpetual store, which a rollback doesn't touch — so the old binary re-signs only the verified→certified band, not an epoch of signatures, and its drain reads the tables its own fold rebuilds, so **MPC participation returns after the re-derivation**. My review of #2073/#2074 repeated the spec's "costs MPC until the next epoch boundary" line into the release-notes obligations; that line was an argument without a measurement, and this PR replaced it with a testable claim: *a re-derivation paid once, scaling with epoch age*. The release notes should carry the corrected form — with the two numbers this scenario's first green run owes (re-emission count, catch-up duration).

**One thing the merge does not do: close the #2064 gate.** The evidence is the first green *dispatch* plus its two numbers, followed by the four prepared fault-validation dispatches (dead subject, blind observer, boundary rollback, no-fresh-sessions) proving the assertions can fail. Merge, dispatch, then the gate closes on the numbers.

PR #2083, opened 2026-08-20
fix(ocs): guard the untagged chain mirrors against layout drift at build time

omersadika:
Closes the untagged-leaf-struct decode non-guarantee found during the dev-docs reconciliation (#2079): `DWalletCoordinatorInnerV1`, `SystemInnerV1`, and `DWalletNetworkEncryptionKey` are plain untagged deserialize structs mirroring deployed Move layouts, and every OCS proof gate verifies **bytes, not shape** — so layout drift between the Move source and the Rust mirror passes all gates and surfaces, at best, as an unattributed bcs error (a trap that has already cost three CI cycles per the pitfalls record).

## What re-verification narrowed before designing

- Decode strictness already holds: every leaf site uses exhausting `bcs::from_bytes` (verified against bcs-0.1.6 — residue is a hard `RemainingInput`). Length-changing drift fails loudly today.
- Both mirrors end in fixed-width tails, so the genuinely silent runtime case is a **same-total-length reshuffle**, not a field-add.
- The real structural hole: Move's coordinator/system are at `const VERSION = 2` with a single unsuffixed inner struct, and both Rust readers match `1 | 2 =>` onto one mirror — an untested, uncommented assertion that two versions share a layout. Mainnet's `compatible` upgrade policy already blocks in-place field-adds, so real drift arrives via VERSION-bump + `migrate()` or fresh publishes — which is why this fix is **build/CI-time prevention**, not runtime defense.

## The three guards

1. **`scripts/check-chain-mirror-layout.sh` + `scripts/chain-mirror-layout.txt`** (CI-wired beside its five sibling checks, with `--list`/`--write` modes): extracts the Move field lists, the Rust mirror field order, and both `const VERSION`s, and diffs them against a committed manifest. The manifest maps Move→Rust **explicitly** — three real name divergences exist (`presign_sessions`→`presigns`, `pricing_and_fee_manager`→`pricing_and_fee_management`, `witnesses_approving_advance_epoch`→`witness_approving_advance_epoch`), so name-equality would have been a lie. Any Move field add/remove/reorder/retype or a VERSION bump fails CI naming everything that must move together, including the spec.
2. **Golden BCS pins** beside each mirror: length + Blake2b digest over a fully-populated value with every field distinct, plus a pin on all four `DWalletNetworkEncryptionKeyState` variant indices (the variant index is the wire discriminant). The struct literal uses no `..Default::default()`, so a Rust field add/remove is a *compile* error before the digest even runs.
3. **Decode-error attribution**: a `map_err`-only wrapper at all five leaf sites that names the Move source struct and states layout drift as a candidate cause — converting "hunt in the wrong struct" into a pointed failure. Strictness untouched.

Deliberately NOT done: splitting the `1 | 2 =>` arm into separate types — an ungated live-path behavior change fabricating a distinction the layouts don't have. The VERSION tripwire is what forces the layout question when the arm is next widened.

## Fault-injection evidence (per `dev-docs/playbooks/test-testing.md`; all injected, caught, reverted, re-greened)

| injected drift | caught by |
|---|---|
| Move field-add | script: field-count mismatch naming both sides |
| Move+Rust coordinated add (counts match) | script: per-field diff naming the exact field |
| Rust-only field-add | compile error (no-default struct literal) |
| `const VERSION` 2→3 | script: version pin (honest note: runtime already fails closed on a `3`; the tripwire's value is forcing the layout question when the arm is widened) |
| Rust retype `u64`→`u32` | digest ONLY — the script reads names, proving the two halves aren't redundant |
| **same-width field swap in `SystemInnerV1` (778 bytes → 778 bytes)** | **both build-time halves — the exact case runtime `bcs::from_bytes` would have accepted silently** |

## What this does NOT guarantee (also in the spec's Residuals and the new convention doc)

Nothing here authenticates shape at runtime — the proof chain proves provenance, never well-formedness, and a same-total-length reshuffle still mis-decodes silently on a running node. The guards bind this repo's releases at build time, not a node pointed at a foreign or hand-patched package. The script proves the Move declaration hasn't moved since a human last checked it against the mirror, not that the mirror was ever correct: the Move and Rust type vocabularies don't correspond mechanically, so Move types are pinned as text and only names and order are tied across. The only true runtime fix is a version tag inside the Move struct — a Move change plus protocol version bump, judged not worth its cost while the upgrade checker blocks the in-place case.

Docs in the same PR: new `dev-docs/conventions/chain-mirror-layout.md`, README index entry, rewritten spec Residuals, pitfalls entry.

Gates: workspace `cargo check --release` clean; clippy `-D warnings` exit 0, no `#[allow]`; `ika-types` 43/43, `ika-sui-client` 39/39, `verified_reader` 45/45; fmt zero unrelated churn; no Move files in the diff.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,6 +104,13 @@ jobs:
         # Every should-never-happen log must increment the fleet-visible
         # invariant violation counter through the shared macro.
         run: ./scripts/check-invariant-violation-markers.sh
+      - name: Chain-mirror layout
+        # The Move structs that Rust mirrors with UNTAGGED structs may not
+        # change shape without the Rust mirror and the version dispatch moving
+        # in the same PR: BCS is positional, so their field order is the wire
+        # format of bytes already on chain, and no OCS proof gate verifies shape.
+        # See dev-docs/conventions/chain-mirror-layout.md.
+        run: ./scripts/check-chain-mirror-layout.sh
       - name: Sui version consistency
         # The Sui pin lives in several places that don't update together;
         # see dev-docs/conventions/sui-version-bump.md.
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,6 +111,15 @@ jobs:
         # format of bytes already on chain, and no OCS proof gate verifies shape.
         # See dev-docs/conventions/chain-mirror-layout.md.
         run: ./scripts/check-chain-mirror-layout.sh
+      - name: Chain-mirror layout (Rust half)
+        # The golden BCS pins beside each mirror are the Rust half of the
+        # check above: the script reads Move declarations and Rust field
+        # NAMES, so a Rust-side retype (u64 -> u32) passes it and only the
+        # pinned length/digest catches it. Nothing else on PR CI executes
+        # ika-types' tests (Cargo Test Check is compile-only), so the pins
+        # must run here or a retype merges green. ika-types has no MPC
+        # crypto, so the debug profile is fine.
+        run: cargo test -p ika-types --lib -- layout_is_pinned variant_indices_are_pinned
       - name: Sui version consistency
         # The Sui pin lives in several places that don't update together;
         # see dev-docs/conventions/sui-version-bump.md.
--- a/crates/ika-core/src/sui_connector/verified_reader.rs
+++ b/crates/ika-core/src/sui_connector/verified_reader.rs
@@ -41,6 +41,7 @@ use ika_network::proof_provider::{
     ProofProvider, VerifiedDynamicFieldsPageRequest, VerifiedObjectResponse,
 };
 
+use ika_types::chain_mirror::{DWALLET_COORDINATOR_INNER_V1, SYSTEM_INNER_V1, decode_chain_mirror};
 use ika_types::sui::system_inner_v1::{DWalletCoordinatorInnerV1, SystemInnerV1};
 use ika_types::sui::{DWalletCoordinator, DWalletCoordinatorInner, System, SystemInner};
 
@@ -1095,10 +1096,9 @@ impl OcsVerifiedReader {
                 let child_id = derive_versioned_child_id(coordinator_id, outer.version)?;
                 let child_obj = self.verified_anchor_object(child_id).await?;
                 let child_bcs = move_object_contents(&child_obj.object)?;
-                let field: Field<u64, DWalletCoordinatorInnerV1> = bcs::from_bytes(child_bcs)
-                    .map_err(|e| {
-                        ReaderError::Decode(format!("Field<u64, DWalletCoordinatorInnerV1>: {e}"))
-                    })?;
+                let field: Field<u64, DWalletCoordinatorInnerV1> =
+                    decode_chain_mirror(child_bcs, DWALLET_COORDINATOR_INNER_V1)
+                        .map_err(ReaderError::Decode)?;
                 Ok((outer, DWalletCoordinatorInner::V1(field.value)))
             }
             v => Err(ReaderError::UnsupportedVersion {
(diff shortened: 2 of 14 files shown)


omersadika (commented):
Reviewed the whole diff, ran the script and all 7 new tests on the branch (green), and re-did two of the fault injections. Design is right and the code is careful — the `$`-anchored lazy Move-field regex, the trailing-space header match (`struct SystemInner ` vs `SystemInnerV1`), the first-colon / first-`->` manifest partition for fully-qualified Move types, the no-`..Default` literals, and the separate enum-discriminant pin all hold up. Error variants at the five leaf sites are preserved (`ReaderError::Decode`, `IkaError::SuiClientSerializationError`, `GrpcSuiClientError::Decode`), so no classification/retry behavior changes.

**One real gap — the Rust half's digest is not enforced by CI.**

On PR/push CI nothing *executes* `cargo test -p ika-types`: `Cargo Test Check` is `cargo check -p ika-core --tests` (compile-only, and ika-core only — it doesn't build ika-types' `#[cfg(test)]` modules at all), and the workspace test runs live only in `workflow_dispatch` workflows (`integration-tests-ci` SCOPE=all, `test-cluster` with `PACKAGE=ika-types`). `cargo clippy --all-targets` does compile the pin tests, so the struct-literal half (Rust field add/remove → compile error) is enforced. The digest half is not.

Reproduced on the branch: retype `SystemInnerV1.stake_subsidy_period_length: u64 → u32` (nothing outside the mirror consumes it):
- `check-chain-mirror-layout.sh` → PASS (reads names only, as documented)
- `cargo check --release -p ika-types --tests` → clean (so Clippy would be green)
- `system_inner_v1_layout_is_pinned` → FAILED, `pinned 778 bytes … got 774`

So today that drift merges green; the fault table's "digest ONLY" row is, in CI terms, "nothing". The convention doc is honest about this (only the Move half is described as "run in CI"), but the PR's "bind this repo's releases at build time" framing isn't true for retypes until the tests run somewhere automatic.

Fix is one step in `ci.yaml`, next to the script step (ika-types has no MPC crypto, so the debug profile is fine and it's seconds once the dep tree is cached):

```yaml
      - name: Chain-mirror layout (Rust half)
        run: cargo test -p ika-types --lib -- chain_mirror layout_is_pinned variant_indices_are_pinned
```

(Putting it in the Clippy job reuses that job's checkout/cache; any job works.) Then the convention doc can say both halves run in CI.

**Nits (take or leave):**
- `assert_pinned_layout` guidance says a pure rename means "only the pin needs updating" — the pin *values* don't change on a rename (BCS is positional); it's the struct literal in the test that needs the new field name. Tiny, but the text is what someone reads at 3am.
- `move_fields`/`rust_fields`: the `startswith("///")` check after `strip_comment` is dead (the `//` split already emptied doc-comment lines). Harmless.

With the CI step added this is merge-ready from my side; without it, it's still a strict improvement over main (the script + attribution + compile guard are all live), just not the full claim.

omersadika:
Review gap closed in 1f393f0c12: added the "Chain-mirror layout (Rust half)" CI step next to the script step, running exactly the pin tests (`cargo test -p ika-types --lib -- layout_is_pinned variant_indices_are_pinned` — libtest ORs the filters; the pair matches all 4 pin tests). Your reproduced retype now fails PR CI instead of merging green. Both nits taken: the rename guidance now says the pin values don't change (it's the test's struct literal + manifest that need the new name), and the dead `///` checks are dropped from both extractors (the `#[` attribute check stays — that one is live). Script re-run green after the extractor edit.

omersadika (commented):
Follow-up verified: the Rust-half step runs in the Format Check job (toolchain + cache present) and the job log shows the 4 pin tests executing and passing, job still ~2.5 min. Both nits addressed. Merge-ready once v131_rollout is green.

PR #2084, opened 2026-08-20
fix(mpc-data): make the pre-freeze wait legible; warn on undecodable blobs; memoize decode verdicts (#2080)

### LINKED ISSUE #2080: Pre-mid-epoch announcement wait is silent for hours (was misdiagnosed as a genesis MPC wedge) — name the wait, WARN on undecodable blobs, memoize decodes
## Nondeterministic genesis wedge: `decide_locally_validated_peers` spins forever, MPC never starts — caught during the #2064 RSS measurement

Found on a local 4-validator network (real binaries at current main `e2fa03f4`, real class-groups crypto, M3 Max) while running the #2064 loaded-RSS measurement. Two identical back-to-back runs of the same harness: the first (smoke) completed network DKG in ~30 minutes; the second wedged at genesis and sat for **3 hours** with no MPC progress before being killed. Same binaries, same flow, different outcome — a race.

## The wedged state, precisely

- Consensus fully healthy: ~197k commits folded, `expect_all_validators_healthy` green, capabilities and **all 4/4 mpc_data announcements recorded on every validator** (`blob_bytes=72864` each) — same counts as the successful smoke run.
- Every validator's **MPC service silent** from the moment of the first consensus commit onward: zero `dwallet_mpc` log lines for 3 hours (the successful run logged 1,302 in the same span). No panics, no ERROR lines, no `mark_drain_gone`.
- Round channel depth 0 the whole time; fold never parked; the #2054 commit-liveness watchdog armed normally and correctly stayed quiet (commits were flowing — this is not isolation).
- **One core pinned at 100%** (6.25% of 16 cores — the measured process CPU) for the full 3 hours.

## Where it spins (live `sample` of the wedged process, twice, minutes apart)

```
IkaNode::monitor_reconfiguration
  → AuthorityPerEpochStore::local_b…/compute…
    → validator_metadata::decide_locally_validated_peers
      → validator_metadata::blob_decodes_to_valid_mpc_data   (validator_metadata.rs:655)
        → bcs::Deserializer … MPCDataV1::deserialize
```

Both samples land in the same frames (second sample: 14 hits). The loop repeatedly BCS-decodes the four 72,864-byte class-groups announcement blobs, forever, without ever advancing — and without logging anything.

## What it is not

- Not missing announcements: all four are recorded on all four validators (verified in each node log), identical to the successful run.
- Not the drain-death path: `mark_drain_gone` never fired; the receiver is alive.
- Not watchdog-catchable **by design**: commits arrive, so arrival-fed liveness is satisfied. This is the *wedged-but-alive* class the #2062 review named as the untested end-to-end remainder — first live specimen, at genesis rather than mid-epoch.

## Hypotheses for whoever picks this up (in order)

1. `blob_decodes_to_valid_mpc_data` returns **false silently** for one blob (a decode-false has no log line), the peer never validates, and the decision loop re-derives from scratch each tick — the spin. Why a byte-identical blob would decode differently run-to-run is then the real question (torn in-memory announcement write? read racing the fold's store? — the announcement store moved in-memory in #2074).
2. The decision succeeds but its consumer in `monitor_reconfiguration` re-invokes without persisting the outcome — a lost-wakeup/latch race won by the smoke run and lost here.

Either way, two hardening observations regardless of root cause: a validation-false on an announcement blob deserves a log line with the digest (three hours of silence is what made this expensive to find), and `decide_locally_validated_peers` re-decoding 4×72KB class-groups blobs per tick is wasted work even on the happy path — decode-once-and-cache is safe since the blobs are content-addressed by digest.

## Evidence bundle (available on request)

Full validator + notifier logs for wedged and successful runs, both process samples, and the 5-second RSS/gauge CSV covering the entire wedge — 448MB, preserved locally.

Release relevance: found *by* the #2064 measurement run; it does not block the measurement (nondeterministic — rerun proceeds), but a genesis-start race that silently disables MPC on a fresh network is a v-next blocker on its own.

omersadika:
# Root cause for #2080: not a race — a legitimate mid-epoch wait, killed ~2 minutes early

**Both hypotheses in the issue are wrong, and so is the "spin" framing.** There is no
race, no torn read, no lost wakeup, and nothing was pinning a core. The 4-validator
network in the wedged run was waiting — correctly, and exactly as designed — for
mid-epoch reconfiguration to publish `V_{e+1}`, and the run was killed **~2 minutes
before that was due**. This PR therefore does not fix a product bug. It fixes the
reason three hours of correct waiting were indistinguishable from a hang, and lands
the two hardening items the issue called for regardless of root cause.

## What actually happened

The mpc_data freeze — and with it network DKG — cannot start until the next-epoch
committee is published on chain. `decide_ready_to_finalize` returns `NotYet` until
`V_{e+1}` is published (or the 3/4-epoch liveness backstop elapses), and `V_{e+1}` is
published by `process_mid_epoch()`, which `system_inner.move::assert_mid_epoch_time`
refuses before `epoch_start_timestamp_ms + epoch_duration_ms / 2`.

The two runs were **not** the same flow. From the launch commands:

| | smoke (succeeded) | full (reported as wedged) |
|---|---|---|
| `EPOCH_DURATION_MS` | `3_600_000` (1 h) | `21_600_000` (6 h) |
| epoch 1 start | ~14:27:00 UTC | ≤ 15:04:39 UTC |
| mid-epoch due | 14:57:00 | **~18:04:30** |
| `Calling process_mid_epoch()` | **14:57:00.032** | never — run killed 18:02:00 |
| freeze | 14:57:09 | — |
| network DKG output | 14:58:04 | — |

The model predicts the smoke run to within 30 ms. For the full run it predicts
mid-epoch at ~18:04:30 against a kill at 18:02:00. The bound is not sensitive to the
exact genesis instant: even taking the earliest conceivable epoch start — 15:03:30,
the moment the Sui localnet came up, before the packages were even published —
mid-epoch reconfiguration was still due 90 seconds *after* the run was killed.

Corroborating, from the evidence bundle:

- `Calling process_mid_epoch()` (INFO, notifier) appears in the smoke notifier log and
  **never** in the wedged one; the wedged notifier's only `sui_executor` line is its
  startup line, which is also all the smoke notifier had before *its* mid-epoch fired.
- Every wedged `end-of-publish gate not yet satisfied` line, for three hours, reports
  `next_epoch_committee_exists=false` — the chain agrees `V_{e+1}` was never created.
- The smoke run's own first MPC activity is at 14:57:14, i.e. **28½ minutes after its
  first consensus commit** and 5 s after its freeze. Its "network DKG in ~30 minutes"
  was itself almost entirely this same wait, just a 1-hour-epoch-sized one.

## Why the stack samples are not a spin

The samples attached to the issue refute hypothesis 1 outright and refute the
"one core pinned at 100%" reading:

- `wedged-v0-sample2.txt` contains **both** `local_blob_coverage_meets_quorum`
  (`authority_per_epoch_store.rs:4135`) and `validated_peers_with_hashes`
  (`:4098`) above `compute_locally_validated_peers`. `validated_peers_with_hashes`
  is only reached *after* `local_blob_coverage_meets_quorum` returns **true** — so
  local blob coverage met the stake quorum. No blob was silently failing to decode.
- The sample is 146,747 thread-samples over a 2.36 s window; **146,513 of them are
  parked** in `__psynch_cvwait`/`kevent`/`__semwait_signal`. The whole process used
  ~234 ms of CPU in 2.36 s (≈0.1 of one core), and the decode path accounts for 5 of
  those samples — about **0.2 % of one core**, not 100 %.

What the frames actually show is the announcement sender's healthy 2 s poll tick
(`mpc_data_announcement_sender.rs::run`, `tokio::time::sleep(2s)` per iteration)
being caught mid-tick, twice, because it is one of the few things still doing any
work at all on an idle-but-waiting node.

## What this PR changes

**1. The emit gate says which wait it is in, at INFO (the actual remedy).**
`ReadyToFinalize::NotYet` now carries a `ReadySignalWait` — is `V_{e+1}` published,
which next-epoch members are uncovered, and how long the liveness backstop still has
to run — and the caller logs it. Previously this was one `debug!` line, so at the
default log level a validator waiting hours for mid-epoch reconfiguration and a
genuinely wedged one produced identical output: none.

Rate-limited to one line per five minutes per validator (first deferral, then every
150th tick), deliberately **not** latched: an operator who attaches mid-wait, or greps
a window of log, has to be able to find out why MPC has not started without hunting
for a line emitted at process start. Half a 24 h mainnet epoch costs ~150 lines.
The below-quorum-coverage branch is promoted from `debug!` to the same rate-limited
INFO for the same reason.

**2. A locally-held but undecodable announcement blob is named at WARN.**
`decide_locally_validated_peers` now takes a tri-state `LocalBlobVerdict`
(`Valid`/`Invalid`/`Absent`) instead of a bool, so the caller can tell "not here yet"
(ordinary, silent) from "here and garbage" (a fault) and reports the latter as
`invalid_blobs`. `compute_locally_validated_peers` warns once per `(announcer,
digest)` with both named. Held-but-invalid has no in-epoch self-heal — the perpetual
table rejects any insert whose digest is not `Blake2b256(bytes)`, so the bytes under a
digest can never change — and it silently costs the announcer its place in this
validator's attestation, which is precisely the failure the issue feared and could not
have seen.

**3. Decode verdicts are memoized by content digest.**
New `MpcDataDecodeCache` on the per-epoch store. The emit gate re-ran
`blob_decodes_to_valid_mpc_data` over every announced ~72 KB class-groups bundle one
to two times per 2 s tick for the entire pre-freeze window, always with the same
answer. Content-addressing makes the verdict a pure function of the digest, so this is
observationally equivalent and turns O(committee × blob size) per tick into one decode
per distinct digest per epoch. `Absent` is deliberately not cached — the blob may
arrive on the next tick, which is the point of the poll. Per-epoch lifetime, so the
map is committee-scale and drops at the boundary.

## Does this bite at epoch boundaries on a live network?

**No — nothing is broken, so nothing can bite.** The gate behaved correctly.

The *legibility* problem, however, recurs at **every** epoch, not just at genesis: the
`NotYet(next_committee_unpublished)` window runs from epoch start to mid-epoch, which
under a 24 h mainnet epoch is a **twelve-hour silent window per epoch, per validator**
during which the off-chain plane looks identical whether it is waiting or wedged. That
is the argument for landing this, and it is an observability argument only.

**This does not block the release.** #2080 should be reclassified from "v-next blocker
/ genesis-start race that silently disables MPC" to "observability gap".

The harness side needs a fix of its own, and it is the reason this looked
nondeterministic. `expect_network_dkg_output_version_at_least` waits `epoch_timeout`
(3 h in the RSS test) from the moment flow 3 starts, while the event it is waiting on
lands `epoch_duration_ms / 2` (3 h) after `system::initialize` — which is ~2 minutes
*earlier*. Network DKG is therefore given only about two minutes of budget after
mid-epoch reconfiguration finally fires, against the 64 seconds the smoke run took to
get from `process_mid_epoch()` to a persisted DKG output. That is a coin flip on a
loaded machine, and the reported run was manually killed at 18:02:00 while its own
flow-3 deadline still had until 18:06:21. Either raise `epoch_timeout` well above
`epoch_duration_ms / 2` or lower `EPOCH_DURATION_MS`.

## Tests

New, all `-p ika-core --release`:

- `validator_metadata::tests::decode_cache_decodes_a_real_blob_once_across_many_polls`
- `validator_metadata::tests::decode_cache_memoizes_the_invalid_verdict_too`
- `validator_metadata::tests::decode_cache_never_caches_absence_and_flips_when_the_blob_lands`
- `validator_metadata::tests::decide_locally_validated_peers_reports_held_but_undecodable_blobs`
- `epoch_tasks::mpc_data_announcement_sender::tests::ready_to_finalize_names_the_wait_it_is_in`
- `epoch_tasks::mpc_data_announcement_sender::tests::ready_signal_deferral_logging_repeats_on_a_bounded_cadence`

The five existing `decide_ready_to_finalize` gate tests now pin the wait reason rather
than accepting any `NotYet`.

**Fault-validated.** Each of the four protected properties was reverted to its pre-fix
behaviour — memoization removed from `MpcDataDecodeCache::verdict`; `invalid_blobs`
left unpopulated; `next_committee_published` hard-coded to `true`; the rate limit
degenerated to a one-shot `deferrals == 0` latch — and the suite re-run: **8 failed,
88 passed**, covering all four targeted tests plus four gate tests that also detect
the wait-reason revert (e.g. `decode_cache_decodes_a_real_blob_once_across_many_polls`
failed with `left: 50` against an expected 1 decode). Restoring the fix returns the
suite to **96 passed, 0 failed**.

`cargo clippy --all-targets --all-features -- -D warnings` and
`cargo simtest clippy --all-targets --all-features -- -D warnings` (the msim
compilation universe) are both clean, as are `cargo fmt -- --check` and the six
repo guard scripts CI runs.

Closes #2080.

---

### DIFF
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -1248,6 +1248,21 @@ pub struct AuthorityPerEpochStore {
     /// The `own_mpc_data_blob_unhealthy` gauge carries the ongoing state.
     self_blob_unhealthy_warned: AtomicBool,
 
+    /// Memoized decode verdicts for announced mpc_data blobs, keyed by
+    /// the content digest they are stored under. The emit gate re-runs
+    /// `compute_locally_validated_peers` on every ~2s tick and each run
+    /// used to BCS-decode every announced ~72KB class-groups bundle
+    /// again for an answer that cannot change (issue #2080). Per-epoch
+    /// lifetime, so it is bounded by the epoch's distinct announced
+    /// digests and drops at the boundary.
+    mpc_data_decode_cache: crate::validator_metadata::MpcDataDecodeCache,
+
+    /// `(announcer, digest)` pairs already warned about for a locally-held
+    /// but undecodable mpc_data blob. Dedupes what would otherwise be the
+    /// same warn on every emit-gate tick; bounded by the epoch's distinct
+    /// announced digests, like the cache above.
+    invalid_peer_blob_warned: parking_lot::Mutex<HashSet<(AuthorityName, [u8; 32])>>,
+
     /// Running max of `commit_timestamp_ms` over every consensus commit
     /// this validator has processed this epoch — the epoch's consensus
     /// clock. In-memory only: it starts at 0 on (re)open and restores
@@ -2322,6 +2337,8 @@ impl AuthorityPerEpochStore {
             handoff_aggregator: parking_lot::Mutex::new(None),
             perpetual_tables_for_handoff: ArcSwapOption::empty(),
             self_blob_unhealthy_warned: AtomicBool::new(false),
+            mpc_data_decode_cache: crate::validator_metadata::MpcDataDecodeCache::new(),
+            invalid_peer_blob_warned: parking_lot::Mutex::new(HashSet::new()),
             max_processed_commit_timestamp_ms: AtomicU64::new(0),
             round_transport: ArcSwapOption::empty(),
             observed_consensus_head_round: AtomicU64::new(0),
@@ -4041,14 +4058,41 @@ impl AuthorityPerEpochStore {
             self.name,
             announcements,
             |digest| {
-                perpetual
-                    .get_mpc_artifact_blob(digest)
-                    .ok()
-                    .flatten()
-                    .map(|bytes| crate::validator_metadata::blob_decodes_to_valid_mpc_data(&bytes))
-                    .unwrap_or(false)
+                self.mpc_data_decode_cache.verdict(digest, || {
+                    perpetual.get_mpc_artifact_blob(digest).ok().flatten()
+                })
             },
         );
+        // A blob that IS held locally but doesn't decode is a fault with
+        // no in-epoch self-heal (the digest names exactly one byte string),
+        // and it silently costs the announcer its place in this validator's
+        // attestation. Before #2080 it was dropped without a word, which is
+        // what made a stalled off-chain plane unreadable from the logs.
+        // Once per (announcer, digest) — the emit gate re-evaluates every
+        // ~2s tick. Self has its own warn plus a gauge just below, so it is
+        // not repeated here.
+        for (authority, digest) in &decision.invalid_blobs {
+            if *authority == self.name {
+                continue;
+            }
+            // Take and drop the dedupe lock before logging — never hold a
+            // mutex across a `tracing` call.
+            let first_sighting = self
+                .invalid_peer_blob_warned
+                .lock()
+                .insert((*authority, *digest));
+            if first_sighting {
+                warn!(
+                    validator = ?authority,
+                    blob_digest = ?digest,
+                    epoch = self.epoch(),
+                    "announced mpc_data blob is held locally but fails the decode gate; \
+                     excluding this validator from our locally-validated set until it \
+                     announces a different blob (its bytes cannot become valid — the \
+                     blob store is content-addressed)"
+                );
+            }
+        }
         if decision.self_blob_unhealthy {
             // Own announcement is in the table but the corresponding
             // perpetual blob is missing or fails decode. Attesting
--- a/crates/ika-core/src/epoch_tasks/mpc_data_announcement_sender.rs
+++ b/crates/ika-core/src/epoch_tasks/mpc_data_announcement_sender.rs
@@ -54,13 +54,51 @@ use std::time::Duration;
 use tokio::sync::watch::Receiver;
 use tracing::{debug, info, warn};
 
+/// Why the emit gate is still waiting, carried on
+/// [`ReadyToFinalize::NotYet`] so the caller can say so out loud.
+///
+/// Before #2080 the gate returned a bare "not yet" and the caller
+/// logged one `debug!` line, so the single longest wait in the
+/// protocol — a fresh network sitting until mid-epoch reconfiguration
+/// publishes `V_{e+1}`, which is HALF THE EPOCH after genesis — was
+/// invisible at the default log level. That silence is what made a
+/// correctly-waiting 4-validator network read as a hang for three
+/// hours.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ReadySignalWait {
+    /// Whether `V_{e+1}` has been published on chain yet. `false` is
+    /// the mid-epoch-reconfiguration wait: nothing this validator
+    /// does can shorten it, and on a fresh network with a long
+    /// genesis epoch it dominates everything else.
+    pub next_committee_published: bool,
+    /// Next-epoch members that are neither locally validated nor
+    /// carried forward by the freeze. Empty while
+    /// `next_committee_published` is false (the member list is not
+    /// known yet).
+    pub uncovered: Vec<AuthorityName>,
+    /// The epoch's consensus clock at evaluation time, and the
+    /// consensus-clock deadline after which the gate emits anyway.
+    /// `deadline_ms == u64::MAX` means no deadline is armed yet (no
+    /// consensus commit processed this epoch).
+    pub now_ms: u64,
+    pub deadline_ms: u64,
+}
+
+impl ReadySignalWait {
+    /// Seconds of consensus clock left before the liveness backstop
+    /// forces an emit; `None` when no deadline is armed.
+    pub fn seconds_until_deadline(&self) -> Option<u64> {
+        (self.deadline_ms != u64::MAX).then(|| self.deadline_ms.saturating_sub(self.now_ms) / 1000)
+    }
+}
+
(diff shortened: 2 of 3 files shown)


omersadika:
## Review (Fable 5, as the author of the refuted issue) — approve

First: this investigation corrected my misdiagnosis, and did it with the model predicting the live system to the second — the currently-running measurement hit `process_mid_epoch()` at 21:04:08 against a ~21:04 prediction, then DKG in 61s against a 64s prior. The correction and reclassification are on #2080 under my name.

Verified in the diff, the three properties the change rests on:

1. **Outcome-invariance** — the tri-state `LocalBlobVerdict` changes *visibility*, not the decision: the validated set excludes corrupt and absent identically to before (the new test asserts both exclusions explicitly). No consensus-relevant behaviour moves.
2. **The memoization's precondition is enforced, not assumed** — `insert_mpc_artifact_blob` (`authority_perpetual_tables.rs:247`) rejects any digest ≠ `Blake2b256(bytes)`, with tests in both directions predating this PR. Content-addressed keys make the cache sound by construction; `Absent` uncached is the right call.
3. **Cache lifetime is per-epoch** — dropped at the boundary, so no cross-epoch staleness surface.

The fault validation (all four properties reverted simultaneously → 8 fail / 88 pass, with the decode-count test reading 50-vs-1) is the standard this repo now expects, met. CI fully green including `v131_rollout`.

The real deliverable is the twelve-silent-hours point: on a 24h mainnet epoch, every validator spends half of every epoch in a wait that, until this PR, produced zero log output — which is precisely the void my wrong narrative grew in. After this lands, that window narrates itself at INFO on a bounded cadence. Cheap insurance against the next 3-hour misdiagnosis, mine or anyone's.

Harness-side margin fix (epoch_timeout vs epoch_duration/2) is already applied to the measurement driver. Ship it.

PR #2087, opened 2026-08-21
fix(consensus): make boot replay run under transaction-voting-off — target the last commit, not the never-written finalized commit (#2085)

### LINKED ISSUE #2085: Boot replay is dead code: with transaction voting off, read_last_finalized_commit is always None — restarts recover through consensus-core, and the boot_replay gauges never move
## What happened

During the #2064 gate-1 measurement (local 4-validator net, real class-groups traffic, two mid-epoch restarts of validator 0), both restarts logged:

```
no finalized commits in the consensus store for this epoch — starting the epoch's derived state from empty unfinalized_tail=72592
```

The store held **72,592 commits** (74,388 on the second restart) — and zero of them were finalized. `boot_replay::replay_epoch_commits` took its `replay_through == 0` early return both times. The batched fold-from-store replay merged in #2074 never ran; state was rebuilt entirely by consensus-core's `CommitObserver::recover_and_send_commits`.

## Root cause

This is structural, not a fixture artifact — it holds on every ika network, mainnet included:

1. ika runs `transaction_voting_enabled = false` at every supported protocol version (deliberately — see the comment in `to_consensus_protocol_config`, `consensus_manager/mod.rs`: flipping it panics the linearizer on commit-sync replay).
2. With voting off, consensus-core's `CommitFinalizer` treats every commit as `already_finalized` and **never writes the `finalized_commits` column family** (`commit_finalizer.rs`, the `already_finalized` branch skips `add_finalized_commit`).
3. Therefore `read_last_finalized_commit()` returns `None` forever, `replay_through` is always 0, and the "unfinalized tail" — which the boot-replay doc-comment assumes is "only ever a few" commits — is always the **entire epoch**.

Unit tests exercise the fold by building finalized commits directly against the store, so CI stays green while production always takes the other branch.

## Why nothing breaks today

Verified against consensus-core `mainnet-v1.77.2` (51d177a) and empirically in the measurement run:

- `recover_and_send_commits` scans the store in **250-commit batches** (same bound as our fold) and yields between commits; the vote-tracker recovery window is GC-bounded. Memory stays bounded.
- With voting off, rejected-transaction sets are vacuously empty, so replaying "unfinalized" commits is semantically identical to replaying finalized ones.
- Measured: 72.6k-commit recovery in ~25–30s; restarted-node RSS peak 1.10–1.43 GiB; catch-up to peers' round ~83.5k within ~38s of process start; subsequent full dwallet lifecycle green.

So the production restart path is correct and memory-bounded — it's just not the path we built, and not the path our observability watches.

## Actual consequences

- `ika_consensus_boot_replay_{target,folded}_commit_index` and `ika_consensus_boot_replay_latency_seconds` read **0 forever** in production. Any dashboard or alert on them is dead on arrival.
- The slow-start log in `consensus_manager/mod.rs` ("…a defect if `ika_consensus_boot_replay_folded_commit_index` is not advancing") points operators at a gauge that never advances — during a long mainnet boot replay (~1.73M commits ≈ 10–12 min at the measured ~2.5–2.9k commits/s) this reads as the very defect it warns about.
- `boot_replay`'s head publication for the catch-up gate (`publish_observed_consensus_head`) is also skipped on the early return, so during replay the gate has no "how far to go" until live commits arrive.
- The gauge that *does* advance during a real boot is consensus-core's, exported as `consensus_ika_commit_observer_last_recovered_commit_index`.

## Fix options

**(a) Make the fold the real path (restores #2074's intent).** With voting off, treat `read_last_commit()` as the replay target when `read_last_finalized_commit()` is `None` — sound because nothing is ever rejected. The fold then returns the true index, `recover_and_send_commits` sees `last_commit_index == replay_after` and cleanly no-ops ("Nothing to recover"), the gauges and head publication come alive, and the doc-comment's invariant is restored. Add a voting-off store fixture test asserting the fold actually runs.

**(b) Minimal: own the recovery path's observability.** Delete (or explicitly deprecate) the `boot_replay_*` gauges, point the slow-start log and dashboards at `consensus_ika_commit_observer_last_recovered_commit_index`, and rewrite the boot-replay doc-comment to say the fold only runs where finalized commits exist (today: nowhere).

(a) is the true fix; (b) is acceptable if we'd rather not touch the boot path before the release. Either way the misleading log hint should not ship as-is.

Found while running the #2064 gate-1 measurement; raw evidence in the measurement logs (both restarts, deterministic).

omersadika:
## What was wrong

`consensus_manager::boot_replay::replay_epoch_commits` picked its replay target from `store.read_last_finalized_commit()`. ika runs consensus with `transaction_voting_enabled = false` at every protocol version (deliberately — see the long comment in `to_consensus_protocol_config`), and with voting off consensus-core's `CommitFinalizer` treats every commit as `already_finalized` and so never reaches `add_finalized_commit` — the `finalized_commits` column family is never written at all. That read therefore returned `None` on every ika node on every boot, `replay_through` was always 0, the fold early-returned, the three `ika_consensus_boot_replay_*` gauges read 0 forever, `publish_observed_consensus_head` was skipped, and the whole epoch was rebuilt by consensus-core's `CommitObserver::recover_and_send_commits` instead — a correct and memory-bounded path, but not the one #2074 built and not the one our observability watches. Every unit test built its fixture with finalized commits, which is exactly why CI stayed green through all of it. Details, measurements and the alternatives considered: #2085.

## What changes

- **The replay target now follows `transaction_voting_enabled`**, not the presence of a finalized-commits row: last **finalized** commit with voting on, last commit with voting off. `replay_epoch_commits` takes a `&ConsensusProtocolConfig`; `ConsensusManager::start` builds it once and hands the *same value* to both the replay and `ConsensusAuthority::start`, so the two cannot read different flags.
- **`load_committed_subdag`** defaults the rejected-transaction set to empty only when voting is off. With voting on it still panics on a missing set, so the finalization-hole guard is unchanged. `decided_with_local_blocks` / `recovered_rejected_transactions` now mirror consensus-core's own `load_committed_subdag_from_store`: true exactly when a stored row was found. (Nothing in ika reads either field — only `rejected_transactions_by_block`, at `consensus_types/consensus_output_api.rs:73`.)
- **The `replay_through == 0` branch** is now reachable for a genuinely empty store, plus the voting-on "commits exist, none finalized yet" case, and logs the two distinctly. `unfinalized_tail` is 0 by construction under voting off, so the tail warning means what it says again.
- Doc-comments rewritten (module, `replay_epoch_commits`, `load_committed_subdag`, the target gauge, and `MysticetiConsensusHandler::new`), plus `dev-docs/specs/event-sourced-epoch.md` — "What the fold covers, and where it stops" now states the two-branch rule, why the gate is the flag, and the handoff.
- The metric **help text** for `ika_consensus_boot_replay_target_commit_index` drops the word "finalized". No metric added or renamed; `scripts/check-metric-names.sh` passes unchanged.

### Why gate on the flag rather than on `read_last_finalized_commit() == None`

The two readings of that `None` are incompatible, so no test against it is right in both worlds:

- **Voting off**: `None` means "this store does not record finalization" — nothing was ever finalized *because nothing ever needs to be*. Reading it as "nothing to replay" is the bug.
- **Voting on**: `None` means "commits exist, none has finalized yet" — a real state after a crash, whose commits have an *unknown* rejected set. Folding those would let this validator accept transactions its peers rejected.

The flag separates them exactly. It is also the right provenance: the consensus store is per-epoch (`get_store_path(epoch)`) and the protocol config is fixed for the epoch, so the flag under which every commit in that store was written is the flag being read.

## Why it is safe

**The empty-rejected-set argument.** With voting off no transaction is ever rejected, so an "unfinalized" commit's rejected set is empty *by construction* rather than unknown, and `CommittedSubDag::new`'s empty `rejected_transactions_by_block` is that commit's true rejected set. Folding it is semantically identical to folding a finalized one. Under voting on nothing changes: the target is still the last finalized commit, and a missing rejected set below it still stops the node.

**The handoff is a no-op.** Consensus is started with `CommitConsumerArgs::new(replayed_through, replayed_through)`. Under voting off `replayed_through == store.read_last_commit().index()`, so `CommitObserver::recover_and_send_commits` (`consensus/core/src/commit_observer.rs`) hits its `last_commit_index == replay_after_commit_index` early return and logs "Nothing to recover for commit observer - replay is requested immediately after last commit index" — before the `assert!(last_commit_index > replay_after_commit_index)` below it. No commit is folded twice and none is skipped: the first commit delivered live is `replay_after + 1`. `CommitConsumerMonitor::new` seeds `highest_handled_commit` with `replay_after_commit_index`, so `ReplayWaiter::wait_for_replay` returns immediately, and `ConsensusHandler::handle_consensus_commit`'s `assert!(self.last_consensus_stats.index < execution_index)` (`consensus_handler.rs:318`) would catch a torn handoff rather than silently double-fold.

**Nothing load-bearing is lost by skipping recovery.** The only side effect `recover_and_send_commits` has beyond re-sending commits is `transaction_vote_tracker.recover_and_vote_on_blocks` for unfinalized commits. Under voting off its results are unreachable: `proposer.rs:543` gates `get_own_votes` on `transaction_voting_enabled()`, and `get_reject_votes` is only called from `CommitFinalizer::process_commit`, which the `already_finalized` branch never enters. Consensus-core already declines the sibling recovery for the same reason (`transaction_vote_tracker.rs:59`, "Skipping vote tracker recovery in non-mysticeti fast path mode").

**Behavioural consequence, stated plainly**: on ika the fold now really runs at boot, so `ConsensusManager::start` takes O(epoch age) before consensus starts. That is #2074's design — the MPC drain runs `drain_while_replaying` first precisely so the bounded round channel cannot park the fold — and the measured cost is the one in #2085: ~25–30 s for 72.6k commits, RSS peak 1.10–1.43 GiB. What changes is which producer feeds the same handler, and that the gauges and the head publication now move.

## Test plan

New, next to the existing tests (which cover the voting-on branch and are unchanged):

- `replay_folds_every_commit_when_transaction_voting_is_off` — builds the store the way production writes it (commits written, **nothing** in the finalized-commits column family; asserted before the replay runs), enough commits to cross a `REPLAY_BATCH_COMMITS` boundary, and asserts the fold actually ran: the returned index equals the last commit index, `boot_replay_folded_commit_index` and `boot_replay_target_commit_index` both equal it, `sub_dag_index` reaches it, the handler saw every commit **exactly once and in order** (leader rounds compared against the store's own scan, so a double fold and a dropped commit are both visible), and the handoff lands on the equality case. This test fails on `main`.
- `replay_folds_nothing_when_voting_is_on_and_no_commit_has_finalized` — the same empty column family under voting on must fold nothing, pinning the other reading of `None`.

Commands run, all green:

```
cargo fmt --all -- --check                                         # FMT OK
cargo clippy -p ika-core --all-targets --all-features -- -D warnings
cargo test --release -p ika-core --lib -- --test-threads=4 boot_replay consensus_manager
  # test result: ok. 12 passed; 0 failed
./scripts/check-metric-names.sh                # 337 literal names OK
./scripts/check-epoch-table-write-discipline.sh
./scripts/check-invariant-violation-markers.sh
```

Closes #2085

### DIFF
--- a/crates/ika-core/src/consensus_handler.rs
+++ b/crates/ika-core/src/consensus_handler.rs
@@ -489,10 +489,15 @@ impl MysticetiConsensusHandler {
     /// Starts the task that folds commits as consensus delivers them.
     ///
     /// There is no already-processed skip here any more, and no watermark to
-    /// drive one. The boot replay has already folded every finalized commit the
-    /// consensus store holds, and consensus is started with that same index as
-    /// its `replay_after_commit_index`, so nothing consensus delivers has been
-    /// folded before. The watermark that used to gate this loop was a second
+    /// drive one. The boot replay has already folded the consensus store up to
+    /// its replay target — the store's last commit, or its last finalized
+    /// commit where transaction voting is on
+    /// (`consensus_manager::boot_replay::replay_epoch_commits`) — and consensus
+    /// is started with that same index as its `replay_after_commit_index`, so
+    /// nothing consensus delivers has been folded before. On ika, where voting
+    /// is off at every protocol version, that target is the store head and
+    /// consensus re-delivers nothing at all. The watermark that used to gate
+    /// this loop was a second
     /// record of the same fact as the consensus store's head, kept in a
     /// different database — and when the two disagreed the node could not start
     /// at all (ika #2057).
--- a/crates/ika-core/src/consensus_manager/boot_replay.rs
+++ b/crates/ika-core/src/consensus_manager/boot_replay.rs
@@ -7,11 +7,12 @@
 //! and no copy of what it derived: that state is in memory, so a start has
 //! nothing to delete and nothing to resume from. This module folds the epoch's
 //! commits back through the same handler that processes live commits, in
-//! bounded batches, until it reaches the store's last finalized commit. Only
-//! then does consensus start, and it starts from exactly that index — so the
-//! index consensus is asked to replay after is READ FROM the consensus store
-//! rather than remembered beside it, and the two can no longer disagree
-//! (ika #2057).
+//! bounded batches, until it reaches the store's replay target — the last
+//! commit, or, where transaction voting is on, the last FINALIZED commit
+//! (`replay_epoch_commits` states the rule). Only then does consensus start,
+//! and it starts from exactly that index — so the index consensus is asked to
+//! replay after is READ FROM the consensus store rather than remembered
+//! beside it, and the two can no longer disagree (ika #2057).
 //!
 //! Batching is what keeps the COMMITS flat on an old epoch: one batch of
 //! commits and their blocks is resident at a time, folded, then dropped. The
@@ -22,9 +23,11 @@
 //! The full model — what "derived" means, why the fold is deterministic, and
 //! what replay must not re-emit — is `dev-docs/specs/event-sourced-epoch.md`.
 
+use std::collections::BTreeMap;
 use std::path::Path;
 use std::time::Instant;
 
+use consensus_config::ConsensusProtocolConfig;
 use consensus_core::storage::Store;
 use consensus_core::storage::rocksdb_store::RocksDBStore;
 use consensus_core::{CommitAPI, CommitIndex, CommittedSubDag, TrustedCommit};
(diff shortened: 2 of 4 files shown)


omersadika (commented):
Reviewed the diff (minimal and correct: flag-gated target built once and handed to both replay and consensus start; empty rejected set only under voting-off; voting-on panic guard untouched; handoff lands on consensus-core's `last_commit_index == replay_after` early return, which I'd verified in the sui source while writing #2085).

Since this changes which code path every production restart takes — and the fold had never run outside unit tests — I also ran this PR's binary through the same restart-focused measurement that produced #2085 (local 4-validator net, real class-groups traffic, 2h epoch, two mid-epoch restarts of validator 0 with a lag phase between). **All 19 phases green.** Same store sizes as the #2085 run, so the numbers are directly comparable:

| | old path (main, consensus-core recovery) | this PR (ika fold) |
|---|---|---|
| restart-a store | 72,592 commits | 72,953 commits |
| replay decision logged | `no finalized commits … unfinalized_tail=72592` | `replaying … replay_through=72953 unfinalized_tail=0 transaction_voting_enabled=false` |
| replay wall time | ~25 s (inside a 33 s restart) | **1.7 s** (inside a 27 s restart) |
| handoff | recovery re-sent all 72k commits | `Nothing to recover for commit observer - replay is requested immediately after last commit index 72953` |
| `boot_replay_{target,folded}_commit_index` | 0 / 0 | **72953 / 72953** (restart-b: 74044 / 74044) |
| `boot_replay_latency_seconds` | 0 | 2 (restart-b: 1) |
| restarted-node RSS peak, restart-a / restart-b | 1.10 / 1.43 GiB | **1.09 / 1.08 GiB** |
| restart-b catch-up to peers' head (~76.8k rounds) | ~38 s from process start | ~10 s after consensus start (−1 → 74,251 → 76,330), then live |
| post-restart full dwallet lifecycle | green | green |

So the fold is both the intended path and the cheaper one: no unbounded consumer-channel enqueue-ahead, no per-commit vote-tracker work, and restart-b's memory peak drops by a third. At this payload density the fold runs at ~40k commits/s; mainnet commits are heavier, so I'd keep the #2085 "≈10–12 min for a 1.73M-commit epoch" figure as a loose upper bound rather than revise it down until the staged rollout's deliberate late-epoch restart measures it at true scale.

Merge-ready from my side. After merge, the infra dashboards can finally use the `boot_replay_*` trio.

PR #2097, opened 2026-08-24
chore(upgrade-test): retarget the cross-binary gate to v1.4.0

omersadika:
Retargets the cross-binary upgrade suite's OLD baseline from the v1.3.1 release to **v1.4.0** (`release/mainnet-v1.4.0` = `release/testnet-v1.4.0` = `ae91dc24bf`, released 2026-08-24 and rolling out across the fleet now): scenario files/gates renamed `v131_*` → `v140_*`, workflow `old_ref` defaults, RUN-flag mapping, must-refuse-mocks and fault-injection guard lists, dev-machine `OLD_BIN` defaults, and the ci-suites playbook + cross-binary-upgrade spec + test-testing playbook in the same change.

Follows the precedent of #2040 ("retarget the cross-binary baseline to the v1.3.1 release") exactly: same rename surface, same one-change-renames-everything discipline the playbook spells out, and the same old-binary mechanism — the harness **builds** the old binary from the git tag in an isolated `git worktree` at that ref's own pinned toolchain. It does not download a release asset, so the only artifact requirement is that the tag exists (verified: both network tags resolve to `ae91dc24bf`, and v1.4.0 pins the same `1.97.1` toolchain as main).

## `mid_epoch_rollback` is deliberately left at v1.3.1

This is the one place this retarget is not a mechanical rename, and it is new since #2040.

`mid_epoch_rollback` (#2077) runs the candidate first and puts the **old** release back on one validator mid-epoch. Its whole subject is an old binary reopening an epoch store in which everything its own fold writes is empty — it asserts the re-derivation through the table-writing path, measures the already-settled work it re-emits, and requires a peer to witness it authoring MPC output again.

v1.4.0 is post-#2074: the event-sourcing change landed in `927f0b3af9`, before the `1.4.0` version bump (#2088) and before the tag. So a v1.4.0 old binary would find its own fold's tables exactly where it expects them, re-derive nothing, and pass all three assertions for free. **Retargeting it would have silently emptied it**, which is precisely what "do not weaken an assertion" rules out. It keeps `old_ref=release/mainnet-v1.3.1` — the last release from before #2074 — and the pin now carries that reasoning at the site, in the workflow, the spec and the playbook.

The general rule is now written down: this scenario's OLD side is pinned to a **capability**, not to "whatever is deployed", and the family sweep must skip it.

## What the retarget costs the forward gate (recorded, not papered over)

The v1.3.1-based `v131_rollout` straddled the storage-model change by accident of timing: three OLD validators writing per-round MPC tables and a `last_consensus_stats` watermark, one NEW validator keeping that state in memory. `v140_rollout` has both sides post-#2074, so that ingredient is gone from it and it stands behind wire and MPC-output agreement only.

No assertion was changed, removed or relaxed anywhere — the assertion lists in all three renamed scenarios are byte-identical modulo the rename. What changed is the prose in `v140_rollout.rs`, `v140_churn.rs`, the workflow header, `ci-suites.md`, `cross-binary-upgrade.md` and `event-sourced-epoch.md`, which now say what the gate does and does not cover. `event-sourced-epoch.md` also records the honest consequence: **nothing exercises a pre-#2074 binary sharing an epoch with a current one in the forward direction any more, and nothing will again** — the fleet is past it, and if that is wanted as a regression record it needs its own pinned scenario.

## Protocol versions

Unaffected. v1.4.0 and main are both `MIN_PROTOCOL_VERSION = MAX_PROTOCOL_VERSION = 7`, so there is no version delta for the scenario to assume: it stays a pure binary swap and the `SupportedProtocolVersions::new_for_testing(7, 7)` pins and `expect_protocol_version_at_*(7)` assertions are unchanged.

## Also fixed in passing

`ci-suites.md`'s must-refuse-mocks list had never picked up `mid_epoch_rollback`, which the workflow has guarded since it landed. That is exactly the drift the list warns about, so it is corrected here rather than left for the next retarget.

## Validation

- `cargo check -p ika-upgrade-test --tests` clean; `cargo fmt --all -- --check` clean; `cargo clippy -p ika-upgrade-test --all-targets --all-features -- -D warnings` clean.
- **The retargeted gate was dispatched for real against the published v1.4.0 assets** and is the evidence for this PR, not a proxy for it: https://github.com/dwallet-labs/ika/actions/runs/32729477864 — **success**.
  The run resolved `old_ref=release/mainnet-v1.4.0` to the immutable commit `ae91dc24bf235fa7d8b5aaca2acf1db9f1ce178f`, built the old validator from it, and logged `v1.4.0 rollout PASSED` after all 62 flow steps — including three real DKG → Presign → Sign lifecycles (`mixed-v7-after-first-reshare`, `mixed-v7-after-second-reshare`, `all-new-v7-after-full-swap`). The workflow's false-green guard (`test result: ok. 1 passed`) is satisfied, so the gate ran rather than skipping.
- This PR's paths auto-trigger `v140_rollout` on the PR itself; that run is also **green** (2957s), resolving the same v1.4.0 commit. All other PR checks pass.

## For the repo admin — required status check

The job name is the check name, so this PR renames the status check that appears on every PR:

| | |
|---|---|
| **old check name** | `v131_rollout` |
| **new check name** | `v140_rollout` |

**Checked, and no branch-protection change is needed.** `main`'s classic protection has both `required_status_checks.contexts` and `required_status_checks.checks` **empty**, and the repo has **no rulesets**, so no check is pinned by name and nothing breaks on merge. (This PR shows as `BLOCKED` only because `main` requires 2 approving reviews.)

Worth knowing anyway if anyone has the old name saved outside branch protection — dashboards, `gh` filters, saved searches, or a personal required-check list. Nothing here touches branch protection.

### DIFF
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -9,18 +9,18 @@ name: Upgrade Test
 # dev-docs/specs/cross-binary-upgrade.md).
 #
 # Manual dispatch can run any scenario. Relevant pull requests run only
-# `v131_rollout`, avoiding the full expensive matrix on unrelated changes.
+# `v140_rollout`, avoiding the full expensive matrix on unrelated changes.
 # (The v3/v4-era rehearsals — cross_binary, v118_*, v121_rollout — were
-# retired when protocol v3/v4 support was removed; v131_rollout is their
-# successor against the current v1.3.1 release.) Pick the scenario with
+# retired when protocol v3/v4 support was removed; v140_rollout is their
+# successor against the current v1.4.0 release.) Pick the scenario with
 # the `test` input:
 #
 #   smoke        — go/no-go plumbing: 4 same-binary processes reach epoch 2.
 #                  Needs only the current build. Fastest.
 #   workload     — full user DKG -> Presign -> Sign on-chain at genesis
 #                  protocol v7. Current build only.
-#   v131_rollout — the deployed-release compatibility gate: boot the literal
-#                  v1.3.1 release (deployed on mainnet AND testnet, which run
+#   v140_rollout — the deployed-release compatibility gate: boot the literal
+#                  v1.4.0 release (deployed on mainnet AND testnet, which run
 #                  protocol v7 — the only version either binary supports, so
 #                  the run genesises there),
 #                  upgrade exactly one validator to the current build,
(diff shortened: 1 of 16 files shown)


omersadika:
**HOLD — do not merge yet** (decision 2026-08-24): this merges only once the fleet is confidently on v1.4.0 and a rollback to v1.3.1 is no longer on the table. Until then the forward gate must keep proving the v1.3.1 → current upgrade path, since v1.3.1 is still where an operator would land on a rollback (and re-upgrade from). Note the PR already keeps `mid_epoch_rollback` pinned at v1.3.1 — the hold is about the forward gate only. Merge conditions: fleet effectively fully upgraded + a few clean epochs of soak; likely alongside the v1.4.1 cut.

omersadika:
HOLD lifted by the repo owner (2026-08-24): with #2098 retiring the v1.3.1 rollback path, v1.3.1 is no longer a supported starting point and the forward gate moves to v1.4.0. Merge order: #2098 first (satisfies the current required check name), then rebase+merge this, then update branch protection required check v131_rollout → v140_rollout.

PR #2104, opened 2026-08-30
refactor: sweep the two retired always-true flag gates (#1909 item 2)

omersadika:
Part of #1909 (item 2). Items 3 and the BLS item are untouched.

## What

`internal_presign_sessions` and `fast_schnorr_supported` are both set in the v4 arm of `get_for_version`. `MIN_PROTOCOL_VERSION == MAX_PROTOCOL_VERSION == 7`, so both are true at every supported version and every branch reading them has exactly one reachable arm. This collapses each gate to that arm.

**No behavior change at v7.** Every edit is either a removed always-true conjunct, a removed unreachable branch, or a comment.

Sites swept:

- `internal_presign_sessions` — `mpc_session.rs` (global-presign routing), `dwallet_mpc_service.rs` (the `IdleStatusUpdate` wire gate and the instantiation call).
- `fast_schnorr_supported` — `mpc_manager.rs` (pool algorithm list) and `mpc_session/input.rs` (the VSS request rejection).

## `bls_checkpoints` is deliberately NOT swept

An earlier revision of this PR also collapsed the seven `bls_checkpoints` gates. That has been reverted: BLS checkpoints are on the deprecation path and the `noa_checkpoints` flag (planned for v8, off everywhere today) supersedes them. So unlike the two flags above, `bls_checkpoints` has a **planned future false arm** — sweeping its gates now would only mean rebuilding those same gates for v8.

All seven sites are restored byte-for-byte to their state on `main` (`authority_per_epoch_store.rs` ×4, `ika-node/src/lib.rs` ×2 checkpoint-service skips, `dwallet_mpc_service.rs` ×1), original comments included. `git diff 905965fc0f` shows no hunk in `authority_per_epoch_store.rs` or `ika-node/src/lib.rs` at all, and no BLS-related hunk in `dwallet_mpc_service.rs`. The flag's `FeatureFlags` comment records that it is still read and why it stays gated.

## FeatureFlags fields stay

The flag set is BCS-serialized into the `ProtocolConfig` digest that rides `AuthorityCapabilitiesV1` through consensus, so dropping a field would move every supported version's digest relative to deployed binaries. The fields keep their slots; the two swept flags' comments are rewritten to say what the flag enabled and why the field remains. The per-version arms in `get_for_version` are untouched for the same reason.

Following the precedent of #1915 (`a025ba0ee0`), accessors go and fields stay — but only one accessor was actually removable here:

- `internal_presign_sessions_enabled()` was a hand-written wrapper. Removed.
- `fast_schnorr_supported()` is **generated by the `ProtocolConfigFeatureFlagsGetters` derive** from the field itself, so it cannot be dropped while the field stays. (The derive offers `#[skip_protocol_config_accessor]`, but it would break the generated `attr_map`/`lookup_arm` code that calls `self.<field>()`, and #1915 did not use it either — the auto-generated `off_chain_validator_metadata()` getter is still there today.)

## Dead plumbing removed as a consequence

- `network_presign_pool_algorithms` loses its `fast_schnorr_supported` parameter and lists the VSS algorithms unconditionally — byte-for-byte the list the flag produced when on, appended in the same order.
- `session_input_from_request` loses its now-unused `protocol_config` parameter (single caller) along with the Rust-side VSS request rejection, which could no longer fire. On-chain `validate_curve_and_signature_algorithm` remains the enforcement for external VSS requests; it was always the primary gate, the Rust one was defense-in-depth.
- Two comments that described the VSS algorithms as "gated on the feature flag" now cite what actually constrains them — the on-chain supported-algorithm map, a property the passing `validate_vss_is_externally_inaccessible` test backs.

## `idle_status_voting.rs` test override

`test_no_idle_status_update_emitted_below_activation` used `set_internal_presign_sessions_for_testing(false)`. I read it: it exists **only** to exercise the false arm. Every assertion in it is that no `IdleStatusUpdate` reaches the wire or any validator's drain while the flag is off — that is precisely the gate this PR removes, so the test asserts a condition the code no longer has and would fail outright. **Deleted**, along with the import it solely used. No other test in the file touches the flag, and nothing else in the repo calls that setter.

## Validation

- `cargo fmt --all --check` — clean.
- `cargo clippy --all-targets --all-features -p ika-core -p ika-node` — 0 errors, 0 warnings.
- `cargo test -p ika-protocol-config` — 6/6 pass, **`snapshot_tests` byte-identical** (no `.snap.new` produced). This is the digest-stability proof: no serialized field moved.
- `cargo test --release -p ika-core dwallet_mpc::integration_tests::idle_status_voting` — 3/3 pass.
- `cargo test --release -p ika-core dwallet_mpc::integration_tests::internal_presign` — 9/9 pass.

(The two release suites above were run before the `bls_checkpoints` revert. The revert only restores code to its `main` state in files those suites do not depend on for these paths; clippy, fmt and the protocol-config snapshot were re-run after it.)

## Docs

`dev-docs/specs/fast-schnorr-vss.md` described `fast_schnorr_supported` as "still a live flag" with an unreachable reject branch — updated to say nothing reads it and why the field remains. `dev-docs/README.md` index line updated to match. Neither doc change mentions BLS checkpoints.

### DIFF
--- a/crates/dwallet-mpc-types/src/dwallet_mpc.rs
+++ b/crates/dwallet-mpc-types/src/dwallet_mpc.rs
@@ -288,10 +288,11 @@ pub enum DWalletSignatureAlgorithm {
 }
 
 impl DWalletSignatureAlgorithm {
-    /// True for the Fast Schnorr (VSS) signature algorithms. These are gated by
-    /// the `fast_schnorr_supported` protocol feature flag, support DKG-created
-    /// keys only (never imported), and do not support the combined
-    /// DKG-and-sign fast path.
+    /// True for the Fast Schnorr (VSS) signature algorithms. These support
+    /// DKG-created keys only (never imported), and do not support the combined
+    /// DKG-and-sign fast path. They are not externally requestable — the
+    /// on-chain supported-algorithm map omits them — but the internal presign
+    /// pool fills them so NOA VSS sign has presigns to consume.
     pub fn is_vss(&self) -> bool {
         matches!(
             self,
--- a/crates/dwallet-mpc-types/src/mpc_protocol_configuration.rs
+++ b/crates/dwallet-mpc-types/src/mpc_protocol_configuration.rs
@@ -202,15 +202,13 @@ pub fn supported_curve_to_signature_algorithms()
 /// This is the dedicated source for the internal pool iteration — deliberately
 /// independent of `SUPPORTED_CURVES_TO_SIGNATURE_ALGORITHMS_TO_HASH_SCHEMES`,
 /// which is the externally-requestable list serialized into the on-chain
-/// `support_config`. Fast Schnorr (VSS) variants are included here when the
-/// feature is enabled at this protocol version, but are deliberately NOT in
-/// the externally-supported map — the on-chain `validate_curve_and_signature_algorithm`
-/// rejects external `request_sign` / `request_presign` for them, while the
-/// internal pool still fills so NOA VSS sign has presigns to consume.
-pub fn network_presign_pool_algorithms(
-    fast_schnorr_supported: bool,
-) -> Vec<(DWalletCurve, DWalletSignatureAlgorithm)> {
-    let mut algorithms = vec![
+/// `support_config`. Fast Schnorr (VSS) variants are included here, but are
+/// deliberately NOT in the externally-supported map — the on-chain
+/// `validate_curve_and_signature_algorithm` rejects external `request_sign` /
+/// `request_presign` for them, while the internal pool still fills so NOA VSS
+/// sign has presigns to consume.
+pub fn network_presign_pool_algorithms() -> Vec<(DWalletCurve, DWalletSignatureAlgorithm)> {
+    vec![
         (
             DWalletCurve::Secp256k1,
             DWalletSignatureAlgorithm::ECDSASecp256k1,
@@ -225,24 +223,19 @@ pub fn network_presign_pool_algorithms(
             DWalletCurve::Ristretto,
             DWalletSignatureAlgorithm::Schnorrkel,
         ),
-    ];
-    if fast_schnorr_supported {
-        algorithms.extend([
-            (
-                DWalletCurve::Secp256k1,
-                DWalletSignatureAlgorithm::TaprootVSS,
-            ),
-            (
-                DWalletCurve::Curve25519,
-                DWalletSignatureAlgorithm::EdDSAVSS,
-            ),
-            (
-                DWalletCurve::Ristretto,
-                DWalletSignatureAlgorithm::SchnorrkelVSS,
-            ),
-        ]);
-    }
-    algorithms
+        (
+            DWalletCurve::Secp256k1,
+            DWalletSignatureAlgorithm::TaprootVSS,
+        ),
+        (
+            DWalletCurve::Curve25519,
+            DWalletSignatureAlgorithm::EdDSAVSS,
+        ),
+        (
+            DWalletCurve::Ristretto,
+            DWalletSignatureAlgorithm::SchnorrkelVSS,
+        ),
+    ]
 }
 
 /// Convert curve u32 to DWalletCurve enum
--- a/crates/ika-core/src/authority/authority_per_epoch_store.rs
+++ b/crates/ika-core/src/authority/authority_per_epoch_store.rs
@@ -5122,7 +5122,10 @@ impl AuthorityPerEpochStore {
                 verified_system_checkpoint_messages.clone(),
             );
         }
-        if make_checkpoint && !verified_system_checkpoint_messages.is_empty() {
+        if self.protocol_config().bls_checkpoints()
+            && make_checkpoint
+            && !verified_system_checkpoint_messages.is_empty()
+        {
             let checkpoint_height = consensus_commit_info.round;
 
             let pending_system_checkpoint =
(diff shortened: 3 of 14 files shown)


omersadika:
Test evidence, as promised in the description. All runs are at `3cf8f7b526` (branch HEAD, including the review-follow-up comment fix).

**Static checks**
```
cargo fmt --all --check                          # clean
cargo check --all-targets \
  -p ika-protocol-config -p dwallet-mpc-types \
  -p ika-core -p ika-node                        # 0 errors, 0 warnings
```
The only other warnings in the build are the pre-existing `serde_yaml 0.9.34+deprecated` semver-metadata notices, unrelated to this branch.

**`cargo test -p ika-protocol-config`** — 6 passed, 0 failed. This is the digest-stability proof:
```
test test::snapshot_tests ... ok
test test::test_getters ... ok
test test::test_setters ... ok
test test::lookup_by_string_test ... ok
test test::limit_range_fn_test ... ok
test test::max_version_test - should panic ... ok
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```
`snapshot_tests` came back **byte-identical** — no `.snap.new` was produced and `git status` showed no snapshot file modified. No serialized `FeatureFlags` field moved, so no supported version's `ProtocolConfig` digest shifts relative to deployed binaries.

**`cargo test --release -p ika-core dwallet_mpc::integration_tests::idle_status_voting -- --test-threads=4`** — 3 passed, 0 failed, in 1117.19s:
```
test ...::test_status_updates_distributed_through_consensus ... ok
test ...::test_split_idle_status_vote_does_not_reach_consensus ... ok
test ...::test_validators_compute_idle_status_correctly ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 537 filtered out
```
This is the file the deleted `test_no_idle_status_update_emitted_below_activation` lived in; the remaining tests, which exercise the `IdleStatusUpdate` path that is now ungated, all pass.

**`cargo test --release -p ika-core dwallet_mpc::integration_tests::internal_presign -- --test-threads=4`** — 9 passed, 0 failed, in 1439.57s:
```
test ...::test_internal_presign_instantiation_at_correct_rounds ... ok
test ...::test_internal_presign_instantiation_normalizes_byzantine_native_placeholder ... ok
test ...::test_internal_presign_continues_when_idle ... ok
test ...::test_internal_presign_multi_key_install_lag_keeps_identifiers_uniform ... ok
test ...::test_internal_presign_stale_batch_expiry ... ok
test ...::test_internal_presign_stops_at_min_pool_size_when_not_idle ... ok
test ...::test_internal_presign_ordinal_stream_rejoins_from_consensus_completions ... ok
test ...::test_internal_presign_vss_parks_until_off_chain_keys_ingested ... ok
test ...::test_mid_epoch_restart_resumes_internal_presign_ordinals_from_pool_high_water ... ok
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 531 filtered out
```
This covers both remaining gates: the now-unconditional `instantiate_internal_presign_sessions` call and the `network_presign_pool_algorithms()` list that lost its `fast_schnorr_supported` parameter. `test_internal_presign_vss_parks_until_off_chain_keys_ingested` in particular confirms the VSS pools still fill and park correctly with the flag argument gone.

Separately, `cargo test --release -p dwallet-mpc-types` passes `validate_vss_is_externally_inaccessible`, which is the on-chain-map property the rewritten `is_vss` comment now cites in place of the retired flag.

omersadika:
Updated in `d85889623f`: the `bls_checkpoints` gates are **kept**, and this PR now sweeps only `internal_presign_sessions` and `fast_schnorr_supported`. Title and description updated to match.

**Why.** BLS checkpoints are on the deprecation path — `noa_checkpoints` (planned for v8, off on every live network today) supersedes them. That gives `bls_checkpoints` a *planned future false arm*, which is exactly what the other two flags lack. Collapsing its gates now would only mean rebuilding the same seven gates for v8.

**All seven sites are byte-for-byte back to `main`**, original comments included:

| file | sites |
|---|---|
| `authority/authority_per_epoch_store.rs` | 4 — pending-system-checkpoint write, checkpoint-service notify, two checkpoint-signature dispatch arms |
| `ika-node/src/lib.rs` | 2 — dwallet and system checkpoint-service skips |
| `dwallet_mpc/dwallet_mpc_service.rs` | 1 — BLS checkpoint path |

Verified mechanically rather than by eye:
```
git diff 905965fc0f --stat -- crates/ika-core/src/authority/authority_per_epoch_store.rs \
                              crates/ika-node/src/lib.rs
# (empty — both files are identical to main)

git diff 905965fc0f -- crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs | grep -i bls
# (no output — no BLS hunk; the file's only remaining hunks are the two
#  internal_presign ones)
```
`grep -rn 'bls_checkpoints()'` finds all 7 call sites present, at their original lines in the two fully-restored files.

The `FeatureFlags` comment is no longer the "nothing reads it anymore" wording — it now records that the flag **is** still read, is true at every supported version today, and stays gated because NOA checkpoints will supersede BLS checkpoints. The `internal_presign_sessions` and `fast_schnorr_supported` changes are untouched.

Neither `dev-docs` change mentions BLS checkpoints — both edits are `fast_schnorr_supported`-only, so no doc fix was needed.

**Re-validation after the revert**
```
cargo fmt --all --check                                        # clean
cargo test -p ika-protocol-config                              # 6/6 pass
cargo clippy --all-targets --all-features -p ika-core -p ika-node   # 0 errors, 0 warnings
```
`snapshot_tests` is again **byte-identical** — no `.snap.new`, and `git status --porcelain --untracked-files=all` is empty. The protocol-config change was comment-only, so no serialized field moved.

CI on `d85889623f`: Clippy SUCCESS, Clippy (msim / simtest) SUCCESS, Format Check SUCCESS, Cargo Test Check SUCCESS.

The `idle_status_voting` (3/3) and `internal_presign` (9/9) release runs reported earlier were made before this revert. They remain valid: the revert only restores files to their `main` state, and it does not touch the internal-presign or VSS paths those suites cover.

PR #2105, opened 2026-08-30
refactor: remove the network-DKG anchor-reconstruction machinery (#1909 item 1)

omersadika:
Part of #1909 (item 1). Items 3 and the BLS item are untouched.

## Why now

The network-DKG canonical anchor migrated once per key, from a V1/V2 anchor to the aggregated V4 `decentralized_party::dkg::PublicOutput`. Every validator reconstructed the full V4 output in memory from its anchor plus the cert-pinned reconfiguration output, mirrored that in place of the stale anchor, and thereafter resolved V4 directly.

**The removal trigger is met.** Verified fleet-wide on 2026-08-30: both network-key version gauges read `4` on every reporting validator of both networks. No key remains on a V1/V2 canonical anchor, so the reconstruction path can no longer fire on either network.

## What is removed

- `NetworkEncryptionKeyPublicData::reconstructed_full_network_dkg_output` — the field and its getter (`dwallet-mpc-types/src/dwallet_mpc.rs`). The struct has exactly one construction site, so this is contained.
- `reconstruct_full_network_dkg_output` — the producer, plus its call in `build_network_encryption_key_public_data` and the `mod tests` that covered only it (`network_dkg.rs`).
- The canonical-mirror **flip** in `mpc_manager.rs`: the `match` that chose the reconstructed V4 bytes over the adopted anchor, and its serialization-failure fallback.
- Both `adopt_cert_verified_keys` tolerance arms.
- `already_adopted_key_survives_dkg_output_migration` (`integration_tests/network_key_adoption.rs`) — it drives a V2→V4 DKG-digest migration that can no longer occur.

## What is deliberately kept

- **`VersionedNetworkDkgOutput` V1..V4 variants.** BCS variant indices are wire format. V1 anchors are never rewritten on chain and must stay decodable; the V1 arm in `reconfiguration.rs` stays for exactly that reason (its comment is updated to say so rather than pointing at the deleted migration).
- **`cache_network_dkg_output` and the durable canonical mirror.** Only the *flip* was migration machinery. The mirror itself is ongoing infrastructure, and it collapses cleanly to caching the adopted anchor — which every validator holding the key agrees on byte-for-byte.
- **Both version gauges.** `ika_dwallet_mpc_network_encryption_key_canonical_dkg_output_version` is asserted on by `ika-upgrade-test/src/cluster.rs`, and it still reports a live fact — its value simply collapses to `key.network_dkg_output().version()`. The reconfiguration-version gauge never touched reconstruction. **No metric is removed, so the generated inventory in `dev-docs/conventions/metrics.md` is unchanged.**
- **General session reconstruction is untouched.** An earlier comment on the issue flagged it as load-bearing. It is a different mechanism (rebuilding sessions from consensus after a restart) that merely shares the word; I confirmed every `dev-docs` and code hit for "reconstruct" outside the anchor path belongs to it and left all of it alone.

## One judgement call worth a reviewer's eye

The two removed tolerance arms are both `} else { debug!(...) }` blocks, so removing them changes **no control flow** — both paths `continue` regardless. Only a debug log goes.

The first arm is squarely migration machinery: its message names the V2→V4 canonical migration, which can no longer happen. I rewrote the comment above it accordingly — a DKG-digest mismatch against the prior epoch's cert is now never expected.

The second arm is **not** migration-related, and its own guarding comment says so: it covers the ongoing, every-epoch case where this epoch's reconfiguration output by design mismatches the prior epoch's cert. It was on the removal list, so I removed it and adjusted the now-stale "(debug)" in its comment — but the behavior it described still happens every epoch, so if the intent was only to retire migration paths, this one arm is worth restoring. Flagging it rather than silently keeping or silently dropping it.

## Validation

- `cargo fmt --all --check` — clean.
- `cargo check --all-targets -p dwallet-mpc-types -p ika-core` — no errors, no new warnings.
- `cargo test --release -p dwallet-mpc-types` and targeted `cargo test --release -p ika-core` over the network-key adoption/DKG modules — results in a follow-up comment.

## Docs

`dev-docs/specs/handoff.md` described the reconstruct-and-flip migration as live behavior of the `NetworkDkgOutput` handoff item; updated to record it as completed and to say the mirror now caches the adopted anchor. The "canonical mirror" reference later in that file is about the durable mirror, which still exists, so it stays.


omersadika:
Test evidence, as promised in the description.

**Static checks**
```
cargo fmt --all --check                                      # clean
cargo check --all-targets -p dwallet-mpc-types -p ika-core   # 0 errors, 0 new warnings
```
The only warnings emitted are the pre-existing `serde_yaml 0.9.34+deprecated` semver-metadata notices, which are unrelated to this branch.

**`cargo test --release -p dwallet-mpc-types`** — 2 passed, 0 failed:
```
test mpc_protocol_configuration::tests::validate_all_supported_curves_to_signature_algorithms_to_hash_schemes_are_correct ... ok
test mpc_protocol_configuration::tests::validate_vss_is_externally_inaccessible ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

**`cargo test --release -p ika-core dwallet_mpc::integration_tests::network_key_adoption -- --test-threads=4`** — 6 passed, 0 failed, in 100.26s:
```
test ...::adopted_keys_always_resolve_a_network_key_id ... ok
test ...::stale_epoch_network_key_data_is_not_spawned ... ok
test ...::empty_overlay_for_prior_epoch_key_flags_stranded ... ok
test ...::empty_reconfiguration_overlay_is_not_adopted_when_cert_pins_reconfiguration ... ok
test ...::this_epoch_reconfiguration_output_is_skipped_and_current_epoch_output_recovers ... ok
test ...::unmapped_cert_referenced_key_defers_and_spawns_derivation ... ok
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 533 filtered out
```

This is the suite that directly covers `adopt_cert_verified_keys` — the function whose two tolerance arms this PR removes. Every remaining adoption test passes, including `this_epoch_reconfiguration_output_is_skipped_and_current_epoch_output_recovers`, which exercises the ongoing (non-migration) reconfiguration-digest skip that the second removed arm used to log. That it still passes confirms the removal was log-only and changed no control flow, as described above.

The deleted `already_adopted_key_survives_dkg_output_migration` is the one test not in this list; it drove the V2→V4 DKG-digest migration that can no longer occur.

omersadika:
Correction to my evidence comment above: it cited `cargo check`, which does **not** run clippy lints, so it missed a real failure. CI caught it and it is now fixed in `e08680f35f`.

**What broke.** Removing the `reconstruct_full_network_dkg_output` test module took the last `#[cfg(test)]` module out of `network_dkg.rs`. That promoted the earlier `network_key_id_derivation_tool` module (also `#[cfg(test)]`) to being the file's last test module — with three items still declared after it (`build_network_encryption_key_public_data`, `timed_sub_call`, `instantiate_dwallet_mpc_network_encryption_key_public_data_from_dkg_public_output`), which trips `clippy::items_after_test_module`. Nothing about those three items changed; the lint became reachable purely because the module below them was deleted.

**Fix.** Moved the test module below those three items so nothing follows it. No `#[allow]`. The move is copy-paste identical — 198 insertions / 198 deletions in one file, and the file's multiset of lines is unchanged before and after, so it is a pure reordering with no edited text.

**Verification**
```
cargo fmt --all --check                                    # clean
cargo clippy --all-targets --all-features -p ika-core      # 0 errors, 0 warnings, exit 0
```
(run in a dedicated target directory, so no shared-artifact contamination)

CI on `e08680f35f`: **Clippy** SUCCESS, **Clippy (msim / simtest)** SUCCESS, **Format Check** SUCCESS, **Cargo Test Check** SUCCESS.

The release test results in the previous comment are unaffected — this change reorders code without altering it.

ycscaly (approved):

PR #2117, opened 2026-09-01
fix(validator): verify the class-groups proof before candidate registration (#488)

### LINKED ISSUE #488: dwallet-network: Verify that the new validator's class groups proof is valid before `request_add_validator_candidate`

omersadika:
Resolves #488.

**Resolution:** on-chain verification is infeasible; the honest-operator path is now
verified client-side; the byzantine path was already rejected at DKG time. The issue
can be closed on merge.

## 1. On-chain verification is infeasible, not merely deferred

Move cannot run class-groups arithmetic. `request_add_validator_candidate` verifies only
the BLS proof-of-possession
(`contracts/ika_system/sources/staking/validator_info.move:117-125`) and stores
`mpc_data_bytes` opaquely (`:135`); there are zero `class_groups` references anywhere
under `contracts/`. This PR does not attempt to change that.

## 2. Discovery: the compensating control, with evidence

Traced the class-groups proof from registration into the pinned inkrypto
(`rev 3ed95fe74e08419b33c181910a4e163169fffc8f`, `class-groups` / `2pc-mpc` crates).

**Party construction does NOT verify.**
`network_dkg_v2_public_input` (`crates/ika-core/src/dwallet_mpc/crytographic_computation/mpc_computations/network_dkg.rs:790-813`)
-> `twopc_mpc::decentralized_party::dkg::PublicInput::new`
(`2pc-mpc/src/decentralized_party/dkg.rs:258`, forwarding at `:322-328`)
-> `class_groups::dkg::PublicInput::new` (`class-groups/src/dkg.rs:291`), which checks
only parameter bounds and then *stores* the keys and proofs verbatim at `:339`. A garbage
proof does not prevent the session from starting.

**Verification happens at DKG round 1, before any share is dealt.**
`advance` -> `class_groups::dkg::Party::prepare_advance` builds the CRT PVSS party
(`class-groups/src/dkg/party.rs:360-384`); its constructor first instantiates each
per-prime key as a group element
(`class-groups/src/publicly_verifiable_secret_sharing/chinese_remainder_theorem/party.rs:224-228`
— a malformed key puts its owner straight into `parties_without_valid_encryption_keys`).
Round 1 then calls `deal_and_encrypt_shares_to_valid_encryption_key_holders`
(`class-groups/src/dkg/first_round.rs:181-186`), which calls
**`verify_knowledge_of_decryption_key_proofs`** at
**`class-groups/src/publicly_verifiable_secret_sharing/chinese_remainder_theorem/deal_shares.rs:106-111`**.
The cryptographic check itself is `proof.verify(...)` at
`class-groups/src/publicly_verifiable_secret_sharing/chinese_remainder_theorem.rs:496`
(function at `:469`).

Construction is therefore not the gate, but the gate still closes *before use*: round 1
verifies before a single share is dealt or encrypted to any of those keys, so an invalid
class-groups key is never actually used by the protocol.

**Fate of the offending validator: convicted and excluded, alone.**
It is added to that round's malicious set
(`chinese_remainder_theorem.rs:504-507`), filtered out of `encryption_keys_per_crt_prime`,
and dealt no PVSS share. Every other party proceeds, provided the surviving valid parties
still form an authorized subset (`deal_shares.rs:116-120`) — only if they do not does the
session fail for everyone. On the ika side that verdict lands in `record_malicious_actors`
(`crates/ika-core/src/dwallet_mpc/mpc_manager.rs:4661-4668`). So one bad registration costs
its owner a useless committee seat; it does not wedge an epoch.

**The same control covers a validator joining an existing network**, which goes through
reconfiguration rather than a fresh network DKG:
`class-groups/src/reconfiguration/party.rs:406-430` builds the upcoming-committee PVSS
party from `upcoming_encryption_key_values_and_proofs_per_crt_prime`, and
`class-groups/src/reconfiguration/first_round.rs:208-217` deals through the same verified
path.

The byzantine path is therefore already covered. What was missing is any way for an
*honest* operator to learn about a bad payload before paying gas and losing an epoch.

## 3. What this PR adds

`dwallet_classgroups_types::verify_class_groups_encryption_key_and_proof` — deliberately
calling the *same* upstream `verify_knowledge_of_decryption_key_proofs` that the DKG and
reconfiguration first rounds call, so the pre-flight verdict and the MPC-time verdict
cannot drift apart.

Wired into both points of the registration path, in `crates/ika/src/validator_commands.rs`:

1. **`make-validator-info`** — verifies the key material derived from `root-seed.key`
   before `validator.info` is written.
2. **`become-candidate`** — `preflight_verify_class_groups_mpc_data` decodes
   `validator_info.mpc_data` (the exact on-chain shape) and verifies it before
   `request_add_validator_candidate` assembles the transaction. This is the last cheap
   point of refusal.

Both failures print an operator-facing message naming the likely cause (a corrupt or
mismatched `root-seed.key`) and the consequence of ignoring it.

The check is not placed inside `ika_sui_client::request_add_validator_candidate`: that
would pull the class-groups crypto into `ika-sui-client` and onto the genesis /
test-cluster registration paths, which use mocked key material and register many
validators at once. The CLI is the path a real operator uses, and it is covered end to end.

Doc comments state plainly what this is: it protects an honest operator from a corrupt or
mismatched root seed; it is **not** an on-chain guarantee — the on-chain-equivalent
guarantee is the DKG-time rejection above.

## 4. Doc update at the structural check

`crates/ika-core/src/validator_metadata.rs` — `blob_decodes_to_valid_mpc_data`'s doc
comment used to say only "that failure surfaces in MPC". It now names both places the
cryptographic check actually happens (the new client-side pre-flight, and
`verify_knowledge_of_decryption_key_proofs` at DKG/reconfiguration round 1 with the
offender's fate). The function itself is unchanged.

## 5. Tests

`crates/dwallet-classgroups-types/src/lib.rs`:

- `preflight_accepts_a_key_and_proof_generated_from_the_same_seed`
- `preflight_rejects_a_proof_generated_for_a_different_key` — encryption key from seed A,
  proof from seed B. The payload BCS round-trips at the on-chain shape (asserted), so it
  passes every structural gate; only the cryptography rejects it.
- `preflight_rejects_proofs_transposed_across_crt_primes` — tampering inside one otherwise
  good payload.
- `dkg_time_verification_convicts_only_the_offending_party` — the compensating control,
  exercised directly: four parties, one with a mismatched payload, run through
  `verify_knowledge_of_decryption_key_proofs` exactly as `deal_shares.rs:106-111` does.
  Asserts the offender alone is convicted and the three honest parties keep their keys.
  Not a full DKG session (that needs a cluster) — it pins the verification-and-exclusion
  step a byzantine registration actually runs into.

`crates/ika/src/validator_commands.rs`:

- `preflight_accepts_the_payload_make_validator_info_produces`
- `preflight_rejects_a_seed_mismatched_payload_before_registration` — at the
  `VersionedMPCData::V1` level, first asserting the bytes still decode.
- `preflight_rejects_mpc_data_that_is_not_a_class_groups_payload`

Run with `--release`; the debug profile runs class-groups math at `opt-level=0`.

```
cargo test --release -p dwallet-classgroups-types -p ika
```

## Files changed

- `crates/dwallet-classgroups-types/src/lib.rs` (+ `Cargo.toml`: `thiserror`)
- `crates/ika/src/validator_commands.rs`
- `crates/ika-core/src/validator_metadata.rs` (doc comment only)
- `Cargo.lock`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5028,6 +5028,7 @@ dependencies = [
  "ika-types",
  "mpc",
  "serde",
+ "thiserror 2.0.20",
  "twopc_mpc",
 ]
 
--- a/crates/dwallet-classgroups-types/Cargo.toml
+++ b/crates/dwallet-classgroups-types/Cargo.toml
@@ -10,6 +10,7 @@ group.workspace = true
 mpc.workspace = true
 crypto-bigint.workspace = true
 serde.workspace = true
+thiserror.workspace = true
 ika-types.workspace = true
 twopc_mpc.workspace = true
 
--- a/crates/dwallet-classgroups-types/src/lib.rs
+++ b/crates/dwallet-classgroups-types/src/lib.rs
@@ -1,18 +1,25 @@
 // Copyright (c) dWallet Labs, Ltd.
 // SPDX-License-Identifier: BSD-3-Clause-Clear
 
+use std::collections::HashMap;
+
 use class_groups::publicly_verifiable_secret_sharing::chinese_remainder_theorem::{
     CRT_FUNDAMENTAL_DISCRIMINANT_LIMBS, CRT_NON_FUNDAMENTAL_DISCRIMINANT_LIMBS, MAX_PRIMES,
-    generate_class_groups_keypair,
+    SecretKeyShareCRTPrimeSetupParameters,
+    construct_knowledge_of_decryption_key_public_parameters_per_crt_prime,
+    construct_setup_parameters_per_crt_prime, generate_class_groups_keypair,
+    verify_knowledge_of_decryption_key_proofs,
 };
 use class_groups::publicly_verifiable_secret_sharing::small_prime::encryption::generate_pvss_keypairs;
 use class_groups::{
-    CompactIbqf, RISTRETTO_FUNDAMENTAL_DISCRIMINANT_LIMBS,
-    SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS, SECP256R1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+    CompactIbqf, DEFAULT_COMPUTATIONAL_SECURITY_PARAMETER, EquivalenceClass,
+    RISTRETTO_FUNDAMENTAL_DISCRIMINANT_LIMBS, SECP256K1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
+    SECP256R1_FUNDAMENTAL_DISCRIMINANT_LIMBS,
 };
 use crypto_bigint::Uint;
 use dwallet_rng::RootSeed;
 use group::GroupElement as _;
+use group::PartyID;
 use ika_types::committee::{
     ClassGroupsEncryptionKeyAndProof, ClassGroupsProof, ValidatorEncryptionKeysAndProofs,
 };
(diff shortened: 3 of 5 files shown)


omersadika:
Closing unmerged — decision, not a defect in the PR.

The discovery half of this PR is the valuable part and stands: an invalid class-groups proof is rejected by the DKG (and reconfiguration) first round before any share is dealt (`class-groups/src/publicly_verifiable_secret_sharing/chinese_remainder_theorem/deal_shares.rs:106-111` → `verify_knowledge_of_decryption_key_proofs`), and the offender alone is convicted. That is the enforcing check, and it is recorded on #488.

The pre-flight half is not being merged: it verifies the bytes bound for the **on-chain** `mpc_data_bytes` field, which is deprecated — the real key material is off-chain (announced blob), and the field is to be ignored and eventually filled with a placeholder (#2119). The check also cannot catch the one common honest mistake (registering with seed A, running seed B — that is the boot-time equality check's job), so its remaining coverage is corrupt/buggy `validator.info`, which the DKG catches anyway. Not worth new code on a path that is going away.

Issue #2119, opened 2026-09-01
Deprecate on-chain mpc_data_bytes: register with a placeholder, move the two remaining readers off the field


omersadika:
## Decision

The on-chain `mpc_data_bytes` field (`validator_info.move:82/88/92`, set at `request_add_validator_candidate`) is **deprecated**. The MPC key material lives off-chain: every validator derives the full `ValidatorEncryptionKeysAndProofs` bundle from its root seed (`derive_mpc_data_blob`), announces it through consensus with a blob hash, and peers fetch it via P2P/OCS. The on-chain field carries only the bare `ClassGroupsEncryptionKeyAndProof`, and nothing about it should be enforced or verified (see #488, closed on that basis, and #2117, closed unmerged).

Direction: **a new validator registers with placeholder bytes** (empty, or any fixed sentinel that lets the Move entry create the object), and the binary ignores the field entirely.

## Why this cannot be done by "just send empty bytes" today

Two readers still decode the on-chain bytes as a real `ClassGroupsEncryptionKeyAndProof`, and both fail closed:

1. **Boot-time seed identity check** — `crates/ika-core/src/dwallet_mpc/dwallet_mpc_service.rs:3225-3245`. Decodes the on-chain bytes and compares them to the locally derived class-groups key; on decode failure or mismatch it returns `MPCManagerError` and the node does not start MPC. A validator registered with a placeholder bricks **itself** at first boot.
2. **Chain-fallback committee build** — `crates/ika-core/src/sui_connector/sui_syncer.rs:~910-945`. The bootstrap-window path (off-chain source not yet installed) decodes *every* committee member's on-chain bytes; one member with missing/undecodable bytes fails the whole build for retry (`MissingOnChainMpcData`) and fires the `chain_fallback_mpc_data_decode` invariant. One placeholder validator breaks this path for **every** node that needs it (joiners, and any node booting before its off-chain source is installed).

Secondary: the rotation machinery still round-trips the field — `set_next_epoch_mpc_data_bytes` (Move: `validator.move:615`, `validator_info.move:243`) and its readers in `crates/ika-sui-client/src/grpc_backend.rs:315-321` / `ika_validator_transactions.rs:1240-1281`.

## Work, in order

1. Re-home the boot identity check (reader 1) onto the off-chain material: compare the locally derived bundle against the validator's own announced blob hash / the off-chain record, or drop the check if the announcement path already guarantees seed identity (state which, with evidence).
[...]

PR #2120, opened 2026-09-01
[merge after Part A] feat(validator): register a placeholder in the deprecated on-chain mpc_data_bytes (#2119)

omersadika:
Part B of #2119 — step 3 and step 4 of the issue's ordered plan.

## Depends on Part A — do not merge first

This PR **must land after the Part A PR** — the one that re-homes the boot-time
seed identity check and the chain-fallback committee build off the field
(steps 1–2 of #2119), so the binary never decodes it.

**Depends on: #2121** (`fix(mpc): Part A of #2119 — stop reading the deprecated
on-chain mpc_data_bytes`). Do not merge this before that one.

Note that #2117 is *not* Part A — it is the closed, unmerged class-groups
preflight PR for #488.

Both PRs edit `dev-docs/specs/validator-mpc-data-announcements.md`, but **the
rebase is clean, not merely resolvable**: #2121 rewrites the invariant-3 clause at
lines 374-399 (plus 324-333), and this PR no longer touches that region at all —
invariant 3 is byte-identical to `main` here, and these hunks are at lines 2-11 and
34-44. #2121 owns that clause outright.

The inventory of readers is written as **history** for the same reason: this PR
merges after #2121 deletes those four sites, so present-tense text would describe
code that no longer exists. It lives in this PR's own section and says "before
#2121, four sites decoded … all four were removed there", keeping the site list
and their failure shapes for the record.

Nothing here changes the binary's readers; this PR only changes what the CLI
**sends**.

### The real constraint is ROLLOUT, not just merge order

Merge order alone is not the safety property. Precisely:

* **`become-candidate` with the placeholder is safe at any time, on any fleet
  version.** A candidate's record is never read — `get_epoch_start_system` walks
  `validator_set.active_committee.members` only (`lib.rs:395-401`), and so does
  the syncer's bootstrap-window fallback.
* **A placeholder validator MUST NOT be staked/activated into the active
  committee** until every node that builds a committee containing it runs a
  Part A binary — all committee validators, plus any fullnode or joiner that can
  still take the bootstrap-window chain fallback.

On v1.4.0 (I diffed `release/mainnet-v1.4.0`, the same commit as the testnet tag;
both reader sites are behaviourally identical to `main`) an activated placeholder
member is a **permanent** failure, not a transient one: `sui_syncer.rs:928`
decodes every member's bytes and hard-fails the whole chain-fallback build on the
first that will not parse (`chain_fallback_mpc_data_decode` +
`MissingOnChainMpcData`, retried forever — no chain state change can satisfy it),
breaking committee construction for every node in the bootstrap window; and the
new validator separately fails its own boot-time seed identity check, so it never
starts MPC.

**In a rolling upgrade, placeholder activation is the LAST step, after the whole
fleet is swapped.** Now stated in the spec doc and, in operator language, in a
warning callout on `become-candidate` in the CLI docs.

### 1. Registration submits a placeholder

`ika validator make-validator-info` no longer BCS-encodes the real
`ClassGroupsEncryptionKeyAndProof` into `validator.info`. It writes a documented
constant, and `become-candidate` submits it verbatim:

```rust
// ika-sui-client::ika_validator_transactions
pub const DEPRECATED_ON_CHAIN_MPC_DATA_PLACEHOLDER_BCS: [u8; 2] = [0x00, 0x00];

pub fn deprecated_on_chain_mpc_data_placeholder() -> VersionedMPCData {
    VersionedMPCData::V1(MPCDataV1 { mpc_data_bytes: Vec::new() })
}
```

On chain that is a `TableVec<vector<u8>>` with **one** entry of **two** bytes:
the `V1` variant index (ULEB128 `0x00`) and a zero-length vector (ULEB128
`0x00`).

**Why this and not a fully empty (zero-chunk) `TableVec`.** An empty `TableVec`
is creatable — `store_mcp_data_in_table_vec` builds the value with
`table_vec::empty` and then pushes chunks, so zero chunks is a well-formed
argument, and `validator_info.move`'s own test constructor builds the field with
`table_vec::empty`. It is the *read* side that rules it out:
`SuiClient::get_epoch_start_system` fails the **entire** epoch-start read when an
active committee member's record does not decode as a `VersionedMPCData` (a gap
there is deliberately treated as a read defect rather than a member to skip, to
avoid an unagreed party-set exclusion). A zero-chunk table reads back as zero
bytes, which is not valid BCS for `VersionedMPCData` — so a single validator
registered that way would break that read for *every* node, and that reader lives
in this crate, not in Part A's. The two-byte envelope decodes cleanly and keeps
the presence/decodability gate meaningful for old and new validators alike.

Two smaller reasons, both in the constant's doc comment: a real payload is never
zero-length, so "decodes, but empty" is already unambiguous and needs no extra
magic marker; and it is the smallest value with that property.

**The Move contract is untouched.** The entry still takes
`mpc_data_bytes: TableVec<vector<u8>>`. Move never indexes into the vector —
`validator_info.move` stores it (`new`, line 135), swaps it
(`set_next_epoch_mpc_data_bytes`, line 243) and drains it on destroy, and
`validate()` does not look at it at all — so the entry count is immaterial to the
contract.

`make-validator-info` still generates/reads `root-seed.key` and still derives the
bundle from it, so a seed that cannot produce usable MPC material fails at
registration rather than at the node's first boot. The derived values are simply
not published any more.

### 2. Rotation plumbing neutered

* Removed `set_next_epoch_mpc_data_bytes`,
  `new_ptb_set_next_epoch_mpc_data_bytes_with_drop` and
  `ptb_set_next_epoch_mpc_data_bytes_inner` from `ika-sui-client`. They had **no
  callers** anywhere in the repo (checked `crates/`, `sdk/`, tests and the
  upgrade-test crate) and were the only remaining way to write the deprecated
  field. A comment at the removal site records what they were and why they went.
* `ika validator set-next-epoch-mpc-data` is **kept, and is not a no-op** — it
  already does the right thing. It regenerates the local `root-seed.key` and
  submits no transaction; rotation completes off chain when the operator installs
  the seed and restarts, and the node re-derives and re-announces its bundle
  through consensus + P2P. Its output now says so explicitly, so an operator who
  expects a transaction digest learns that its absence is correct:
  *"No Sui transaction was submitted: on-chain MPC data is deprecated and ignored
  (ika#2119). Rotation happens off chain — install this file at the validator's
  root-seed path and restart it…"*
* `grpc_backend::get_mpc_data_from_validators_pool` no longer selects between
  `next_epoch_mpc_data_bytes` / `previous_mpc_data_bytes` and always reads the
  current record. With no writer left, the staging slots are permanently `None`,
  and the `grpc_validator_mpc_data_overlap` invariant guarded a state nothing can
  produce (that site had no alert or dashboard depending on it;
  `scripts/check-invariant-violation-markers.sh` still passes). The
  `read_next_epoch_mpc_data` flag is retained in the trait so `ika-core`'s call
  site — Part A's file — is untouched, and is documented as ignored.
* Move functions left in place, as required.

### 3. Docs

* `dev-docs/specs/validator-mpc-data-announcements.md` — new section stating the
  on-chain field is a deprecated placeholder, with the exact encoding, the
  "contract unchanged" note, and the fact that there is no on-chain rotation.
  Invariant 3 now says explicitly that its completeness gate is about the
  record's presence and decodability, not its contents, and that the payload
  readers in `ika-core` are moved off the field by Part A. Added "add no new
  readers of this field".
* `docs/content/docs/cli/validator-commands.mdx` — `make-validator-info` now
  documents the files it writes and that `mpc_data` in `validator.info` is not
  key material; `become-candidate` explains the placeholder and links the issue.
  Also corrected a stale claim under `set-next-epoch-mpc-data`: the docs said it
  replaces an existing `root-seed.key`, but the command refuses to overwrite one.

### 4. Tests

New, all in the two owned crates:

* `ika-sui-client`: `placeholder_encodes_to_the_documented_two_bytes`,
  `placeholder_round_trips_as_versioned_mpc_data`,
  `empty_table_vec_would_not_decode` (pins the reason the empty table was
  rejected), and `registration_ptb_pushes_exactly_the_placeholder_chunk`, which
  builds the real PTB and asserts exactly one `table_vec::empty`, exactly one
  `table_vec::push_back`, and one pure input equal to `[0x02, 0x00, 0x00]`.
* `ika`: `registration_payload_is_exactly_the_documented_placeholder`, which
  round-trips a `ValidatorInfo` through the `validator.info` YAML the operator
  actually carries between `make-validator-info` and `become-candidate` and pins
  the serialized field to `AAA=`; and
  `set_next_epoch_mpc_data_output_says_rotation_is_off_chain`, which pins the
  operator-facing "no transaction was submitted" message.

No existing test in either crate asserted on real on-chain `mpc_data` content, so
nothing had to be adapted.

**These pins now actually run on PR CI.** A named step in the `Cargo Test Check`
job executes the four `ika-sui-client` pins with the `ci.yaml:119-140` convention
(`set -o pipefail`, `tee`, and the "N passed" grep from #2113). I tightened the
vacuity guard to the exact count — `ok\. 4 passed` rather than a `[1-9][0-9]*`
floor — so renaming *some* of the pins is caught too, with a comment saying to
bump the number. That job already compiles `ika-sui-client` through its
`ika-core` check, so the added cost is the test run itself. Dry-run locally with
the step's exact command: `test result: ok. 4 passed`, guard exits 0. Byte-format
pins that never execute are decoration.

## Verification

* `cargo fmt --check` clean.
* `cargo clippy --all-targets --all-features -- -D warnings` clean on `ika` and
  `ika-sui-client`.
* `cargo test --release -p ika -p ika-sui-client` green.
* `scripts/check-invariant-violation-markers.sh` passes.

**Live chain check (read-only).** The one assumption the repo could not settle is
whether any validator currently has a staged `next_epoch_mpc_data_bytes`. It does
not. I walked every validator record on both networks over the public Sui GraphQL
endpoint — no keys, no transaction, read-only — and found
`next_epoch_mpc_data_bytes` and `previous_mpc_data_bytes` **`None` on all 115
mainnet validators (ika epoch 398) and all 111 testnet validators (epoch 401)**,
with a current `mpc_data_bytes` record present on every one. So collapsing the
staging-slot branch in `grpc_backend` changes nothing observable today. The code
comment there now claims only what the code establishes, and cites this
separately, as an observation about chain state rather than a guarantee.

**On-chain creatability is verified by construction, not by a live transaction.**
The PTB shape is unchanged (`table_vec::empty` followed by `push_back` per chunk);
only the chunk count drops from many to one. The Move entry accepts any
`TableVec`, `validate()` does not inspect it, and the field's own test
constructor uses an empty one. A live `request_add_validator_candidate` carrying
the placeholder is left to Part A's cluster acceptance run, since a cluster run
on this branch alone would fail on the `ika-core` readers Part A removes.

## Deliberately out of scope

* `crates/ika-swarm-config` genesis registration still writes real bytes. It does
  not have to change for the object to be creatable, and flipping it here would
  make local clusters fail on this branch before Part A lands. Post-Part-A the
  field is ignored either way.
* `sdk/typescript/src/tx/system.ts:370-385` still exposes
  `setNextEpochMpcDataBytes`. Out of this PR's file scope, and **dead rather than
  dangerous**: it passes the argument as
  `tx.pure(bcs.vector(bcs.vector(bcs.u8())))` where the Move entry types it as
  `TableVec<vector<u8>>`, an object — so the call cannot execute. Worth removing
  in a TS-SDK follow-up. Now recorded in the spec doc together with the other
  half of that caveat: `validator_info.move:299-319` `rotate_next_epoch_info`
  **does** swap a staged slot into the current record at an epoch change, so a
  staged slot would not be inert if one ever appeared.
* Validators registered before #2119 keep their real bytes on chain. Post-Part-A
  nothing reads them, so there is no migration.

Closes nothing on its own — #2119 closes when Part A and this both land.


omersadika:
Part A: #2121

